PR #6, opened 2021-12-14
add jest, unit-test action

glekner:
structure
image


glekner:
@vojtechszocs @yaacov please review.
notice the new `Node CI / unit-test (push)` check

PR #59, opened 2022-02-03
Add nav items

yaacov:
## 📝 Description
Add virtual machine instance and template links

## 🎥 Demo

nav-items

### DIFF
--- a/console-extensions.json
+++ b/console-extensions.json
@@ -4,7 +4,11 @@
     "properties": {
       "id": "virtualization",
       "name": "%plugin__kubevirt-plugin~Virtualization%",
-      "insertAfter": "workloads"
+      "insertAfter": "workloads",
+      "dataAttributes": {
+        "data-quickstart-id": "qs-nav-sec-virtualization",
+        "data-test-id": "virtualization-nav-item"
+      }
     }
   },
   {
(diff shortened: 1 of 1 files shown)


console-extensions.json
```diff
@@ -17,7 +21,44 @@
...
+        "data-quickstart-id": "qs-nav-virtualmachineinstances",
+        "data-test-id": "virtualmachineinstances-nav-item"
+      }
+    }
+  },{
+    "type": "console.navigation/href",
+    "properties": {
+      "id": "templates",
+      "section": "virtualization",
+      "name": "%plugin__kubevirt-plugin~Teamplates%",
```
glekner, line 52: Teamplates -> Templates

glekner (changes requested):

console-extensions.json
yaacov: +1 Thanks !

console-extensions.json
```diff
@@ -17,7 +21,44 @@
...
+      }
+    }
+  },{
+    "type": "console.navigation/href",
+    "properties": {
+      "id": "templates",
+      "section": "virtualization",
+      "name": "%plugin__kubevirt-plugin~Templates%",
+      "href": "/templates",
+      "namespaced": true,
```
metalice, line 54: how will this handle this kind of bug: https://bugzilla.redhat.com/show_bug.cgi?id=2047186
with current architecture? something we should consider.

yaacov: This is the fix for that bug.
no more special namespaces for templates, you only see the namespace you are in

PR #62, opened 2022-02-07
VMI list

metalice:
Signed-off-by: Matan Schatzman <mschatzm@redhat.com>

## 📝 Description

> Added VMI list.
TODO: more filters, missing columns (??) , actions(??), ??

## 🎥 Demo

> image

PR #70, opened 2022-02-10
Add Templates list

hstastna:
## 📝 Description

Add VM Templates list. The list shows the useful information about the template configuration. The list columns are:
- _Name_
- _Namespace_ (support all namespaces)
- ~~_Template provider_ (+ support level?)~~
- ~~_Source_ (to monitor availability and status of boot source)~~
- _Boot source availability_
  - check if DataSource is ready or if PVC exists
- _CPU | Memory_
- _Actions_ (will be added in a followup PR)
  - Clone Template
  - Edit boot source
  - Delete Template

## 🎥 Demo
redhat

---

TODO:
- [x] display only the **VM** Templates,
- [x] add Templates filtering
- [x] align columns properly
- [x] make the columns sortable
- [x] _Name_ column
- [x] _Namespace_ column
- [x] ~~_Template provider_ column~~
- [x] ~~_Source_ column~~
- [x] _Boot source availability_ column
- [x] _CPU | Memory_ column
- [x] fix filtering issues


src/models/index.ts
```diff
@@ -0,0 +1,67 @@
+import { K8sModel } from '@openshift-console/dynamic-plugin-sdk/lib/api/common-types';
```
avivtur, line 1: We recently upgraded the kubevirt-api, so you should be able to import all those models from `@kubevirt-ui/kubevirt-api/console/models/` directory

src/views/templates/list/TemplatesList.tsx
```diff
@@ -0,0 +1,120 @@
...
+    title: t('Template provider'),
+    id: 'provider',
+  },
+  {
+    title: t('Source'),
+    id: 'source',
+  },
+];
+
+const VMRow: React.FC<RowProps<V1VirtualMachine, { kind: string }>> = ({
```
avivtur, line 38: please consider change name to `VirtualMachineTemplateRow` :)

avivtur (commented):
I think it will also be nice to split each copmonent to a separated file, for better maintainability,
you can see this PR as an example: https://github.com/kubevirt-ui/kubevirt-plugin/pull/62/files

src/views/templates/list/VirtualMachineTemplatesList.tsx
```diff
@@ -0,0 +1,88 @@
+import * as React from 'react';
+import {
+  TEMPLATE_TYPE_BASE,
+  TEMPLATE_TYPE_LABEL,
+  TEMPLATE_TYPE_VM,
+} from 'src/views/catalog/templatescatalog/utils/constants';
```
yaacov, line 6: we should not use information from another extension.
we may want to remove the catalog, change it, remove/change this consts in the catalog some day.

If you need this consts for this extension, add them in this extension

src/views/templates/list/VirtualMachineTemplatesList.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const { t } = useKubevirtTranslation();
+
+  const [templates, loaded, loadError] = useK8sWatchResource<V1Template[]>({
+    kind: 'Template',
+    selector: {
+      matchExpressions: [
+        {
+          operator: 'In',
+          key: TEMPLATE_TYPE_LABEL,
+          values: [TEMPLATE_TYPE_BASE, TEMPLATE_TYPE_VM],
```
yaacov, line 62:
```suggestion
          operator: 'Exists',
          key: 'template.kubevirt.io/type',
```
the reason I like using `exist` is for currecness, we will support any template type, not just `base` and `vm`
the reason I like clear text instead of const is because it's clearer to read and debug in case you want to know what your filtering by, you don't need to look up the value of the const.

src/views/templates/list/VirtualMachineTemplatesColumns.tsx
```diff
@@ -0,0 +1,28 @@
...
+    id: 'provider',
+  },
+  {
+    title: t('Source'),
+    id: 'source',
+  },
+  {
+    title: t('Actions'),
+    id: 'actions',
+  },
```
yaacov, line 25:
```suggestion
   {
    title: '',
    id: '',
    props: { className: 'dropdown-kebab-pf pf-c-table__action'}, // once we have Kebab{ className: Kebab.columnClass }
  },
```
actions column doesn't have a title

yaacov: cc:// @vojtechszocs about the Kebab ^^

src/views/templates/list/VirtualMachineTemplatesList.tsx
hstastna: We will support any template type, not just base and vm, but I was told that for now, we want to display only the VM Templates... so now I am confused. Seems that things have changed...

I would move those constants somewhere to some more generic place, for example _src/utils/constants.ts_, as those are already useful in more places. Copying the same strings representing the same info doesn't seem to me as the best for providing more reusable and understandable code. And looking up the value of the const takes me one click. IMO using the same constant in more places also provides the info about the fact that the string represents the same information, also shows the connections between different parts of the code.

src/views/templates/list/VirtualMachineTemplatesList.tsx
hstastna: Thank you for pointing this, you are right, not a good practice - importing something from some other extension. And I would better move those constants to some more generic place, for the reasons I've already mentioned in my another comment.

src/views/templates/list/VirtualMachineTemplatesList.tsx
yaacov: > We will support any template type, not just base and vm, but I was told that for now, we want to display only the VM Templates... so now I am confused. Seems that things have changed...

we started supporting any template since 4.10, the main reason for the template list redesign is to better align with the beck-end who is supporting common templates and non common templates from any namespace.

in cases when design is not clear ping @tnisan or me

yaacov: > I would move those constants somewhere to some more generic place,

once the plugin is stable and we can go over it and find commonly used patterns, we may choose some constants used all over the place, document them (so every one can re-use them safely and currectly) version them (so changes done for one components can be tracked in case they affect other components) and put them in the shared directory

until them, create constants inside the extension you are working on, we will have time to do optimizations and collect shared components and constants once the plugin get stable. 

the lable selector currently used by the template list should not be shared by other extinctions, you do not want an implementation  details in the templates list to affect the virtualmachine extension or the catalog, they may have other needs, even if currently they may use the same label, it does not mean they will continue to use the same value.
you will end up with two similar const one for template extension and one for the other extension, or worse a bug in one of them if the const is changed to fix one of the extensions.

lets say the template list use TEMPLATE_TYPE_LABEL and the catalog use the same label, but in some future we have a bug and change the value of TEMPLATE_TYPE_LABEL for the list view, suddenly the catalog will fail, and we will not know why, it uses TEMPLATE_TYPE_LABEL, it will not be trivial to understand the the value was changed to fix a bug in the template list,

yaacov: > Catalog contains the same templates

no, we don't know that now, catalog may have other needs, the design of the catalog is not finished, and we certainly don't know what requirements we will get in the future.

once we have a stable code we can look for common patterns, and do other optimizations, until then, please only work within the code space of the extension you are working on.

as @vojtechszocs suggested we will have to have clear guidelines how to identify common patterns, how many times a common code should appear in the code, and size the repeatable code should be, what is ok to repeat in the name of readability and correctness, until we have stable code we can look for patterns in, and until we agree on what is a repeatable pattern, lets be safe and use the extension space.

yaacov:
needs i18 update

hstastna:

src/views/templates/list/components/VirtualMachineTemplatesColumns.tsx
```diff
@@ -0,0 +1,34 @@
+import { TFunction } from 'i18next';
+
+import { K8sResourceCommon, TableColumn } from '@openshift-console/dynamic-plugin-sdk';
+import { sortable } from '@patternfly/react-table';
+
+const VirtualMachineTemplatesColumns: (t: TFunction) => TableColumn<K8sResourceCommon>[] = (t) => [
```
metalice, line 6: it can be exported as custom hook so you don't need to pass t into it and just consume t :)

src/views/templates/list/components/VirtualMachineTemplatesRowFilter.ts
```diff
@@ -0,0 +1,37 @@
...
+];
+
+const includeFilter = (compareData: FilterValue, compareString: string) => {
+  return (
+    compareData.selected?.length === 0 ||
+    compareData.selected?.includes(compareString) ||
+    !compareData?.all?.find((s: string) => s === compareString)
+  );
+};
+
```
metalice, line 23: can be exported to utils file :)

src/views/templates/list/components/VirtualMachineTemplatesSource.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { getProcessedTemplateBootSource } from '../../utils/selectors';
+
+const fillRequiredParameters = (template: V1Template): V1Template => {
+  const parameters = template?.parameters;
+  parameters.filter((p) => p.required && !p.value).forEach((p) => (p.value = 'blank'));
+  return {
+    ...template,
+    parameters,
+  };
+};
```
metalice, line 18: nothing is returned from this function or I'm missing something?

metalice: also can be exported to utils file

metalice (commented):
left a few feedbacks, overall seems nice :)

src/views/templates/list/components/VirtualMachineTemplatesSource.tsx
hstastna: @metalice do you mean to export `fillRequiredParameters` somewhere to _src/utils/resources/template/_ or somewhere else?

src/views/templates/utils/virtualMachineTemplatesRowFilter.ts
```diff
@@ -0,0 +1,37 @@
...
+import { getTemplateProviderName } from '@kubevirt-utils/resources/template/utils';
+import { FilterValue, RowFilter } from '@openshift-console/dynamic-plugin-sdk';
+
+export const templateProviders = [
+  {
+    id: 'KubeVirt',
+    title: 'KubeVirt',
+  },
+  {
+    id: 'User', // TODO translate?
```
hstastna, line 11: @metalice @glekner Do we need translation for this?

glekner: no

src/views/templates/utils/virtualMachineTemplatesRowFilter.ts
```diff
@@ -0,0 +1,37 @@
...
+
+const includeFilter = (compareData: FilterValue, compareString: string) => {
+  return (
+    compareData.selected?.length === 0 ||
+    compareData.selected?.includes(compareString) ||
+    !compareData?.all?.find((s: string) => s === compareString)
+  );
+};
+
+export const useFilters = (): RowFilter[] => {
```
avivtur, line 24: since this is a hook, it will be best if you transfer this into a separate file, under hooks directory IMO

src/views/templates/utils/virtualMachineTemplatesCPUMemory.ts
```diff
@@ -0,0 +1,42 @@
...
+    value = combinedVal;
+  } else {
+    value = combinedVal.slice(0, index);
+    unit = combinedVal.slice(index);
+  }
+
+  return [value, toIECUnit(unit)];
+};
+
+export const useVirtualMachineTemplatesCPUMemory = (template: V1Template): string => {
```
avivtur, line 36: same here, since this is a hook, it will be best if you transfer this into a separate file, under hooks directory IMO

src/views/templates/list/components/VirtualMachineTemplatesColumns.tsx
```diff
@@ -0,0 +1,37 @@
+import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
```
avivtur, line 1: this file should also exist under hooks directory

src/views/templates/utils/virtualMachineTemplatesCPUMemory.ts
```diff
@@ -0,0 +1,42 @@
+import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
```
avivtur, line 1: The logic in this file was implemented already and exists here: https://github.com/kubevirt-ui/kubevirt-plugin/blob/main/src/utils/resources/vm/utils/disk/size.ts

src/utils/resources/template/hooks/useVmTemplateSource/useVmTemplateSource.ts
```diff
@@ -47,16 +47,21 @@ export const useVmTemplateSource = (template: V1Template): useVmTemplateSourceVa
...
+    return getDataSource(name, namespace)
+      .then((dataSource) => {
+        if (
+          dataSource?.status?.conditions?.find((c) => c.type === 'Ready' && c.status === 'True')
+        ) {
+          setIsBootSourceAvailable(true);
+          setTemplateBootSource({
+            type: BOOT_SOURCE.PVC_AUTO_UPLOAD,
+            source: {
+              pvc: dataSource.spec.source.pvc,
```
avivtur, line 61: please add optional chaining => `pvc: dataSource?.spec?.source?.pvc`

src/views/templates/list/VirtualMachineTemplatesList.tsx
```diff
@@ -0,0 +1,50 @@
...
+        />
+        <VirtualMachineTemplateSupport />
+        <VirtualizedTable<K8sResourceCommon>
+          data={filteredData}
+          unfilteredData={data}
+          loaded={loaded}
+          loadError={loadError}
+          columns={useVirtualMachineTemplatesColumns()}
+          Row={VirtualMachineTemplatesRow}
+          rowData={{ kind: 'Template' }}
```
avivtur, line 43: I think it would be better to use `TemplateModel.kind` instead of hard coded const

src/views/templates/utils/virtualMachineTemplatesCPUMemory.ts
hstastna: I'm not sure if all the logic from this file could be replaceable by what already exists...

hstastna: Or maybe you meant logic in https://github.com/kubevirt-ui/kubevirt-plugin/pull/151/files#diff-6384a372b7cedbc72c535dc614e0c3ebcb1f8937c7f6638028d30082dcb300b1 ?

PR #82, opened 2022-02-16
Add hook that checks if Template supports QuickCreate

glekner:
## 📝 Description

Added `useIsTemplateSupportsQuickCreate` that detects if QuickCreate is supported by a given template.
The Drawer will show a skeleton until the hook detection completes

## 🎥 Demo
Here is a Demo with Network Throttled to demonstrate the loading experience

https://user-images.githubusercontent.com/24938324/154291800-e38f7d6c-22c1-4db2-8b20-60464f853b98.mov

### DIFF
--- a/webpack.config.ts
+++ b/webpack.config.ts
@@ -57,7 +57,7 @@ const config: Configuration = {
     rules: [
       {
         test: /\.(jsx?|tsx?)$/,
-        exclude: /node_modules/,
+        exclude: /node_modules\/(?!(@kubevirt-ui)\/kubevirt-api).*/,
         use: [
           {
             loader: 'ts-loader',
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -15,7 +15,9 @@ const config: Config.InitialOptions = {
   testURL: 'http://localhost/',
   testRegex: '.*\\.test\\.(ts|tsx|js|jsx)$',
   testPathIgnorePatterns: ['<rootDir>/node_modules/', '<rootDir>/dist/'],
-  transformIgnorePatterns: ['<rootDir>/node_modules/(?!(@patternfly|@openshift-console\\S*?)/.*)'],
+  transformIgnorePatterns: [
+    '<rootDir>/node_modules/(?!(@kubevirt-ui/kubevirt-api|@patternfly|@openshift-console\\S*?)/.*)',
+  ],
   setupFilesAfterEnv: ['<rootDir>/jest-setup.ts'],
   moduleNameMapper: {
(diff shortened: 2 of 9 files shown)


src/views/catalog/templatescatalog/hooks/useIsTemplateSupportsQuickCreate.ts
```diff
@@ -0,0 +1,47 @@
+import * as React from 'react';
+
+import { V1Template } from '@kubevirt-ui/kubevirt-api/console';
+import { k8sCreate } from '@openshift-console/dynamic-plugin-sdk';
+import { K8sModel } from '@openshift-console/dynamic-plugin-sdk/lib/api/common-types';
+
+export const ProcessedTemplatesModel: K8sModel = {
```
glekner, line 7: @yaacov do we need this model?

yaacov: do the sdk expose any models ?

glekner: the sdk doesn't expose models

yaacov: than we will add it to the API repo :+1:

webpack.config.ts
```diff
@@ -57,7 +57,7 @@ const config: Configuration = {
     rules: [
       {
         test: /\.(jsx?|tsx?)$/,
-        exclude: /node_modules/,
+        exclude: /node_modules\/(?!(@kubevirt-ui)\/kubevirt-api).*/,
```
yaacov, line 60: shouldn't this be in a separate PR ?

src/views/catalog/templatescatalog/hooks/useIsTemplateSupportsQuickCreate.ts
```diff
@@ -0,0 +1,47 @@
...
+  labelPlural: 'Processed Templates',
+  apiVersion: 'v1',
+  apiGroup: 'template.openshift.io',
+  plural: 'processedtemplates',
+  abbr: 'PT',
+  namespaced: true,
+  kind: 'Template',
+};
+
+export const useIsTemplateSupportsQuickCreate = (template: V1Template): [boolean, boolean, any] => {
```
yaacov, line 18: It will be nice to have this hook return the VM to be crated instead of just if an error happened
the name can then be more informative, it will tell what we actullay doing ( useProcessedTemplate ? )
and the user will use the return error as indicator if the template is ready to be created as is, WDYT ?

glekner:
please lgtm @yaacov

PR #83, opened 2022-02-16
bump sdk versions

glekner:
## 📝 Description

bump console dynamic sdk versions

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,8 +53,8 @@
   "devDependencies": {
     "@cypress/webpack-preprocessor": "^5.11.0",
     "@kubevirt-ui/kubevirt-api": "^0.0.34",
-    "@openshift-console/dynamic-plugin-sdk": "0.0.3",
-    "@openshift-console/dynamic-plugin-sdk-internal": "^0.0.2-wp5",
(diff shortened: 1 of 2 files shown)


src/views/catalog/templatescatalog/TemplatesCatalog.tsx
```diff
@@ -25,7 +25,7 @@ const TemplatesCatalog: React.FC<RouteComponentProps<{ ns: string }>> = ({
   },
 }) => {
   const { t } = useKubevirtTranslation();
-  const [isAdmin, isAdminLoaded] = useIsAdmin();
+  const isAdmin = useIsAdmin();
```
yaacov, line 28:
```suggestion
  const canListNS =  useFlag('CAN_LIST_NS');
```

IMHO using `CAN_LIST_NS` is more correct because we don't actually test for admin user, we test for any user that can list namespaces, I also like using the canListNS approach because it's reduces the need for the isAdmin wrapper.

yaacov: btw/ this should actually be two PRs, one for version bumb, and one for using the new hook ...

PR #84, opened 2022-02-17
[WIP] Create Virtual Machine Page

upalatucci:


src/views/catalog/customize/CustomizeVirtualMachine.scss
```diff
@@ -0,0 +1,37 @@
+.customize-vm {
```
glekner, line 1: there is no need for any of this css.
you should just copy classnames from similar forms.

check http://localhost:9000/catalog/instantiate-template?template=mariadb-persistent
for example

src/views/catalog/customize/CustomizeVirtualMachine.tsx
```diff
@@ -0,0 +1,185 @@
...
+  Stack,
+  StackItem,
+  TextInput,
+  Title,
+} from '@patternfly/react-core';
+import ExternalLinkAltIcon from '@patternfly/react-icons/dist/esm/icons/external-link-alt-icon';
+
+import { CustomizeVirtualMachineScheleton } from './components/CustomizeVirtualMachineScheleton';
+
+import '@patternfly/patternfly/patternfly.css';
```
glekner, line 26: why is this needed?

src/views/catalog/customize/CustomizeVirtualMachine.tsx
```diff
@@ -0,0 +1,185 @@
...
+import './CustomizeVirtualMachine.scss';
+
+type RightHeadingProps = {
+  template: V1Template;
+};
+
+const RightHeader: React.FC<RightHeadingProps> = ({ template }) => {
```
glekner, line 33: copy the classnames from the link I put above

src/views/catalog/customize/CustomizeVirtualMachine.tsx
```diff
@@ -0,0 +1,185 @@
...
+const CustomizeForm: React.FC<CustomizeFormProps> = ({ template }) => {
+  const { t } = useKubevirtTranslation();
+  const [optionalFieldsExpanded, setOptionalFieldsExpanded] = React.useState(true);
+  const optionalFields = template.parameters?.filter((parameter) => !parameter.required);
+  const requiredFields = template.parameters?.filter((parameter) => parameter.required);
+
+  return (
+    <Form>
+      {requiredFields?.map((field) => (
+        <FieldGroup
```
glekner, line 103: @yaacov @metalice @avivtur  we need to implement form validation infrastructure.
I'll try to think of the best way , let me know what you think

src/views/catalog/customize/CustomizeVirtualMachine.tsx
```diff
@@ -0,0 +1,185 @@
...
+  );
+};
+
+type CustomizeVirtualMachineProps = {
+  match: match<{ ns?: string }>;
+};
+
+const CustomizeVirtualMachine: React.FC<CustomizeVirtualMachineProps> = () => {
+  const { t } = useKubevirtTranslation();
+  const { params } = useURLParams();
```
glekner, line 150: nice

glekner (changes requested):

src/views/catalog/customize/CustomizeVirtualMachine.tsx
upalatucci: Yes. in this case it's not complicated. I have only to be sure that the field is not empty. But in the future, we could have more complicated validations

src/views/catalog/customize/CustomizeVirtualMachine.tsx
upalatucci: Ah sorry i didn't want to commit this :D

src/views/catalog/customize/CustomizeVirtualMachine.scss
upalatucci: Oh thanks!!

src/views/catalog/customize/CustomizeVirtualMachine.tsx
upalatucci: okay thanks

src/views/catalog/customize/CustomizeVirtualMachine.scss
upalatucci: It's a little different as the buttons should be in a sticky position but it can help for sure. thanks again

src/views/catalog/customize/CustomizeVirtualMachine.tsx
yaacov: > But in the future, we could have more complicated validations

Validation in the UI should be expicitly `truthy`, so apart from checking for valid names (see documentaion: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/ ) we can't do much without sending to the backend.

src/views/catalog/customize/CustomizeVirtualMachine.tsx
```diff
@@ -0,0 +1,98 @@
...
+  const loading = !loaded && !error;
+
+  if (error) return <CustomizeError />;
+
+  return (
+    <Sidebar isPanelRight hasGutter className="customize-vm__container">
+      <SidebarContent className="customize-vm__content">
+        <Title headingLevel="h1">{t('Create VirtualMachine from template')}</Title>
+        {template && <CustomizeForm template={template} />}
+        {loading && <CustomizeVirtualMachineScheleton />}
```
avivtur, line 88: `CustomizeVirtualMachineScheleton` ==> `CustomizeVirtualMachineSkeleton`

src/views/catalog/customize/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const VIRTUAL_MACHINE_KIND = 'VirtualMachine';
```
avivtur, line 1: this const is redundant, please import `VirtualMachineModel` like this:
`import VirtualMachineModel from '@kubevirt-ui/kubevirt-api/console/models/VirtualMachineModel';`
and use the kind property instead of `VIRTUAL_MACHINE_KIND` const

src/views/catalog/customize/CustomizeVirtualMachine.tsx
```diff
@@ -0,0 +1,98 @@
...
+      version: 'v1',
+      kind: 'Template',
+    },
+    isList: false,
+    namespaced: true,
+    name,
+    namespace: templateNamespace,
+  });
+
+  const loading = !loaded && !error;
```
avivtur, line 79: can you tell me why the check for the error is also needed? isn't `loaded` enough?

src/views/catalog/customize/components/CustomizeForm.tsx
```diff
@@ -0,0 +1,152 @@
...
+} from '@patternfly/react-core';
+
+import { DEFAULT_NAMESPACE, VIRTUAL_MACHINE_KIND } from '../constants';
+
+type FieldGroupProps = {
+  field: TemplateParameter;
+  showError?: boolean;
+};
+
+const FieldGroup: React.FC<FieldGroupProps> = ({ field, showError }) => {
```
avivtur, line 28: nice :) 
can you extract this component to another file for better readability?

src/views/catalog/customize/constants.ts
upalatucci: okay thanks

src/views/catalog/customize/CustomizeVirtualMachine.tsx
upalatucci: i had some problems with errors and loaded at the start :D . I don't need this anymore. thanks

src/views/catalog/customize/components/CustomizeForm.tsx
upalatucci: okay

src/views/catalog/customize/CustomizeVirtualMachine.scss
yaacov: No stick things please

PR #87, opened 2022-02-18
update template catalog to latest design

glekner:
## 📝 Description

update template catalog and drawer to latest design

## 🎥 Demo

Screen Shot 2022-02-18 at 13 40 16
Screen Shot 2022-02-18 at 13 40 24


src/views/catalog/templatescatalog/utils/helpers.ts
```diff
@@ -86,6 +88,23 @@ export const getTemplatePVCName = (template: V1Template): string =>
   getTemplateParameterValue(template, TEMPLATE_BASE_IMAGE_NAME_PARAMETER) ||
   getTemplateParameterValue(template, TEMPLATE_DATA_SOURCE_NAME_PARAMETER);
 
+export const getTemplateBootSourceType = (template: V1Template): BOOT_SOURCE => {
```
glekner, line 91: @yaacov please review this function as a starting point.
We now search for `${NAME}` but we will switch to `rootdisk` in the future.
right now the supported types are PVC, URL, Registry

yaacov: a posible flow described below, I think is should not be under "utils" it's a big algorithm with a possibility to change and evolve over time, I would be happier if we have a directory dedicated to implementing how to find the source, one of the exported hooks/functions will be "getTemplateRootdiskDataVolumeSource(template)"


a. we support only templates that have `datavolumetemplates`
b. we look for a datavolumespec that is linked to the `rootdisk`
c. in datavolumespec we support:
source:
https://kubevirt-ui.github.io/kubevirt-api/interfaces/kubevirt_models.V1beta1DataVolumeSpec.html#source
and sourceRef:
https://kubevirt-ui.github.io/kubevirt-api/interfaces/kubevirt_models.V1beta1DataVolumeSpec.html#sourceRef
d. in the case of source:
https://kubevirt-ui.github.io/kubevirt-api/interfaces/kubevirt_models.V1beta1DataVolumeSource.html
we support:
http, pvc and registry
e. in the case of sourceRef:
https://kubevirt-ui.github.io/kubevirt-api/interfaces/kubevirt_models.V1beta1DataVolumeSourceRef.html
we support:
kind = "DataSource"
need to talk to ronen because it's a pvc, but it's auto-upload, we may want to call it something else "?pvc - auto upload?"

f: in the case of "pvc" and "pvc-auto-upload" we may need to test if the pvc exist.
if "pvc": 
check pvc exist
if "pvc-auto-upload"
check data-source exist -> check spec.source of datasource -> if pvc check if exist

codecov.yml
```diff
@@ -1,8 +1,7 @@
 codecov:
   require_ci_to_pass: false
   notify:
-    require_ci_to_pass: no
-    after_n_builds: 1
+    wait_for_ci: no
```
yaacov, line 4: that is a different PR

PR #93, opened 2022-02-22
Add initial Context Provider for VM Wizard

glekner:
## 📝 Description

Adding a global context provider extension with the following values:

```ts
type WizardTemplateContextType = {
  /** the template used for the wizard */
  template?: V1Template;
  /**
   * updates the template vm object. The provided vm object will be merged with the existing vm object
   * @param vmObject V1VirtualMachine
   */
  updateVmObject?: (vmObject: V1VirtualMachine) => void;
};
```


glekner:
@metalice @vojtechszocs @yaacov plz review

src/views/catalog/utils/WizardTemplateContext.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { getTemplateVirtualMachineObject } from './vm-template-source/utils';
+
+type WizardTemplateContextType = {
+  /** the template used for the wizard */
+  template?: V1Template;
+  /**
+   * updates the template vm object. The provided vm object will be merged with the existing vm object
+   * @param vmObject V1VirtualMachine
+   */
+  updateVmObject?: (vmObject: V1VirtualMachine) => void;
```
yaacov, line 15: why not just provide the new template  ?

looks more correct to me to have a template object in the session state
and then replace it each time it changes

e.g.
```
session state:
{
    templae-<name>-<namespace>-<uid>: the-template-obj
    templae-<name>-<namespace>-<uid>: template...
}
```
clean after a vm is created, or let it clean up each session

src/views/catalog/utils/WizardTemplateContext.tsx
```diff
@@ -0,0 +1,43 @@
...
+  };
+};
+
+function setSessionStorage(key: string, value) {
+  window.sessionStorage.setItem(key, JSON.stringify(value));
+}
+
+function getSessionStorage(key: string) {
+  const value = window.sessionStorage.getItem(key);
+  return value ? JSON.parse(value) : undefined;
```
metalice, line 42: JSON is not JavaScript. It uses a limited subset of the JavaScript syntax, but the only data types it can encode are:

null
boolean
number
string
object: map from string to any of the types
array: list of values
(see http://json.org/)

JSON is not capable of serializing functions. Any functions will be omitted (like undefined), prototypes will be destroyed, etc.

If we going save or pass any function we should avoid it

src/views/catalog/utils/WizardTemplateContext.tsx
```diff
@@ -0,0 +1,43 @@
...
+   */
+  updateTemplate?: (template: V1Template) => void;
+};
+
+export const WizardTemplateContext = React.createContext<WizardTemplateContextType>({});
+
+export const WizardTemplateContextProvider = WizardTemplateContext.Provider;
+
+export const useWizardTemplate = (): WizardTemplateContextType => {
+  const [template, setTemplate] = React.useState<V1Template>(() => getSessionStorage('template'));
```
metalice, line 20: sessionStorage my guess isnt reactive and its async. Have you tested to see if data really update and cause rerender from changes in 2 different components?

glekner:
please review @vojtechszocs @metalice @yaacov

glekner:
I decided to implement session storage only for the current template. The storage is reset after VM creation. 
This helps the wizard survive a full page reload. 
Also helps us develop the wizard faster as we don't need to fill the customize form each time we update the code. 
  
@metalice @yaacov @vojtechszocs @avivtur 
I'm still not sure about processing the template after each change. Processing the template does not give any validation to the VM object. It may be best to just update the template state without processing it (backend call) and build forms with great validation.

yaacov:
> It may be best to just update the template state without processing it (backend call) and build forms with great validation.

Our requirements are:
a - no client logic
b - only one source of truth and it is the back-end 

requirement a means that we should not implement validation at all (not complicated, anyways)
requirement b means that any validation we do on the UI side is not valid, we will need to validate using the back end each time we want to show the user indication about validity of input (unless it's trivial like missing required field,  integers, positive numbers and names validations)

PR #97, opened 2022-02-25
Combine Catalog routes to one Switch

glekner:
## 📝 Description

Combine Catalog routes into a single component with a nested Switch.
This will help us share state between these routes in a more efficient and isolated way.
Next step is to wrap the new Switch with a custom Context Provider

### DIFF
--- a/console-extensions.json
+++ b/console-extensions.json
@@ -100,16 +100,6 @@
   {
     "type": "console.page/route",
     "properties": {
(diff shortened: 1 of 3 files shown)


yaacov:
> Next step is to wrap the new Switch with a custom Context Provider

tileview extention
customize-template extention 
and review extention

are three separate extensions, giving them a shared context will prevent a good separation, it may tempt users to share data or utilities between this services.
the "contract" between this extensions is that they pass and recive `name`+`namespace` of the template, leaving each extension to add/patch/delete the template. sharing a common context break the extension signature.

PR #116, opened 2022-03-02
Add loading component

avivtur:
## 🎥 Demo

image

### DIFF
--- a/src/utils/components/Loading/Loading.tsx
+++ b/src/utils/components/Loading/Loading.tsx
@@ -1,0 +1,11 @@
+import * as React from 'react';
+
+const Loading: React.FC = () => (
+  <div className="co-m-loader co-an-fade-in-out" data-test="loading-indicator">
+    <div className="co-m-loader-dot__one" />
+    <div className="co-m-loader-dot__two" />
+    <div className="co-m-loader-dot__three" />
+  </div>
+);
+
+export default React.memo(Loading);


yaacov:
Patternfly suggest using "spinner" component
`import { Spinner } from '@patternfly/react-core';`

for example:
https://www.patternfly.org/v4/components/table/react-demos/#loading

we need to to go by patternfly, unless the console has a different rule
a. why did you decide to go with a different component then patternfly ?
b. why add it in the shared space, do you plan on other components use it instead of the patternfly spinner ?
c. if you want other components to use, you need to add it to the components repo.

avivtur:
a. because the `Console` does use this loading component [1]
b. yes
c. AFAIK the components repo is not reday for consumption, for now it would be best IMO to store components at the shared directory in our project so it can ease the development process while gathering all components that we want to export to the components repo to one location, and once the components repo will be ready for consumption, we can ship these components to the repo for external usage.

[1] https://github.com/openshift/console/blob/master/frontend/public/components/utils/status-box.tsx#L55

yaacov:
> AFAIK the components repo is not reday for consumption

In the mtg last week, I said it's open, and invited people to start using it (I also said that I didn't finish everything I wanted, but we can't wait so it's open as it is)

> for now it would be best IMO to store components at the shared directory in our project so it can ease the development process 

Using shared directory creates code conflicts, when you want to move fast, you use your tree branch, and try to minimize impact on other code users, e.g. not forcing them the use specific methods or components, and not coding on top on there code.

> while gathering all components that we want to export to the components repo to one location, and once the components repo will be ready for consumption, we can ship these components to the repo for external usage.

exactly, while we push forward fast each developer creates the components they need, and once we have some time to breath we can consolidate, refactor, review look for common patterns and components, take the best code and move it to a shared space, for components it's the components repo.

yaacov:
btw/
@glekner @avivtur it's a console component [1] you should move it to the sdk and convince the console guys to maintain it, if they don't want it, we may end up using an old design the the console stopped using and moved to use the new patternfly design.

[1] https://github.com/openshift/console/blob/master/frontend/public/components/utils/status-box.tsx#L55

PR #117, opened 2022-03-03
Storage Section Part I: Moving getters functions 

upalatucci:
## 📝 Description

> from views/catalog/templatescatalog/utils to /views/catalog/utils


yaacov:
@glekner the costomize-template, review-vm and the template-tile-view are three different components that should work independently, they should not share functionality.

please don't do optimization (e.g. consolidating logic method between components) before we have some information about the components and what is the best way to split/consolidate them.

for example: if you think that methods for the template-information panel can be shared, it may means it's a 4th component that can be used by the tile-view and the customization page.

glekner:
@yaacov they are three different **components** under `views/catalog` . they are not different **views**

we need a shared `utils` folder under `catalog`. its not going in the root folder of the project.

yaacov:
> they are not different views

they don't need to share any data or functionality, we share the vm object only as a workaround until we have better way, this component should not share more things, sharing functionality will make it harder to separate them when we have time to refactor and re-arrange the code.

PR #118, opened 2022-03-03
Initial Wizard overview tab

glekner:
## 📝 Description

Add Initial Wizard overview tab

## 🎥 Demo
overview


src/views/catalog/wizard/tabs/overview/components/WizardOverviewDisksTable/WizardOverviewDisksTable.tsx
```diff
@@ -0,0 +1,57 @@
...
+import {
+  DescriptionList,
+  DescriptionListDescription,
+  DescriptionListGroup,
+  DescriptionListTerm,
+  Stack,
+  StackItem,
+} from '@patternfly/react-core';
+
+import useDisksTableData from '../../../../../../virtualmachines/details/components/disk/hooks/useDisksTableData';
```
glekner, line 14: @yaacov as you can see I want to share logic with a hook that already does what I need because it tackled the problem before me, in this case getting a disk details.

This import is wrong, so we have two ways to tackle this:

1. Copy the code to my context, having to maintain duplicate code in multiple places (very bad IMO)
2. Create a shared utils/hooks folder for each resource (VM, VMI, Templates)

src/utils/resources/template/utils/flavor.ts
```diff
@@ -34,3 +34,30 @@ export const getFlavorData = (
...
+};
+
+export const getTemplateFlavorData = (
+  template: V1Template,
+): {
+  flavor: string;
+  cpuCount: number;
+  memory: string;
+} => {
+  const flavor = getTemplateFlavor(template);
```
yaacov, line 60: flavors are detracted, does the design show flavors, it look like a bug ?

src/views/catalog/templatescatalog/components/TemplatesCatalogTile.tsx
```diff
@@ -31,7 +31,7 @@ export const TemplateTile: React.FC<TemplateTileProps> = React.memo(({ template,
   const workload = getTemplateWorkload(template);
   const displayName = getTemplateName(template);
   const bootSourceType = getTemplateBootSourceType(template)?.type;
-  const { memory, cpuCount } = getFlavorData(template);
+  const { memory, cpuCount } = getTemplateFlavorData(template);
```
yaacov, line 34: looks like you can have more information about the CPU:
http://kubevirt.io/api-reference/main/definitions.html#_v1_cpu

removing flavor allows us to show more correct data here, e.g. cores and threads instead of the flavor code number
not sure showing `const cpu = core * thread * socket;` makes sense

@stu-gott hi, does it make sense for the UI to show "CPU: < core * thread * socket >" , should we move to something more verbose, like "Cores: <core> Threads: <thread>, Sockets: <socket>" , should we show only `Cores` and ignore threads and sockets ?

PR #125, opened 2022-03-07
Add shared resources logic structure

glekner:
## 📝 Description

Adding  shared resources logic
We need a place where we can store our shared resources logic (VM, VMI, Templates)
Logic stands for Utils (Selectors, Helpers) and Hooks. As of now we will not share components

There are multiple views that needs shared logic (e.g. Details page of VM and VM Wizard Review Overview Tab)
The shared logic will be under `src/utils/resources` and will be split into three directories - `vm, vmi, template` and a `shared.ts` file for generic methods like `getLabel, getAnnotation, getResourceURL`

each resource directory will have two directories: `utils, hooks`

Structure
Screen Shot 2022-03-07 at 13 03 26

Example of a resource folder (template):
Screen Shot 2022-03-07 at 13 03 37


glekner:
@tnisan @yaacov @metalice @pcbailey @avivtur 
please review and share your thoughts.

metalice:
I like this, it prevents us from duplicate unnecessary code

yaacov:
I like the idea of shared functionality, shared functionality helps create a unified/"official" way of doing something, it also help reduce code.
 
my concerns are that it's too soon to start looking for repeating patterns we can extract,
at this phase of development we may make bad common functionality that may not be the common functionality we actually need.

when we have more time, e.g. when the plugin will work, we will have time to extract common patterns, pick the best versions and document the best use case for each.

IMHO optimizing the code now is too early

PR #128, opened 2022-03-07
Add Storage CustomizeForm Part 1: Rethinking CustomizeForm

upalatucci:


src/views/catalog/customize/components/CustomizeForms/useCustomizeFormSubmit.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  const onSubmit = async (event) => {
+    event.preventDefault();
+    setTemplateLoaded(false);
+    try {
+      const formData = new FormData(event.currentTarget as HTMLFormElement);
+
+      const processedTemplate = await processTemplate(template, ns, formData);
+
+      const vm = getTemplateVirtualMachineObject(processedTemplate);
```
glekner, line 32: @upalatucci can you copy the `display-name` annotation from the template to the VM?
`annotations.openshift.io/display-name` before `updateVM`

src/views/catalog/customize/components/CustomizeForms/useCustomizeFormSubmit.ts
```diff
@@ -0,0 +1,48 @@
...
+      updateVM(vm);
```
glekner, line 33: we should `await` updateVM before `history.push`
Are you displaying the error that comes from `useWizardVMContext` somewhere?

metalice:
> @glekner @metalice @avivtur @hstastna @yaacov @pcbailey @vojtechszocs guys can you please review this pr? I followed some suggestions from the last pr and this is a rethinking. Sorry for so many prs on the same thing. Hope this will go smoother. It will follow another pr to actually add the storage section in a different Form component inside the `CustomizeForms` folder.

Looking much much better 👍 🙃🙂 I'll try to find time to go over it, but I see gilad already started. Just wanted to say that I like it much better now.

upalatucci:
Thank you @metalice this was what i need at this point :grin:

src/views/catalog/customize/components/CustomizeForms/useCustomizeFormSubmit.ts
upalatucci: Okay, thanks. No, I'll do a different pr to show the errors.

src/views/catalog/customize/components/CustomizeForms/useCustomizeFormSubmit.ts
upalatucci: Okay

src/views/catalog/customize/components/CustomizeForms/CustomizeForm.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const [requiredFields, optionalFields] = buildFields(template, [parameterForName], t);
+
+  return (
+    <Form onSubmit={onSubmit}>
+      {requiredFields?.map((field) => (
+        <FieldGroup key={field.name} field={field} showError={error} />
+      ))}
+
+      <ExpandableOptionsFields optionalFields={optionalFields} />
+
```
glekner, line 33: We can add an `Alert` here if `error` exists?
something like 
```tsx
        {error && (
          <StackItem>
            <Alert variant="danger" title={t('Create VirtualMachine error')} isInline>
              {error.message}
            </Alert>
          </StackItem>
        )}
```

glekner (commented):
reviewed just the hook again

src/views/catalog/customize/components/FormError.tsx
```diff
@@ -0,0 +1,24 @@
...
+    return (
+      <StackItem>
+        <Alert variant="danger" title={t('Create VirtualMachine error')} isInline>
+          {error.message}
+        </Alert>
+      </StackItem>
+    );
+  else {
+    return null;
+  }
```
metalice, line 23:
```suggestion

  if (!error?.message) return null;
  
    return (
      <StackItem>
        <Alert variant="danger" title={t('Create VirtualMachine error')} isInline>
          {error.message}
        </Alert>
      </StackItem>
    );
```

src/views/catalog/customize/components/CustomizeForms/useCustomizeFormSubmit.ts
```diff
@@ -0,0 +1,56 @@
...
+import { useWizardVMContext } from '../../../utils/WizardVMContext';
+import { processTemplate } from '../../utils';
+
+type useCustomizeFormSubmitType = [
+  onSubmit: (event: any) => Promise<void>,
+  loaded: boolean,
+  error: any,
+];
+
+export const useCustomizeFormSubmit = (template: V1Template): useCustomizeFormSubmitType => {
```
yaacov, line 18: why do you need a use for submit ?

it should be a synchronous operation

src/views/catalog/customize/components/CustomizeForms/useCustomizeFormSubmit.ts
```diff
@@ -0,0 +1,56 @@
...
+    event.preventDefault();
+    setTemplateLoaded(false);
+    try {
+      const formData = new FormData(event.currentTarget as HTMLFormElement);
+
+      const processedTemplate = await processTemplate(template, ns, formData);
+
+      const vm = getTemplateVirtualMachineObject(processedTemplate);
+
+      vm.metadata.annotations[ANNOTATIONS.displayName] = getAnnotation(
```
yaacov, line 36: that looks like a bug, the vm display name is not the template display name
a - the requirement of this Task dose not talk abou it
b - did you check with design about that ?

src/views/catalog/customize/components/CustomizeForms/useCustomizeFormSubmit.ts
upalatucci: just to have a clean code inside the component and why not maybe reuse some logic for the other forms.

src/views/catalog/customize/components/CustomizeForms/useCustomizeFormSubmit.ts
upalatucci: @glekner

yaacov: IMHO @glekner just copied this bug from our old code ...
it does not make sense to me that all VMs created from the same template will have the same display name, am I missing something ?

yaacov: btw/ the reason we need a new code instead of just copy the old one is all this little bugs that are hiding in helpers ans setters in the old code

src/views/catalog/customize/overrides.tsx
```diff
@@ -0,0 +1,46 @@
...
+      ...dataVolumeStorage?.resources,
+      requests: {
+        ...dataVolumeStorage?.resources?.requests,
+        storage: newStorage,
+      },
+    },
+  };
+};
+
+export const overrideVirtualMachineObject = (
```
yaacov, line 20:
```suggestion
export const overrideVirtualMachineNameAndNamespace = (
```

src/views/catalog/customize/overrides.tsx
```diff
@@ -0,0 +1,46 @@
...
+    ...virtualMachine,
+    metadata: {
+      ...virtualMachine.metadata,
+      name,
+      namespace,
+    },
+  };
+};
+
+export const overrideTemplate = (
```
yaacov, line 37: That looks strange, why do you need this ?

PR #130, opened 2022-03-08
update eslint and rules

glekner:
## 📝 Description

following Tuesday meeting, here are some eslint patches

1. Bump versions
2. add `"no-nested-ternary": "error",` rule cc @pcbailey 
3. @yaacov @metalice , adding `"@typescript-eslint/no-explicit-any"` is problematic, how do we deal with Errors being `any` type? [Example](https://github.com/kubevirt-ui/kubevirt-plugin/blob/main/src%2Futils%2Fresources%2Ftemplate%2Fhooks%2FuseVmTemplateSource%2FuseVmTemplateSource.ts#L17)


src/views/virtualmachines/details/components/disk/utils/VirtualMachineDisksTabUtils.ts
```diff
@@ -60,11 +60,10 @@ export const getDiskInterface = (disk: V1Disk): string => disk[getDiskDrive(disk
...
   const diskInterface = getDiskInterface(disk);
-  return diskInterface === 'virtio'
-    ? diskInterface
-    : diskInterface === 'scsi' || diskInterface === 'sata'
-    ? diskInterface.toUpperCase()
-    : '';
+  if (diskInterface === 'scsi' || diskInterface === 'sata') {
+    return diskInterface.toUpperCase();
+  }
+  return diskInterface ?? '';
```
avivtur, line 66: this change will conflict have with [1], I can insert this change to [1] instead to have this here, and it's also not really related to this PR :)

avivtur:
@glekner @yaacov @metalice @pcbailey @upalatucci @hstastna @vojtechszocs 
maybe we should try to enforce it when reviewing the PRs and try to avoid using `any` as a type, wdyt?

yaacov:
@glekner @yaacov @metalice @pcbailey @upalatucci @hstastna @vojtechszocs

now that we have a utils directory with no SDK component, we can re-add unit testing + jdoc rules to the utils directory, this may make the utils less dangerous to use.

glekner:
added jsdoc to existing functions/hooks under shared utils and a nested eslintrc file for enforcing jsdoc in that directory

PR #134, opened 2022-03-09
Adding storage form

upalatucci:
## 📝 Description

> Storage Expandable section for disk source customization

## 🎥 Demo
Screenshot from 2022-03-09 16-34-26

Screenshot from 2022-03-09 16-34-34

Screenshot from 2022-03-09 16-34-49


upalatucci:
@glekner @yaacov @metalice @avivtur @pcbailey @hstastna @vojtechszocs Can you please review this pr? :grin:

src/views/catalog/customize/components/PersistentVolumeClaimSelect/PersistentVolumeSelectProject.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  return (
+    <FormGroup
+      label={t('Persistent Volume Claim project')}
+      fieldId={fieldId}
+      id={fieldId}
+      isRequired
+      className="pvc-selection-formgroup"
+    >
+      <Select
```
glekner, line 47: add `maxHeight={400}` to make the inner select scrollable.
Add it to all `Select` instances actually

src/views/catalog/customize/components/CustomizeSource/CustomizeSource.tsx
```diff
@@ -0,0 +1,176 @@
...
+          label={t('Image URL')}
+          fieldId={`disk-source-required-${selectedSourceType}`}
+          isRequired
+          className="disk-source-form-group"
+        >
+          <TextInput
+            value={httpURL}
+            type="text"
+            onChange={setHTTPURL}
+            aria-label={t('Image URL')}
```
glekner, line 144: we need to have better validation here. at least `startsWith('https://...)` cc @yaacov

src/views/catalog/customize/components/CustomizeSource/quantity.ts
```diff
@@ -0,0 +1,63 @@
+import byteSize, { ByteSizeResult } from 'byte-size';
+
+const multipliers: Record<string, number> = {};
```
glekner, line 3: @avivtur didn't you implement this already in the disks shared files?
we need a share util for units, sizes

glekner (commented):
looks good @upalatucci.
we need to think about shared util for units, sizes

src/views/catalog/customize/components/PersistentVolumeClaimSelect/PersistentVolumeSelectProject.tsx
upalatucci: okay thanks

src/views/catalog/customize/components/CustomizeSource/CustomizeSource.tsx
upalatucci: Okay for me. Is there also for the container URL some simple validation that we can use?

yaacov: the validation is "nice to have", the requirement is to have a text input.

lets start with something working, and then after we have it working add the validation, in the next step

src/views/catalog/customize/components/CustomizeSource/quantity.ts
yaacov: we are currently in the phase of getting things done for each plugin extension,
it means that @avivtur may have his own version of the same functionlity,

once we have a working extension, we may go over it, choose the best implementation and move it to a shared directory

src/views/catalog/customize/components/CustomizeForms/CustomizeFormWithDisk.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  const [requiredFields, optionalFields] = buildFields(template, [parameterForName], t);
+
+  const onDiskSourceChange = React.useCallback((newDiskSource) => {
+    setCustomDiskSource(newDiskSource);
+  }, []);
+
+  return (
+    <Form onSubmit={onSubmit}>
+      {requiredFields?.map((field) => (
```
yaacov, line 39: name override inputs are above strorage section
required fields are below the storage section

---
order of inputs is:

name
storage (disk or cd)
add drivers disk check box
required parameters
optional parameters

src/views/catalog/customize/components/CustomizeForms/CustomizeFormWithDisk.tsx
```diff
@@ -0,0 +1,53 @@
...
+        <FieldGroup key={field.name} field={field} showError={error} />
+      ))}
+
+      <ExpandableCustomizeSourceSection
+        onChange={onDiskSourceChange}
```
yaacov, line 44: you can't have a different onChange for the different sections, 
it's always
a - pass parameters
b - overide name
c - overide storage
d - add drivers disk
in one pass
there is no meaning to doing on change without first processing the parameters

yaacov: to clearify
you should have one onChange method
this method should
a - cakk a method that takes a template name+namespace + parameters and return a processed templates
b - take the vm object
c - update vm name and namespace
d - update main disk
e - add installCD if needed
f - add driversCD if needed
return the vm json ( do with it what needed for example create the VM / store it in some config map as json string / store it in session storage ... )

src/views/catalog/customize/components/CustomizeForms/CustomizeFormWithDisk.tsx
upalatucci: really? So strange

src/views/catalog/customize/components/CustomizeForms/CustomizeFormWithDisk.tsx
upalatucci: okay, so the problem is that this process should not be in the onSubmit but in the onChange right?

src/views/catalog/customize/components/CustomizeForms/CustomizeFormWithDisk.tsx
upalatucci: fixed now

PR #138, opened 2022-03-13
Update templates catalog to latest design

glekner:
## 📝 Description

Update templates catalog to latest design
- Updated sidebar filters
- Added `Boot source available` checkbox
- Added `useAvailableSourceTemplates` hook that handles fetching templates with available boot source
- Simplified `useVmTemplatesFilters`

A Loading state is shown until at least one template with available source has been fetched.
once available templates are fetched they are cached, so the loading state won't show again.
## 🎥 Demo

https://user-images.githubusercontent.com/24938324/158055476-4ff4e375-4510-4101-928e-f85b93f92088.mov


glekner:
@yaacov  @avivtur please re-review

PR #155, opened 2022-03-16
decouple vm context and validation

glekner:
## 📝 Description

decouple vm context and validation to two files


upalatucci (approved):
/lgtm

PR #172, opened 2022-03-22
[WIP] Wizard Yaml TAB with Code Editor

upalatucci:
## 📝 Description

> Code Editor for yaml

## 🎥 Demo
Screenshot from 2022-03-22 12-03-30


src/views/catalog/wizard/tabs/yaml/components/YAMLEditor/index.ts
```diff
@@ -0,0 +1,3 @@
+import * as React from 'react';
+
+export default React.lazy(() => import('./YAMLEditor'));
```
avivtur, line 3: why not under src/utils/components? this could be used by VM,VMI,Templates as well

upalatucci: Yes absolutely right

PR #177, opened 2022-03-24
Consume description modal for VM details tab

avivtur:
## 📝 Description

> Add a brief description

## 🎥 Demo

### DIFF
--- a/src/utils/components/DescriptionModal/DescriptionModal.tsx
+++ b/src/utils/components/DescriptionModal/DescriptionModal.tsx
@@ -6,7 +6,7 @@ import TabModal from '../TabModal/TabModal';
 
 type DescriptionModalProps = {
   obj: K8sResourceCommon;
-  onSubmit: (description: string) => Promise<void>;
+  onSubmit: (description: string) => Promise<void | K8sResourceCommon>;
   isOpen: boolean;
   onClose: () => void;
 };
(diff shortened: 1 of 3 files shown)


src/utils/components/DescriptionModal/DescriptionModal.tsx
```diff
@@ -6,7 +6,7 @@ import TabModal from '../TabModal/TabModal';
 
 type DescriptionModalProps = {
   obj: K8sResourceCommon;
-  onSubmit: (description: string) => Promise<void>;
```
upalatucci, line 9: Here I think can remain only void as in this component you don't `await` the promise and don't use the promise result

src/views/virtualmachines/details/tabs/details/components/grid/leftGrid/VirtualMachineDetailsLeftGrid.tsx
```diff
@@ -75,6 +80,32 @@ const VirtualMachineDetailsLeftGrid: React.FC<VirtualMachineDetailsLeftGridProps
...
           descriptionData={getAnnotation(vm, DESCRIPTION_ANNOTATION) || None}
           descriptionHeader={t('Description')}
+          isEdit
+          onEditClick={() => setDescriptionModalOpen(true)}
+        />
+        <DescriptionModal
+          obj={vm}
+          isOpen={descriptionModalOpen}
+          onClose={() => setDescriptionModalOpen(false)}
+          onSubmit={(updatedDescription) => {
```
upalatucci, line 90: In my opinion, would be more readable if this function is extracted and moved above.

upalatucci (commented):
Great :+1:

src/utils/components/DescriptionModal/DescriptionModal.tsx
avivtur: the k8sUpdate function I'm using as `onSubmit`, requires this change since it returns a new VM object, and without it, I get an error from TS about wrong type

PR #120, opened 2022-03-03
Adding console tab VMI page - VNC console

metalice:
Signed-off-by: Matan Schatzman <mschatzm@redhat.com>

## 📝 Description

> Add a brief description

## 🎥 Demo
image

PR #179, opened 2022-03-24
Adding serial console

metalice:
## 📝 Description

Serial console + actions + console selector dropdown

Based on https://github.com/kubevirt-ui/kubevirt-plugin/pull/120 - should be merged first

## 🎥 Demo


src/views/virtualmachinesinstance/details/tabs/console/components/AccessConsoles/AccessConsoles.tsx
```diff
@@ -0,0 +1,126 @@
...
+      } else {
+        const typeText = getChildTypeName(child);
+        acc.push(
+          <SelectOption
+            key={typeText}
+            id={typeText}
+            value={{ value: typeText, toString: () => typeText } as SelectOptionObject}
+          />,
+        );
+      }
```
upalatucci, line 92: Can this be done like this?
```js
        const typeText = getChildTypeName(child);
        acc.push(
          <SelectOption
            key={typeText}
            id={typeText}
            value={{ value: typeText, toString: () => typeMap[typeText] || typeText } as SelectOptionObject}
          />,
        );
```

upalatucci (commented):
Don't know if this can help :-) Little suggestion

src/views/virtualmachinesinstance/details/tabs/console/components/AccessConsoles/AccessConsoles.tsx
metalice: great suggestion

yaacov: why do we maintain a copy of patternflys component?
https://github.com/patternfly/patternfly-react/blob/main/packages/react-console/src/components/AccessConsoles/AccessConsoles.tsx

if we need to add new features should we add them in patternfly?
if we actually want to maintain our own version, we should do it in a more public place, maybe notify the patently guys we are doing it?

metalice: If we will need a new feature we can simply add it to our own code base, pattternfly/console codebase isn't up-to-date with the rest of the console code and causes conflicts. 
If we will want the far future to do it in a public place we can think over the approach.

PR #181, opened 2022-03-28
Add vm create dropdown

glekner:
## 📝 Description

Add vm create dropdown

## 🎥 Demo
Screen Shot 2022-03-28 at 9 02 35


glekner:
@avivtur please review

PR #184, opened 2022-03-28
fix cloudinit disk creation if missing

glekner:
## 📝 Description

if cloudinitdisk is deleted or doesn't exist, we need to re-create it when editing in the Scripts tab


upalatucci (approved):
/lgtm

PR #187, opened 2022-03-29
add test for vm catalog filter

gouyang:
test vm catalog filter


gouyang:
Looks templates are different between upstream CI and downstream, will add this tests in downstream test only.

PR #193, opened 2022-03-31
update wizard modals with useModal

glekner:
## 📝 Description

update wizard modals with useModal
Also added new Add disk modal to Disks tab


upalatucci (approved):
/lgtm

PR #194, opened 2022-03-31
Add confirm prompts to wizard exit

glekner:
## 📝 Description

Add confirm prompts to wizard exit
Add template's name and namespace to context so the user can go back to the customize form.
Add template's display name for the overview tab OS section


upalatucci (approved):
/lgtm

PR #198, opened 2022-04-01
Wizard - Edit Disk modal

upalatucci:
## 📝 Description

> Add a brief description

## 🎥 Demo


src/views/catalog/wizard/tabs/disks/hooks/useEditDiskState.ts
```diff
@@ -0,0 +1,70 @@
...
+        volumeTypes.DATA_VOLUME,
+        volumeTypes.PERSISTENT_VOLUME_CLAIM,
+      ].includes(key),
+    );
+
+    if (volumeSource === sourceTypes.EPHEMERAL) {
+      initialDiskSourceState.ephemeralSource = volume.containerDisk?.image;
+      return { diskSource: sourceTypes.EPHEMERAL, diskSize: DYNAMIC };
+    }
+    return { diskSource: OTHER, diskSize: null };
```
glekner, line 50: why is diskSize null here?
we need different logic here, we're in the Wizard.
The VM is not yet created, we can edit disks, volumes with much more freedom

src/views/catalog/wizard/tabs/disks/components/DiskRowActions.tsx
```diff
@@ -23,17 +26,27 @@ const DiskRowActions: React.FC<DiskRowActionsProps> = ({ diskName }) => {
   const { t } = useKubevirtTranslation();
   const [isDropdownOpen, setIsDropdownOpen] = React.useState(false);
   const [isDeleteModalOpen, setIsDeleteModalOpen] = React.useState(false);
-  const label = t('Delete {{diskName}} disk', { diskName: diskName });
-  const submitBtnText = t('Delete');
+  const [isEditModalOpen, setIsEditModalOpen] = React.useState(false);
```
glekner, line 29: we can switch these `isOpen` states with `useModal`

PR #199, opened 2022-04-01
Add Actions column to VM Templates list

hstastna:
## 📝 Description
Add _Actions_ column to VM Templates list with the Kebab menu, containing the following options for each of the Templates:
- _Edit Template_
  - leads to the VM Template Details page, where the user will be able to edit the specific properties if possible
- _Clone Template_
  -  _Clone Template_ modal is opened
- _Edit boot source_
  - _Edit Boot source to template_ modal is opened (followup PR)
  - disabled for ~~Red Hat~~ common templates
- _Delete Template_
  - _Delete VM template_ confirmation modal is opened
  - disabled for ~~Red Hat~~ common templates

Also add the dash '-' the Boot source column if the Boot source is not available. 

## 🎥 Demo
Kebab menu with the available actions for a common template:
actions1
Kebab menu for other than a common template:
actions2
When the Boot source is available:
actions3
Clone Template modal:
clonee
Delete template confirmation modal:
delete_modal

---

TODO:
- [x] add Actions column with Kebab menu to the list
- [x] disable Edit boot source and Delete Template options for ~~Red Hat~~ common templates
- [x] redirect to the VM Template details nav page, after clicking on Edit Template action
- [x] add Clone Template modal
- [ ] ~~add Edit boot source to template modal~~
- [x] add Delete template confirmation modal
- [x] check the label `template.kubevirt.io/type` and make all the common templates not ~~editable~~ possible to be deleted


src/views/templates/actions/hooks/useVirtualMachineTemplatesActions.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+import { getTemplateProviderName } from '../../utils/selectors';
+
+type useVirtualMachineTemplatesActionsProps = (template: V1Template) => Action[];
+
+const useVirtualMachineTemplatesActions: useVirtualMachineTemplatesActionsProps = (
+  template: V1Template,
+) => {
+  const { t } = useKubevirtTranslation();
+  const isRedHatTemplate = getTemplateProviderName(template) === 'Red Hat';
```
avivtur, line 17: we need to disable the delete option for all common templates,
this check will not catch all common templates since in some cases the annotation `openshift.io/provider-display-name` can hold a value `KubeVirt` and the delete will be enabled.
I suggest to change this check to something a bit more accurate,
maybe checking if the label `app.kubernetes.io/name: common-templates` exists, @yaacov wdyt?

yaacov: > disable the delete option for all common templates

We should not create/edit/delete templates that are controlled and deleted by SSP operator,
so when disabling actions on this template we need to test for templates controlled by SSP operator.
The label to use is `template.kubevirt.io/type: base` [1], and the name we call this templates is "common templates"
( maybe we should rename them "base templates" instead ? )

[1]https://github.com/kubevirt/ssp-operator/blob/master/internal/operands/common-templates/constants.go#L6
    https://github.com/kubevirt/ssp-operator/blob/master/internal/operands/common-templates/constants.go#L5

common template is defined by:
`const isCommonTemplate = metadate.labels['template.kubevirt.io/type'] === "base"`
so if we want to know if it's common template this is the test

provided by RedHat is defined by:
`const isProvidedByRedHat = metadata.annotations['template.kubevirt.io/provider'] === 'Red Hat'`
but the provider is not checked by SSP operator, and it is not indicative for actions being available or not on the back end.

hstastna:
_Note:_
Implementation of "Edit template" and "Edit boot source" actions for non common templates will be added in followup PRs.

src/views/templates/actions/VirtualMachineTemplatesActions.tsx
```diff
@@ -0,0 +1,58 @@
...
+    if (typeof action?.cta === 'function') {
+      action?.cta();
+      setIsOpen(false);
+    }
+  };
+
+  const disabledActionInfo = (actionId: string): string => {
+    switch (actionId) {
+      case 'edit-boot-source':
+        return t('Boot source cannot be edited for Red Hat templates');
```
yaacov, line 29: edit boot source is not disabled on redhat templates, it is disabled by the automatic upload mechanism

src/views/templates/actions/VirtualMachineTemplatesActions.tsx
```diff
@@ -0,0 +1,58 @@
...
+      case 'delete-template':
+        return t('Red Hat templates cannot be deleted');
```
yaacov, line 31: Red Hat provided templates can be deleted, the templates you can't delete are the common templates (owned by SSP operator)

src/views/templates/actions/components/CloneTemplateModal.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const { t } = useKubevirtTranslation();
+  const [templateName, setTemplateName] = React.useState('');
+  const [templateProvider, setTemplateProvider] = React.useState('');
+
+  const onSubmit = async () => {
+    await k8sCreate<V1Template>({
+      model: TemplateModel,
+      data: {
+        apiVersion: obj?.apiVersion,
+        kind: TemplateModel.kind,
```
yaacov, line 25: - no need to change the `kind`, obj is a template, it should have the currect kind
- if you only want to change the provider, you should take `obj` and change only the provider
- you also need to change `template.kubevirt.io/type` label from `base` to `vm` because after clonning it's not a common template anymore
- do you need to change the `namespace` ?

to change only one element in an obj we have `immer` you can look how it's used in the wizard
another option is to use some `rfc6902` package that allow to patch an object, but I don't think we have one @glekner @upalatucci ?

src/views/templates/actions/VirtualMachineTemplatesActions.tsx
```diff
@@ -0,0 +1,58 @@
...
+      position={DropdownPosition.right}
+      toggle={<KebabToggle onToggle={setIsOpen} />}
+      dropdownItems={actions?.map((action) => (
+        <DropdownItem
+          key={action?.id}
+          onClick={() => handleClick(action)}
+          isDisabled={action?.disabled}
+        >
+          {action?.label}
+          {action?.disabled && <p>{disabledActionInfo(action?.id)}</p>}
```
yaacov, line 51: - action label does not change, the help text that help user know when this action is available and when not, is always displayed
- for the help under the action you can should use a description ( https://www.patternfly.org/v4/components/dropdown#dropdown-with-descriptions ) not primary action text

src/views/templates/actions/hooks/useVirtualMachineTemplatesActions.tsx
```diff
@@ -0,0 +1,72 @@
...
+) => {
+  const { t } = useKubevirtTranslation();
+  const isCommonTemplate = template.metadata.labels['template.kubevirt.io/type'] === 'base';
+  const { createModal } = useModal();
+
+  const actions = [
+    {
+      id: 'edit-template',
+      label: t('Edit Template'),
+      cta: () => console.log('Edit Template'), // TODO lead to the template details page
```
yaacov, line 24: the template details link is available now, you can fill it in in this PR

src/views/templates/actions/hooks/useVirtualMachineTemplatesActions.tsx
```diff
@@ -0,0 +1,72 @@
...
+          <DeleteModal
+            obj={template}
+            isOpen={isOpen}
+            onClose={onClose}
+            headerText={t('Delete Virtual Machine Template?')}
+            onDeleteSubmit={() =>
+              k8sDelete({
+                model: TemplateModel,
+                resource: template,
+                json: undefined,
```
yaacov, line 60: this is very strange, why do you need `json` and `requestInit` to delete the template here ?

src/views/templates/actions/VirtualMachineTemplatesActions.tsx
hstastna: I followed the design I was given. If we need to change this, we need the design update, too.

src/views/templates/actions/VirtualMachineTemplatesActions.tsx
hstastna: I followed the design I was given. If we need to change this, we need the design update, too.

src/views/templates/actions/VirtualMachineTemplatesActions.tsx
yaacov: > If we need to change this, we need the design update, too.
 
great, we will.

src/views/templates/actions/VirtualMachineTemplatesActions.tsx
yaacov: makes sense to me to

src/views/templates/actions/VirtualMachineTemplatesActions.tsx
```diff
@@ -0,0 +1,59 @@
...
+      isPlain
+      isOpen={isOpen}
+      position={DropdownPosition.right}
+      toggle={<KebabToggle onToggle={setIsOpen} />}
+      dropdownItems={actions?.map((action) => (
+        <DropdownItem
+          key={action?.id}
+          onClick={() => handleClick(action)}
+          isDisabled={action?.disabled}
+          description={disabledActionInfo(action?.disabled, action?.id)}
```
yaacov, line 50: description should not change, we can fix it in a followup.
```suggestion
          description={t(action?.description)}
```

src/views/templates/actions/components/CloneTemplateModal.tsx
```diff
@@ -0,0 +1,72 @@
...
+            ...obj?.metadata?.annotations,
+            'template.kubevirt.io/provider': templateProvider,
+          },
+          labels: { ...obj?.metadata?.labels, 'template.kubevirt.io/type': 'vm' },
+          name: templateName,
+          namespace: obj?.metadata?.namespace,
+        },
+      },
+    });
+  };
```
yaacov, line 36:
```suggestion
const cloneTemplate = rfc6902.applyPatch(obj, [
  {op: 'replace', path: '/metadata/labels/template.kubevirt.io/type', value: 'vm'},
  {
      op: 'replace',
      path: '/metadata/annotations/template.kubevirt.io/provider',
      value: templateProvider
  },
  ]);

  const onSubmit = async () => {
    await k8sCreate<V1Template>({
      model: TemplateModel,
      data: cloneTemplate,
    });
  };
```

hstastna: I am not sure if we really need to add a new package to our project just because of this...

yaacov: @hstastna hi, you don't need to add a new package, see:
https://github.com/kubevirt-ui/kubevirt-plugin/pull/199#discussion_r850073925

> to change only one element in an obj we have immer you can look how it's used in the wizard
> another option is to use some rfc6902 package that allow to patch an object, but I don't think we have one

in the code snippet here you use the spread operator to mutate the template object. using the spread operator is a good option to do light mutations, but it becomes less appealing when the object being mutated is deep, or you have more then a few changes you want to do in the object.

In https://github.com/kubevirt-ui/kubevirt-plugin/pull/199#discussion_r850073925 I suggested `immer` or some rfc6902 complaint library not necessary `rfc6902`, any library that conform to the rfc6902 standard will do.

spread operator:
pros:
light and built into the language, no need to import anything, easy to read and maintain on shallow object. 
cons:
we have to carefully shallow copy every level of the state structure, make is hard to understand and maintain on deep objects.

Immer:
pros:
works nice with typescript and react
cons:
require external library

using rfc6902 standard:
pros:
using a standard, works well with k8s and openshift that use this standard
cons:
require external library

if you wish to use a helper library instead of using the spread operator as suggested in https://github.com/kubevirt-ui/kubevirt-plugin/pull/199#discussion_r850073925, you can try other options like `immer` or implement an rfc6902 standard complient helper or try some other solution, you do not need to add a new package.

hstastna: Thank you very much, I will reconsider pros and cons you've mentioned above and will update the PR.

PR #201, opened 2022-04-01
Add filtering by OS and Workload profile to VM Templates list

hstastna:
## 📝 Description
Add filtering by OS and _Workload profile_ column to VM templates list.

Rename the existing _Boot source availability_ column name to shorter _Boot source_, according to the design updates.

## 🎥 Demo
Filtering by OS, _Boot source availability_ renamed to _Boot source_:
filter_os
Make new _Workload profile_ column sortable: first _Desktop_, then _High performance_, then _Server_ (when ascending order)
workload_sort


hstastna:

src/views/templates/list/hooks/useVirtualMachineTemplatesColumns.ts
```diff
@@ -19,7 +19,13 @@ const useVirtualMachineTemplatesColumns = (): TableColumn<K8sResourceCommon>[] =
       sort: 'metadata.namespace',
     },
     {
-      title: t('Boot source availability'),
+      title: t('Workload profile'),
+      id: 'workload',
+      transforms: [sortable],
+      sort: 'objects[0].spec.template.metadata.annotations.["vm.kubevirt.io/workload"]',
```
yaacov, line 25: hmm, here you assume only one object and that it is the vm, but elsewhere we use a method to get the vm object, ( can we can assume one object and that it is the vm in other places too ? )

src/views/templates/list/hooks/useVirtualMachineTemplatesFilters.ts
```diff
@@ -24,39 +24,71 @@ const useTemplateProviders = (): TemplateProviderProps[] => {
...
-  const compareString = getTemplateProviderNameWithOther(template, providers);
+  const compareString = getItemNameWithOther(itemName, items);
 
   return compareData.selected?.length === 0 || compareData.selected?.includes(compareString);
 };
 
 const useVirtualMachineTemplatesFilters = (): RowFilter[] => {
   const { t } = useKubevirtTranslation();
   const providers = useTemplateProviders();
+  const osNames = useOsNames();
```
yaacov, line 73: this actually return a constant ( + the others translated string)
do you actually need a use method here ?
```suggestion
const OS_NAMES = [
    {
      id: OS_NAME_TYPES.rhel,
      title: OS_NAME_LABELS.rhel,
    },
    {
      id: OS_NAME_TYPES.fedora,
      title: OS_NAME_LABELS.fedora,
    },
    {
      id: OS_NAME_TYPES.centos,
      title: OS_NAME_LABELS.centos,
    },
    {
      id: OS_NAME_TYPES.windows,
      title: OS_NAME_LABELS.windows,
    },
    {
      id: OS_NAME_TYPES.other,
      title: t('Other'),
    },
  ];
  
  const osNames = [
      ...OS_NAMES,
      {
          id: OS_NAME_TYPES.other,
          title: t('Other'),
      },
];
```

src/views/templates/list/hooks/useWorkloadProfile.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  return {
+    desktop: t('Desktop'),
+    server: t('Server'),
+    highperformance: t('High performance'),
+  };
+};
+
+const useWorkloadProfile = (template: V1Template): string => {
+  return useWorkloadLabels()?.[getTemplateWorkload(template)];
```
yaacov, line 22: in the list view you sort by a simple `object[0]...` to get the workload, here you use a method, this is an opening for bugs in cases the `getTemplateWorkload`  will get a different workload.

src/views/templates/list/hooks/useVirtualMachineTemplatesColumns.ts
hstastna: I am not sure what we can assume in other places. In which places... ? But if you wonder about the line 25 for sorting the workload profile, I haven't found any better way to define which info we want to sort by. If you have some another/better working solution, I am happy to see/try it. Thank you! :)

src/views/templates/list/hooks/useVirtualMachineTemplatesFilters.ts
```diff
@@ -1,15 +1,16 @@
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
-import { V1Template } from '@kubevirt-utils/models';
+import { getTemplateOS, OS_NAME_TYPES } from '@kubevirt-utils/resources/template';
 import { FilterValue, RowFilter } from '@openshift-console/dynamic-plugin-sdk';
 
+import { OS_NAMES } from '../../utils/constants';
```
yaacov, line 5: `@kubevirt-utils` ?

PR #218, opened 2022-04-05
Approximated disk quantity in list

upalatucci:
## 🎥 Demo
Before
Screenshot from 2022-04-05 10-24-34

After
Screenshot from 2022-04-05 12-09-32


avivtur:
why is the size rounded to 33Gi instead of ~33.59Gi like it would have shown on the disks list after VM creation?

upalatucci:
right I'll add a precision parameter to the function

avivtur (approved):
looks good nice work :)

PR #222, opened 2022-04-06
add test script for downstream CI

gouyang:
## 📝 Description

add a test script for downstream tests


test-cypress.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
```
glekner, line 1: @gouyang I dont see this file `test-cypress.sh` used anywhere?

gouyang: it will be used by downstream CI which has the kubevirt-plugin deployed, `test-prow-e2e` is not suitable for downstream as it deploys the plugin.

it also can be used for triggering tests locally.

PR #223, opened 2022-04-06
remove vm error from disks tab table

glekner:
## 📝 Description

The Disks table will show an error even if it's not related to the Disks tab.
The `loadError` prop is meant for resource fetch errors, the wizard's vm is not fetched, so we should remove it from the Disks tab table.


glekner:
@upalatucci please review

PR #228, opened 2022-04-07
Add console.flag extension for kubevirt dynamic plugin

glekner:
## 📝 Description

Add console.flag extension for 4.11 Cluster.

If plugin is installed, we will add: `"disallowed": ["KUBEVIRT_DYNAMIC"]`
to the static extensions that we are upgrading with the dynamic ones that we currently have


glekner:
@yaacov please review

locales/en/plugin__kubevirt-plugin.json
```diff
@@ -439,5 +439,8 @@
   "MAC Address": "MAC Address",
   "masquerade": "masquerade",
   "VMI": "VMI",
-  "Virtual Machines Instances": "Virtual Machines Instances"
+  "Virtual Machines Instances": "Virtual Machines Instances",
```
glekner, line 442: turns out we are missing some extractions in `main`.
hijacking this PR to fix

yaacov: should be `VirtualMachinesInstances` it's a CRD name.

PR #256, opened 2022-04-12
Add helper text to Customization form

upalatucci:
## 📝 Description

> Use the annotations `template.kubevirt.io/images` and `template.kubevirt.io/containerdisks` to provide helper text in the form storage section

## 🎥 Demo

Screenshot from 2022-04-12 17-10-29

### DIFF
--- a/src/utils/resources/template/utils/selectors.ts
+++ b/src/utils/resources/template/utils/selectors.ts
@@ -53,6 +53,22 @@ export const getTemplateSupportLevel = (template: V1Template): string =>
   getAnnotation(template, ANNOTATIONS.supportLevel);
 
 /**
+ * A selector that returns the containerDisks of a given template
+ * @param {V1Template} template - template
+ * @returns {string[]} array of containerDisks
+ */
+export const getTemplateContainerDisks = (template: V1Template): string[] | undefined =>
+  getAnnotation(template, ANNOTATIONS.containerDisks)?.split('\n');
+
+/**
+ * A selector that returns the import URLs of a given template
+ * @param {V1Template} template - template
+ * @returns {string[]} array of import URLS
+ */
+export const getTemplateImportURLs = (template: V1Template): string[] | undefined =>
+  getAnnotation(template, ANNOTATIONS.importURLs)?.split('\n');
+
+/**
  * A selector that returns the flavor of a given template
  * @param {V1Template} template - template
  */
--- a/src/views/catalog/customize/components/CustomizeSource/utils.ts
+++ b/src/views/catalog/customize/components/CustomizeSource/utils.ts
@@ -1,4 +1,12 @@
+import { TFunction } from 'react-i18next';
+
+import { V1Template } from '@kubevirt-ui/kubevirt-api/console';
 import { V1beta1DataVolumeSpec } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import {
+  getTemplateContainerDisks,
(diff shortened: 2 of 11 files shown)


src/views/catalog/customize/components/CustomizeSource/utils.ts
```diff
@@ -52,3 +60,27 @@ export const getPVCSource = (
...
+  const dataVolumeTemplates = getTemplateVirtualMachineObject(template)?.spec?.dataVolumeTemplates;
+
+  return dataVolumeTemplates?.[0]?.spec?.storage?.resources?.requests?.storage;
+};
+
+export const getHTTPHelperText = (template: V1Template, t: TFunction) => {
+  const importUrls = getTemplateImportURLs(template);
+
+  if (importUrls)
+    return t('Enter URL to download (for example: {{importUrls}}', {
```
yaacov, line 74: it's a list (same for registry), you should pick only one example out of the list ( first one? )

src/utils/resources/template/utils/selectors.ts
```diff
@@ -52,6 +52,20 @@ export const getTemplateProviderName = (template: V1Template): string =>
 export const getTemplateSupportLevel = (template: V1Template): string =>
   getAnnotation(template, ANNOTATIONS.supportLevel);
 
+/**
+ * A selector that returns the containerDisks of a given template
+ * @param {V1Template} template - template
+ */
```
yaacov, line 58: missing return value and type

src/views/catalog/customize/components/CustomizeSource/utils.ts
upalatucci: Curious... The metadata attribute of template has the type `ObjectMetadata`.
```js
export declare type ObjectMetadata = {
    annotations?: {
        [key: string]: string;
    };
    ...
};
```
So every annotation has type `string` and not `string[]`. Do we have to change this type? Or maybe the annotation is a string formatted in a particular way?

src/utils/resources/template/utils/selectors.ts
upalatucci: okay thanks

src/views/catalog/customize/components/CustomizeSource/utils.ts
yaacov: intersting, we will need to check that, did you find a template with this annotations ?

src/utils/resources/template/utils/selectors.ts
```diff
@@ -52,6 +52,22 @@ export const getTemplateProviderName = (template: V1Template): string =>
...
+  getAnnotation(template, ANNOTATIONS.containerDisks);
+
+/**
+ * A selector that returns the import URLs of a given template
+ * @param {V1Template} template - template
+ * @returns array of import URLS
+ */
+export const getTemplateImportURLs = (template: V1Template): string | undefined =>
+  getAnnotation(template, ANNOTATIONS.importURLs);
+
```
yaacov, line 70: yes, it's a list encoded as a string with EOL seperators

```suggestion
 * @returns {[]string} array of import URLS
 */
export const getTemplateImportURLs = (template: V1Template): []string =>
  template.metadata?.annotations['template.kubevirt.io/images']?.split('\n') || [];

```

upalatucci: Okay thanks :-)

PR #257, opened 2022-04-12
create vm from catalog

gouyang:
add test to create vm from catalog


gouyang:
@yaacov, The test cannot pass because the VM could not turn into running, the cdi does not import vm disk at all in prow.

yaacov:
/retitle [e2e] test create vm from catalog

yaacov:
> The test cannot pass because the VM could not turn into running, the cdi does not import vm disk at all in prow.

If this tests fail on prow, can we move them to Teir2 and run them on QE env only ?

PR #266, opened 2022-04-13
Fix handling of undefined template annotations

glekner:
## 📝 Description

- Documentation link will now show `N/A` if link is not available
- Catalog Template will show `Other` if workload is not available
- Wizard Overview Tab - Operating system will show `Not available` if os is not available instead of template's name


upalatucci (approved):
/lgtm

PR #275, opened 2022-04-14
Template page Header

upalatucci:
## 🎥 Demo
Screenshot from 2022-04-14 14-20-14


src/views/templates/details/hooks/useTemplateActions.tsx
```diff
@@ -0,0 +1,34 @@
...
+        label: t('Create Virtual Machine'),
+        cta: () => console.log(template),
+      },
+      {
+        id: 'template-action-create-new-template',
+        label: t('Create new Template'),
+        cta: () => console.log(template),
+      },
+      {
+        id: 'Customize boot source',
```
yaacov, line 23: we dropped this action in 4.11

src/views/templates/details/hooks/useTemplateActions.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
+import { Action } from '@openshift-console/dynamic-plugin-sdk';
+
+export const useTemplateActions = (template: V1Template): Action[] => {
+  const { t } = useKubevirtTranslation();
+
+  const actions: Action[] = React.useMemo(
+    () => [
+      {
+        id: 'template-action-create-vm',
```
yaacov, line 13: we dropeed this action in 4.11

src/views/templates/details/hooks/useTemplateActions.tsx
```diff
@@ -0,0 +1,34 @@
...
+        label: t('Create Virtual Machine'),
+        cta: () => console.log(template),
+      },
+      {
+        id: 'template-action-create-new-template',
```
yaacov, line 18: we now call it "clone template"
see: https://github.com/kubevirt-ui/kubevirt-plugin/pull/199/files#diff-f2aabfe993673642dbeb30394a3d068805d7be2d290c372f68067273da386577R27

src/views/templates/details/hooks/useTemplateActions.tsx
hstastna: Maybe we need here the same actions as in the list view? https://user-images.githubusercontent.com/13417815/161285378-aae38ca9-c861-46cf-a14e-355e2035fe2d.png, PR https://github.com/kubevirt-ui/kubevirt-plugin/pull/199

yaacov: yes, we should have "edit boot source" in the actions, we should have an extension that add actions per CRD (in the case "Template") and get the SDK to render the actions when needed, so we will have the same actions in the list, details page, topology, and any other place it's needed, using drop down to render actions by our self is a bad idea.

note I: "edit template" does not make sense here, we are in the edit template page, maybe we should drop it from the list actions too
note II: "customize boot source" is a different action from "edit boot source" and is is dropped in 4.11

hstastna: > note I: "edit template" does not make sense here, we are in the edit template page, maybe we should drop it from the list actions too

Regarding "edit template" in the list view, I was informed that this action redirects to the VM template details page/Details tab, where the user will be able to edit the specific options that they can/is allowed to edit. Yes, we will reach this place also by clicking on the template's name in the list view, so maybe we don't need "edit template" in the Actions drop down. But maybe it is nice to have more options how to get to the template's details page, maybe it is nice to have all the possible actions in the Actions drop down.. Depends on how the users use the UI. Some of them may like to click on the template's name, some of them may look into the Actions first. We should ask the UX before dropping this from Actions drop down.

yfrimanm:
When a user is in the Templates Details page there is no need show the Edit Template action from the actions because it will link to the Template details page anyhow.

src/views/templates/details/hooks/useTemplateActions.tsx
yaacov: > Maybe we need here the same actions as in the list view?

Makes sense to me too, when using the SDK we only define the actions and the SDK should render the actions in all the places needed, for example the list view, the details page, the topology page, generic resource pages and any other places an action drop down is required, the plugin should not render the actions, only define them, and when we define the actions we need to take into account that they may be rendered in different use cases.

> Regarding "edit template" in the list view

If we use the same actions in the list view and the details page (also topology and generic resource page) we need to choose the actions carefully to fit all scenarios, while "edit template" makes sense in the list view, it does not make sense in the details page. If we want to use the same actions in both the list view and the details page we need to consider that "edit template" does not make sense in the details page, and consider the pros and cons of having it in the list page taking into account that if we leave it in the list view we can't have the same actions list in the details page.

src/views/templates/details/TemplateActions.tsx
```diff
@@ -0,0 +1,48 @@
...
+  template: V1Template;
+};
+
+const TemplateActions: React.FC<TemplateActionsProps> = ({ template }) => {
+  const { t } = useKubevirtTranslation();
+  const [isDropDownOpen, setDropdownOpen] = React.useState(false);
+
+  const actions = useVirtualMachineTemplatesActions(template).filter(
+    (action) => action.id !== 'edit-template',
+  );
```
yaacov, line 20: :-(

PR #287, opened 2022-04-18
fix boot source selector

glekner:
## 📝 Description
fix boot source selector for quick create and wizard

### DIFF
--- a/src/utils/resources/vm/utils/selectors.ts
+++ b/src/utils/resources/vm/utils/selectors.ts
@@ -1,4 +1,4 @@
-import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import { V1Disk, V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
 
 /**
  * A selector for the virtual machine's networks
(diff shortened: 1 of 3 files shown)


src/utils/resources/vm/utils/selectors.ts
```diff
@@ -104,3 +104,17 @@ export const getTolerations = (vm: V1VirtualMachine) => vm?.spec?.template?.spec
...
+ * If disks with bootOrder are available, returns the disk with lowest bootOrder integer value.
+ * If not, returns the first disk
+ * @param {V1VirtualMachine} vm the virtual machine
+ * @returns the virtual machine boot disk
+ */
+export const getBootDisk = (vm: V1VirtualMachine): V1Disk =>
+  (getDisks(vm) || [])
+    .filter((d) => d.bootOrder)
+    .reduce((acc, disk) => {
+      return acc.bootOrder < disk.bootOrder ? acc : disk;
```
yaacov, line 119:
```suggestion
      return acc?.bootOrder < disk.bootOrder ? acc : disk;
```

`acc` can be undefined if `getDisks(vm)` is empty

PR #307, opened 2022-04-20
Common Template No Editable

upalatucci:
## 🎥 Demo

Screenshot from 2022-04-20 14-41-34
Screenshot from 2022-04-20 14-41-25


yaacov:
@upalatucci hi, the design is to:
a - have a top warning that common templates are not aditable
b - disable all edit options, [ except "edit boot source" that actually edit a dataVolume and not the template ]

PR #313, opened 2022-04-22
VM SSH Access and User Credentials

upalatucci:
## 🎥 Demo
Screenshot from 2022-04-22 17-25-10
Screenshot from 2022-04-22 17-25-02


glekner:
@metalice please review

src/utils/components/SSHAccess/SSHAccessModal.tsx
```diff
@@ -0,0 +1,133 @@
...
+  isOpen: boolean;
+  onClose: () => void;
+  sshService: IoK8sApiCoreV1Service;
+};
+
+export const NODE_PORTS_LINK =
+  'https://access.redhat.com/documentation/en-us/openshift_container_platform/4.9/html/networking/configuring-ingress-cluster-traffic#nw-using-nodeport_configuring-ingress-cluster-traffic-nodeport';
+
+export const PORT = 22000;
+export const TARGET_PORT = 22;
```
metalice, line 29: can be in a constants file

src/utils/components/SSHAccess/SSHAccessModal.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+const getSSHServiceFromVM = (vm: V1VirtualMachine): IoK8sApiCoreV1Service => ({
```
metalice, line 31: can be in a util file

src/utils/components/SSHAccess/SSHAccessModal.tsx
```diff
@@ -0,0 +1,133 @@
...
+          resource: serviceResource,
+          json: undefined,
+          requestInit: undefined,
+        });
+      }
+    } catch (e) {
+      // eslint-disable-next-line no-console
+      console.log(`Failed to ${isEnabled ? 'create' : 'delete'} SSH service:`, e.message);
+    }
+  };
```
metalice, line 84: im guessing that in a scenrio when a user just open modal and click onsubmit without changing anything there will be still unneccessry calls, I think we should harden the conditions for the call submits, such as isEnabled && !!sshService , and same goes for delete, !isEnabled && !sshService

src/utils/components/UserCredentials/useSSHCommand.ts
```diff
@@ -0,0 +1,46 @@
...
+import { getCloudInitCredentials } from '@kubevirt-utils/resources/vmi';
+import { K8sResourceCommon, useK8sWatchResource } from '@openshift-console/dynamic-plugin-sdk';
+
+export type useSSHCommandResult = {
+  command: string;
+  user: string;
+  sshServiceRunning: boolean;
+};
+
+const getInfrastructureAPIURL = (infrastructure: any): string =>
```
metalice, line 13: can be in a util file , also do we really don't have any other options than : any ?

src/utils/components/UserCredentials/useSSHCommand.ts
```diff
@@ -0,0 +1,46 @@
...
+  infrastructure && infrastructure?.status ? infrastructure?.status.apiServerURL : undefined;
```
metalice, line 14: I'm guessing, infrastructure?.status is enough

metalice: and also infrastructure?.status?.apiServerURL

metalice: hmm, maybe even just return: infrastructure?.status.apiServerURL, as u return undefined if its don't exist.. and if  infrastructure?.status?.apiServerURL don't exist, it already undefined.

src/utils/components/UserCredentials/useSSHCommand.ts
```diff
@@ -0,0 +1,46 @@
...
+      isList: false,
+    });
+
+  const infrastuctureApiUrl =
+    infrastructureLoaded && !infrastructureError && getInfrastructureAPIURL(infrastructure);
+  const apiHostname = infrastuctureApiUrl && new URL(infrastuctureApiUrl).hostname;
+  const consoleHostname = window.location.hostname; // fallback to console hostname
+
+  const { user } = getCloudInitCredentials(vmi);
+  const sshServicePort = sshService?.spec?.ports?.find(
```
metalice, line 34: i think i saw this function in another file, so it can be exported to util

src/utils/resources/vmi/utils/credentials.ts
```diff
@@ -0,0 +1,18 @@
+import { V1VirtualMachineInstance } from '@kubevirt-ui/kubevirt-api/kubevirt';
```
metalice, line 1: please name this file with a better meaning, such as cloud-init.ts , its a share util file, so everything belong to cloud-int can rest here.

src/views/virtualmachines/details/tabs/details/hooks/useSSHService.tsx
```diff
@@ -1,37 +1,25 @@
...
-  const sshServiceModal = React.useMemo(
-    () => ({
-      kind: ServiceModel.kind,
-      isList: true,
-      namespace: vmi?.metadata?.namespace,
-    }),
-    [vmi],
-  );
-
-  const [services, isServicesLoaded] = useK8sWatchResource<K8sResourceCommon[]>(sshServiceModal);
```
metalice, line 22: why not using useK8sWatchResource to get services? why u using k8sGet, you won't be updated with services changes..

metalice:
Hi @upalatucci , thank you for the PR. left some comments, also please connect it as well to VMI details tab, in the appropriate section.

src/utils/components/UserCredentials/useSSHCommand.ts
upalatucci: I've found only `K8sResourceKind` and `any` in the console code. But with `K8sResourceKind` i can't do `infrastructure.status`

src/utils/components/UserCredentials/useSSHCommand.ts
upalatucci: Do you think utils/utils.ts can be a good file?

upalatucci:
Thanks, @metalice. Sure I'll work on it.

src/utils/components/UserCredentials/useSSHCommand.ts
metalice: yes can be :)

src/utils/components/UserCredentials/useSSHCommand.ts
metalice: if we have InfrastructureModel we don't have a matching API? for example V1Infrastructure or something like that?

PR #323, opened 2022-04-26
Fixed a bug in vmi creation

DanaOrr:

## 📝 Description
Fixed a bug in vmi creation

## 🎥 Demo
image
image

### DIFF
--- a/src/views/virtualmachinesinstance/list/components/VirtualMachineInstanceEmptyState/VirtualMachineInstanceEmptyState.tsx
+++ b/src/views/virtualmachinesinstance/list/components/VirtualMachineInstanceEmptyState/VirtualMachineInstanceEmptyState.tsx
@@ -20,14 +20,16 @@ import { RocketIcon, VirtualMachineIcon } from '@patternfly/react-icons';
 const VirtualMachineInstanceEmptyState: React.FC = ({}) => {
   const { t } = useKubevirtTranslation();
   const history = useHistory();
-  const namespace = useActiveNamespace();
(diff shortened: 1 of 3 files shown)


glekner (commented):
Great job.
I dont think we need `useActiveNamespace` here.

You should implement this like `VirtualMacineEmptyState`
it receives the URL as a prop from `VirtualMachineList` which has access to namespace`
then you can leave it as ``/k8s/ns/${namespace || 'default'}/templatescatalog`;`

just inject the namespace from `VirtualMachineInstancesList` to `VirtualMacineInstancesEmptyState`

src/views/virtualmachinesinstance/list/components/VirtualMachineInstanceEmptyState/VirtualMachineInstanceEmptyState.tsx
```diff
@@ -17,17 +16,20 @@ import {
...
   const history = useHistory();
-  const namespace = useActiveNamespace();
-  const catalogURL = `/k8s/ns/${namespace || 'default'}/templatescatalog`;
 
   return (
     <EmptyState variant={EmptyStateVariant.large}>
       <EmptyStateIcon icon={VirtualMachineIcon} />
       <Title headingLevel="h4" size="lg">
-        {t('No VirtualMachinesInstaces found')}
+        {t('No VirtualMachinesInstances found')}
```
yaacov, line 32: VirtualMachinesInstances -> VirtualMachineInstances

https://kubevirt.io/api-reference/master/definitions.html#_v1_virtualmachineinstance

PR #339, opened 2022-04-27
Changing the events tab with the disks tab in vmi

DanaOrr:

## 📝 Description
Changing the events tab location with the disks tab location in vmi

## 🎥 Demo

image

image


yaacov:
@DanaOrr can you add more verbose description, for example "the current tab order ... not align with ... need to align with design", and some description for the screenshots, for example "before fixing the bug" "after fixing the bug" ?

PR #340, opened 2022-04-27
Bug 2065019: 'Running VMs per template' in the new overview tab counts VMs that are not running

avivtur:
## 📝 Description

Fix donut card title

## 🎥 Demo

before:

Screenshot (24)

after:

Screenshot (25)


yaacov:
@avivtur hi, are you sure the correct fix is not to actually count running vms ?
this fix sound a little "too easy" for me :-)

yaacov:
/lgtm

(I see it is the correct way in bugzilla)

PR #344, opened 2022-04-28
Bug 2079783: Actions are broken in topology view

glekner:
## 📝 Description

Bug 2079783: Actions are broken in topology view

## 🎥 Demo

image


glekner:
/hold

the bug is not fixed.

PR #348, opened 2022-04-29
Add Descheduler to VM Template's Scheduling tab

hstastna:
## 📝 Description

Add support to opt-in/opt-out of descheduler integration for VMs created from a Template in order to benefit from its load balancing capabilities, to the VM Template details page - _Scheduling and resources requirements_ tab.

## 🎥 Demo
Descheduler is installed:
d1
d2
d3
d4
Descheduler not installed:
not


hstastna:

src/views/templates/details/tabs/scheduling/components/Descheduler.tsx
```diff
@@ -0,0 +1,82 @@
...
+              </div>
+            </>
+          }
+        >
+          <DescriptionListTermHelpTextButton>{t('Descheduler')}</DescriptionListTermHelpTextButton>
+        </Popover>
+      </DescriptionListTermHelpText>
+
+      <DescriptionListDescription>
+        {useDeschedulerInstalled && (
```
avivtur, line 61: you can use here `isDeschedulerInstalled` instead :)

src/views/templates/details/tabs/scheduling/components/Descheduler.tsx
```diff
@@ -0,0 +1,82 @@
+import React from 'react';
+import { TemplateSchedulingGridProps } from 'src/views/templates/details/tabs/scheduling/components/TemplateSchedulingLeftGrid';
```
avivtur, line 2: using the full path can be a problem `src/...` , it is better to use relative paths instead like `../../......`
please change it for all your imports

avivtur (commented):
do we really want to disable it when the Descheduler is not installed?
maybe we should provide some details about why it's disabled?

glekner:
/hold
aviv left review

avivtur:
please fix comments in a follow-up PR as needed :)

src/views/templates/details/tabs/scheduling/components/Descheduler.tsx
hstastna: I cannot name it `isDeschedulerInstalled` because our rules force me to name it with "use"... 

Also it is by design to disable the option when the Descheduler is not installed.. yes, we are missing some toolip that it is not installed. It is not perfect

src/views/templates/details/tabs/scheduling/components/Descheduler.tsx
hstastna: Well, I was told by Kobi that I should not use relative paths..

src/views/templates/details/tabs/scheduling/components/Descheduler.tsx
avivtur: I meant using the const you decalred on line 26

avivtur: since you are using the hook instead it's return value it will always be true

hstastna: Yes, I can see that the `{useDeschedulerInstalled && ...` check is unnecessary. Thank you very much, Aviv!

src/views/templates/details/tabs/scheduling/components/Descheduler.tsx
yaacov: what i said was the using `../` is not a nice solution, it indicate that the item you try to import is not the item you need.
here you have the same problem, using `src/...` indicate the same problem, you are trying to import something you should not need to import.
here you are trying to use a prop from a different component on the same level (`../` or `src/...` is has the same problem).

so if `Deschedualer` need props for `TemplateSchedualer...`  they should not be brothers (e.g. on the same level)
it means that  `TemplateSchedualer...` should be in some place other components that need it's prop can use without needing to use `../` or `src/...`.

PR #351, opened 2022-05-01
Fixing the conditions column in vm

DanaOrr:

## 📝 Description
VM conditions column had wrong value. 
I changed the reason for type.
And I added in the pophover if there is no message that will appear the reason.

## 🎥 Demo
Before:
image
After:
image

### DIFF
--- a/src/views/virtualmachines/list/components/VMStatusConditionLabel/VMStatusConditionLabel.tsx
+++ b/src/views/virtualmachines/list/components/VMStatusConditionLabel/VMStatusConditionLabel.tsx
@@ -19,7 +19,7 @@ export const VMStatusConditionLabel: React.FC<V1VirtualMachineCondition> = React
             e.preventDefault();
           }}
         >
-          {condition?.reason}={condition?.type}
+          {condition?.type}={condition?.status}
         </Label>
       </Popover>
     );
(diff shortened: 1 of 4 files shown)


src/views/virtualmachines/list/components/VMStatusConditionLabel/VMStatusConditionLabel.tsx
```diff
@@ -19,7 +19,7 @@ export const VMStatusConditionLabel: React.FC<V1VirtualMachineCondition> = React
             e.preventDefault();
           }}
         >
-          {condition?.reason}={condition?.status}
+          {condition?.reason}={condition?.type}
```
avivtur, line 22: we want to show the condition's type and the condition's status
so please change this line to `{condition?.type}={condition?.status}`

yaacov:
can you fix the description ?

what is the problem ?
why it happen ?
how you fixed it ?

screenshots:
"the list before fix, see the condition in red triangle has ... "
"this is the list after fix, see ..."

yaacov:
a - the after image still show the bug
b - the bug was that we show "<reason for failing> is  false/true" instead of "<type or failing> is false/true" so users could not know what is wrong, for example instead of saying "ready = false" letting the user the VM is not ready, we showed "pod fail = false" , so users may thing "ahh, pod is not failing" all is OK, but what we needed to say is "vm is not ready"
c - fix is to show "<type or failing> is false/true" instead of "<reason for failing> is  false/true"

avivtur:
/retitle Bug 2078709: VM conditions column have wrong key/values

PR #352, opened 2022-05-01
Add filter by template to VM list view

avivtur:
## 📝 Description

Adding a filter to VM list view to filter according to the template

## 🎥 Demo

templates-filter


src/views/virtualmachines/utils/virtualMachineRowFilter.ts
```diff
@@ -1,10 +1,15 @@
+import { TFunction } from 'i18next';
+
+import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import { OTHER } from '@kubevirt-utils/components/DiskModal/DiskFormFields/utils/constants';
```
yaacov, line 4: not sure we want to use OTHER from disks :-)

avivtur: changed to using const other = t('Other') instead

PR #364, opened 2022-05-02
Display Tooltip when VM Template Descheduler not installed

hstastna:
## 📝 Description
Add the tooltip to the _Descheduler_ to inform the user that:
```
"To enable the descheduler, you must install the Kube Descheduler Operator from OperatorHub and enable one
or more descheduler profiles.", 
```
if the Descheduler is not installed, according to the new design.

Include some minor changes/fixes, regarding Descheduler:

 - Remove the unnecessary check [here](https://github.com/kubevirt-ui/kubevirt-plugin/compare/main...hstastna:Template_Descheduler_tooltip?expand=1#diff-162e38124fe78686c5785f0692a79308cbd1f8218295b61c53ea4551428de71eL61), as this check was already present as the prop for the `Button` component [here](https://github.com/kubevirt-ui/kubevirt-plugin/compare/main...hstastna:Template_Descheduler_tooltip?expand=1#diff-162e38124fe78686c5785f0692a79308cbd1f8218295b61c53ea4551428de71eL64).

 - Update the Descheduler component's props: use the `DeschedulerProps` instead of `TemplateSchedulingGridProps`.

 - Change the `useDeschedulerOn` to `isDeschedulerOn`, and check **only** the appropriate annotation.

 - Update displaying the Pencil icon (`PencilAltIcon` PF component) the way that if the field is not editable (operator not installed), the icon is not displayed - according to the design updates.

Related PR: https://github.com/kubevirt-ui/kubevirt-plugin/pull/348

## 🎥 Demo
_Before:_
desch_before

_After:_
Tooltip with the useful info and no any Pencil icon in case that the operator is not installed:
afterr_d

### DIFF
--- a/src/views/templates/details/tabs/scheduling/components/Descheduler.tsx
+++ b/src/views/templates/details/tabs/scheduling/components/Descheduler.tsx
@@ -1,10 +1,10 @@
 import React from 'react';
-import { TemplateSchedulingGridProps } from 'src/views/templates/details/tabs/scheduling/components/TemplateSchedulingLeftGrid';
-import { useDeschedulerInstalled, useDeschedulerOn } from 'src/views/templates/utils';
+import { isDeschedulerOn, useDeschedulerInstalled } from 'src/views/templates/utils';
 import { DESCHEDULER_URL } from 'src/views/templates/utils/constants';
 
 import { useModal } from '@kubevirt-utils/components/ModalProvider/ModalProvider';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
+import { V1Template } from '@kubevirt-utils/models';
 import {
   Button,
   DescriptionListDescription,
@@ -12,6 +12,7 @@ import {
   DescriptionListTermHelpText,
   DescriptionListTermHelpTextButton,
   Popover,
+  Tooltip,
 } from '@patternfly/react-core';
 import { ExternalLinkAltIcon, PencilAltIcon } from '@patternfly/react-icons';
 
@@ -19,10 +20,13 @@ import DeschedulerModal from './DeschedulerModal';
 
 import 'src/views/templates/details/tabs/scheduling/TemplateSchedulingTab.scss';
 
-const Descheduler: React.FC<TemplateSchedulingGridProps> = ({ template }) => {
+type DeschedulerProps = {
+  template: V1Template;
+};
+
+const Descheduler: React.FC<DeschedulerProps> = ({ template }) => {
   const { t } = useKubevirtTranslation();
   const { createModal } = useModal();
-  const isDeschedulerOn = useDeschedulerOn(template);
   const isDeschedulerInstalled = useDeschedulerInstalled();
 
   return (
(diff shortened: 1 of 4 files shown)


hstastna:

yaacov:
/hold

- the help for the descheduler should be when users press the title with doted lines
- you should not have conditional help messages, e.g. if a user can't do some action in some condition, the help message should describe when the user can do the action whether they can or can't do it now.
- when an item is not editable, the pencil should be removed

src/views/templates/details/tabs/scheduling/components/Descheduler.tsx
```diff
@@ -1,25 +1,30 @@
...
 
-const Descheduler: React.FC<TemplateSchedulingGridProps> = ({ template }) => {
+type DeschedulerProps = {
+  template: V1Template;
+};
+
+const Descheduler: React.FC<DeschedulerProps> = ({ template }) => {
   const { t } = useKubevirtTranslation();
   const { createModal } = useModal();
   const isDeschedulerOn = useDeschedulerOn(template);
```
yaacov, line 30: this hook is a bug, when checking for descheduler on/off you should only check for annotation, not looking for operators.

src/views/templates/details/tabs/scheduling/components/Descheduler.tsx
```diff
@@ -58,22 +63,33 @@ const Descheduler: React.FC<TemplateSchedulingGridProps> = ({ template }) => {
...
-            variant="link"
-            iconPosition={'right'}
-          >
-            {isDeschedulerOn ? t('ON') : t('OFF')}
-            <PencilAltIcon className="co-icon-space-l pf-c-button-icon--plain" />
-          </Button>
-        )}
+        <Tooltip
+          content={
+            !isDeschedulerInstalled &&
```
yaacov, line 68: you should not have conditional help messages

src/views/templates/details/tabs/scheduling/components/Descheduler.tsx
```diff
@@ -58,22 +63,33 @@ const Descheduler: React.FC<TemplateSchedulingGridProps> = ({ template }) => {
...
+            t(
+              'To enable the descheduler, you must install the Kube Descheduler Operator from OperatorHub and enable one or more descheduler profiles.',
+            )
+          }
+          position="right"
+        >
+          <span>
+            <Button
+              isInline
+              isDisabled={!isDeschedulerInstalled}
```
yaacov, line 78: why, a user can't set descheduler and later install the operator ?

src/views/templates/details/tabs/scheduling/components/Descheduler.tsx
```diff
@@ -58,22 +63,33 @@ const Descheduler: React.FC<TemplateSchedulingGridProps> = ({ template }) => {
...
+              onClick={() =>
+                createModal(({ isOpen, onClose }) => (
+                  <DeschedulerModal template={template} isOpen={isOpen} onClose={onClose} />
+                ))
+              }
+              variant="link"
+              iconPosition={'right'}
+            >
+              {isDeschedulerInstalled && isDeschedulerOn ? t('ON') : t('OFF')}
```
yaacov, line 87: if the vm has Descheduler annotation set to on, a user should see that it is on

for example:
if a user get from cluster admin a template that has Descheduler on,
then check in the UI and see that it's off
this is not the correct behavior

we can do something like:
a - no descheduler annotations => 'OFF'
b - deschedualer annotation + operator installed => 'ON'
c - deschedualer annotation + operator not installed => 'ON (deschedualer operator not installed, learn more ^ )' with a learn more link to kubevirt doc about deschedualer

yaacov: p.s.
did you check that non-privilage user (e.g. test user) can watch operator,
can a non-privlage user know if the operator installed or not ?

src/views/templates/details/tabs/scheduling/components/Descheduler.tsx
hstastna: What is not correct about the fact that the descheduler may not be installed? Of course it can be installed later, but it does not make sense to have this option enabled, if it is not installed.

src/views/templates/details/tabs/scheduling/components/Descheduler.tsx
hstastna: The descheduler has to be installed, only then it makes sense to look at the annotation. I was told to implement it so also in our older repo... Also in the design doc, there is:
_If the operator isn’t installed, the Descheduler is disabled as well._
+ the picture with descheduler set to OFF

src/views/templates/details/tabs/scheduling/components/Descheduler.tsx
hstastna: So what is the correct way to display different messages for different scenarios?

src/views/templates/details/tabs/scheduling/components/Descheduler.tsx
hstastna: We need to check for both, the operator and the annotation.

hstastna:
>  - the help for the descheduler should be when users press the title with doted lines
  >  - you should not have conditional help messages, e.g. if a user can't do some action in some condition, the help message should describe when the user can do the action whether they can or can't do it now.
>   - when an item is not editable, the pencil should be removed

I am sorry that I followed the design doc I got.

src/views/templates/details/tabs/scheduling/components/Descheduler.tsx
hstastna: > did you check that non-privilage user (e.g. test user) can watch operator,
can a non-privlage user know if the operator installed or not ?

No, because I was told not to do too many things in one PR. Also I think we are already mixing too much stuff in this PR. This PR is not about the Descheduler itself, it's just about adding the missing tooltip, according to the design. The discussion about how the Descheduler should work, should have been done here: https://github.com/kubevirt-ui/kubevirt-plugin/pull/348 If it is buggy, I suggest to make sure what is buggy or which design doc to follow, and then to make another PR especially for the purpose of fixing the Descheduler's bugs.

yaacov:
/lgtm

Note: we will need to fix the way we handle deschedualer on + no-operator in a future PR

PR #365, opened 2022-05-02
Display the popovers in VM Template Details tab correctly

hstastna:
## 📝 Description
Display the content of the popovers for _Name_, _Namespace_, _Labels_, _Annotations_, _Owner_, _Created at_ correctly - with the additional empty line above the breadcrumbs in the popover's content.

**Before:**
Name popover:
name_before
Owner popover:
owner_before

**After:**
Name popover:
name_after
Owner popover:
owner_after


hstastna:

src/views/templates/details/tabs/details/components/Annotations.tsx
```diff
@@ -12,6 +12,8 @@ import {
   Popover,
 } from '@patternfly/react-core';
 
+import 'src/views/templates/details/tabs/details/TemplateDetailsPage.scss';
```
yaacov, line 15: the annotations component should not need this scss import,
this scss should be called from TemplateDetailsPage.ts
if you need it here, you need to look why, and adress the issue some other way

yaacov:
/lgtm

@hstastna  hi, can you look why you need the scss imports and fix that issue ?

btw/ a hint that this import is problematic is that you needed to use import from non-trivial path (using full path is the same as using relative path, it's a way to import things that may not be planned to imported from your component) sometimes it is necessary to do import from relative paths, but it's a hint that something may be wrong

hstastna:
@yaacov the issue is fixed, PTAL again. Thanks!

PR #379, opened 2022-05-04
Add migration tool page

upalatucci:
Documentation: https://docs.google.com/document/d/1Oi17nG4zGr05zBwuCPvq63SmCwvDVUPwgyC5XICfml8/edit#

## 🎥 Demo
1. Operator has been found but no subscription: link to add the subscription
Screenshot from 2022-05-04 14-46-05

1. Operator and Subscription have been found: link to the MTV
Screenshot from 2022-05-04 14-46-58


src/views/migration/components/MigrationToolPhrase.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { ExternalLinkAltIcon } from '@patternfly/react-icons';
+
+import { PackageManifestKind, SubscriptionsKind } from '../constant';
+
+type MigrationToolPhraseProps = {
+  operator: PackageManifestKind;
+  subscription: SubscriptionsKind;
+};
+
+const MigrationToolPhrase: React.FC<MigrationToolPhraseProps> = ({ operator, subscription }) => {
```
yaacov, line 17: we should only detect if it's installed and running, or if it's not,;
if it's installed we send users to the MTV application
if it's not installed completely we send users to the operator hub.

the differentiation between operator/subscription and instance depend on MTV and it's the responsibility of the MTV team to have documentation for that.
The best we can do is send user to the MTV docs.

src/views/migration/constant.ts
```diff
@@ -0,0 +1,36 @@
+import { K8sModel } from '@openshift-console/dynamic-plugin-sdk';
+import { K8sResourceCommon } from '@openshift-console/dynamic-plugin-sdk';
+
+export const MTV_OPERATOR = 'mtv-operator';
+
+export type PackageManifestKind = K8sResourceCommon & { status: any };
+
+export type SubscriptionsKind = K8sResourceCommon & { spec: any; status: any };
+
+export const SubscriptionModel: K8sModel = {
```
yaacov, line 10: we should not add models in the plugin code, they get stale and we can't update them manually each time upstream update a version.

src/views/migration/utils.ts
```diff
@@ -0,0 +1,7 @@
+import { PackageManifestKind, SubscriptionsKind } from './constant';
+
+export const createInstallUrl = (operator: PackageManifestKind, namespace?: string) =>
+  `/operatorhub/subscribe?pkg=${operator?.metadata?.name}&catalog=${operator?.status?.catalogSource}&catalogNamespace=${operator?.status?.catalogSourceNamespace}&targetNamespace=${namespace}`;
+
+export const createOperatorURL = (subscription: SubscriptionsKind) =>
```
yaacov, line 6: that does not look right, the url should look something like `mtv.apps.my-cluster.com` starting with `/k8s...` does not look good.

see:
https://access.redhat.com/documentation/en-us/migration_toolkit_for_virtualization/2.3/html/installing_and_using_the_migration_toolkit_for_virtualization/installing-the-operator#obtaining-console-url_mtv

src/views/migration/constant.ts
upalatucci: Should I use any or maybe we can create these models in the kubevirt-api in some ways?  Same thing for the SubscriptionsKind

src/views/migration/components/MigrationToolPhrase.tsx
upalatucci: Can you define `not installed completely`?
We have 3 cases:
- not installed ( cluster has operator but not subscription ) -> Button is a link to the subscription page
- installed (operator and subscription ) -> Button is a link to the MTV page  something like `virt-openshift-mtv.apps.CLUSTER` 
-alert (no operator ) -> Alert

yaacov: we should identify if we have a path to a working MTV installation,
if no link available we assume MTV is not installed or not installed correctly (or completely, e.g. while installing)

if we have a link to local install (or in the future to an external install) we can have a link to the MTV applicatin
o/w we tell users we did not identify a working MTV installation and guide them to the documentation and to the operator hub in case they want to create a local install.

src/views/migration/constant.ts
```diff
@@ -0,0 +1,12 @@
...
+export const MTV_OPERATOR = 'mtv-operator';
+
+export const MTV_ROUTE_NAME = 'virt';
+
+export type PackageManifestKind = K8sResourceCommon & { status: any };
+
+export type SubscriptionsKind = K8sResourceCommon & { spec: any; status: any };
+
+export const MIGRATION_TOOL_DOCUMENTATION_URL =
+  'https://access.redhat.com/documentation/en-us/migration_toolkit_for_virtualization/2.3/html/installing_and_using_the_migration_toolkit_for_virtualization/index#:~:text=The%20Migration%20Toolkit%20for%20Virtualization%20(MTV)%20enables%20you%20to%20migrate,running%20on%20OpenShift%20Container%20Platform';
```
yaacov, line 12:
```suggestion
  'https://access.redhat.com/documentation/en-us/migration_toolkit_for_virtualization/2.3/html/installing_and_using_the_migration_toolkit_for_virtualization';
```

src/views/migration/constant.ts
```diff
@@ -0,0 +1,12 @@
+import { K8sResourceCommon } from '@openshift-console/dynamic-plugin-sdk';
+
+export const MTV_OPERATOR = 'mtv-operator';
+
+export const MTV_ROUTE_NAME = 'virt';
```
yaacov, line 5: according to [1] the correct way is to look for a label "service: forklift-ui"
" The URL for the forklift-ui service that opens the login page for the MTV web console is displayed."
 
( if we are using the documentation we are safer when breaking changes are done ) 

[1] https://access.redhat.com/documentation/en-us/migration_toolkit_for_virtualization/2.3/html/installing_and_using_the_migration_toolkit_for_virtualization/installing-the-operator#obtaining-console-url_mtv

src/views/migration/components/MigrationToolPhrase.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const { t } = useKubevirtTranslation();
+
+  if (!operator)
+    return (
+      <Alert
+        variant={AlertVariant.danger}
```
yaacov, line 27: it's perfectly fine to not install MTV on this cluster, you can still use it if it's installed on a different cluster.

src/views/migration/components/MigrationToolPhrase.tsx
```diff
@@ -0,0 +1,58 @@
...
+        <Link to={'/operatorhub'}>
+          OperatorHub. <ExternalLinkAltIcon />
+        </Link>
+      </Trans>
+    );
+
+  return (
+    <>
+      {t(
+        'The Migration Toolkit operator is successfully installed, please deploy Migration Tool instance in order to use the Migration Toolkit',
```
yaacov, line 52: see https://github.com/kubevirt-ui/kubevirt-plugin/pull/379/files#r867438065
the installation of MTV is more complex then just installing the MTV instance, we can leave the "how to install" to the official documentation.

src/views/migration/constant.ts
yaacov: kubevirt-api, in the short term we can add it as hard coded modals, in the longer term we will need some automation to auto update the models

src/globals.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module '*.png' {
```
glekner, line 1: why do we need this? can't you just inline the png path in https://github.com/kubevirt-ui/kubevirt-plugin/pull/379/files#diff-52a107f248365357ba6e5dbb6bb3f3617058f9a112552af68f2df2d479ba0df3R57

upalatucci: Typescript without this line give me this error
``` Cannot find module './migration-illustration.png' or its corresponding type declarations.```

src/views/migration/constant.ts
upalatucci: Instead [here](https://access.redhat.com/documentation/en-us/migration_toolkit_for_virtualization/2.3/html/installing_and_using_the_migration_toolkit_for_virtualization/installing-the-operator#obtaining-console-url_mtv) the command to get the MTV web console URL is 
```$ oc get route virt -n openshift-mtv -o custom-columns=:.spec.host``` .
This means that I have to get the resource `route` named `virt` in the namespace `openshift-mtv` and use the `spec.host` attribute.

Maybe these two methods are equivalent?

PR #385, opened 2022-05-08
fix template details page crash if no description available

glekner:
## 📝 Description
fix template details page crash if no description available


hstastna (approved):
Thank you for fixing this, Gilad! Good catch! 👍🏽

PR #388, opened 2022-05-09
Fix template's page to load templates for non-priv user

avivtur:
## 📝 Description

when selecting a namespace as non-privileged user, and navigating to templates list, we get an error message for fetching templates if one of all namespaces that a user can list, has recieved an error.

## 🎥 Demo

before:

templates-page-load-error

after:

https://user-images.githubusercontent.com/67270715/167402693-c4812c89-9732-4c6a-811c-8b3c9cf84497.mp4

### DIFF
--- a/src/utils/resources/template/hooks/useVmTemplates.tsx
+++ b/src/utils/resources/template/hooks/useVmTemplates.tsx
@@ -16,6 +16,10 @@ import { Operator } from '@openshift-console/dynamic-plugin-sdk-internal/lib/api
 
 import { TEMPLATE_TYPE_BASE, TEMPLATE_TYPE_LABEL, TEMPLATE_TYPE_VM } from '../utils/constants';
 
+const OPENSHIFT_NS = 'openshift';
+const KUBEVIRT_OS_IMAGES_NS = 'kubevirt-os-images';
+const OPENSHIFT_OS_IMAGES_NS = 'openshift-virtualization-os-images';
+
 /** A Hook that returns VM Templates from allowed namespaces
  * @param namespace - The namespace to filter the templates by
  */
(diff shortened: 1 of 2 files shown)


src/utils/resources/template/hooks/useVmTemplates.tsx
```diff
@@ -35,10 +39,16 @@ export const useVmTemplates = (namespace?: string): useVmTemplatesValues => {
...
   // Bug fix: TODO open BZ for SSP oc get projects don't show on projects when non-priv users
   const projectNames = projects?.map((proj) => proj?.metadata?.name);
-  if (projectNames && !projectNames.includes('openshift')) {
-    projectNames.push('openshift');
+  if (projectNames && !projectNames.includes(OPENSHIFT_NS)) {
+    projectNames.push(OPENSHIFT_NS);
   }
 
+  // Bug fix: the project with the images can't access templates which cause
+  // the template list view to show error instead of existing templates
```
yaacov, line 47: the correct fix should be not to fail if can't access templates for this namespace,
with correct fix if user can list templates from this namespaces they will not see them

src/utils/resources/template/hooks/useVmTemplates.tsx
```diff
@@ -35,10 +39,16 @@ export const useVmTemplates = (namespace?: string): useVmTemplatesValues => {
 
   // Bug fix: TODO open BZ for SSP oc get projects don't show on projects when non-priv users
```
yaacov, line 40: BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2081295

yaacov:
/hold
btw/ why are we allowing all namespace for templates at all ?

@avivtur can you remove the "all namespace" support for non admins ?
it's a regular page, if user don't have permissions for "all namespace" they should see an error "user not allowed to ... "
templates are just an object like config maps and pods

src/utils/resources/template/hooks/useVmTemplates.tsx
avivtur: once we try to fetch templates from namespace `openshift-virtualization-os-images`/`kubevirt-os-images` we would fail, that is the behavior from the backend as well but this is a template we can list when list for projects, that is the reason I removed that namespace (`openshift-virtualization-os-images`/`kubevirt-os-images`). 
do you have another suggestion how to go around it?

PR #392, opened 2022-05-10
[e2e] vmi tabs should be loaded

gouyang:
add test to check vmi tabs can be loaded normally


gouyang:
the VMI pages will probably be removed, hence close this one

PR #398, opened 2022-05-10
Improve performance of template boot source availability

glekner:
## 📝 Description

Improve performance of template boot source availability


metalice:
lgtm

PR #404, opened 2022-05-11
Display Template Descheduler ON according to annotation

hstastna:
## 📝 Description
Display Template Descheduler's value "ON"/"OFF" according to annotation, no matter the operator installed or not, according to the design.
Only display the value for non admin user (cannot edit no matter the operator installed or not).

## 🎥 Demo
Before: Descheduler operator not installed but the annotation present:
off
After: The operator not installed but the annotation present:
on
Non admin can only see the value: (annotation not present)
off_non_admin


hstastna:

PR #408, opened 2022-05-11
Bug 2071488: "Migrate Node to Node" is confusing.

avivtur:
## 📝 Description

change VM migrate action label

## 🎥 Demo

before:

Screenshot (31)

after:

Screenshot (35)

### DIFF
--- a/src/views/virtualmachines/actions/VirtualMachineActionFactory.tsx
+++ b/src/views/virtualmachines/actions/VirtualMachineActionFactory.tsx
@@ -113,6 +113,7 @@ export const VirtualMachineActionFactory = {
       label: t('Migrate'),
       cta: () => migrateVM(vm),
       accessReview: asAccessReview(VirtualMachineModel, vm, 'patch'),
+      description: t('Migrate to a different Node'),
     };
   },
   cancelMigration: (
--- a/src/views/virtualmachines/actions/VirtualMachineActionFactory.tsx
+++ b/src/views/virtualmachines/actions/VirtualMachineActionFactory.tsx
@@ -110,7 +110,7 @@ export const VirtualMachineActionFactory = {
         !!vm?.status?.conditions?.find(
           ({ type, status }) => type === 'LiveMigratable' && status === 'False',
         ),
-      label: t('Migrate VirtualMachine to Node'),
+      label: t('Migrate'),
       cta: () => migrateVM(vm),
       accessReview: asAccessReview(VirtualMachineModel, vm, 'patch'),
     };
--- a/locales/en/plugin__kubevirt-plugin.json
+++ b/locales/en/plugin__kubevirt-plugin.json
@@ -371,6 +371,7 @@
   "Memory used": "Memory used",
   "Metrics are collected by the OpenShift Monitoring Operator.": "Metrics are collected by the OpenShift Monitoring Operator.",
   "Migrate": "Migrate",
+  "Migrate to a different Node": "Migrate to a different Node",
   "Migration Toolkit for Virtualization": "Migration Toolkit for Virtualization",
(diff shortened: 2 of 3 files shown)


src/views/virtualmachines/actions/VirtualMachineActionFactory.tsx
```diff
@@ -110,7 +110,7 @@ export const VirtualMachineActionFactory = {
         !!vm?.status?.conditions?.find(
           ({ type, status }) => type === 'LiveMigratable' && status === 'False',
         ),
-      label: t('Migrate VirtualMachine to Node'),
```
yaacov, line 113: it's missing the description [1]:
"Migrate the virtual machine to a different node" 

[1] https://bugzilla.redhat.com/show_bug.cgi?id=2071488#c3

avivtur: I was changing according to comment #7:
https://bugzilla.redhat.com/show_bug.cgi?id=2071488#c7 
as it is should be obvious that it is nodes.
also IMO, it will not look good if only the `Migrate` option will have a description.
we should consider if we want to add descriptions to all actions and if so we follow this practice and be consistent to all of the plugin e.g. Templates/VMI views should also have descriptions in their actions and basically every dropdown the plugin have, for now I think we can close this as is

yaacov: https://bugzilla.redhat.com/show_bug.cgi?id=2071488#c7 was about the title being "Migrate" it did not ask for the removal of the subtitle description.
https://bugzilla.redhat.com/show_bug.cgi?id=2071488#c8  and https://bugzilla.redhat.com/show_bug.cgi?id=2071488#c9 comming after #c7 specifically quote #c3 and #c7 , my understanding is that the agreement is about title + description as suggested in #c3

> also IMO, it will not look good if only the Migrate option will have a description.

Other item may need description too, but that will be addressed in different bugs according to design.

avivtur: > https://bugzilla.redhat.com/show_bug.cgi?id=2071488#c7 was about the title being "Migrate" it did not ask for the removal of the subtitle description.

since we never had a subtitle description [#7](https://bugzilla.redhat.com/show_bug.cgi?id=2071488#c7) is just asking for main title to be `Migrate` and not mentioning any subtitle description is as we don't want description for now

> https://bugzilla.redhat.com/show_bug.cgi?id=2071488#c8 and https://bugzilla.redhat.com/show_bug.cgi?id=2071488#c9 comming after #c7 specifically quote #c3 and #c7 , my understanding is that the agreement is about title + description as suggested in #c3

I was also looking at [#4](https://bugzilla.redhat.com/show_bug.cgi?id=2071488#c4) and [#5](https://bugzilla.redhat.com/show_bug.cgi?id=2071488#c5) and understood we need PM/UX approval,
do we have any design by UX that I'm not aware of for this issue?

> > also IMO, it will not look good if only the Migrate option will have a description.
> 
> Other item may need description too, but that will be addressed in different bugs according to design.

+1

yaacov: > since we never had a subtitle description [#7](https://bugzilla.redhat.com/show_bug.cgi?id=2071488#c7) is just asking for main title to be Migrate and not mentioning any subtitle description is as we don't want description for now as I understand it

the agreement in https://bugzilla.redhat.com/show_bug.cgi?id=2071488#c3 since https://bugzilla.redhat.com/show_bug.cgi?id=2071488#c7 comes after #c3, #c7 assume subtitle exist, and it did not talk about changing the agreement about subtitle.

PR #418, opened 2022-05-15
display default templates for non priv user

avivtur:
## 📝 Description

when navigating to Virtualization -> Catalog as non-privileged user we are displaied with all template items instead of default templates 

## 🎥 Demo

before:

https://user-images.githubusercontent.com/67270715/168478328-2c3dd5e0-408e-4a98-93d8-c3235b322bb8.mp4

after:

https://user-images.githubusercontent.com/67270715/168478344-771a5984-3ec4-4dc4-a9ff-c8efb633c7a5.mp4


yaacov:
/hold to stop re tests

@avivtur can you check the test fails ?

PR #433, opened 2022-05-17
Make Template's Dedicated resources and Eviction strategy editable

hstastna:
## 📝 Description
Make _Dedicated resources_ and _Eviction strategy_ fields editable, in VM Template's _Scheduling_ tab.
Add the appropriate modals to edit the values - in case that the VM Template is not a common template.

## 🎥 Screenshots
**Before:**
grid_before

**After:**
grid_after1
grid_after2
grid_after3


hstastna:

PR #435, opened 2022-05-17
Make Template's Description field editable

hstastna:
## 📝 Description
Make _Description_ field editable, in VM Template's _Details_ tab.
Add using the appropriate modal to edit the description - in case that the VM Template is not a common template.

## 🎥 Demo
**Before:**
Dash displayed in case there was no any description + no editability:
desc_before

**After:**
"None" string displayed in case there is no any description + editability present (only the pencil icon clickable): 
d_1
Modal window to add/edit the description, after clicking on "None" or pencil icon:
desc_after2
Description updated after changes in the modal saved (only the pencil icon clickable):
d_2

### DIFF
--- a/src/views/templates/details/tabs/details/components/Description.tsx
+++ b/src/views/templates/details/tabs/details/components/Description.tsx
@@ -1,0 +1,75 @@
+import * as React from 'react';
+import produce from 'immer';
+
+import { TemplateModel, V1Template } from '@kubevirt-ui/kubevirt-api/console';
+import { DescriptionModal } from '@kubevirt-utils/components/DescriptionModal/DescriptionModal';
+import { useModal } from '@kubevirt-utils/components/ModalProvider/ModalProvider';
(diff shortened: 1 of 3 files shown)


hstastna:

src/views/templates/details/tabs/details/components/Description.tsx
```diff
@@ -0,0 +1,75 @@
...
+      />
+    ));
+
+  return (
+    <DescriptionListGroup>
+      <DescriptionListTerm>{t('Description')}</DescriptionListTerm>
+      <DescriptionListDescription>
+        {editable ? (
+          <Button type="button" isInline onClick={onEditClick} variant="link">
+            {templateDescription}
```
yaacov, line 64: description should not be clickable
a. it may be a long text
b. it should be black

hstastna: Thanks, I will update the PR

PR #436, opened 2022-05-17
Bug 2086479: Use "Save" submit button text in all modals

hstastna:
## 📝 Description

**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2086479

Use "Save" string as the default submit button text for all of the modal windows we use, instead of "Submit", to achieve consistency in the UI.
Remove unnecessary optional `submitBtnText` prop when using `TabModal` component, as "Save" will be the default submit button text.

## 🎥 Screenshots
**Before:**
btn_before

**After:**
btn_after


hstastna:

PR #450, opened 2022-05-19
Bug 2087232: all-namespace back to template list

upalatucci:
before: click on template => template detail (in the template namespace) => click on the template menu -> template list in the template namespace
now clicking on template menu you get redirected always to template list in all-namespaces

## 🎥 Demo

TemplateListAllNamespace

### DIFF
--- a/console-extensions.json
+++ b/console-extensions.json
@@ -57,37 +57,15 @@
     }
   },
   {
-    "type": "console.navigation/href",
+    "type": "console.navigation/resource-ns",
     "properties": {
       "id": "templates",
       "section": "virtualization",
       "name": "%plugin__kubevirt-plugin~Templates%",
-      "href": "/templates",
-      "prefixNamespaced": true,
+      "resource": "templates",
       "dataAttributes": {
         "data-quickstart-id": "qs-nav-templates",
         "data-test-id": "templates-nav-item"
-      }
-    },
-    "flags": {
-      "required": ["OPENSHIFT"]
-    }
-  },
-  {
-    "type": "console.page/route",
-    "properties": {
-      "path": ["/k8s/ns/:ns/templates/:name", "/k8s/all-namespaces/templates/:name"],
-      "component": {
-        "$codeRef": "TemplateNavPage"
-      }
-    }
-  },
-  {
-    "type": "console.page/route",
-    "properties": {
-      "path": ["/k8s/ns/:ns/templates", "/k8s/all-namespaces/templates"],
-      "component": {
-        "$codeRef": "VirtualMachineTemplatesList"
       }
     }
   },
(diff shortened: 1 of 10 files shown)


src/views/templates/list/components/VirtualMachineTemplatesRow.tsx
```diff
@@ -10,12 +12,27 @@ import useWorkloadProfile from '../hooks/useWorkloadProfile';
...
-  RowProps<V1Template, { kind: string; availableTemplatesUID: Set<string> }>
-> = ({ obj, activeColumnIDs, rowData: { kind, availableTemplatesUID } }) => {
+  RowProps<V1Template, { availableTemplatesUID: Set<string> }>
+> = ({ obj, activeColumnIDs, rowData: { availableTemplatesUID } }) => {
+  const history = useHistory();
+
   return (
     <>
       <TableData id="name" activeColumnIDs={activeColumnIDs}>
-        <ResourceLink kind={kind} name={obj.metadata.name} namespace={obj.metadata.namespace} />
```
avivtur, line 18: why not using the `onClick` prop to push the url?

console-extensions.json
```diff
@@ -57,26 +57,22 @@
...
   {
-    "type": "console.navigation/href",
+    "type": "console.navigation/resource-ns",
     "properties": {
       "id": "templates",
       "section": "virtualization",
       "name": "%plugin__kubevirt-plugin~Templates%",
-      "href": "/templates",
-      "prefixNamespaced": true,
+      "resource": "virtualmachinetemplates",
```
avivtur, line 65: is there a reason for using `"resource"` instead of `"model"` property?
if it should be `"resource"`, should it be `"virtualmachinetemplates"` and not simply `"templates"`?

upalatucci: This is a bit of a hack and not sure about that cc: @vojtechszocs . We are using a custom page to show the virtual machine templates (`console.navigation/href`) . It's not a resource page like virtual machines or pods, jobs, etc (`console.navigation/resource-ns`). But the nav item of these resources has a different namespace mechanism (that fixes the bug if applied also to the templates)

src/views/templates/list/components/VirtualMachineTemplatesRow.tsx
upalatucci: Ah great!

console-extensions.json
avivtur: Template is a resource, we simply want to filter only the Templates resource with a VM object when we fetch them, I think we should follow same logic as we do on VM or VMI, @yaacov wdyt?

yaacov: a. Is it better to use "resource": "templates", @vojtechszocs WDYT ?
b. can we use "console.page/resource/list" and "console.page/resource/details" to override an internal resource view ?

src/views/templates/details/TemplatePageTitle.tsx
```diff
@@ -26,7 +30,7 @@ const TemplatePageTitle: React.FC<TemplatePageTitleTitleProps> = ({ template })
           <Button
             variant="link"
             isInline
-            onClick={() => onBreadcrumbClick(`/k8s/ns/${namespace}/${VirtualMachineModelRef}`)}
+            onClick={() => onBreadcrumbClick(`/k8s/${namespacePath}/${VirtualMachineModelRef}`)}
           >
             {t('Virtualization')}
```
yaacov, line 35: this will create wrong breadcrumbs:
should be:
Pod > pod details
here we have:
Workloads > Pod > pod details

@upalatucci  can you remove the "virtualization" link ?

PR #456, opened 2022-05-23
Bug 2087589: Fix customize form errors

glekner:
## 📝 Description

Bug 2087589: Fix customize form errors

1. Added `react-hook-form`
2. Fix unwanted errors before submit inside the customize form

Finally found a way to make `react-hook-form` work with Patternfly.
we should add it to more forms in the future


upalatucci (approved):
/lgtm

PR #457, opened 2022-05-23
Bug 2087559: template in "VMs per template" should take user to vm list

glekner:
## 📝 Description
Bug 2087559: template in "VMs per template" should take user to vm list


avivtur (approved):
/lgtm
:1st_place_medal:

PR #459, opened 2022-05-23
update dockerfile - improve security and reduce size

glekner:
## 📝 Description

update dockerfile - improve security and reduce size


upalatucci (approved):
/lgtm

PR #460, opened 2022-05-23
Bug 2088407: Create Template button

upalatucci:
## 🎥 Demo

TemplateCreation

### DIFF
--- a/src/templates/vm-template-yaml.ts
+++ b/src/templates/vm-template-yaml.ts
@@ -16,15 +16,25 @@ objects:
   - apiVersion: kubevirt.io/v1
     kind: VirtualMachine
     metadata:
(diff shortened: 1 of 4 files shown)


src/templates/vm-template-yaml.ts
```diff
@@ -33,41 +43,50 @@ objects:
...
                   password: '\${CLOUD_USER_PASSWORD}'
                   chpasswd: { expire: False }
               name: cloudinitdisk
-          hostname: '\${NAME}'
 parameters:
   - name: NAME
     description: Name for the new VM
     required: true
-  - password: CLOUD_USER_PASSWORD
+  - name: CLOUD_USER_PASSWORD
```
yaacov, line 89: ?

PR #469, opened 2022-05-24
Bug 2089327: When trying to add network using network modal when no network found …

metalice:
…is crashing

## 📝 Description

When removing all networks from VM and then trying to add a new network is causing the app to crash. It caused as trying to spread undefined. 

## 🎥 Demo

After:

network-crash-after

Before:

network-crash-before


upalatucci (approved):
/lgtm

PR #477, opened 2022-05-24
Bug 2094323: Make Template Workload profile editable

hstastna:
## 📝 Description
**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2094323

Make the _Workload profile_ field in the VM Template's _Details_ tab editable.
Add the modal to edit the label, to choose from _Desktop_, _Server_ and _High performance_ profiles.

## 🎥 Demo
**After:**
Click on the pencil icon to edit the workload profile:
profile-after1
profile-after2
profile-after3


hstastna:

src/views/templates/details/tabs/details/components/WorkloadProfile.tsx
```diff
@@ -0,0 +1,77 @@
...
+  DescriptionListGroup,
+  DescriptionListTerm,
+} from '@patternfly/react-core';
+import { PencilAltIcon } from '@patternfly/react-icons';
+
+import { TemplateDetailsGridProps } from '../TemplateDetailsPage';
+
+import WorkloadProfileModal from './WorkloadProfileModal';
+
+const WorkloadProfile: React.FC<TemplateDetailsGridProps> = ({ template, editable }) => {
```
yaacov, line 25: can you add a WorkloadProfile type, e.g. `type WorkloadProfileNames = 'Desktop' | 'Server' | 'High performance'`; ?

src/views/templates/details/tabs/details/components/WorkloadProfile.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const updateWorkload = (updatedWorkload: string) => {
+    const updatedTemplate = produce<V1Template>(template, (templateDraft: V1Template) => {
+      if (!templateDraft.objects[0].spec.template.metadata.annotations)
+        ensurePath(templateDraft, 'objects[0].spec.template.metadata.annotations');
+
+      delete templateDraft.objects[0].spec.template.metadata.annotations[VM_WORKLOAD_ANNOTATION];
+
+      if (Object.values(workloadLabels).includes(updatedWorkload))
+        templateDraft.objects[0].spec.template.metadata.annotations[VM_WORKLOAD_ANNOTATION] =
+          Object.keys(workloadLabels).find((key) => workloadLabels[key] === updatedWorkload);
```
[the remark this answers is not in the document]
yaacov: btw:// vm may not be always object 0, best is to find the vm object
```
templateDraft.objects.find((obj) => { return obj.kind === 'VirtualMachine'; })
```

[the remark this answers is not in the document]
avivtur: we have a util function for it called `getTemplateVirtualMachineObject` under `utils` directory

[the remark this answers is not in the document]
hstastna: From what I was told, we want to display only **VM** Templates in the list, so I've implemented it this way, so we work only with such Templates here. Such Templates have only one object by design, so this check is unnecessary here. I am not a fan of "what if it changes sometimes in the future... " approach, because it can lead to producing lots of code that will never be really useful. So this is why I don't spend too much time here checking if the object is really a VM object.

Also I am not very happy using `getTemplateVirtualMachineObject` function, because it is a long name, IMHO it makes the code less readable, comparing to directly using `objects[0]`which is super clear, but I understand your point of view, folks. Thanks for that! 👍🏽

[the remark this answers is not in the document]
yaacov: > From what I was told, we want to display only VM Templates 

we should show by default only templates that have a vm object, but this is a templates list (all templates), and a template page (any template), we should be able to show all, and play nicely with all templates, including templates that more objects then just the vm object.

> Also I am not very happy using getTemplateVirtualMachineObject function

my vote is to do the find inline :-)
``` ts
const obj = templateDraft.objects.find((obj) => { return obj.kind === 'VirtualMachine'; });
// vs
const obj =  getTemplateVirtualMachineObject(templateDraft);
```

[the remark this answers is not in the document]
hstastna: I had discussion with the UX and there is no any reason to display other than VM templates.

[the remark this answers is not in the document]
yaacov: not displaying templates that don't have a vm object, does mean that you should crash if the template include other objects.

for example the template:
```
kind: Template
...
objects:
- kind: Service
  ...
- kind: VirtualMachine
  ...
```

is a template that include a VM object, gut the code you suggest will crash, it will take `object[0]` and assume it's a VM while it's a Service object.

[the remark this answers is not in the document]
hstastna: Here we are working with templates that were already filtered out from other than VM Templates.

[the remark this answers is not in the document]
yaacov: > Here we are working with templates that were already filtered out from other than VM Templates.

the example above it for a template the has a vm object, it should pass the current filtering

src/views/templates/details/tabs/details/components/WorkloadProfile.tsx
hstastna: Done!

src/views/templates/details/tabs/details/components/WorkloadProfile.tsx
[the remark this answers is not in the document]
hstastna: I've updated the PR according to the suggestions.

src/views/templates/details/tabs/details/components/WorkloadProfile.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const updateWorkload = (updatedWorkload: WorkloadProfileNames) => {
+    const updatedTemplate = produce<V1Template>(template, (templateDraft: V1Template) => {
+      ensurePath(getVMObject(templateDraft), 'spec.template.metadata.annotations');
+
+      delete getVMObject(templateDraft).spec.template.metadata.annotations[VM_WORKLOAD_ANNOTATION];
+
+      if (updatedWorkload)
+        getVMObject(templateDraft).spec.template.metadata.annotations[VM_WORKLOAD_ANNOTATION] =
+          Object.keys(workloadLabels).find((key) => workloadLabels[key] === updatedWorkload);
+
```
yaacov, line 41:
```suggestion
      getVMObject(templateDraft).spec.template.metadata.annotations['vm.kubevirt.io/workload'] = updatedWorkload

```

src/views/templates/details/tabs/details/components/WorkloadProfile.tsx
```diff
@@ -0,0 +1,78 @@
...
+import WorkloadProfileModal from './WorkloadProfileModal';
+
+const WorkloadProfile: React.FC<TemplateDetailsGridProps> = ({ template, editable }) => {
+  const { createModal } = useModal();
+  const { t } = useKubevirtTranslation();
+  const workload = useWorkloadProfile(template);
+
+  const updateWorkload = (updatedWorkload: WorkloadProfileNames) => {
+    const updatedTemplate = produce<V1Template>(template, (templateDraft: V1Template) => {
+      ensurePath(getVMObject(templateDraft), 'spec.template.metadata.annotations');
```
yaacov, line 34:
```suggestion
      const vm = getVMObject(templateDraft);
      ensurePath(vm, 'spec.template.metadata.annotations');
```

src/views/templates/details/tabs/details/components/WorkloadProfile.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+const WorkloadProfile: React.FC<TemplateDetailsGridProps> = ({ template, editable }) => {
+  const { createModal } = useModal();
+  const { t } = useKubevirtTranslation();
+  const workload = useWorkloadProfile(template);
+
+  const updateWorkload = (updatedWorkload: string) => {
+    const updatedTemplate = produce<V1Template>(template, (templateDraft: V1Template) => {
+      if (!templateDraft.objects[0].spec.template.metadata.annotations)
+        ensurePath(templateDraft, 'objects[0].spec.template.metadata.annotations');
```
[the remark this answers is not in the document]
hstastna: I haven't found how to simply check the whole path within the existing `ensurePath`, while avoiding adding more _for_-loops. Any suggestion?

src/views/templates/details/tabs/details/components/WorkloadProfile.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const workload = useWorkloadProfile(template);
+
+  const updateWorkload = (updatedWorkload: WorkloadProfileNames) => {
+    const updatedTemplate = produce<V1Template>(template, (templateDraft: V1Template) => {
+      const vm = getVMObject(templateDraft);
+      ensurePath(vm, 'spec.template.metadata.annotations');
+
+      vm.spec.template.metadata.annotations[VM_WORKLOAD_ANNOTATION] = Object.keys(
+        workloadLabels,
+      ).find((key) => workloadLabels[key] === updatedWorkload);
```
yaacov, line 39:
```suggestion
      const workloadAnnotions: Record<WorkloadProfileNames, string> = {
        'Desktop': 'desktop',
        'Server': 'server',
        'High performance': 'highperformance',
      };
      vm.spec.template.metadata.annotations[VM_WORKLOAD_ANNOTATION] = workloadAnnotions[updatedWorkload];
```

workloadLabels

src/views/templates/details/tabs/details/components/WorkloadProfileModal.tsx
```diff
@@ -0,0 +1,69 @@
...
+  onClose: () => void;
+  onSubmit: (workload: WorkloadProfileNames) => Promise<void | K8sResourceCommon>;
+};
+
+const WorkloadProfileModal: React.FC<WorkloadProfileModalProps> = React.memo(
+  ({ obj, isOpen, onClose, onSubmit }) => {
+    const { t } = useKubevirtTranslation();
+    const [isDropdownOpen, setIsDropdownOpen] = React.useState<boolean>(false);
+    const [workload, setWorkload] = React.useState<WorkloadProfileNames>(
+      useWorkloadProfile(obj) || 'Desktop',
```
yaacov, line 27:
```suggestion
    const [workload, setWorkload] = React.useState<WorkloadProfileKeys>(
      useWorkloadProfile(obj) || 'desktop',
```

src/views/templates/list/hooks/useWorkloadProfile.ts
```diff
@@ -6,18 +6,19 @@ import { getTemplateWorkload } from '../../utils/selectors';
...
 
-const workloadLabels = {
+export const workloadLabels = {
   desktop: 'Desktop',
   server: 'Server',
   highperformance: 'High performance',
 };
 
-const useWorkloadProfile = (template: V1Template): string => {
+export type WorkloadProfileNames = 'Desktop' | 'Server' | 'High performance';
```
yaacov, line 16:
```suggestion
export type WorkloadProfileNames = 'Desktop' | 'Server' | 'High performance';
export type WorkloadProfileKeys = 'desktop' | 'server' | 'highperformance';

```

src/views/templates/list/hooks/useWorkloadProfile.ts
```diff
@@ -6,18 +6,19 @@ import { getTemplateWorkload } from '../../utils/selectors';
...
+
+const useWorkloadProfile = (template: V1Template): WorkloadProfileNames => {
```
yaacov, line 18: can we do the translation in the object that use this method ?

e.g. 
the consumer run `t`
```tsx
const useWorkloadProfile = (template) => getVmObj(template).spec.template.metadata.annotations['vm.kubevirt.io/workload']
<some component label={t(useWorkloadProfile(template))}>
```
vs this method run `t`
```tsx
const useWorkloadProfile = (template) => t(getVmObj(template).spec.template.metadata.annotations['vm.kubevirt.io/workload'])
<some component label={useWorkloadProfile(template)}>
```

src/views/templates/details/tabs/details/components/WorkloadProfile.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { PencilAltIcon } from '@patternfly/react-icons';
+
+import { TemplateDetailsGridProps } from '../TemplateDetailsPage';
+
+import WorkloadProfileModal from './WorkloadProfileModal';
+
+const WorkloadProfile: React.FC<TemplateDetailsGridProps> = ({ template, editable }) => {
+  const { createModal } = useModal();
+  const { t } = useKubevirtTranslation();
+  const workload = useWorkloadProfile(template, t);
```
yaacov, line 29:
```suggestion
  const workload = getTemplateWorkload(template);
```

src/views/templates/details/tabs/details/components/WorkloadProfile.tsx
```diff
@@ -0,0 +1,74 @@
...
+        onClose={onClose}
+        onSubmit={updateWorkload}
+      />
+    ));
+
+  return (
+    <DescriptionListGroup>
+      <DescriptionListTerm>{t('Workload profile')}</DescriptionListTerm>
+      <DescriptionListDescription>
+        {workload || t('Other')}
```
yaacov, line 63:
```suggestion
        { t(workloadLabels[workload] || 'Other') }
```

src/views/templates/list/components/VirtualMachineTemplatesRow.tsx
```diff
@@ -42,7 +44,7 @@ const VirtualMachineTemplatesRow: React.FC<
         <ResourceLink kind="Namespace" name={obj.metadata.namespace} />
       </TableData>
       <TableData id="workload" activeColumnIDs={activeColumnIDs} className="pf-m-width-10">
-        {useWorkloadProfile(obj)}
+        {useWorkloadProfile(obj, t) || t('Other')}
```
yaacov, line 47:
```suggestion
         { t(workloadLabels[workload] || 'Other') }
```
you whould also need to get the workload key `const workload = getTemplateWorkload(obj);`

yaacov: maybe a helper here may be nice ?
``` ts
const getWorkloadLabel = (workload: WorkloadProfileKeys): WorkloadProfileDisplayNames  => 
  workloadLabels[workload] || 'Other';
```

src/views/templates/list/hooks/useWorkloadProfile.ts
```diff
@@ -1,23 +1,24 @@
...
   server: 'Server',
   highperformance: 'High performance',
 };
 
-const useWorkloadProfile = (template: V1Template): string => {
-  const { t } = useKubevirtTranslation();
-
-  return t(workloadLabels[getTemplateWorkload(template)] || 'Other');
-};
+export type WorkloadProfileNames = 'Desktop' | 'Server' | 'High performance';
```
yaacov, line 17:
```suggestion
// t('plugin__kubevirt-plugin~Desktop')
// t('plugin__kubevirt-plugin~Server')
// t('plugin__kubevirt-plugin~High performance')
// t('plugin__kubevirt-plugin~Other')
export type WorkloadProfileNames = 'Desktop' | 'Server' | 'High performance' | 'Other';
```

src/views/templates/details/tabs/details/components/WorkloadProfile.tsx
```diff
@@ -0,0 +1,74 @@
...
+    const updatedTemplate = produce<V1Template>(template, (templateDraft: V1Template) => {
+      const vm = getTemplateVirtualMachineObject(templateDraft);
+      ensurePath(vm, 'spec.template.metadata.annotations');
+
+      vm.spec.template.metadata.annotations[VM_WORKLOAD_ANNOTATION] = updatedWorkload;
+
+      return templateDraft;
+    });
+
+    return k8sUpdate({
```
yaacov, line 41: maybe `k8sPatch` makes more sense here ?

src/views/templates/list/hooks/useWorkloadProfile.ts
hstastna: We don't need the prefix `plugin__kubevirt-plugin~` because its already the default namespace.

Regarding 'Other', I would better avoid adding 'Other' to workload profiles, as 'Other' is not a workload profile, but an exception, edge case. Its purpose is to display 'Other' in the templates list or Details tab for the situations when there's other than Desktop, Server, High performance set as template's workload profile.

src/views/templates/list/hooks/useWorkloadProfile.ts
hstastna: Choosing `WorkloadProfileNames` as the name of the type. No need to use `WorkloadProfileDisplayNames` because it is unnecessary long name, and it can be messed with the template's "Display name" which is something different.

src/views/templates/list/components/VirtualMachineTemplatesRow.tsx
hstastna: Adding a helper here is not necessary. But let me think again about how the translation is implemented here. Thanks!

src/views/templates/details/tabs/details/components/WorkloadProfile.tsx
hstastna: It is better not to use constants inside `t` function. I already discussed that within the team.

yaacov: we use constants in `t` all the time
for example `t('Other')`,
In this case you want to translate the flavor, that is prop,
you have several ways of solving this issue:
you can have a dict that return the already translated values:
``` ts
const workloadProfiles = {
  desktop: t('Desktop'),
  server:t('Server'),
  highperformance: t('High performance'),
};
```
or a dict hodding the translation keys:
```ts
// t('plugin__kubevirt-plugin~Desktop')
// t('plugin__kubevirt-plugin~Server')
// t('plugin__kubevirt-plugin~High performance')
// t('plugin__kubevirt-plugin~Other')
const workloadProfiles = {
  desktop: 'Desktop',
  server: 'Server',
  highperformance: 'High performance',
};
```
and then translate them when using them, in this case it nicer to use the tranlation keys, and translate only when rendering,
a god reason to use the translation keys, is that we get the label using a method `getWorkloadLabes` such a method is getting simpler if you don't need to do translations inside the method.
for example:
```ts
const getWorkloadLabes = (workload, t) =>
   t(workloadProfiles(workload) || 'Other');
```
vs
```ts
const getWorkloadLabes = (workload) =>
   workloadProfiles(workload) || 'Other';
``` 
a. if you use the keys instead of the translations, you do need to pass `t` arround,
b. someone calling this method may use it as a component in a bigger string and re-translate it, for example
```ts
t(`this is the workload: ${workload}`)
```

src/views/templates/list/hooks/useWorkloadProfile.ts
hstastna: And `WorkloadProfileKeys` type [already added](https://github.com/kubevirt-ui/kubevirt-plugin/pull/477/files#diff-7e0e1100c332e878944a89ad9bb14ab21a3a7622063d562320fe14e05cab4423R18).

src/views/templates/details/tabs/details/components/WorkloadProfile.tsx
hstastna: Updated the code to achieve [your suggestion](https://github.com/kubevirt-ui/kubevirt-plugin/pull/477/files#r889840996) and without creating any extra helper like `getWorkloadLabes` as suggested just above.

src/views/templates/details/tabs/details/components/WorkloadProfile.tsx
hstastna: Maybe, but works fine also with the `k8sUpdate`.

yaacov:
/retitle Bug 2094323: Make Template Workload profile editable

src/views/templates/details/tabs/details/components/WorkloadProfile.tsx
```diff
@@ -0,0 +1,74 @@
...
+        onClose={onClose}
+        onSubmit={updateWorkload}
+      />
+    ));
+
+  return (
+    <DescriptionListGroup>
+      <DescriptionListTerm>{t('Workload profile')}</DescriptionListTerm>
+      <DescriptionListDescription>
+        {workload}
```
yaacov, line 63:
```suggestion
        { t(workloadProfiles[workload] || 'Other')}
```
you can avoid using `t` as a method parameter ,outside a component, by sending the label key instead of the translated label, see https://github.com/kubevirt-ui/kubevirt-plugin/pull/477/files#r889840969

src/views/templates/list/hooks/useWorkloadProfile.ts
```diff
@@ -1,23 +1,21 @@
...
   highperformance: 'High performance',
 };
 
-const useWorkloadProfile = (template: V1Template): string => {
-  const { t } = useKubevirtTranslation();
+export type WorkloadProfileKeys = 'desktop' | 'server' | 'highperformance';
 
-  return t(workloadLabels[getTemplateWorkload(template)] || 'Other');
-};
+const useWorkloadProfile = (template: V1Template): string =>
```
yaacov, line 18: it's not a hook anymore, no need to start with a 'use...'

src/views/templates/details/tabs/details/components/WorkloadProfile.tsx
hstastna: When using _k8sPatch_, I need to know the exact path, where I want to change the data -  that can be a bit difficult to solve when I change something in the Template's VM object. Recently you were explaining to me that the VM object can be anywhere, not only/always at the first place of the _template.objects_ array. Using _k8sPatch_ I'd have to solve this somehow - I consider this as unnecessary, and so `k8sUpdate` is much easier and straightforward to use here.

yaacov: yes, you will need to first find the index of the vm object and only then use patch

src/views/templates/utils/constants.ts
```diff
@@ -27,3 +27,16 @@ export type SOURCE_OPTIONS_IDS =
   | typeof SOURCE_TYPES.uploadSource;
 
 export const SUPPORT_URL = 'https://access.redhat.com/articles/4234591';
+
+// t('Desktop')
+// t('Server')
+// t('High performance')
+// t('Other')
```
yaacov, line 34: you don't need `Other` because it's not a string in `workloadProfiles`

yaacov: you do need it in https://github.com/kubevirt-ui/kubevirt-plugin/pull/477/files#diff-c05bbb0bae554062323840c4381c34d6cc141eafbf6d0b2b406462c5f5a774b5R27 because it's the fallback value if translation key is missing in `workloadProfiles`, so maybe add a comment
``` ts
// Other is the default value in case translation key is missing in workloadProfiles
// t('Other')
```

src/views/templates/details/tabs/details/components/WorkloadProfile.tsx
hstastna: There are more issues to solve than only to find the index of the VM object...

src/views/templates/details/tabs/details/components/WorkloadProfile.tsx
```diff
@@ -0,0 +1,97 @@
...
+    // const updatedTemplate = produce<V1Template>(template, (templateDraft: V1Template) => {
+    //   const vm = getTemplateVirtualMachineObject(templateDraft);
+    //   ensurePath(vm, 'spec.template.metadata.annotations');
+
+    //   vm.spec.template.metadata.annotations[VM_WORKLOAD_ANNOTATION] = updatedWorkload;
+
+    //   return templateDraft;
+    // });
+
+    // return k8sUpdate({
```
vojtechszocs, line 40: @hstastna @yaacov Reading up on the docs, I think we could use `k8sUpdate` (replace a resource) when we have the full object and a potential optimistic lock failure (due to resource being modified between the read and write) is acceptable. If these conditions are true, it's easier to use `k8sUpdate` than `k8sPatch`.

Patching is more complex because we need to compute the patch meta object and IIRC there were some slight differences between regular resources vs. some specific sub resources. With multiple `k8sPatch` calls for the given resource and field combo, the last write will win. So if we're really concerned about potential optimistic lock failures and we don't have the full object, we should use `k8sPatch` method.

src/views/templates/utils/constants.ts
yaacov: you should translate "Other" you can do it here 
or near
https://github.com/kubevirt-ui/kubevirt-plugin/pull/477/files#diff-c05bbb0bae554062323840c4381c34d6cc141eafbf6d0b2b406462c5f5a774b5R27
where it is defined

I like to translating it here, because it's part of the strings describing a workload, `['Desk ... , 'Other']`, it also makes sense to tranlate it near https://github.com/kubevirt-ui/kubevirt-plugin/pull/477/files#diff-c05bbb0bae554062323840c4381c34d6cc141eafbf6d0b2b406462c5f5a774b5R27 because it's defined there.

src/views/templates/details/tabs/details/components/WorkloadProfile.tsx
yaacov: in this case `k8sPatch` is straight forward to implement, IMHO it's worth the extra time to implement here.

for example the path can be:
```ts 
const workloadPath = `/objects/${vmObjectIndex}/spec/template/metadata/annotations/vm.kubevirt.io~1workload`;
```
and the patch comment can also be small:
```ts
 return k8sPatch({
      model: TemplateModel,
      resource: template,
      data: [
        {
          op: hasWorkload ? 'replace' : 'add',
          path: workloadPath,
          value: updatedWorkload,
        },
      ],
    });
```

PR #478, opened 2022-05-25
Bug 2081182: change to use node IP in ssh command

avivtur:
## 📝 Description

changing the hook that creates the string for SSH access command to follow doc [1] and using the VM's node IP address

[1] https://docs.openshift.com/container-platform/4.10/virt/virtual_machines/virt-accessing-vm-consoles.html#virt-accessing-vm-yaml-ssh_virt-accessing-vm-consoles

### after:
ssh-command-before

### DIFF
--- a/src/utils/components/UserCredentials/useSSHCommand.ts
+++ b/src/utils/components/UserCredentials/useSSHCommand.ts
@@ -1,9 +1,7 @@
-import { InfrastructureModel, modelToGroupVersionKind } from '@kubevirt-ui/kubevirt-api/console';
 import { IoK8sApiCoreV1Service } from '@kubevirt-ui/kubevirt-api/kubernetes';
 import { V1VirtualMachineInstance } from '@kubevirt-ui/kubevirt-api/kubevirt';
 import { getCloudInitCredentials } from '@kubevirt-utils/resources/vmi';
 import { getSSHNodePort } from '@kubevirt-utils/utils/utils';
-import { useK8sWatchResource } from '@openshift-console/dynamic-plugin-sdk';
 
 export type useSSHCommandResult = {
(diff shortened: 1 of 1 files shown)


src/utils/components/UserCredentials/useSSHCommand.ts
```diff
@@ -34,7 +34,7 @@ const useSSHCommand = (
 
   if (user) command += `${user}@`;
 
-  command += `${apiHostname || consoleHostname} -p ${sshServicePort}`;
+  command += `${nodeInternalIPAddress || consoleHostname} -p ${sshServicePort}`;
```
yaacov, line 37:
```suggestion
  command += `${consoleHostname} -p ${sshServicePort}`;
```

the bug is to use "console host name" instead of "api name"
the internal ip is actually a bug, because a user can't access internal ip addresses

avivtur: changed, all redundant code was removed as well

src/utils/components/UserCredentials/useSSHCommand.ts
```diff
@@ -15,26 +13,12 @@ const useSSHCommand = (
...
 
   const { user } = getCloudInitCredentials(vmi);
   const sshServicePort = getSSHNodePort(sshService);
 
-  let command = 'ssh ';
-
-  if (user) command += `${user}@`;
-
-  command += `${apiHostname || consoleHostname} -p ${sshServicePort}`;
+  const command = `ssh ${user && `${user}@`}${consoleHostname} -p ${sshServicePort}`;
```
upalatucci, line 21: if user is undefined, it will return `ssh undefinedconsoleHostname`

PR #480, opened 2022-05-25
Changes mistakes

DanaOrr:

## 📝 Description
Changes VirtualMachineInstance to VMI.


src/views/clusteroverview/overview/components/running-vms-per-template-card/RunningVMsPerTemplateCard.tsx
```diff
@@ -50,7 +50,7 @@ const RunningVMsPerTemplateCard = () => {
           right: 20,
           top: 20,
         }}
-        subTitle={t('VirtualMachines')}
+        subTitle={t('VMs')}
```
yaacov, line 53: here it actually `VirtualMachines` right ?

src/views/virtualmachinesinstance/details/VirtualMachinesInstancePageHeader.tsx
```diff
@@ -23,9 +23,7 @@ const VirtualMachinesInstancePageHeader: React.FC<VirtualMachinesInstancePageHea
       <VirtualMachineInstanceBreadcrumb />
       <span className={'co-m-pane__heading'}>
         <h1 className="co-resource-item__resource-name">
-          <span className={`co-m-resource-icon co-m-resource-icon--lg`}>
-            {t('VirtualMachineInstance')}
-          </span>
+          <span className={`co-m-resource-icon co-m-resource-icon--lg`}>{t('VMI')}</span>
```
yaacov, line 26:
```suggestion
          <span className={`co-m-resource-icon co-m-resource-icon--lg`}>{t('VirtualMachineInstances')}</span>
```
?

DanaOrr: Before:
image
After:
image
Also here, this is the change, What do you think?

yaacov: the kind label is not translatable, so it should just be 'VMI'
oyyy, it's also wrong here ...
https://github.com/kubevirt-ui/kubevirt-plugin/blob/main/src/views/virtualmachines/details/VirtualMachineNavPageTitle.tsx#L34

but it's more problematic, the kind label should not be hardcoded it should come from the object modal 
https://github.com/kubevirt-ui/kubevirt-api/blob/main/console/models/VirtualMachineInstanceModel.ts#L11

src/views/clusteroverview/overview/components/running-vms-per-template-card/RunningVMsPerTemplateCard.tsx
yaacov: ok,
makes sense to me too

PR #484, opened 2022-05-26
Bug 2089963: Add VM os label and annotations from template

upalatucci:
## 📝 Description

>We have to take this info from the template

## 🎥 Demo

Without guestagent

Screenshot from 2022-05-26 11-38-52

With guestagent

Screenshot from 2022-05-26 11-38-41


yaacov:
/hold

please use https://kubevirt.io/api-reference/master/definitions.html#_v1_virtualmachineinstanceguestagentinfo
and add a comment about guest agent needed if data is not available

PR #493, opened 2022-05-29
Bug 2090054: Services “kubernetes" and "openshift" should be hidden

glekner:
## 📝 Description

> Add a brief description

## 🎥 Demo


yaacov:
@glekner 
the method that filter services is here:
https://github.com/kubevirt-ui/kubevirt-plugin/blob/f6b67164a6a59871d61f9ad4b1a249a412e332d0/src/utils/resources/vmi/utils/services.ts
filter fail when selectors list is empty :-( 
because {} is truthy

PR #496, opened 2022-05-29
Bug 2091406: Fixing the namespace label when customize new vm

DanaOrr:

## 📝 Description
When creating a vm using customize wizard, default is referring to namespace of newly created vm instead of template namespace.
This fix will use the template namespace for openshift.

## 🎥 Demo
Before:
image

After:
image


yaacov:
/retitle Bug 2091406: Fixing the namespace label when customize new vm

PR #497, opened 2022-05-30
jest and cypress conflict

gouyang:
demo about jest and  cypress conflict


gouyang:
the test failed because of Compilation Error:
```
Error: Webpack Compilation Error
[tsl] ERROR in /go/src/github.com/kubevirt-ui/kubevirt-plugin/cypress/tests/overview.spec.ts(12,20)
      TS2551: Property 'toEqual' does not exist on type 'Assertion'. Did you mean 'equal'?
```

@yaacov @glekner Could you take a look at this?
There is a thread about this assertion error: https://stackoverflow.com/questions/58999086/cypress-causing-type-errors-in-jest-assertions
I've tried the workarounds there but no luck, please help.

gouyang:
@glekner thanks, I'm just moving the tests from 4.10 to 4.11 and see this problem. So I will use the cypress `expect`  to solve the problem.

PR #507, opened 2022-05-31
Bug 2090042: Unify edit button on VM details tab

metalice:

## 📝 Description

Unify edit buttons on VM details tab

## 🎥 Demo

After:
image

Before:
image


hstastna (commented):
Hi Matan, this PR looks interesting to me as I am not aware of this design change, but the code looks ok to me :)

And what about moving the Pencil icon to the same line for _CPU | Memory_ field? Like [this](https://user-images.githubusercontent.com/13417815/166683684-46475cdc-98c3-4431-9fd8-5b1aa3363be6.png). IMO the Pencil icon shouldn't be placed on the separate line under the values ;)

hstastna:
I've checked this with @yfrimanm, and the correct design is what we actually have - the "Before" picture. I suggest to leave those Pencil icons as they are + to fix the position of the icon for _CPU | Memory_ field as I've suggested in my previous comment. Thank you! ;)

metalice:
Great thanks @hstastna  !!

PR #510, opened 2022-05-31
Bug 2090048: Boot mode should be editable while VM is running

avivtur:
## 📝 Description

Enabling boot mode edit when VM is in `Running` status

### before:
bootmode_editable_before

### after:

https://user-images.githubusercontent.com/67270715/171174907-500d63c0-9d3e-44f4-bdf0-00547b27f662.mp4


yaacov:
/hold

you also need to show the "restart needed" thing that we show when we edit something while running :-(

PR #523, opened 2022-05-31
Bug 2091945: delete a template should lead to templates list

avivtur:
## 📝 Description

when we are at Template's details page and we delete the template, we should be led to template's list

### before:

https://user-images.githubusercontent.com/67270715/171168625-ffb2efec-02f8-4b06-8ba6-3e74e23319ca.mp4

### after:

https://user-images.githubusercontent.com/67270715/171168669-3d157cc8-bf58-4141-95d1-4cba1dcdb441.mp4


hstastna (commented):
Please, check my comment below. Otherwise it's perfect! Thank you, Aviv, for fixing the bug I've created when implementing a template deletion 😅 👼🏽 👍🏽

PR #530, opened 2022-06-01
Bug 2089979: adjust CPU|Memory pencil icon location

avivtur:
## 📝 Description

using span instead of div

### before:
pencil-wrong-location

### after:
pencil-wrong-location-after


hstastna (approved):
Awesome! This PR is fixing the issue I saw just yesterday and wanted to fix it. Thanks, Aviv! 😊 👍🏽
 
Just a note that we are using span instead of div element in this PR, not the opposite ;)

yaacov:
/approve
:tada:

PR #535, opened 2022-06-01
Bug 2091758: Rename OS to Operating system in filtering

hstastna:
## 📝 Description
**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2091758

Change the "OS" string to correct "Operating system", in the VM Template list filtering.
Change it the same way also in the Catalog and VirtualMachine list filtering, for consistency reasons.

## 🎥 Demo
**Before:**
os_before3
os_before1
os_before2
**After:**
os_after3
os_after1
os_after2


hstastna:

PR #544, opened 2022-06-02
Bug 2092306: VM is stucking with WaitingForVolumeBinding

glekner:
## 📝 Description

Bug 2092306: VM is stucking with WaitingForVolumeBinding
## 🎥 Demo

Screen Shot 2022-06-02 at 11 50 57


src/views/catalog/customize/cd.ts
```diff
@@ -32,10 +28,15 @@ export const addInstallationCDRom = (
       },
     };
   } else if (cdSource.source?.http || cdSource?.source?.pvc) {
```
yaacov, line 30: @upalatucci hi, in the case of PVC do we need a datavolumetemplate entry ?

PR #553, opened 2022-06-03
[WIP] Bug 2092796: Display CPU | Memory field consistently in the UI

hstastna:
## 📝 Description
**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2092796

Display _CPU | Memory_ field in the template card same as in the template drawer.
Display it the same also in other places like VM Templates list/_Details_ tab, VirtualMachines _Overview_/_Details_ tab.
Display the vertical divider line in the grey color.
Also display the values after "CPU" or "Memory" string and the units after the CPU value.

## 🎥 Demo
**Before:**
cpu-before

**After:**
cpu-after

---

TODO:
- [ ] refactor/unify the code, use one component in all the places
- [ ] check displaying the units with the UX: which ones to use? Gi vs. Gi**B**

### DIFF
--- a/locales/en/plugin__kubevirt-plugin.json
+++ b/locales/en/plugin__kubevirt-plugin.json
@@ -20,7 +20,6 @@
   "{{count}} Tolerations rules": "{{count}} Tolerations rules",
   "{{count}} Tolerations rules_plural": "{{count}} Tolerations rules",
   "{{cpu}} CPU | {{memory}} Memory": "{{cpu}} CPU | {{memory}} Memory",
-  "{{cpuCount}} CPU | {{memory}} Memory": "{{cpuCount}} CPU | {{memory}} Memory",
   "{{dsLabel}} name": "{{dsLabel}} name",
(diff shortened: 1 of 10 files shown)


yaacov:
/hold

@hstastna hi,
https://bugzilla.redhat.com/show_bug.cgi?id=2092796 is only about the wizard card, 
how to fix the cpu/mem in other places is still under discussion

hstastna:
@yaacov This PR is WIP anyway, no need to put on hold this PR.
Of course I discussed fixing this also in other places, and there was clear 'yes, of course'. There was no any discussion about if to make this consistently in the UI. This is pretty straightforward. Also there are no any obstacles or reasons why not to implement it consistently across the UI. I was also told that making the UI consistent should even be automatic from our developers' side, when fixing such bugs in the UI. It is very ineffective (for everybody, not just QE) to open bug for each and every place that needs the same fix.

hstastna:
Closing the PR, as I'm not gonna work on this anymore. If needed, a new PR with the correct fixes will be opened eventually.

PR #557, opened 2022-06-06
Bug 2093691: Add left padding to VM create drawer content

hstastna:
## 📝 Description
**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2093691

When creating a VM, the drawer left padding in upper part disappeared. Fix the missing padding to display the content correctly.

## 🎥 Screenshots
**Before:**
padding_before
**After:**
padding_after

---

_Note:_
When investigating the problem, the relevant component had the following variable set:
`--pf-c-modal-box__body--PaddingLeft: 0 !important;`
Because of this, we cannot use the Patternfly `--pf-c-modal-box__body--PaddingLeft`, nor even `--pf-c-modal-box--PaddingLeft` to achieve the expected result, because this is actually not working correctly, doesn't set the expected padding when used.

Related PF doc:
https://www.patternfly.org/2020.06/documentation/core/components/modalbox


hstastna:
@glekner @avivtur @metalice @pcbailey @vojtechszocs please review

PR #559, opened 2022-06-07
Bug 2086294: [Dark mode] - The number of vms per template in the donuts is visible in overview page

DanaOrr:

## 📝 Description
Installing the latest version of react-charts made the number inside the donuts visible in virtual machines per template card on overview page.

## 🎥 Demo
Before:
image

After:
image


yaacov:
@DanaOrr it looks like the new version breaks things ?
can you look at the unit test fails ? maybe they are easy to fix ?

yaacov:
cc:// @pcbailey

hstastna:
I've tested this PR and it fixes the bug. Looks good to me 👍🏽

PR #564, opened 2022-06-07
Bug 2094405: missing env resource from disks list

avivtur:
## 📝 Description

when adding a resource from the environment tab when VM is running the resource is not shown on the disks list

### before:
env-resource-missing

### after:
env-resource-missing-after

### DIFF
--- a/src/utils/components/DiskModal/utils/helpers.ts
+++ b/src/utils/components/DiskModal/utils/helpers.ts
@@ -11,6 +11,7 @@ import {
   V1Disk,
   V1RemoveVolumeOptions,
   V1VirtualMachine,
+  V1VirtualMachineInstance,
   V1Volume,
 } from '@kubevirt-ui/kubevirt-api/kubevirt';
 import { buildOwnerReference } from '@kubevirt-utils/resources/shared';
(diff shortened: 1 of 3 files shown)


src/utils/components/DiskModal/utils/helpers.ts
```diff
@@ -226,3 +228,23 @@ export const getRemoveHotplugPromise = (vm: V1VirtualMachine, diskName: string)
...
+export const getDisksFromVMAndVMI = (
+  vm: V1VirtualMachine,
+  vmi: V1VirtualMachineInstance,
+): V1Disk[] => {
+  const vmiDisksNames = (vmi?.spec?.domain?.devices?.disks || [])?.map((disk) => disk?.name);
+  const missingDisksFromVMI = getDisks(vm)?.filter((disk) => !vmiDisksNames?.includes(disk?.name));
+  return [...(vmi?.spec?.domain?.devices?.disks || []), ...(missingDisksFromVMI || [])];
+};
+
+export const getVolumesFromVMAndVMI = (
```
yaacov, line 241: `get<something>FromVMAndVMI` is a slippery slope, looks good here but next PR it will be `get<X>FromVMLikeObjcts` :-)
Can we do a `getMissingDisksFromVMI` and add them at  the bottom of the disks list ?
it will be the same thing, that is what you do in line 238, add the missing to the end of the list, but we will avoid the need for a `fromVMandVMI` method

avivtur: changed, can you take another look?

PR #570, opened 2022-06-08
Template Action: Add Edit boot source reference

upalatucci:
## 📝 Description

'Edit boot source' -> go to disks tab
'Edit boot source reference' (old 'Edit boot source')

## 🎥 Demo

Disabled on red hat templates
Screenshot from 2022-06-08 10-01-44

Screenshot from 2022-06-08 10-01-01

EditBOotSource


src/views/templates/actions/hooks/useVirtualMachineTemplatesActions.tsx
```diff
@@ -81,9 +81,19 @@ const useVirtualMachineTemplatesActions: useVirtualMachineTemplatesActionsProps
...
+        history.push(
+          `/k8s/ns/${template.metadata.namespace}/templates/${template.metadata.name}/disks`,
+        ),
+    },
+    {
+      id: 'edit-boot-source-ref',
       label: (
         <>
-          {t('Edit boot source')} {editableBootSource === null && <Loading />}
+          {t('Edit boot source reference')} {editableBootSource === null && <Loading />}
```
metalice, line 96: i think editableBootSource === null can be replaced with !editableBootSource, please check

upalatucci: Thanks, Matan :-). `editableBootSource` starts with null (not defined) and after a request, it can become true or false. Maybe I can use the `??` operator.
Currently, in another pr I'm going to edit this logic so let me know if this logic is a little tricky

PR #578, opened 2022-06-08
Bug 2094495: Descheduler modal has same text in two lines

avivtur:
## 📝 Description

checkbox label and description are the same in the descheduler modal.

### before:
descheduler-text-before

### after:
descheduler-text-after


apinnick:
LGTM

PR #591, opened 2022-06-09
Bug 2095129: Display disk size units in Clone VM modal

hstastna:
## 📝 Description
**Fixes**:
https://bugzilla.redhat.com/show_bug.cgi?id=2095129

Display the units instead of 'undefined', in the _Clone VirtualMachine_ modal, for the _Disks_ field in the _Configuration_ section.

Related PR: https://github.com/kubevirt-ui/kubevirt-plugin/pull/558

## 🎥 Demo
**Before:**
clone-before
**After:**
clone-after

### DIFF
--- a/src/utils/resources/vm/utils/disk/size.ts
+++ b/src/utils/resources/vm/utils/disk/size.ts
@@ -30,7 +30,7 @@ export const formatBytes = (rawSize: string, unit?: string): string => {
   }
   const size = hasNumber(rawSize);
   const sizeUnit = hasSizeUnit(rawSize) || unit;
-  const sizeUnits = ['B', 'KiB', 'MiB', 'GiB', 'TiB'];
+  const sizeUnits = ['', 'Ki', 'Mi', 'Gi', 'Ti'];
   let unitIndex = (sizeUnit && sizeUnits.findIndex((sUnit) => sUnit === sizeUnit)) || 0;
   let convertedSize = size;
   while (convertedSize >= 1024) {


hstastna:

yaacov:
/approve
@upalatucci please review

upalatucci:
For me, it's okay to use Mi/Gi/Ti in `formatBytes`. 
The function's goal would be to give an approximation and select the best prefix (doesn't matter the unit).

But this change will bring back Gi quantities in the VM disk and filesystem tables. 
Can you add the B unit there? 

Screenshot from 2022-06-10 09-40-02

src/utils/resources/vm/utils/disk/size.ts
```diff
@@ -30,7 +30,7 @@ export const formatBytes = (rawSize: string, unit?: string): string => {
   }
   const size = hasNumber(rawSize);
   const sizeUnit = hasSizeUnit(rawSize) || unit;
-  const sizeUnits = ['B', 'KiB', 'MiB', 'GiB', 'TiB'];
+  const sizeUnits = ['B', 'Ki', 'Mi', 'Gi', 'Ti'];
```
upalatucci, line 33:
```
  const sizeUnits = ['', 'Ki', 'Mi', 'Gi', 'Ti'];
```

upalatucci, line 33:
```suggestion
  const sizeUnits = ['', 'Ki', 'Mi', 'Gi', 'Ti'];
```

hstastna: `formatBytes` is implemented to use raw data, which should be without "B" suffix for the units, so it must have not been working nowhere (incl. _Disks_ tab), where used, if used properly. I'll check what's going on in the _Disks_ tab anyway, but adding "B" suffix to the units where needed, should be done in a followup PR, as this PR has nothing to do with it, and as the changes can affect more places than mentioned in this PR.

@upalatucci can you please, explain, why to remove 'B' from the `sizeUnits` array? I think we need it there.

upalatucci: I think we should remove 'B' as it represents the unit byte and is not a prefix like Ki, Mi, Gi.

hstastna: You are right that "B" is not a prefix, but can't for example "1000B" come to this function as a raw size?

upalatucci: No. According to [k8s](https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/), you'll receive only '1000' without the B.

hstastna: > I think we should remove 'B' as it represents the unit byte and is not a prefix like Ki, Mi, Gi.

Ok, I will remove  'B' from that array, but not because it is not a prefix, but because we don't need it if we receive a raw size in bytes - a pure value without any unit, in `sizeUnits`. I really hope we won't receive anything like "30B" or so, in `sizeUnits`, because in that case it would not work.

hstastna: > But this change will bring back Gi quantities in the VM disk and filesystem tables.
> Can you add the B unit there?

Yes, I verified, that this PR will affect the units displayed in the VM _Disks_ tab - it is because `formatBytes` is used there, but this function's role is not displaying GiB instead of Gi (for example), but converting the raw value in bytes to something more user friendly. It was not fixed correctly, in one of the recent PRs, when trying to fix missing "B" when displaying units.

Anyway, fixing displaying units with "B" needs to be fixed separately. We need to add it when needed, differently than putting it to the `sizeUnits` array, because, as we can see, it causes bugs in other places. I have a solution but let's do it in a followup PR because this is not related to the BZ I am fixing in this PR. I'd like to use a nice [function](https://github.com/kubevirt-ui/kubevirt-plugin/pull/70/files#diff-4ebbb7ad9ddc308e1be2d3506502b5bed3bc53ed698243e291fbd840ea76838aR13) for adding "B", a function that also nicely checks if there is already "B" as part of the units, and if not, then adds it

upalatucci (approved):
/lgtm

PR #598, opened 2022-06-14
Add catalog nav item id

yaacov:
Add catalog nav item id


yaacov:
@upalatucci please review

cc:// @nunnatsa

PR #600, opened 2022-06-14
Bug 2096620: Update boot source reference modal title

yaacov:
Signed-off-by: yzamir <yzamir@redhat.com>

## 📝 Description

Update boot source reference modal title

a. The action "Edit boot source reference" now opens a modal with the same name
b. Opening a new tab for upload is deprecated and should be replaced with inline upload, removed the deprecated options to upload using a different tab.

## 🎥 Demo

Before:
boot-reference-modal-before

After:
boot-reference-modal


yaacov:
@upalatucci please review

PR #614, opened 2022-06-15
Bug 2095953: Moving consoles buttons

metalice:

## 📝 Description

Moving consoles buttons

## 🎥 Demo

After:
image

Before:
image


yaacov:
@metalice the bug description say:
"""
A. The VNC console dropdown is currently too wide and should be narrower.
B. The Send key dropdown should move to be next to the VNC dropdown, on the right.
C. Send Key should be Send key (lowercase in key)
Version-Release number of selected component (if applicable):
"""

yaacov:
houl

metalice:
> you should not move the "disconnect" button

cool, moved only send keys

PR #621, opened 2022-06-15
Bug 2095573: Add label for env disk

avivtur:
## 📝 Description

Adding a label to identify environment disks to DiskRow component

### Before:
env-disk-label-before

### After:
env-disk-label
env-disk-label-wizard
env-disk-label-template


upalatucci (approved):
/lgtm

PR #622, opened 2022-06-15
Bug 2092787: No operators links for no-priv user

upalatucci:
## 🎥 Demo

**Before**
Screenshot from 2022-06-15 17-26-25

**After** 
Only for no priv users
With external links to the relative operator documentation 

image

### DIFF
--- a/src/views/clusteroverview/overview/components/getting-started-card/related-operators-section/RelatedOperatorsSection.tsx
+++ b/src/views/clusteroverview/overview/components/getting-started-card/related-operators-section/RelatedOperatorsSection.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 
+import { useIsAdmin } from '@kubevirt-utils/hooks/useIsAdmin';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
 
 import GettingStartedSectionContents from '../utils/getting-started-content/GettingStartedSectionContents';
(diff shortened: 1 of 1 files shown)


avivtur:
the `Learn more about Operators` link is also broken I'd suggest checking if there is another url to use,
if not, we will probably need to hide that link or that entire section hidden as well

src/views/clusteroverview/overview/components/getting-started-card/related-operators-section/RelatedOperatorsSection.tsx
```diff
@@ -56,7 +58,7 @@ const RelatedOperatorsSection: React.FC = () => {
       title={t('Related operators')}
       titleColor={'var(--co-global--palette--orange-400)'}
       description={t('Ease operational complexity with virtualization by using Operators.')}
-      links={links}
+      links={isAdmin ? links : []}
```
yaacov, line 61:
```suggestion
      links={operatorDescriptions}
```
this are not the actual links, they include the description and the links :-)
the descritions should stay, only the links should be removed, something like:
``` ts
{
      id: 'openshift-data-foundation',
      title: t('OpenShift Data Foundation'),
      href: isAdmin ? '/operatorhub/all-namespaces?keyword=OCS' : null,
    },
```

yaacov: ( and check that `GettingStartedSectionContents` can handle null links gracefully )

yaacov: yes, if no `href` we can remove the '->' and the link behaviour

PR #629, opened 2022-06-16
Bug 2090042: Move pencil icon for BootOrder

upalatucci:
## 🎥 Demo

**Before**
image

**After**
image


yaacov:
should look like other places, for example "description" and "boot method"
a - no "Edit" text
b - if possible on the line of the thing being edited
     ( not sure if applicable in this case )

upalatucci:
I think in the case of Boot Order is not possible to have the pencil inline. 
If I remove the `Edit` label there will be only the pencil. 
The label `Labels` has also a popover on click so I can't do what we do with GPU devices 

Screenshot from 2022-06-16 14-38-52


Screenshot from 2022-06-16 14-38-31

PR #634, opened 2022-06-16
Bug 2092783: Hide the top consumers card for non-admins

pcbailey:
This PR hides the top consumers card in the cluster overview page for non-admins.

### Admin:
Selection_109

### Non-admin:
Selection_110


pcbailey:
@avivtur @glekner @hstastna @metalice @upalatucci @vojtechszocs @yaacov PTAL

PR #641, opened 2022-06-20
auto upload sholuld be auto import

yaacov:
## 📝 Description

auto upload should be auto import

## 🎥 Demo
Before:
import-upload-before

After:
import-upload-after


src/utils/resources/template/utils/constants.ts
```diff
@@ -96,7 +96,7 @@ export enum BOOT_SOURCE {
 export const BOOT_SOURCE_LABELS = {
   [BOOT_SOURCE.PVC]: 'PVC',
   [BOOT_SOURCE.DATA_SOURCE]: 'PVC',
-  [BOOT_SOURCE.DATA_SOURCE_AUTO_UPLOAD]: 'PVC (auto upload)',
+  [BOOT_SOURCE.DATA_SOURCE_AUTO_UPLOAD]: 'PVC (auto import)',
```
metalice, line 99: Consider changing key value AUTO_UPLOAD to AUTO_IMPORT

yaacov: nice +1

yaacov: @metalice updated, please re-review

PR #644, opened 2022-06-20
Bug 2092654: Add ssh auth key name with remove button

upalatucci:
## 📝 Description

Allow the user to select an existing secret for the vm

## 🎥 Demo

VM create a new Service then delete the key 
SSHModalPart1

attach an existing service to the VM (shared service between two vms) then detach the secret and add a secret not owned by any VM (it will allow edit of the secret)
SSHModalPart2


yaacov:
/wip
need a way to add secrets as well as remove them

PR #649, opened 2022-06-21
Bug 2094202: cloudinit username default

metalice:
## 📝 Description

Remove default user message, making field required.

## 🎥 Demo

disable-cloud


yaacov:
Things to do:
```
a - remove the help hint
b - make the username field required
c - in places the UI assume a username and it's missing, we will show some "Not available" or empty as needed
    ( do not guess it's "cloud-user", if missing )
```

PR #668, opened 2022-06-23
Revert "Bug 2097586: fix storage profile settings checkbox"

avivtur:
Reverts kubevirt-ui/kubevirt-plugin#652

reverting as it is not the right solution and causes more bugs


yaacov:
@avivtur note, this fix is for 4.12, we will need to back-port it to 4.11

avivtur:
closing in favor of https://github.com/kubevirt-ui/kubevirt-plugin/pull/670

PR #669, opened 2022-06-24
Expose qemu machine type read-only in VM Details tab

hstastna:
## 📝 Description

Display qemu machine type in the VM/Template _Details_ tab, the field is not editable. 

_Related doc:_
https://kubevirt.io/user-guide/virtual_machines/virtual_hardware/#machine-type

_Issue:_
https://issues.redhat.com/browse/CNV-12598

## 🎥 Demo
_Machine type_ field in the VM _Details_ tab:
machine1
_Machine type_ field in the VM Template's _Details_ tab:
machine2

### DIFF
--- a/locales/en/plugin__kubevirt-plugin.json
+++ b/locales/en/plugin__kubevirt-plugin.json
@@ -398,6 +398,7 @@
   "Local storage (LSO)": "Local storage (LSO)",
   "MAC address": "MAC address",
   "MAC Address": "MAC Address",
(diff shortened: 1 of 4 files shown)


yaacov:
/approve

a - should be near "boot mode" and "cpu / meme" (like cpu, mem and boot mode it describe the virtulization h/w)
     ( maybe even create a section for h/w section and put the devices and "boot order" too )
b - should also be in templates, wizard and vmis views

PR #676, opened 2022-06-27
Bug 2096780: Template Auth SSH Service

upalatucci:
## 📝 Description

Adding secret selection in template scripts tab

## 🎥 Demo
image

image
Screenshot from 2022-06-27 16-32-22


yaacov:
/lgtm

next steps:
a - show only secrets that have sshkey data
b - allow to create a new secret is tricky, you will need to solve:
     - secret can't be used on a different namespace (so you can't use it when creating a vm on different namespace)
     - secret can't be automatically deleted when template is deleted, because some vm may use it

PR #678, opened 2022-06-27
Use getTemplateVirtualMachineObject in template utils

hstastna:
## 📝 Description
This is a small refactoring PR related to:
using `getTemplateVirtualMachineObject` function to get the VM object where needed, instead of hardcoced `template.objects[0]`, in the template utils/selectors.

### DIFF
--- a/src/views/templates/utils/selectors.ts
+++ b/src/views/templates/utils/selectors.ts
@@ -1,29 +1,34 @@
 import { V1Template } from '@kubevirt-utils/models';
 import { getAnnotation } from '@kubevirt-utils/resources/shared';
 import { getLabel } from '@kubevirt-utils/resources/shared';
-import { WORKLOADS, WORKLOADS_LABELS } from '@kubevirt-utils/resources/template';
+import {
+  getTemplateVirtualMachineObject,
+  WORKLOADS_LABELS,
+} from '@kubevirt-utils/resources/template';
 import { VM_WORKLOAD_ANNOTATION } from '@kubevirt-utils/resources/vm/utils';
 
 import { ANNOTATIONS, LABELS } from './constants';
 
 export const getAffinity = (template: V1Template) =>
-  template?.objects[0]?.spec?.template?.spec?.affinity || {};
+  getTemplateVirtualMachineObject(template)?.spec?.template?.spec?.affinity || {};
 
 export const getEvictionStrategy = (template: V1Template): string =>
-  template?.objects[0]?.spec?.template?.spec?.evictionStrategy;
+  getTemplateVirtualMachineObject(template)?.spec?.template?.spec?.evictionStrategy;
 
 export const getNodeSelector = (template: V1Template) =>
-  template?.objects[0]?.spec?.template?.spec?.nodeSelector;
+  getTemplateVirtualMachineObject(template)?.spec?.template?.spec?.nodeSelector;
 
 export const getTemplateProviderName = (template: V1Template): string =>
   getAnnotation(template, ANNOTATIONS.providerName, null) ||
   getAnnotation(template, ANNOTATIONS.providerDisplayName, null);
 
-export const getTemplateWorkload = (template: V1Template): WORKLOADS =>
-  template?.objects[0]?.spec?.template?.metadata?.annotations?.[VM_WORKLOAD_ANNOTATION];
+export const getTemplateWorkload = (template: V1Template): string =>
+  getTemplateVirtualMachineObject(template)?.spec?.template?.metadata?.annotations?.[
+    VM_WORKLOAD_ANNOTATION
+  ];
 
 export const getTolerations = (template: V1Template) =>
-  template?.objects[0]?.spec?.template?.spec?.tolerations;
+  getTemplateVirtualMachineObject(template)?.spec?.template?.spec?.tolerations;
 
 // t('Other')
 export const getWorkloadProfile = (template: V1Template): string =>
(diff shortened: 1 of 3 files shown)


yaacov:
/lgtm

nice !

src/views/templates/utils/selectors.ts
```diff
@@ -1,29 +1,34 @@
...
-  template?.objects[0]?.spec?.template?.spec?.nodeSelector;
+  getTemplateVirtualMachineObject(template)?.spec?.template?.spec?.nodeSelector;
 
 export const getTemplateProviderName = (template: V1Template): string =>
   getAnnotation(template, ANNOTATIONS.providerName, null) ||
   getAnnotation(template, ANNOTATIONS.providerDisplayName, null);
 
-export const getTemplateWorkload = (template: V1Template): WORKLOADS =>
-  template?.objects[0]?.spec?.template?.metadata?.annotations?.[VM_WORKLOAD_ANNOTATION];
+export const getTemplateWorkload = (template: V1Template): string =>
```
yaacov, line 25:
```suggestion
export const getTemplateWorkload = (template: V1Template): WORKLOADS =>
```
you should not change the signature of a common lib method
( if you do, you need to go over the code and make sure it's ok )

hstastna: I changed that because it didn't work, compilation was failing which is obvious if you think deeper about the change we did here. Tried many things and wasn't able to use the better type. But I haven't tried the force cast yet...

src/views/templates/utils/selectors.ts
```diff
@@ -1,29 +1,35 @@
...
 
 export const getTemplateProviderName = (template: V1Template): string =>
   getAnnotation(template, ANNOTATIONS.providerName, null) ||
   getAnnotation(template, ANNOTATIONS.providerDisplayName, null);
 
 export const getTemplateWorkload = (template: V1Template): WORKLOADS =>
-  template?.objects[0]?.spec?.template?.metadata?.annotations?.[VM_WORKLOAD_ANNOTATION];
+  getTemplateVirtualMachineObject(template)?.spec?.template?.metadata?.annotations?.[
+    VM_WORKLOAD_ANNOTATION
+  ] as WORKLOADS;
```
yaacov, line 29:
```suggestion
  {
  const workload = getTemplateVirtualMachineObject(template)?.spec?.template?.metadata?.annotations?.[
    VM_WORKLOAD_ANNOTATION
  ];
  return workload in WORKLOADS ? workload as WORKLOADS : null;
  }
```

hstastna: I'd prefer the actual version, as it is shorter and self explanatory enough. Returns _undefined_ if other workload profile than one of the three values present, and then "Other" is displayed in this case.

yaacov: > Returns `undefined` if other workload

can you check, `as` does not check the value is actually a workload [1].

another option to force `WORKLOADS` is to do
``` ts
return WORKLOADS[workload as WORKLOADS]
```

[1] https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#type-assertions

yaacov: we currently do 'Other', which is fine ...
do you think showing the verbatim value from the yaml makes more sense ?

hstastna: Actually, we show 'Other' in the details, and in the modal we display the exact value from the yaml which makes sense to me. But with the change you are suggesting earlier above, it won't be possible anymore. And so now it makes more sense to me to return any `string`, not anything more exact like any of `WORKLOADS`. IMO, the check if the value is one of WORKLOADS, becomes important later in the code.

yaacov: > What do you think about it?

I'm ok with both, i like returning `string` better, but ok with `WORKLOADS` too

hstastna: So I suggest to leave this as it is now.

yaacov: > So I suggest to leave this as it is now

now the signature say `WORKLOADS` and the return value is `string` that is not nice. how do you suggest fixing it ?

yaacov: > Why is it "not nice"?

it's not correct, it will create a variable with type `A` but the value will be of type `B`, this is not correct,
types are used to help you assert that the value is of a type, if you use `as` without checking the value
you break the typing mechanism. 

for example:
``` ts
enum WORKLOADS {
  desktop = 'desktop',
  server = 'server',
  highperformance = 'highperformance',
}

const a = 'hi'
const b = 'desktop'

const returnsWorlads = (input: string): WORKLOADS => input as WORKLOADS;

const aWorkload: WORKLOADS = returnsWorlads(a);
// do something that assume aWorkload is of type WORKLOADS
// and it may crash or not depending on the input
console.log( returnsWorlads(aWorkload) ) // hi

const bWorkload: WORKLOADS = returnsWorlads(b);
// do something that assume aWorkload is of type WORKLOADS
// and it may crash or not depending on the input
console.log( returnsWorlads(bWorkload) ) // desktop
```

hstastna: Yes it works like that and IMO this is the behavior we want here now. I don't see here any conflict with what I've written earlier.
Btw, originally, at the beginning of this conversation, there was `string` return type, but you didn't like it...

hstastna: The tests failed because I forgot to change type to string in some related places, but you had a problem with not using WORKLOADS type, originally.

PR #683, opened 2022-06-29
Bug 2072797: Add duration to top consumers card

pcbailey:
This PR adds the option to select a duration to the top consumers card.

### Screenshots

Selection_114

Selection_115


pcbailey:
@avivtur @metalice @glekner @vojtechszocs @upalatucci @hstastna @yaacov PTAL

src/views/clusteroverview/overview/components/top-consumers-card/utils/DurationDropdown.tsx
```diff
@@ -0,0 +1,47 @@
...
+      variant={SelectVariant.single}
+    >
+      {DurationOption.getAll().map((durationOption) => {
+        const dropdownLabel = durationOption.getDropdownLabel();
+        const durationValue = durationOption.getValue();
+
+        return (
+          <SelectOption
+            data-test-id={`${durationValue}`}
+            key={`${durationValue}`}
```
avivtur, line 36: can these two lines be replaced with `prop={propVal}`?

src/views/clusteroverview/overview/components/top-consumers-card/utils/DurationDropdown.tsx
```diff
@@ -0,0 +1,47 @@
+import * as React from 'react';
+
+import { Select, SelectOption, SelectVariant } from '@patternfly/react-core';
+
+import DurationOption from './DurationOption';
+
+export type DurationDropdownProps = {
+  selectedDuration: any;
+  selectHandler: any;
```
avivtur, line 9: can you please change the type from `any`?

src/views/clusteroverview/overview/components/top-consumers-card/utils/DurationOption.ts
```diff
@@ -0,0 +1,53 @@
+import DropdownEnum from '@kubevirt-utils/utils/dropdownEnum';
+import { ObjectEnum } from '@kubevirt-utils/utils/ObjectEnum';
+
+class DurationOption extends DropdownEnum<string> {
+  static readonly FIVE_MIN = new DurationOption('5m', {
+    // t('5 minutes')
```
avivtur, line 6: I think this should be `// t('plugin__kubevirt-plugin~5 minutes')` and should be moved to src/utils/i18n.ts as well as the other similar comments in this file

avivtur (commented):
Hey Phillip, nice work :smile: 
I left you some comments

src/views/clusteroverview/overview/components/top-consumers-card/utils/DurationDropdown.tsx
pcbailey: Done. Good catch. I was using a combination of values and changed my mind, but obviously didn't clean up after myself.

src/views/clusteroverview/overview/components/top-consumers-card/utils/DurationDropdown.tsx
pcbailey: Done. Another good catch. I need to start adding TODOs when I mean to go back and fix something. My memory isn't good enough to just remember them, anymore. =)

src/views/clusteroverview/overview/components/top-consumers-card/utils/DurationOption.ts
pcbailey: Why? `yarn i18n` picked them up. Is there something I'm missing?

avivtur: AFAIK src/utils/i18n.ts is to make some file that contains all these kind of comments to the translation parser, not as a strict rule but more of a convention to us :)
and Im almost sure when you put it in a comment like you did, we need to prefix `plugin__kubevirt-plugin~` to the string we want to translate.
adding @glekner to this since he has more knowledge of this than me :)

pcbailey: I went ahead and changed it, but further explanation would be helpful so I'll understand the difference. =)

glekner: no need for prefix

src/utils/utils/dropdownEnum.ts
```diff
@@ -0,0 +1,18 @@
+import { ObjectEnum } from '@kubevirt-utils/utils/ObjectEnum';
```
glekner, line 1: @pcbailey I feel like using `ObjectEnum` is a pain from the past that we should try to avoid at all cost.
I do understand that you have portions of your feature relying on it, so IMO its ok to use it for porting features like the Virtualization dashboard. But I strongly recommend to move over to plain objects and hooks when implementing new features from now on. cc @avivtur @metalice @yaacov

PR #686, opened 2022-06-29
[release-4.11] Bug 2101430: Fix customize form crash

upalatucci:
**Cause**
No optional parameters. All required. (name parameter does no count)

**Solution**
Fix optional fields component


yaacov:
/retitle  [release-4.11] Bug 2102076: Fix customize form crash

PR #709, opened 2022-06-30
Bug 2101454: cant load PVCs for non-priv user

avivtur:
## 📝 Description

this fix includes:

- fetching the PVCs data from each namespace instead of fetching for 'all-namespaces'
- When a non priv user is trying to change a boot source reference for a DataSource that points to a PVC in a namespace where the user has no permissions to re-create the PVC, it will fail, so disabled this action in case the user can not re-create the PVC in the same namespace the older PVC was

### before:
edit-boot-source-ref-before-pvc-stuck

### after:

https://user-images.githubusercontent.com/67270715/176730575-2ce1bf0d-56b9-4e5b-8db5-4b81e97cf518.mp4


upalatucci (approved):
Great :-)

PR #715, opened 2022-07-01
Bug 2102256: Move Add buttons to left in VM/Template details tabs

hstastna:
## 📝 Description
**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2102256

Move the 'Add ...' buttons to the left side of the page where they were originally, in the VM _Network interfaces, Disks, Snapshots_ tabs, also in the VM Template _Network interfaces, Disks_ tabs.
Remove unnecessary `ListPageHeader` component used there for the empty title of those tabs, that styled the buttons incorrectly to the right.

## 🎥 Demo
**Before:**
VM _Network interfaces_ tab and the button on the left:
before1
before4
before6
**After:**
after1
after4
after6


hstastna:

hstastna:
> @hstastna can you check the Tabs of the Wizard aswell? I believe Disks/Network tabs has those buttons aswell

Good point, Gilad! Thnaks, I'll check that.

hstastna:
> @hstastna can you check the Tabs of the Wizard aswell? I believe Disks/Network tabs has those buttons aswell

PR updated 👍🏽

src/views/catalog/wizard/tabs/network/WizardNetworkTab.tsx
```diff
@@ -3,24 +3,23 @@ import * as React from 'react';
...
-  ListPageBody,
-  ListPageCreateButton,
-  ListPageHeader,
-} from '@openshift-console/dynamic-plugin-sdk';
+import { ListPageBody, ListPageCreateButton } from '@openshift-console/dynamic-plugin-sdk';
 
 import NetworkInterfaceList from './components/list/NetworkInterfaceList';
 import NetworkInterfaceModal from './components/modal/NetworkInterfaceModal';
 
+import 'src/utils/styles/ListPageCreateButton.scss';
```
glekner, line 11: we should import absolute css files like this, only relatively.
please copy the css to a new file in this directory and `import './file.css`

src/utils/styles/ListPageCreateButton.scss
```diff
@@ -0,0 +1,3 @@
+.list-page-create-button-margin {
+  margin: 1.5rem 0 1rem 0;
```
glekner, line 2: are we sure we need this specific css?
is there a problem with the sdk components? we should let the sdk team know.
we should use PF or SDK Components, not overwrite css for basic components like the create button.

src/views/catalog/wizard/tabs/network/WizardNetworkTab.tsx
hstastna: Is it optimal to copy the same file many times?

src/utils/styles/ListPageCreateButton.scss
hstastna: Yes there is a problem with sdk component `ListPageHeader`, but solving that problem would take some time.. I think we don't have much time for waiting for somebody else to solve it. And I am worried that the change was intentional, because the component is made for displaying header. And when you display some header, there is some string usually, and when there is some string and some button, you usually display the string in the left, and the button in the right.

PR #725, opened 2022-07-04
Bug 2100682: Delete only the appropriate HW device in the modal

hstastna:
## 📝 Description
**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2100682

Delete only the appropriate hardware device (GPU device/Host device)
from the devices list, when clicking on the icon to delete
on the right in the modal.

## 🎥 Demo
Having 2 devices with the same _Name_ leading to the buggy scenario:
delete-a
**Before:**
Deletion of the one of the devices caused deletion of all the devices with the same _Name_:
a_before
**After:**
Only the appropriate device deleted:
a_after


hstastna:

avivtur (approved):
Nice :)

PR #746, opened 2022-07-07
CNV-19275: Add virtctl to overview

pcbailey:
## 📝 Description

This PR adds a link to the overview header that displays a popup with information on downloading `virtctl` when clicked.

### Popup closed

Without the `Show getting started resources` button
Selection_121

With the `Show getting started resources` button
Selection_122

### Popup open

Without the `Show getting started resources` button
Selection_124

With the `Show getting started resources` button
Selection_125


pcbailey:
@avivtur @glekner @hstastna @metalice @upalatucci @vojtechszocs PTAL

src/views/clusteroverview/utils/VirtctlPopup.tsx
```diff
@@ -0,0 +1,69 @@
...
+  });
+
+  const tools = Array.isArray(cliTools) ? cliTools : [...[cliTools]];
+  // skipcq: JS-0349
+  const virtctlObj = tools?.find(
+    (tool) => (tool as K8sResourceCommon)?.metadata?.name === VIRTCTL_DOWNLOADS,
+  ) as K8sResourceKind;
+
+  const displayName = virtctlObj?.spec?.displayName;
+  const links = virtctlObj?.spec?.links;
```
metalice, line 29: im guessing you can deconstruct displayName and link in same line
const {displayName , links } = virtctlObj?.spec || {}

src/views/clusteroverview/utils/VirtctlPopup.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+const VirtctlPopup: React.FC = () => {
+  const { t } = useKubevirtTranslation();
+  const [cliTools] = useK8sWatchResource({
+    groupVersionKind: ConsoleCLIDownloadModelGroupVersionKind,
+    isList: true,
+    namespaced: false,
+  });
+
+  const tools = Array.isArray(cliTools) ? cliTools : [...[cliTools]];
```
metalice, line 22: i think its the same as = Array.isArray(cliTools) ? cliTools : [cliTools];

src/views/clusteroverview/utils/VirtctlPopup.tsx
```diff
@@ -0,0 +1,69 @@
...
+  // skipcq: JS-0349
```
metalice, line 23: what does it mean?

src/views/clusteroverview/utils/VirtctlPopup.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const virtctlObj = tools?.find(
+    (tool) => (tool as K8sResourceCommon)?.metadata?.name === VIRTCTL_DOWNLOADS,
```
metalice, line 25: u sure K8sResourceCommon and K8sResourceKind are mandatory here as hard casting?

src/views/clusteroverview/utils/VirtctlPopup.tsx
```diff
@@ -0,0 +1,69 @@
...
+      <div className="virtctl-popup__body--message">
+        {t(
+          'The virtctl client is a supplemental command-line utility for managing virtualization resources from the command line.',
+        )}
+      </div>
+      <div>
+        {links?.length > 1 && (
+          <ul>
+            {links.map((link) => (
+              <li key={`${link.text}-${link.href}`}>
```
metalice, line 42: does .href and .text are guaranteed? is no, consider adding '?. '

metalice:
Hi @pcbailey thanks for the PR. I left some small comments :)

src/views/clusteroverview/utils/VirtctlPopup.tsx
glekner: does this skip DeepSource checks? why?

src/views/clusteroverview/utils/VirtctlPopup.tsx
pcbailey: Good call. Done!

src/views/clusteroverview/utils/VirtctlPopup.tsx
pcbailey: Ah, true. Done!

src/views/clusteroverview/utils/VirtctlPopup.tsx
pcbailey: @metalice @glekner Exactly, it's the equivalent of `eslint-disable`. It was throwing an error because of the `K8sResourceKind` cast. If I don't cast it I get a type error because `spec` isn't on `K8sResourceCommon`. I'm open to a better way to handle that.

src/views/clusteroverview/utils/VirtctlPopup.tsx
pcbailey: I typically don't hard cast unless I get errors, but I double-checked and was able to remove the hard cast to `K8sResourceCommon`, but the `K8sResourceKind` is needed to avoid an error as described in the DeepSource comment above.

src/views/clusteroverview/utils/VirtctlPopup.tsx
pcbailey: Good catch. Done!

src/views/clusteroverview/utils/VirtctlPopup.tsx
metalice: useK8sWatchResource<K8sResourceKind> won't help in this case?

PR #747, opened 2022-07-07
Bug 2102074: Improve VM Creation time on VM Overview Details card

hstastna:
## 📝 Description

**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2102074

Add "ago" and plural form to VM creation time string to be grammatically correct.

## 🎥 Demo
**Before:**
Singular and without "ago":
minute_before
**After:**
0minutes
1minute


hstastna:

metalice:
Hi @hstastna I think other components are using Timestamp component. Can you please check? if needed add this fix to other places where needed? So everything will be on same schema of timing with pluriliza.

hstastna:
@metalice I've checked that and the Timestamp component displays the date+time, when something was created. Here it is different: it does not show when it was created, but how many minutes/hours/days **ago** it was created.

metalice:
Cool 👍 

/lgtm

PR #751, opened 2022-07-11
Bug 2099923: Merge SSH access and SSH command

upalatucci:
## 📝 Description

- Delete ssh command label
- move command on ssh access
- Rename 'port: ..' to 'Service port...'
- label 'Edit' near 'SSH access' 

## 🎥 Demo
Before
Screenshot from 2022-07-11 16-13-25
Screenshot from 2022-07-11 16-13-19

After
Screenshot from 2022-07-11 16-11-19
Screenshot from 2022-07-11 16-11-13


src/views/virtualmachinesinstance/details/tabs/details/components/Details/SSHAccess/SSHAccess.tsx
```diff
@@ -20,34 +27,47 @@ type SSHAccessProps = {
 const SSHDetails: React.FC<SSHAccessProps> = ({ vmi, vm, sshService, sshServiceLoaded }) => {
```
pcbailey, line 27: The file name is `SSHAccess.tsx` and the props type is `SSHAccessProps`. Why is the component name `SSHDetails`? I see there is already an `SSHAccess` component, so I assume the file name and props types should be renamed to match the component name?

src/views/virtualmachines/details/tabs/details/utils/gridHelper.tsx
```diff
@@ -83,11 +81,13 @@ export const getRunningVMRightGridPresentation = (
       ? GuestAgentIsRequiredText
       : guestAgentData?.timezone?.split(',')[0],
     node: <ResourceLink kind={NodeModel.kind} name={nodeName} />,
-    userCredentials: sshServiceLoaded ? (
-      <UserCredentials vmi={vmi} sshService={sshService} />
+    sshAccess: sshServiceLoaded ? (
```
pcbailey, line 84: This logic is duplicated in the `SSHDetails` component. You could create a small component for it that takes `sshServiceLoaded` as a prop and then use it in both places. Not necessary, but could make it a bit simpler.

src/views/virtualmachinesinstance/details/tabs/details/components/Details/SSHAccess/SSHAccess.tsx
upalatucci: Yes okay. I'll change the file name to `SSHDetails`

PR #752, opened 2022-07-12
Bug 2106175: All pages are crashed after visit Virtualization -> Overview

glekner:
## 📝 Description

> Add a brief description

## 🎥 Demo


src/views/clusteroverview/MonitoringTab/top-consumers-card/utils/DurationDropdown.tsx
```diff
@@ -27,8 +27,8 @@ const DurationDropdown: React.FC<DurationDropdownProps> = ({ selectedDuration, s
       variant={SelectVariant.single}
     >
       {DurationOption.getAll().map((durationOption) => {
-        const dropdownLabel = durationOption.getDropdownLabel();
-        const durationValue = durationOption.getValue();
+        const dropdownLabel = durationOption?.getDropdownLabel() || '';
```
pcbailey, line 30: @glekner All members of `DurationOption` are predefined with both `value` and `dropdownLabel` fields. How is it they could be `undefined` or `null`?

PR #753, opened 2022-07-12
CNV-19414: Templates project tab

metalice:

## 📝 Description

Adding template project tab - this gives the user option to move common templates from one name space to another.

## 🎥 Demo

templates-project


src/utils/components/CreateProjectModal/CreateProjectModal.tsx
```diff
@@ -0,0 +1,117 @@
...
+          model: ProjectRequestModel,
+          data,
+        }).then((value) => createdProject(value))
+      }
+      headerText={t('Create project')}
+      submitBtnText={t('Create')}
+      isDisabled={isEmpty(name)}
+    >
+      <Text>
+        {t('An OpenShift project is an alternative representation of a Kubernetes namespace.')}
```
avivtur, line 54: I think we can simply omit the 'An' at the beginning

src/utils/components/CreateProjectModal/CreateProjectModal.tsx
```diff
@@ -0,0 +1,117 @@
...
+      <br />
+      <br />
+      <Form>
+        <FormGroup
+          label={t('Name')}
+          labelIcon={
+            <Popover
+              bodyContent={
+                <Trans t={t} ns="plugin__kubevirt-plugin">
+                  A Project name must consist of lower case alphanumeric characters or &apos;, and
```
avivtur, line 69: I think we can omit the 'A' at the beginning

src/views/clusteroverview/SettingsTab/TemplatesProjectTab/utils/utils.ts
```diff
@@ -0,0 +1,60 @@
...
+    handleLoading(true);
+    try {
+      await k8sPatch<HyperConverged>({
+        model: HyperConvergedModel,
+        resource: hyperConverged,
+        data: [
+          {
+            op: 'replace',
+            path: `/spec/commonTemplatesNamespace`,
+            value: newNamespace === OPENSHIFT ? null : newNamespace,
```
avivtur, line 32: why do you need to check here if the newNamespace is openshift? what will happen if we changed it once from `openshift` to some other namespace and than I want to change it back? won't we lose all templates?

avivtur (commented):
I think we need to change in all the text shown to user all resources like `Templates` `Projects` `Namespaces` to be PascalCase, please fix them as well
other than that looks good to me :)

PR #755, opened 2022-07-12
Bug 2104402: Envs Save button fix

upalatucci:
## 📝 Description

 **Cause**
Save button enabled and disabled at the wrong moment
 - vm update return environments with a different order.
 - Comparison of new and old envs based on the index

**Solution**
Does not base the comparison on the index

### DIFF
--- a/.github/workflows/ci_checks.yml
+++ b/.github/workflows/ci_checks.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Setup Yarn
         uses: actions/setup-node@v2
         with:
-          node-version: '14'
+          node-version: '16'
           cache: 'yarn'
       - run: yarn install
 
@@ -44,7 +44,7 @@ jobs:
       - name: Setup Yarn
         uses: actions/setup-node@v2
         with:
-          node-version: '14'
+          node-version: '16'
           cache: 'yarn'
       - run: yarn install
 
@@ -64,7 +64,7 @@ jobs:
       - name: Setup Yarn
         uses: actions/setup-node@v2
         with:
-          node-version: '14'
+          node-version: '16'
           cache: 'yarn'
       - run: yarn install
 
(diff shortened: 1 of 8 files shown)


src/utils/components/EnvironmentEditor/tests/mosks.ts
```diff
@@ -0,0 +1,114 @@
+import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
+
+export const exampleVirtualMachineWithEnvironments: V1VirtualMachine = {
+  apiVersion: 'kubevirt.io/v1',
+  kind: 'VirtualMachine',
+  metadata: {
+    annotations: {
+      'vm.kubevirt.io/validations':
+        '[\n  {\n    "name": "minimal-required-memory",\n    "path": "jsonpath::.spec.domain.resources.requests.memory",\n    "rule": "integer",\n    "message": "This VM requires more memory.",\n    "min": 1610612736\n  }\n]\n',
```
metalice, line 9: Literal strong won't work here instead of new line escape?

upalatucci: It works! pushing...

.github/workflows/ci_checks.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       - name: Setup Yarn
         uses: actions/setup-node@v2
         with:
-          node-version: '14'
+          node-version: '16'
```
upalatucci, line 22: one test failed because `environment.at(-1)` was not recognized by node 14. This is a feature of node 16. But because node 16 is LTS now we can change this.

package.json
```diff
@@ -142,6 +142,7 @@
     "mochawesome-merge": "^4.2.1",
     "pluralize": "^8.0.0",
     "prettier": "^2.5.1",
+    "reselect": "^4.1.6",
```
upalatucci, line 145: `reselect` as peer dep breaks new tests. Now dev dep. I'll remove it with the new SDK version, I promise

upalatucci (commented):
Some notes for reviewers

PR #756, opened 2022-07-12
Bug 2100442: show ssh service even without running vm

upalatucci:
## 📝 Description

Show services with stopped VM using `vm.spec.template.metadata.labels` instead of `vmi.metadata.labels` when vmi is not available (because vm not running)

Do you guys think we need both vm and vmi? Or we can just use `vm.spec.template.metadata.labels` ? 
## 🎥 Demo

image

### DIFF
--- a/src/utils/components/SSHAccess/useSSHService.tsx
+++ b/src/utils/components/SSHAccess/useSSHService.tsx
@@ -1,6 +1,6 @@
 import { ServiceModel } from '@kubevirt-ui/kubevirt-api/console';
 import { IoK8sApiCoreV1Service } from '@kubevirt-ui/kubevirt-api/kubernetes';
-import { V1VirtualMachineInstance } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
 import { getServicesForVmi } from '@kubevirt-utils/resources/vmi';
 import { useK8sWatchResource } from '@openshift-console/dynamic-plugin-sdk';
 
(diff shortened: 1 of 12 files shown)


hstastna (commented):
Looks good to me, I am okay with what you did. Just I am not sure about always enabling editing the SSH access field [here](https://github.com/kubevirt-ui/kubevirt-plugin/pull/756/files#diff-c55e2d99d4fdce563632cf997cb1986e233933cf4096895a190889edfd665d8aR155), but that should be fine if we just edit `vm.spec.template.metadata.labels`.
Regarding your question if we need both vm and vmi, I'm not sure. IMHO vm could be enough, but I'd better ask Kobi about this. But IMHO you could already `useSSHService(vm)` with just `vm` as an argument, and in that function you'd just
`const { vmi } = useVMIAndPodsForVM(vm?.metadata?.name, vm?.metadata?.namespace);`
so you wouldn't need both `vm` and `vmi` as the arguments of the `useSSHService` function.

src/utils/components/SSHAccess/utils.ts
```diff
@@ -31,20 +31,14 @@ const buildSSHServiceFromVM = (vm: V1VirtualMachine, sshLabel: string) => ({
   },
 });
 
-export const createSSHService = async (vmi: V1VirtualMachineInstance, vm: V1VirtualMachine) => {
-  const { namespace, labels, name } = vmi?.metadata;
+export const createSSHService = async (vm: V1VirtualMachine, vmi: V1VirtualMachineInstance) => {
+  const { namespace, labels, name } = vm.metadata;
```
avivtur, line 35: const { namespace, labels, name } = vm.metadata || {};

src/utils/components/SSHAccess/utils.ts
```diff
@@ -31,20 +31,14 @@ const buildSSHServiceFromVM = (vm: V1VirtualMachine, sshLabel: string) => ({
...
     await k8sPatch({
       model: VirtualMachineModel,
-      resource: {
-        kind: VirtualMachineModel.kind,
-        metadata: {
-          name: vmi?.metadata?.ownerReferences?.[0]?.name,
-          namespace: namespace,
-        },
-      },
+      resource: vm,
```
avivtur, line 41: nice :)

src/utils/components/SSHAccess/useSSHService.tsx
```diff
@@ -1,21 +1,31 @@
...
-    kind: ServiceModel.kind,
-    isList: true,
-    namespace: vmi?.metadata?.namespace,
-  });
+const useSSHService = (
+  vmi: V1VirtualMachineInstance,
+  vm?: V1VirtualMachine,
+): UseSSHServiceReturnType => {
+  const [services, servicesLoaded, servicesError] = useK8sWatchResource<IoK8sApiCoreV1Service[]>(
+    vm
```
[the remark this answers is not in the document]
upalatucci: I'm afraid that if the k8sWatch receives `namespace: undefined` as the vm is not yet defined, the hook will start requesting services for all the namespaces

PR #761, opened 2022-07-15
Bug 2103807: Name PVC source same as VM name when creating VM quickly

hstastna:
## 📝 Description

Name the PVC source name the same as the VM name, according to the parameter NAME in the yaml used also for the PVC source name.

The problem was that the part of the processing template was omitted, when quick creating VM, comparing creating VM with the customization process. The parameter NAME was not properly set when quick creating VM.

**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2103807

## 🎥 Demo
**Before:**
quick_before
**After:**
quick_after


hstastna:

PR #778, opened 2022-07-20
CNV-19844: Adding utilization charts to vm metrics tab

metalice:

## 📝 Description

Moving charts to shared locations.
Add a responsive hook, so charts resize will be dynamically
Add charts to the metrics tab

## 🎥 Demo

image


src/utils/components/Charts/hooks/useResponsiveCharts.tsx
```diff
@@ -0,0 +1,32 @@
...
+        setWidth(node.getBoundingClientRect().width);
+        setHeight(node.getBoundingClientRect().height);
+      };
+      sizeSetter();
+      setListener(getResizeObserver(containerRef?.current, sizeSetter));
+      containerRef.current = node;
+    }
+  }, []);
+
+  useEffect(() => () => listener && listener(), [listener]);
```
glekner, line 27: can you make this effect more readable with blocks?.
also, don't you need a cleanup here?

src/views/virtualmachines/details/tabs/metrics/virtual-machine-metrics-tab.scss
```diff
@@ -40,3 +40,7 @@
     }
   }
 }
+.util-threshold-chart {
+  width: 100%;
+  height: 150px;
```
glekner, line 45: why hardcoded height?
can't we use flex here?

src/utils/components/Charts/hooks/useResponsiveCharts.tsx
metalice: cool 😎

PR #784, opened 2022-07-24
Bug 2100290: Fixing the error that doesn't dismissed on Catalog review page

DanaOrr:

## 🎥 Demo

Error


src/views/catalog/customize/components/CustomizeForms/useCustomizeFormSubmit.ts
```diff
@@ -178,6 +178,12 @@ export const useCustomizeFormSubmit = ({
     }
   };
 
+  React.useEffect(() => {
+    if (templateError) {
+      setTemplateError(undefined);
+    }
```
glekner, line 184: I think you need to disable the eslint warning here

`    // eslint-disable-next-line react-hooks/exhaustive-deps`

DanaOrr: Ok, thank you :)

upalatucci (approved):

gouyang:
@DanaOrr the problem is not fixed completely, if you click the "review and create" button, the error still pops up.

PR #786, opened 2022-07-25
Bug 2103806: PVC 404

upalatucci:


hstastna (approved):

avivtur (commented):
Hi Ugo thanks for this PR :) 
The usage of `kind` prop should be depracated, please use instead like so
 `groupVersionKind={modelToGroupVersionKind(PersistentVolumeClaimModel)}` instead of `kind={PersistentVolumeClaimModel.kind}` in all relevant places

upalatucci:
Thanks @avivtur nice catch!

avivtur:
/lgtm
:100: :100:

PR #787, opened 2022-07-25
Bug 2101501: Display VM name in Create VM drawer according to the NAME

hstastna:
## 📝 Description
**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2101501

Display value for _VirtualMachine name_ field in the Create VirtualMachine drawer according to the template `NAME` parameter.
After clicking on _Quick create VirtualMachine_, the VM is created with the correct name - same as in the field above the button.
Also the same corresponding correct name is displayed when user clicks on _Customize VirtualMachine_ button in the drawer.

## 🎥 Demo
**Before:**
No matter the template `NAME` parameter, the VM name is generated the same way:
a
**After:**
If the VM name is generated from a given regex:
a
If the VM name is a specific string:
aa
If the VM name is set to "None", we generate the name as "Before":
aaa

---

TODO:
- [x] fix the name displayed when clicking on "Customize VirtualMachine":
buggy


src/views/catalog/customize/components/CustomizeForms/CustomizeForm.tsx
```diff
@@ -23,7 +23,14 @@ export const CustomizeForm: React.FC<CustomizeFormProps> = ({ template }) => {
...
   const { onSubmit, loaded, error } = useCustomizeFormSubmit({ template });
   const [requiredFields, optionalFields] = buildFields(template);
-  const nameField = getVirtualMachineNameField(template, t);
+
+  const [vmName, setVmName] = React.useState('');
+  React.useMemo(() => {
+    getvmName(template).then((name) => {
+      setVmName(name);
+    });
+  }, [template]);
```
avivtur, line 32: the use state here can be removed, you can use instead:
`const vmName = useMemo(() => getVMName(template) , [template])`
- we recently made a change that allows the following import statement: `import { useMemo } from 'react'`
- I'd think it is better to change getvmName to getVMName since VM is an acronym and it's a convention to write them as upper case.

src/views/catalog/customize/components/CustomizeForms/CustomizeFormWithStorage.tsx
```diff
@@ -64,7 +65,15 @@ const CustomizeFormWithStorage: React.FC<CustomizeFormWithStorageProps> = ({ tem
...
   const [requiredFields, optionalFields] = React.useMemo(() => buildFields(template), [template]);
-  const nameField = React.useMemo(() => getVirtualMachineNameField(template, t), [template, t]);
+
+  const [vmName, setVmName] = React.useState('');
+  React.useMemo(() => {
+    getvmName(template).then((name) => {
+      setVmName(name);
+    });
+  }, [template]);
+
```
avivtur, line 75: same here :)

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
```diff
@@ -43,13 +43,20 @@ export const TemplatesCatalogDrawerCreateForm: React.FC<TemplatesCatalogDrawerCr
...
     const [quickCreateError, setQuickCreateError] = React.useState(undefined);
     const [models, modelsLoading] = useK8sModels();
     const [defaultStorageClass, defaultStorageClassLoaded] = useDefaultStorageClass();
 
+    React.useMemo(() => {
+      getvmName(template).then((name) => {
+        setVmName(name);
+      });
+    }, [template]);
+
```
avivtur, line 59: here too :)

src/views/catalog/customize/components/CustomizeForms/CustomizeForm.tsx
hstastna: Thanks for the suggestions, they all make sense to me :)

hstastna: Note that my _getvmName_ returns a promise, so I am not sure if we can really get rid of using state to get the string for the VM name...

src/views/catalog/customize/utils.ts
```diff
@@ -127,16 +127,31 @@ export const processTemplate = async ({
...
+export const getvmName = async (template: V1Template): Promise<string> => {
+  return await k8sCreate<V1Template>({
+    model: ProcessedTemplatesModel,
+    data: template,
+    queryParams: {
+      dryRun: 'All',
+    },
+  }).then(
+    (simpleProcessedTemplate) =>
+      simpleProcessedTemplate?.parameters.filter(
```
pcbailey, line 139: Is the `parameters` property guaranteed to exist?

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
```diff
@@ -43,13 +50,22 @@ export const TemplatesCatalogDrawerCreateForm: React.FC<TemplatesCatalogDrawerCr
   React.memo(({ namespace, template, canQuickCreate, isBootSourceAvailable, onCancel }) => {
     const history = useHistory();
     const { t } = useKubevirtTranslation();
-    const [vmName, setVmName] = React.useState(generateVMName(template));
+    const { updateVM } = useWizardVMContext();
+    const { ns } = useParams<{ ns: string }>();
+
+    const [vmName, setVmName] = React.useState('');
```
pcbailey, line 56: nitpick: It should be `setVMName`. All letters in an abbreviation should be capitalized.

pcbailey (commented):
It looks very nice! Just a few small comments.

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
```diff
@@ -74,6 +90,35 @@ export const TemplatesCatalogDrawerCreateForm: React.FC<TemplatesCatalogDrawerCr
...
+            processedTemplate.metadata.namespace;
+        });
+        updateVM(updatedVM)
+          .then(() => {
+            history.push(
+              `templatescatalog/customize?name=${template.metadata.name}&namespace=${template.metadata.namespace}`,
+            );
+          })
+          .catch((e) => {
+            console.error(e);
```
avivtur, line 117: please try to avoid using console.error() to display errors and follow the pattern of `quickCreateError` state

cypress/tests/catalog/create-vm-from-catalog.spec.ts
```diff
@@ -2,7 +2,7 @@ import { K8S_KIND, TEST_NS } from '../../utils/const/index';
 import { defaultSourceVM } from '../../utils/const/testVM';
 import { vm } from '../../views/vm';
 
-describe('Create VM from catalog', () => {
+xdescribe('Create VM from catalog', () => {
```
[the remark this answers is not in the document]
hstastna: Yes, I was consulting this with Guohua. This is just a temporary change, the test will be updated by Guohua after this PR will be merged.

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
hstastna: Okay, I'll take a look, thank you! :)

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
```diff
@@ -74,6 +92,40 @@ export const TemplatesCatalogDrawerCreateForm: React.FC<TemplatesCatalogDrawerCr
...
+              `templatescatalog/customize?name=${template.metadata.name}&namespace=${template.metadata.namespace}`,
+            );
+          })
+          .catch((err) => {
+            setIsPreparingToCoustomizeVM(false);
+            setCustomizeVMError(err);
+          });
+      });
+    };
+
```
upalatucci, line 128: thinking more about this approach.
Instead of injecting the `useWizardVMContext` can't we just add an URL param like vmName or something like that to pass the generated VM name to the customizeForm?

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
```diff
@@ -95,7 +147,7 @@ export const TemplatesCatalogDrawerCreateForm: React.FC<TemplatesCatalogDrawerCr
                         name="vmname"
                         aria-label="virtualmachine name"
                         value={vmName}
-                        onChange={(v) => setVmName(v)}
+                        onChange={(v) => setVMName(v)}
```
upalatucci, line 150:
```suggestion
                        onChange={setVMName}
```

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
```diff
@@ -43,13 +50,24 @@ export const TemplatesCatalogDrawerCreateForm: React.FC<TemplatesCatalogDrawerCr
...
     const [quickCreateError, setQuickCreateError] = React.useState(undefined);
+    const [isPreparingToCoustomizeVM, setIsPreparingToCoustomizeVM] = React.useState(false);
+    const [customizeVMError, setCustomizeVMError] = React.useState(undefined);
     const [models, modelsLoading] = useK8sModels();
     const [defaultStorageClass, defaultStorageClassLoaded] = useDefaultStorageClass();
 
+    React.useMemo(() => {
+      getVMName(template).then((name) => {
+        setVMName(name);
+      });
```
upalatucci, line 68:
```suggestion
      getVMName(template).then(setVMName);
```

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
hstastna: I was discussing this with other folks and we came up with using `useWizardVMContext` as the only way.

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
```diff
@@ -43,13 +50,24 @@ export const TemplatesCatalogDrawerCreateForm: React.FC<TemplatesCatalogDrawerCr
...
+    const [vmName, setVMName] = React.useState('');
     const [startVM, setStartVM] = React.useState(true);
     const [isQuickCreating, setIsQuickCreating] = React.useState(false);
     const [quickCreateError, setQuickCreateError] = React.useState(undefined);
+    const [isPreparingToCoustomizeVM, setIsPreparingToCoustomizeVM] = React.useState(false);
+    const [customizeVMError, setCustomizeVMError] = React.useState(undefined);
     const [models, modelsLoading] = useK8sModels();
     const [defaultStorageClass, defaultStorageClassLoaded] = useDefaultStorageClass();
 
+    React.useMemo(() => {
```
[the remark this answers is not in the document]
hstastna: Thank you very much, that's a good point!

gouyang:
The test failure caused by VM cannot be scheduled, improve the tests via #911

src/utils/utils/utils.ts
```diff
@@ -3,24 +3,27 @@ import { IoK8sApiCoreV1Service } from '@kubevirt-ui/kubevirt-api/kubernetes';
...
 
-export const isString = (val) => val !== null && typeof val === 'string';
+export const isString = (val: unknown) => val !== null && typeof val === 'string';
 
 export const getSSHNodePort = (sshService: IoK8sApiCoreV1Service) =>
   sshService?.spec?.ports?.find((port) => parseInt(port.targetPort, 10) === 22)?.nodePort;
 
-export const isTemplateParameter = (value: string): boolean => !!/^\${[A-z0-9_]+}$/.test(value);
+export const isTemplateParameter = (value: string): boolean =>
+  Boolean(/^\${[A-z0-9_]+}$/.test(value));
```
pcbailey, line 26: Does `Boolean` do something different than `!!`? `!!` converts any other type to a boolean type. I assume that's all that's being done by `Boolean` here, right?

src/views/catalog/customize/utils.ts
```diff
@@ -97,20 +98,21 @@ export const processTemplate = async ({
...
+    draftVMTemplate.metadata.namespace = namespace;
+
     const parameterForName = extractParameterNameFromMetadataName(template);
 
-    draftTemplate = setTemplateParameters(draftTemplate, formData);
+    draftVMTemplate = setTemplateParameters(draftVMTemplate, formData);
 
-    draftTemplate = replaceTemplateParameterValue(
-      draftTemplate,
+    draftVMTemplate = replaceTemplateParameterValue(
```
pcbailey, line 109: +1 on the clearer name. =)

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
```diff
@@ -74,6 +92,49 @@ export const TemplatesCatalogDrawerCreateForm: React.FC<TemplatesCatalogDrawerCr
         });
     };
 
+    const onCustomizeVM = () => {
+      setIsPreparingToCustomizeVM(true);
+      setCustomizeVMError(undefined);
+
+      // check for the required params and specify some default value for them to get the template processed successfully
+      // required parameters' values will be specified later  by the user anyway, during customization process
```
pcbailey, line 100: You have an extra space between 'later' and 'by'.

pcbailey (commented):
Just a question and a nitpick. I'm not going to hold this up any longer. We can fix the small stuff in a separate PR if needed.

/lgtm

src/utils/utils/utils.ts
hstastna: @pcbailey Right. Unfortunately, something has changed the way that the whole files are tested here on github, not only diffs, and from now it does not like `!!`. Not sure why... And it always suggests me to use `Boolean(...)` instead of `!!`. Without such changes, the PR would be unmergeable.

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
hstastna: Oh, you are right! Will fixed that in some another PR! :)

src/utils/utils/utils.ts
pcbailey: @hstastna That's strange since it's a valid JavaScript operator. Maybe @yaacov will have an idea of why it's blocking it.

PR #794, opened 2022-07-27
Bug 2060499: Quick create - template with multiple objects

upalatucci:
## 📝 Description

In a follow-up pr, I'll solve this bug also in the wizard using the same function `createAllTemplateObjects`   

In this solution, I use `Promise.all`. This means that if any object creation request fails, it will follow the fail-fast behavior and throw an exception without waiting for the other promise's resolution.

What to do if one object fail?
Maybe some objects have been created and at least one failed. What can we do? We can try to delete the created objects? WDYT?  I can do it in a separate pr

### DIFF
--- a/src/utils/resources/template/utils/helpers.ts
+++ b/src/utils/resources/template/utils/helpers.ts
@@ -4,6 +4,12 @@ import { V1Template } from '@kubevirt-ui/kubevirt-api/console';
 import VirtualMachineModel from '@kubevirt-ui/kubevirt-api/console/models/VirtualMachineModel';
 import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
 import { getAnnotation } from '@kubevirt-utils/resources/shared';
+import {
+  getGroupVersionKindForResource,
+  k8sCreate,
+  K8sModel,
+  K8sResourceCommon,
+} from '@openshift-console/dynamic-plugin-sdk';
 
 import { ANNOTATIONS } from './annotations';
 
@@ -33,3 +39,27 @@ export const replaceTemplateVM = (template: V1Template, vm: V1VirtualMachine) =>
     draftTemplate.objects.splice(vmIndex, 1, vm);
   });
 };
+
+export const createAllTemplateObjects = (
+  template: V1Template,
+  models: { [key: string]: K8sModel },
+  namespace?: string,
+): Promise<K8sResourceCommon[]> =>
+  Promise.all(
+    template.objects.map((object) => {
+      const { group, version, kind } = getGroupVersionKindForResource(object);
+
+      const ref = [group || 'core', version, kind].join('~');
+
+      const model = models[ref] || models[object.kind];
+
+      return k8sCreate<K8sResourceCommon>({
+        model: model,
+        data: produce(object as K8sResourceCommon, (draftObject) => {
+          if (!draftObject.metadata.namespace && namespace) {
+            draftObject.metadata.namespace = namespace;
+          }
+        }),
+      });
+    }),
+  );
--- a/src/views/catalog/utils/quick-create-vm.ts
+++ b/src/views/catalog/utils/quick-create-vm.ts
@@ -1,38 +1,60 @@
+import produce from 'immer';
+
 import { ProcessedTemplatesModel, V1Template } from '@kubevirt-ui/kubevirt-api/console';
 import VirtualMachineModel from '@kubevirt-ui/kubevirt-api/console/models/VirtualMachineModel';
 import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
 import {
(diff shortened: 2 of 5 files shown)


src/utils/resources/template/utils/helpers.ts
```diff
@@ -33,3 +39,23 @@ export const replaceTemplateVM = (template: V1Template, vm: V1VirtualMachine) =>
...
+    template.objects.map((object) => {
+      const { group, version, kind } = getGroupVersionKindForResource(object);
+
+      const ref = [group || 'core', version, kind].join('~');
+
+      const model = models[ref] || models[object.kind];
+
+      return k8sCreate<K8sResourceCommon>({
+        model: model,
+        data: namespace ? { ...object, metadata: { ...object.metadata, namespace } } : object,
```
glekner, line 58: use `produce` here

src/utils/resources/template/utils/helpers.ts
```diff
@@ -33,3 +39,23 @@ export const replaceTemplateVM = (template: V1Template, vm: V1VirtualMachine) =>
...
+export const createAllTemplateObjects = (
+  template: V1Template,
+  models: { [key: string]: K8sModel },
+  namespace?: string,
+): Promise<K8sResourceCommon[]> =>
+  Promise.all(
+    template.objects.map((object) => {
+      const { group, version, kind } = getGroupVersionKindForResource(object);
+
+      const ref = [group || 'core', version, kind].join('~');
```
glekner, line 52: you can use `transformGroupVersionKindToReference` here

src/views/catalog/utils/quick-create-vm.ts
```diff
@@ -2,37 +2,48 @@ import { ProcessedTemplatesModel, V1Template } from '@kubevirt-ui/kubevirt-api/c
...
   LABEL_USED_TEMPLATE_NAME,
   LABEL_USED_TEMPLATE_NAMESPACE,
+  replaceTemplateVM,
 } from '@kubevirt-utils/resources/template';
 import { getTemplateVirtualMachineObject } from '@kubevirt-utils/resources/template/utils/selectors';
-import { k8sCreate } from '@openshift-console/dynamic-plugin-sdk';
+import { k8sCreate, K8sModel } from '@openshift-console/dynamic-plugin-sdk';
 
-export const quickCreateVM = (
+export const quickCreateVM = async (
```
glekner, line 13: please change the signature of the function to receive one object

src/views/catalog/utils/quick-create-vm.ts
```diff
@@ -2,37 +2,48 @@ import { ProcessedTemplatesModel, V1Template } from '@kubevirt-ui/kubevirt-api/c
...
+  vm.metadata.labels[LABEL_USED_TEMPLATE_NAMESPACE] = processedTemplate.metadata.namespace;
 
-    return k8sCreate<V1VirtualMachine>({
-      model: VirtualMachineModel,
-      data: vm,
-    });
-  });
+  if (startVM) {
+    vm.spec.running = true;
+  }
```
glekner, line 36: can you add `produce` here aswell?

src/views/catalog/templatescatalog/TemplatesCatalog.tsx
```diff
@@ -24,7 +24,6 @@ const TemplatesCatalog: React.FC<RouteComponentProps<{ ns: string }>> = ({
   },
 }) => {
   const [selectedTemplate, setSelectedTemplate] = React.useState<V1Template | undefined>(undefined);
-
```
glekner, line 27: redundant

glekner (changes requested):
nice!

src/utils/resources/template/utils/helpers.ts
upalatucci: It's deprecated. Can I use it anyway?

src/utils/resources/template/utils/helpers.ts
```diff
@@ -33,3 +40,32 @@ export const replaceTemplateVM = (template: V1Template, vm: V1VirtualMachine) =>
...
+          if (!draftObject.metadata.namspace && namespace)
+            draftObject.metadata.namespace = namespace;
+        });
+      }
+
+      return k8sCreate<K8sResourceCommon>({
+        model: model,
+        data: objectWithNamespace,
+      });
+    }),
```
glekner, line 70:
```suggestion
      return k8sCreate<K8sResourceCommon>({
        model: model,
        data: produce(object, (draftObject) => {
          if (!draftObject.metadata.namspace && namespace) {
            draftObject.metadata.namespace = namespace;
            }
        }),
      });
    }),
```

src/utils/resources/template/utils/helpers.ts
upalatucci: Can't import this function...

PR #795, opened 2022-07-28
Bug 2060499: getTemplateVirtualMachineObject instead of objects[0]

upalatucci:
## 📝 Description

Template with multiple objects can have the vm at any index

### DIFF
--- a/src/views/catalog/customize/drivers.ts
+++ b/src/views/catalog/customize/drivers.ts
@@ -1,6 +1,9 @@
 import { produceVMDisks } from '@catalog/utils/WizardVMContext/utils/vm-produce';
 import { ConfigMapModel, V1Template } from '@kubevirt-ui/kubevirt-api/console';
-import { getTemplateVirtualMachineObject } from '@kubevirt-utils/resources/template';
+import {
+  getTemplateVirtualMachineObject,
+  replaceTemplateVM,
+} from '@kubevirt-utils/resources/template';
 import { k8sGet } from '@openshift-console/dynamic-plugin-sdk';
 
 import {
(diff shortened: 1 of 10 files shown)


src/views/templates/details/tabs/scheduling/components/AffinityRulesModal.tsx
```diff
@@ -91,14 +92,15 @@ const AffinityRulesModal: React.FC<AffinityModalProps> = ({
 
   const updatedTemplate = React.useMemo(() => {
     return produce<V1Template>(template, (templateDraft: V1Template) => {
+      const draftVM = getTemplateVirtualMachineObject(templateDraft);
       if (!getAffinity(templateDraft)) {
-        templateDraft.objects[0].spec.template.spec.affinity = [];
```
upalatucci, line 95: Now it's explicit that we are using a vm and not a random k8sCommon object so typescript triggered about affinity. Affinity property is not an array so I defaulted to null here

src/views/catalog/customize/utils.ts
```diff
@@ -109,21 +109,22 @@ export const processTemplate = async ({
...
-      draftTemplate = await mountWinDriversToTemplate(draftTemplate);
-    }
+    const draftVM = getTemplateVirtualMachineObject(draftTemplate);
+
+    draftVM.metadata.name = virtualMachineName;
   });
 
-  const processedTemplate = await k8sCreate<V1Template>({
+  if (withWindowsDrivers) {
+    templateToProcess = await mountWinDriversToTemplate(templateToProcess);
```
upalatucci, line 118: Now `mountWinDriversToTemplate` already create another template copy. Inside the `produce` function was not working anymore.

src/views/catalog/customize/drivers.ts
```diff
@@ -64,5 +67,5 @@ export const mountWinDriversToTemplate = async (template: V1Template): Promise<V
     });
   });
 
-  return template;
+  return replaceTemplateVM(template, newVM);
```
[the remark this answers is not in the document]
upalatucci: Can't do that. VM is in an array of objects. I think I have to replace it anyways...

PR #797, opened 2022-07-28
Bug 2060499: Replace vm object correctly into template

upalatucci:

### DIFF
--- a/src/views/catalog/customize/cd.ts
+++ b/src/views/catalog/customize/cd.ts
@@ -1,5 +1,3 @@
-import produce from 'immer';
-
 import { produceVMDisks } from '@catalog/utils/WizardVMContext';
 import { V1Template } from '@kubevirt-ui/kubevirt-api/console';
 import {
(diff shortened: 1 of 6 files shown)


src/views/catalog/customize/cd.ts
```diff
@@ -70,11 +71,7 @@ export const addCDToTemplate = (
   template: V1Template,
   cdSource: V1beta1DataVolumeSpec,
 ): V1Template => {
-  return produce(template, (draftTemplate) => {
-    if (!cdSource) return draftTemplate;
-    let virtualMachine = getTemplateVirtualMachineObject(draftTemplate);
-    virtualMachine = addInstallationCDRom(virtualMachine, cdSource);
+  const virtualMachine = addInstallationCDRom(getTemplateVirtualMachineObject(template), cdSource);
```
hstastna, line 74: I asked myself what if `!cdSource`, as the line
`if (!cdSource) return draftTemplate;`
was removed. And so I've checked `addInstallationCDRom`, and I think that in [this line](https://github.com/kubevirt-ui/kubevirt-plugin/blob/main/src/views/catalog/customize/cd.ts#L30) we should better make a change from
`cdSource.source?.http` to
`cdSource?.source?.http`
WDYT? Could you add this change to this PR to make sure nothing in there will break?

hstastna (commented):
Check my comment below. Otherwise great work! 👍🏽

src/views/catalog/customize/cd.ts
upalatucci: Yes sure :-) thanks

PR #799, opened 2022-07-31
[CNV-20217] Migrate various static extensions to dynamic

glekner:
## 📝 Description

EPIC - https://issues.redhat.com/browse/CNV-20213
Subtask - https://issues.redhat.com/browse/CNV-20217
Console extensions removal PR - https://github.com/openshift/console/pull/11905

Hardware devices page
Dashboard overview health
Snapshot Details page
Dev view create vm action


console-extensions.json
```diff
@@ -357,5 +357,46 @@
       },
       "component": { "$codeRef": "MigrationPoliciesList" }
     }
+  },
+  {
+    "type": "dev-console.add/action",
+    "properties": {
+      "id": "dev-catalog-virtualization",
+      "groupId": "developer-catalog",
+      "href": "/k8s/ns/:namespace/templatescatalog",
```
glekner, line 366: The Add VM action from the dev perspective now leads to the new catalog.
What should we do with the old dev form,catalog @yaacov ?

yaacov: AFAU the catalog is part of the admin perspective ? is it ok to send users from the dev perspective to the admin one ?
will it work for users that can only see the dev perspective ?

PR #801, opened 2022-07-31
Bug 2111240: Fix units display bug

pcbailey:
## 📝 Description

This PR fixes a bug in the display of disk size units that was present in both the 'Edit boot source' modal and anywhere `DiskSizeNumberInput` was used. This bug would cause the units to change from its existing value to `B` if the existing value was deleted or the value was changed to zero.

#### Before:
Selection_135

#### After:
Selection_134

#### Before:
Selection_136

#### After:
Selection_137


src/views/templates/actions/components/SelectSource.tsx
```diff
@@ -178,7 +178,11 @@ export const SelectSource: React.FC<SelectSourceProps> = ({
       )}
 
       {withSize && selectedSourceType !== SOURCE_TYPES.defaultSource && (
-        <VolumeSize quantity={volumeQuantity} onChange={setVolumeQuantity} />
+        <DiskSizeNumberInput
```
metalice, line 181: I think DiskSizeNumberInput can be moved to more general location instead of nested under disk modal, also when extracting can be rename to more general name such as SizeUnitsNumberInput, same goes to prop diskSize = size

src/views/catalog/utils/quantity.ts
```diff
@@ -1,5 +1,8 @@
 import byteSize, { ByteSizeResult } from 'byte-size';
 
+const fixUnitSuffix = (unit: string): string =>
+  unit.charAt(unit.length - 1) === 'B' ? unit : unit + 'B';
+
 const multipliers: Record<string, number> = {};
 multipliers.B = 1;
 multipliers.Ki = multipliers.B * 1024;
```
metalice, line 8: why declare such a way? 
multipliers = {
B: 1,
Ki: 1024
};

src/views/catalog/utils/quantity.ts
```diff
@@ -1,5 +1,8 @@
 import byteSize, { ByteSizeResult } from 'byte-size';
 
+const fixUnitSuffix = (unit: string): string =>
+  unit.charAt(unit.length - 1) === 'B' ? unit : unit + 'B';
```
metalice, line 4: please try unit[length -1] i think we can skip chartAt

src/views/catalog/utils/quantity.ts
```diff
@@ -50,11 +53,16 @@ export const bytesFromQuantity = (
...
 
-    if (ISUnit?.length || bytesUnit?.length || decimalUnit?.length) {
+    if (originalUnit?.length) {
       const bytes = value * multipliers[bytesUnit?.[0] || ISUnit?.[0] || decimalUnit?.[0]];
-
       byteSizeResult = bytesToIECBytes(bytes, precision);
+
+      // Prevents units from changing to 'B' when user enters 0 or erases existing value
+      if (value === 0) {
+        byteSizeResult.unit = fixUnitSuffix(originalUnit[0]) || byteSizeResult.unit;
```
metalice, line 64: please add ?. instead of just .

metalice:
thanks for this PR @pcbailey , i left some comments :)

pcbailey:
@avivtur @metalice @glekner @vojtechszocs @upalatucci @hstastna PTAL

src/views/catalog/utils/quantity.ts
hstastna: We don't need this `fixUnitSuffix`, IMO `toIECUnit` implemented [here](https://github.com/kubevirt-ui/kubevirt-plugin/blob/main/src/utils/utils/units.ts#L14) already does the thing 😏

metalice: Also a suggestion is to used xbytes package which is already into our code base. You can sample from the charts :)

pcbailey: @hstastna Thanks for the tip on `toIECUnit`. I didn't know about it. I replaced `fixUnitSuffix` with it for this PR.

@metalice `xbytes` looks very clean. I'll open a separate PR to clean up this file where possible.

src/views/catalog/utils/quantity.ts
pcbailey: Good question. Ask Ugo... he wrote it. =)

src/views/templates/actions/components/SelectSource.tsx
pcbailey: Done. I went with `MemorySizeInput` since RAM and storage are both types of memory.

src/views/catalog/utils/quantity.ts
pcbailey: Done!

pcbailey:
@metalice @hstastna Please re-review.

src/views/catalog/utils/quantity.ts
metalice: @upalatucci

src/utils/components/MemorySizeInput/MemorySizeInput.tsx
```diff
@@ -98,4 +97,4 @@ const DiskSizeNumberInput: React.FC<DiskSizeNumberInputProps> = ({ diskSize, onC
   );
 };
 
-export default DiskSizeNumberInput;
+export default MemorySizeInput;
```
metalice, line 100: in a quick look, i can see this component can be used a memory, CPU, PVC, and migration sizing, so memoryinput as a name cant work for all cases, need a more general name, in my opinion.
This component used PVC_SIZE_FORMAT I think this should be changed to a more appropriate name? maybe also moved? The same goes for fieldID prop, which is hardcoded for PVC

src/views/catalog/utils/quantity.ts
upalatucci: Yes we can follow the matan suggestion.
Was more complicated before 😂 I thought could be much more readable this way

src/views/catalog/utils/quantity.ts
metalice: cool, 😎  so if already cleaning, I think that toIECUnit can be replaced with xbytes so that we will be united within all our code base. looking forward to this PR. Thank you!

src/views/catalog/utils/quantity.ts
upalatucci: We can use the `units: 'iec'` instead of this custom unit at this point

src/utils/components/MemorySizeInput/MemorySizeInput.tsx
pcbailey: This component is only set to handle memory-type units as-is. It would need to be refactored to handle additional unit types. Is that what you're suggesting?

metalice: as I see it, this component is handling input number with a dropdown for unit TIB, GIB, MIB, which can be used for disk size, memory size, etc.. so calling it only memory isn't really matching the usage for disk size or others... The same goes for fieldId and PVC_SIZE_FORMAT, as it isn't only PVC. should be something like UNITS_FORMATS

pcbailey: I'm a bit confused because your original comment said it could be used for CPUs as well. So, just to clarify, you do mean to leave it where it works anywhere where base 2 capacity units (MiB, GiB, TiB, etc) are being used, correct? 

If so, I suggest `CapacityInput`.

metalice: CapacityInput is cool, just make sure to take care of everything mention with pvc-* in this file :)

pcbailey: I think I've already done that part. I just haven't pushed it yet. =)

metalice:
Cool. Let's merge :)

> @metalice I pushed the changes. I'm going to open a separate PR to replace what we can in `src/views/catalog/utils/quantity.ts` so we can get rid of byteSize entirely and use only xbytes.

PR #803, opened 2022-08-01
Fix vm list stuck loading

avivtur:
## 📝 Description

When a non-priv user reaches VM list page with no namespace selected the page is looks like it's stuck on loading

### before:

https://user-images.githubusercontent.com/67270715/182141872-68702a23-0767-4b3c-b517-e7edbb88df71.mp4

### after:

Screenshot (86)

### DIFF
--- a/src/utils/hooks/useKubevirtWatchResource.ts
+++ b/src/utils/hooks/useKubevirtWatchResource.ts
@@ -1,18 +1,35 @@
 import { useEffect, useState } from 'react';
 
-import { useK8sWatchResource, WatchK8sResource } from '@openshift-console/dynamic-plugin-sdk';
+import {
+  useAccessReview,
+  useK8sWatchResource,
+  WatchK8sResource,
+} from '@openshift-console/dynamic-plugin-sdk';
 
 const useKubevirtWatchResource = (watchOptions: WatchK8sResource) => {
   const [data, setData] = useState([]);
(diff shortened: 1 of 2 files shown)


src/utils/hooks/useKubevirtWatchResource.ts
```diff
@@ -1,18 +1,36 @@
...
 
+  const [getListAllowed] = useAccessReview({
+    verb: 'get',
+    resource: watchOptions?.groupVersionKind?.kind || watchOptions?.kind,
+    namespace: watchOptions?.namespace,
+  });
+
   useEffect(() => {
+    if (!getListAllowed) {
+      setData([]);
```
metalice, line 23: already empty array

src/utils/hooks/useKubevirtWatchResource.ts
```diff
@@ -1,18 +1,36 @@
...
   const [resource, loaded, loadError] = useK8sWatchResource<any>(watchOptions);
 
+  const [getListAllowed] = useAccessReview({
+    verb: 'get',
+    resource: watchOptions?.groupVersionKind?.kind || watchOptions?.kind,
+    namespace: watchOptions?.namespace,
+  });
+
   useEffect(() => {
+    if (!getListAllowed) {
```
metalice, line 22: i think is better to add loaded of useAccessReview to this condition

PR #806, opened 2022-08-02
Bug 2101423: Fixing the wrong user name when using ignition

DanaOrr:

## 📝 Description
Cloudinit can have a few config types. This fix should be able to present password and user if present for the current supported configs.

## 🎥 Demo
Before: 
wrongUserName1

After:
wrongUserName


hstastna (commented):
Thank you very much, Dana, for this PR! 👍🏽  It gave me some ideas about how to improve my fix for the same BZ (as I was assigned to it), for example adding an extra and optional parameter `passwd` - maybe I would return this in case of cloud-init ConfigDrive data source. 
Your PR does not deal with some cases for example what if there are more users, but your work definitely did worth! Thank you! 🎇

PR #807, opened 2022-08-02
Bug 2101423: Display user name on using ignition

hstastna:
## 📝 Description

**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2101423

Display the user name correctly in the _Guest login credentials_ in the VM's _Console_ tab.

KubeVirt supports the cloud-init NoCloud and ConfigDrive data sources. The implementation for ConfigDrive data source was missing. The fix supports also displaying multiple usernames. Note that `users` (list of objects) is optional in the `passwd` object.

_Useful docs:_
https://coreos.github.io/ignition/configuration-v3_3/
https://kubevirt.io/user-guide/virtual_machines/startup_scripts/#cloud-init

Thanks to @DanaOrr for her suggestions on https://github.com/kubevirt-ui/kubevirt-plugin/pull/806.

## 🎥 Demo
**Before:**
Displaying "[" as the _User name_ not corresponding to the VM's yaml:
before_user
**After:**
Displaying the _User name_ correctly:
afterr
If more users specified:
users
If no any credentials specified:
after_no_credentials
If both the user name and password specified (not using ignition):
fedora_no_ignition
If `username` or `name` missing (and other credentials present):
no_username

### DIFF
--- a/src/utils/components/Consoles/components/CloudInitCredentials/CloudInitCredentialsContent.tsx
+++ b/src/utils/components/Consoles/components/CloudInitCredentials/CloudInitCredentialsContent.tsx
@@ -4,9 +4,8 @@ import { Trans } from 'react-i18next';
 import { V1VirtualMachineInstance } from '@kubevirt-ui/kubevirt-api/kubevirt';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
 import { getCloudInitCredentials } from '@kubevirt-utils/resources/vmi';
+import { isEmpty } from '@kubevirt-utils/utils/utils';
 import { ClipboardCopy } from '@patternfly/react-core';
-
-import { CLOUD_INIT_MISSING_USERNAME } from '../../utils/constants';
 
 type CloudInitCredentialsContentProps = {
   vmi: V1VirtualMachineInstance;
@@ -14,9 +13,10 @@ type CloudInitCredentialsContentProps = {
 
 const CloudInitCredentialsContent: React.FC<CloudInitCredentialsContentProps> = ({ vmi }) => {
   const { t } = useKubevirtTranslation();
-  const { user, password } = getCloudInitCredentials(vmi);
+  const { users } = getCloudInitCredentials(vmi);
+  const usernameTitle = t('User name: ', { count: users?.length });
 
-  if (!user && !password) {
+  if (isEmpty(users)) {
     return <>{t('No credentials, see operating system documentation for the default username.')}</>;
   }
 
@@ -28,23 +28,28 @@ const CloudInitCredentialsContent: React.FC<CloudInitCredentialsContentProps> = 
         for more information.
       </Trans>
       <div>
-        <strong>{t('User name: ')}</strong>
-        {user || CLOUD_INIT_MISSING_USERNAME}
-
-        {password && (
+        <strong>{usernameTitle}</strong>
+        {users.map((user, index) => (
           <>
-            <strong>{t(' Password: ')} </strong>
+            {user?.name}
+            {index + 1 < users?.length ? ', ' : ''}
 
-            <ClipboardCopy
-              variant="inline-compact"
-              isCode
-              clickTip={t('Copied')}
-              hoverTip={t('Copy to clipboard')}
-            >
-              {password}
-            </ClipboardCopy>
+            {user?.password && (
+              <>
+                <strong>{t(' Password: ')} </strong>
+
+                <ClipboardCopy
+                  variant="inline-compact"
+                  isCode
+                  clickTip={t('Copied')}
+                  hoverTip={t('Copy to clipboard')}
+                >
+                  {user.password}
+                </ClipboardCopy>
+              </>
+            )}
           </>
-        )}
+        ))}
       </div>
     </>
   );
--- a/src/utils/resources/vmi/utils/cloud-init.ts
+++ b/src/utils/resources/vmi/utils/cloud-init.ts
@@ -1,23 +1,47 @@
+import { load } from 'js-yaml';
+
 import { V1VirtualMachineInstance } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import { CLOUD_INIT_MISSING_USERNAME } from '@kubevirt-utils/components/Consoles/utils/constants';
+
+type CloudinitUserDataObject = {
+  user?: string;
+  password?: string;
+  passwd?: { users: { name: string }[] };
+};
 
 export const getCloudInitCredentials = (
   vmi: V1VirtualMachineInstance,
-): { user: string; password: string } => {
+): { users: { name?: string; password?: string }[] } => {
   const cloudInitVolume = vmi?.spec?.volumes?.find(
     (volume) => volume?.cloudInitNoCloud || volume?.cloudInitConfigDrive,
   );
+  const cloudInitDataSource =
+    cloudInitVolume?.cloudInitNoCloud || cloudInitVolume?.cloudInitConfigDrive;
 
-  const cloudInit = cloudInitVolume?.cloudInitNoCloud || cloudInitVolume?.cloudInitConfigDrive;
+  try {
+    const userDataObject: CloudinitUserDataObject = load(cloudInitDataSource?.userData);
 
-  const cloudInitValuesStrings = cloudInit?.userData
-    ?.split('\n')
-    ?.filter((row) => !row?.includes('#cloud-config'));
+    if (userDataObject?.user || userDataObject?.password) {
+      return {
+        users: [
+          {
+            name: userDataObject?.user || CLOUD_INIT_MISSING_USERNAME,
+            password: userDataObject?.password,
+          },
+        ],
+      };
+    }
 
-  const password = cloudInitValuesStrings
-    ?.find((row) => row?.includes('password'))
-    ?.split(': ')?.[1];
+    if (userDataObject?.passwd?.users) {
+      return {
+        users: userDataObject?.passwd?.users?.map((userobject) => {
+          return { name: userobject?.name || CLOUD_INIT_MISSING_USERNAME };
+        }),
+      };
+    }
+  } catch (e) {
+    console.error(e);
+  }
 
-  const user = cloudInitValuesStrings?.find((row) => row?.includes('user'))?.split(': ')?.[1];
-
-  return { user, password };
+  return { users: [] };
 };
--- a/locales/en/plugin__kubevirt-plugin.json
+++ b/locales/en/plugin__kubevirt-plugin.json
@@ -890,6 +890,7 @@
   "User": "User",
   "User Name": "User Name",
   "User name: ": "User name: ",
+  "User name: _plural": "User name: ",
   "User permissions": "User permissions",
   "Utilization": "Utilization",
   "Value": "Value",
--- a/src/utils/components/UserCredentials/useSSHCommand.ts
+++ b/src/utils/components/UserCredentials/useSSHCommand.ts
@@ -15,7 +15,8 @@ const useSSHCommand = (
 ): useSSHCommandResult => {
   const consoleHostname = window.location.hostname; // fallback to console hostname
 
-  const { user } = getCloudInitCredentials(vmi);
+  const { users } = getCloudInitCredentials(vmi);
+  const user = users?.[0]?.name;
   const sshServicePort = getSSHNodePort(sshService);
 
   let command = 'ssh ';


hstastna:

src/utils/resources/vmi/utils/cloud-init.ts
```diff
@@ -1,23 +1,38 @@
...
 
-  const cloudInit = cloudInitVolume?.cloudInitNoCloud || cloudInitVolume?.cloudInitConfigDrive;
+  // KubeVirt supports the cloud-init NoCloud and ConfigDrive data sources
+  const cloudInitDataSource =
+    cloudInitVolume?.cloudInitNoCloud || cloudInitVolume?.cloudInitConfigDrive;
 
-  const cloudInitValuesStrings = cloudInit?.userData
-    ?.split('\n')
-    ?.filter((row) => !row?.includes('#cloud-config'));
+  const userDataObject: CloudinitUserDataObject = load(cloudInitDataSource?.userData);
```
metalice, line 22: I think its missing the try-catch block as the load can throw an exception when failed to parse

src/utils/resources/vmi/utils/cloud-init.ts
```diff
@@ -1,23 +1,38 @@
...
 export const getCloudInitCredentials = (
   vmi: V1VirtualMachineInstance,
-): { user: string; password: string } => {
+): { user?: string; password?: string; users?: string[] } => {
   const cloudInitVolume = vmi?.spec?.volumes?.find(
     (volume) => volume?.cloudInitNoCloud || volume?.cloudInitConfigDrive,
   );
 
-  const cloudInit = cloudInitVolume?.cloudInitNoCloud || cloudInitVolume?.cloudInitConfigDrive;
+  // KubeVirt supports the cloud-init NoCloud and ConfigDrive data sources
```
metalice, line 18: im not sure this comment is needed :)

src/utils/resources/vmi/utils/cloud-init.ts
```diff
@@ -1,23 +1,38 @@
...
+type CloudinitUserDataObject = {
+  user?: string;
+  password?: string;
+  passwd?: { users: { name: string }[] };
+};
+
 export const getCloudInitCredentials = (
   vmi: V1VirtualMachineInstance,
-): { user: string; password: string } => {
+): { user?: string; password?: string; users?: string[] } => {
```
metalice, line 13: users?: {name: string}[]

src/utils/resources/vmi/utils/cloud-init.ts
```diff
@@ -1,23 +1,38 @@
...
+    return {
+      user: userDataObject?.user,
+      password: userDataObject?.password,
+    };
+  }
 
-  const user = cloudInitValuesStrings?.find((row) => row?.includes('user'))?.split(': ')?.[1];
+  if (userDataObject?.passwd?.users) {
+    return {
+      users: userDataObject?.passwd?.users?.map((userobject: { name: string }) => userobject?.name),
```
metalice, line 33: { name: string } - i think can be removed as the declaration should be in the function type

src/utils/components/Consoles/components/CloudInitCredentials/CloudInitCredentialsContent.tsx
```diff
@@ -28,8 +30,12 @@ const CloudInitCredentialsContent: React.FC<CloudInitCredentialsContentProps> =
         for more information.
       </Trans>
       <div>
-        <strong>{t('User name: ')}</strong>
-        {user || CLOUD_INIT_MISSING_USERNAME}
+        {usernames.map((username) => (
```
metalice, line 33: it's nice to show all the possible users, but if, for example, we will have 10 users, it will be a long list on separate lines... I'm not sure if we want to show them all or simply the first one. could you please consult with Yifat? 

Another thing, currently, the appearance is user: XXX password: XXX, in a single line from the last screenshot in the PR I can see its now separate in 2 lines. please check with her also this. - or just keep it consist that if its a single user it will be single line as its now

metalice (commented):
Hi @hstastna , nice, left some comments :) 👍

src/utils/resources/vmi/utils/cloud-init.ts
hstastna: It is not needed but I liked to add it as it makes clearer what the following code is about ;)

src/utils/resources/vmi/utils/cloud-init.ts
hstastna: Yes. I was not sure if I should convert `user` to array of strings, it would be confusing IMHO. Because we can get user and passowrd, but also list of users from here.. and so we need to deal with that somehow. Or I could rename `user` to `users`? And then simply to return` {users: string[], passord: string}`. WDYT? But I am not sure, it looks confusing that there can be many users but only one password...

src/utils/components/Consoles/components/CloudInitCredentials/CloudInitCredentialsContent.tsx
hstastna: Haven't contacted Yifat yet but also I think we don't have a design for this, and this is why we need to discuss the possible solutions.

Regarding another thing, username and password were already displayed the same way before I made any code changes, How to display them, that's another question for @yfrimanm. But I think we are fine because when both the username and password are present, there are no more any usernames, so no any problem about how to display it, username and password can be in a separate lines - better readability. Only when there is `passwd` object specified in the yaml, there can be more users and various credentials specified for them in the yaml (but not `password`).

src/utils/resources/vmi/utils/cloud-init.ts
metalice: when looking on line 15, i think this self explanatory :) volume?.cloudInitNoCloud || volume?.cloudInitConfigDrive

src/utils/resources/vmi/utils/cloud-init.ts
metalice: if you want to only use users , I would suggest doing users: {name: string, password: string | null} , and on rendering, if no password, don't render password field

src/utils/resources/vmi/utils/cloud-init.ts
hstastna: Well, the implementation that followed that line was not so self explanatory

src/utils/resources/vmi/utils/cloud-init.ts
hstastna: So do you suggest me to rename `user` to `users` and so to return only `users` as the array with objects that contain name and password, right? I can do it so. Just for me personally I'd better go for the way it is now, because it represents the yaml structure/data much better: there can be `user` and `password`.. or `users` - array of objects with `name`, and the rest is optional there.

src/utils/resources/vmi/utils/cloud-init.ts
hstastna: 👍🏽 Thanks I'll checked that. I think I just added { name: string } there because of some eslint warning I got or something like that.

src/utils/resources/vmi/utils/cloud-init.ts
metalice: yaml structure is nice to keep if we don't need to do hacks/workarounds on logic. to use only users, not as optional is a much better solution; it prevents us from making login assumptions on .jsx files and makes a complete separation between login and simple rendering files. so yes, much prefer to use only single users[{name: string; password?: string}] :-)

metalice: we can prevent from this logic on rendering side , const usernames = users || [user || CLOUD_INIT_MISSING_USERNAME]; 

you can also move CLOUD_INIT_MISSING_USERNAME , usage into this function here

hstastna: I think I will leave `CLOUD_INIT_MISSING_USERNAME` in `CloudInitCredentialsContent` component where already is to be able to render "User name: " only when there really is some. Because IMO displaying this would not be nice:
`User name: No credentials, see operating system documentation for the default username. `

src/utils/components/Consoles/components/CloudInitCredentials/CloudInitCredentialsContent.tsx
hstastna: Found that I was mistaken and you were correct: originally for one user and password, both strings were displayed on one line. So I am fixing this, too. Thanks!

src/utils/resources/vmi/utils/cloud-init.ts
```diff
@@ -1,23 +1,47 @@
...
+    const userDataObject: CloudinitUserDataObject = load(cloudInitDataSource?.userData);
 
-  const password = cloudInitValuesStrings
-    ?.find((row) => row?.includes('password'))
-    ?.split(': ')?.[1];
+    if (userDataObject?.user || userDataObject?.password) {
+      return {
+        users: [
+          {
+            name: userDataObject?.user || CLOUD_INIT_MISSING_USERNAME,
```
metalice, line 28: nice :) 👍

src/utils/resources/vmi/utils/cloud-init.ts
```diff
@@ -1,23 +1,47 @@
...
+          },
+        ],
+      };
+    }
 
-  const user = cloudInitValuesStrings?.find((row) => row?.includes('user'))?.split(': ')?.[1];
+    if (userDataObject?.passwd?.users) {
+      return {
+        users: userDataObject?.passwd?.users?.map((userobject) => {
+          return { name: userobject?.name };
```
metalice, line 38: i think we should also have here || CLOUD_INIT_MISSING_USERNAME

src/utils/resources/vmi/utils/cloud-init.ts
```diff
@@ -1,23 +1,47 @@
...
+type CloudinitUserDataObject = {
+  user?: string;
+  password?: string;
+  passwd?: { users: { name: string }[] };
+};
 
 export const getCloudInitCredentials = (
   vmi: V1VirtualMachineInstance,
-): { user: string; password: string } => {
+): { users: { name?: string; password?: string }[] } => {
```
metalice, line 14: once you add CLOUD_INIT_MISSING_USERNAME to line 38, the name isn't optional but guaranteed, which is much better,  and can remove the?

src/utils/resources/vmi/utils/cloud-init.ts
hstastna: I am not sure, I'd better not to do it, as the `users` array is optional in the `passwd` object, and also because `CLOUD_INIT_MISSING_USERNAME` was originally used for the case when the username was missing while the password was present - which is not the case for the `passwd` object.

src/utils/resources/vmi/utils/cloud-init.ts
hstastna: I don't think so. if we end up [here](https://github.com/kubevirt-ui/kubevirt-plugin/pull/807/files#diff-3876c1f1ef780f9cb2697888d5543d57d66c5f3ea540bfccb3d9c2e70d5f2a16R35) we don't have it guaranteed, and I also think we don't want, because as I've mentioned in some other comment: `CLOUD_INIT_MISSING_USERNAME` was originally used for the case when the username was missing while the password was present - which is not the case for the `passwd` object.

src/utils/resources/vmi/utils/cloud-init.ts
metalice: Is it possible to have only sshkey without a name? If so we will present a nullish value.

hstastna: I think it is not possible. If already some user present in `users` array, the `name` is required, the other stuff is optional:
https://coreos.github.io/ignition/configuration-v3_3/

metalice: I tried to run a machine with the same YAML in the bug and just removed the name, as you can see, the machine is running without any issue, which means there isn't a check for name guaranteed. We should ad a default value in this case, as we will get a User name: with an empty value... 
image

src/utils/components/Consoles/components/CloudInitCredentials/CloudInitCredentialsContent.tsx
```diff
@@ -28,23 +27,27 @@ const CloudInitCredentialsContent: React.FC<CloudInitCredentialsContentProps> =
...
-              variant="inline-compact"
-              isCode
-              clickTip={t('Copied')}
-              hoverTip={t('Copy to clipboard')}
-            >
-              {password}
-            </ClipboardCopy>
-          </>
-        )}
+        {users.map((username) => (
```
metalice, line 30: I think more suitiable is a user instead of a username as the username is the actual name. and the user?.name is better understood then username?.name

yfrimanm:
@hstastna I think that in case of more thank 1 user, we can use this
image

gouyang:
> @hstastna I think that in case of more thank 1 user, we can use this. Thanks. image

How often it will be a user array? If it's a array, in my opinion, just show the array there would be better than a new design here. eg:  `username: [coreos, cloud-user, centos]

src/utils/resources/vmi/utils/cloud-init.ts
hstastna: Interesting. I don't understand the reason of why this issue should make the VM not running normally. Maybe it does not matter because the user cannot login without proper credentials anyway. But I'd expect some check from the backend side once the user changes the yaml this way, because it definitely is not correct like this, as the `name` is required there according to the doc (or the doc is not correct?). It shouldn't be possible to have there ssh key without any username. Thank you very much for checking it!

To solve this, IMO we should ignore such setting in the UI and to display that there are no credentials provided, as we usually do. Or to add some new text that some credentials are there but not set correctly. This is a question on someone who knows perfectly, how exactly this part of the UI should work in such cases. Or the UX. And we should raise this issue of not checking the yaml properly also to backend. WDYT?

src/utils/components/Consoles/components/CloudInitCredentials/CloudInitCredentialsContent.tsx
```diff
@@ -4,19 +4,19 @@ import { Trans } from 'react-i18next';
...
-
 type CloudInitCredentialsContentProps = {
   vmi: V1VirtualMachineInstance;
 };
 
 const CloudInitCredentialsContent: React.FC<CloudInitCredentialsContentProps> = ({ vmi }) => {
   const { t } = useKubevirtTranslation();
-  const { user, password } = getCloudInitCredentials(vmi);
+  const { users } = getCloudInitCredentials(vmi);
+  const usernameTitle = users?.length > 1 ? t('User names: ') : t('User name: ');
```
metalice, line 17: i think we can use the auto plurals from i18n , some reference: https://www.i18next.com/translation-function/plurals

the result should be something like this:
image

image

src/utils/resources/vmi/utils/cloud-init.ts
metalice: Opening a bug for BE is always recommended :). Feel free to do so.

We can tackle it in 2 ways, as I see.
1. adding  userobject?.name || CLOUD_INIT_MISSING_USERNAME
2. on line 37, before running the map function over the array, run a filter function and remove all the "users" without names.

In both cases, we are guaranteed to have a name in place, so it results in removing the optional from the name field at line 14.

PR #808, opened 2022-08-02
Bug 2114516: Open website after clicking on Fedora cloud image list link

hstastna:
## 📝 Description

**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2114516

Make the link in the "Example" for Fedora work, when editing template's disk and choosing "URL (creates PVC)" as _Source_ in the _Edit disk_ modal.

The core of the problem was missing some default value for the `os` (in some edge cases when it is `undefined`), comparing to what is returned for `getURLSourceHelpertText` function (used [in the affected component](https://github.com/kubevirt-ui/kubevirt-plugin/blob/main/src/utils/components/DiskModal/DiskFormFields/DiskSourceFormSelect/components/DiskSourceUrlInput.tsx#L14)): Fedora OS related helper text. So the same OS was used as the default one also in the affected `URLSourceHelperText` component.

## 🎥 Demo
_Before:_
Nothing happens after clicking on **_Fedora cloud image list_** link.
fedora_before
_After:_
The corresponding website is opened:
fedora_after


hstastna:

PR #824, opened 2022-08-09
Bug 2096780: template create secret object

upalatucci:
## 📝 Description

Same thing as sysprep. 
Currently, the user can attach an existing secret (in the same namespace).
With this pr, the user would be able to add an ssh secret key into the template objects array.
A new secret will be created along side the vm 

## 🎥 Demo

Screenshot from 2022-08-09 11-59-38
Screenshot from 2022-08-09 11-59-21
Screenshot from 2022-08-09 11-59-16


package.json
```diff
@@ -91,6 +91,7 @@
     "react-i18next": "^11.14.3",
     "react-router-dom": "5.x",
     "react-tagsinput": "^3.19.0",
+    "reselect": "^4.1.6",
```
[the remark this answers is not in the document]
metalice: have u tried to install it as peer dep or dev dep?

src/utils/components/AuthorizedSSHKeyModal/AuthorizedSSHKeyModal.tsx
```diff
@@ -36,9 +36,9 @@ export const AuthorizedSSHKeyModal: React.FC<{
   const [isValidatedKey, setIsValidatedKey] = React.useState<boolean>(true);
 
   const submitHandler = React.useCallback(async () => {
-    if (selectedSecretName) await onSubmit(selectedSecretName);
-    else await onSubmit(undefined, value);
-  }, [onSubmit, selectedSecretName, value]);
+    if (createSecretOpen) await onSubmit(undefined, value);
+    else await onSubmit(selectedSecretName);
```
metalice, line 40: `createSecretOpen) ? await onSubmit(undefined, value) : onSubmit(selectedSecretName)`

src/views/templates/details/tabs/scripts/components/SSHKey/sshkey-utils.ts
```diff
@@ -0,0 +1,106 @@
...
+    name: secretName,
+    namespace: namespace,
+  },
+  data: { key: btoa(sshKey) },
+});
+
+export const updateSecretName = (template: V1Template, secretName: string) => {
+  const vm = getTemplateVirtualMachineObject(template);
+
+  template.objects = template.objects.map((object) =>
```
metalice, line 30: please add ?.

src/views/templates/details/tabs/scripts/components/SSHKey/sshkey-utils.ts
```diff
@@ -0,0 +1,106 @@
...
+export const removeSecretObject = (template: V1Template, secretName: string) => {
+  template.objects = (template?.objects || []).filter(
+    (object) => object?.metadata?.name !== secretName,
+  );
+};
+
+export const removeCredential = (template: V1Template, secretName: string) => {
+  const vm = getTemplateVirtualMachineObject(template);
+
+  vm.spec.template.spec.accessCredentials = vm.spec.template.spec.accessCredentials.filter(
```
metalice, line 64: please add ?.

src/views/templates/details/tabs/scripts/tests/mocks.ts
```diff
@@ -0,0 +1,5 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+import { V1Template } from '@kubevirt-ui/kubevirt-api/console';
+
+export const getMockTemplate = (): V1Template =>
+  JSON.parse(JSON.stringify(require('./data/template.json')));
```
metalice, line 5: if i remember correctly, json.* may throw exception, if so, please surround with try catch block

src/views/templates/details/tabs/scripts/tests/sshkey-utils.test.ts
```diff
@@ -0,0 +1,105 @@
...
+import { getMockTemplate } from './mocks';
+
+jest.mock('@openshift-console/dynamic-plugin-sdk', () => ({
+  k8sUpdate: jest.fn(({ data: template }) => Promise.resolve(template)),
+}));
+
+const template = getMockTemplate();
+const someExternalSecretName = 'some-external-secret';
+
+const sshkey = 'testtest';
```
metalice, line 16: ?

upalatucci: Does not matter what string is. I just throw it some text

src/views/templates/details/tabs/scripts/tests/mocks.ts
upalatucci: Yes. But it's a function for testing. If thrown an exception because there is some issue, is a good thing right?

package.json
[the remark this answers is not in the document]
upalatucci: No, I just used `yarn add reselect`. 
But should work also as peer dep or dev dep.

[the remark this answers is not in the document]
upalatucci: The issue here is that the SDK uses this module but is not declared in its package json. With the new version, this bug is solved but the plugin does not compile for other reasons. So as long as the package `reselect` is in the node_modules, it's fine.

src/views/templates/details/tabs/scripts/tests/mocks.ts
metalice: the test should be checking the logic of components if JSON fails to parse because data integrity that is hard coded locally isn't related to the test itself :) 
but also, because its belongs to the test and it's hard coded I don't mind leaving it as it.. :) so up to u...

package.json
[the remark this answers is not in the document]
metalice: maybe u can try to add this to peer deps? i will be easier for us to remember to remove it once its not needed

src/views/templates/details/tabs/scripts/components/SSHKey/SSHKey.tsx
```diff
@@ -0,0 +1,91 @@
...
+                    createModal((modalProps) => (
+                      <AuthorizedSSHKeyModal
+                        {...modalProps}
+                        sshKey={secretKey}
+                        vmSecretName={externalSecretName}
+                        onSubmit={onSSHChange}
+                      />
+                    ))
+                  }
+                  variant="link"
```
metalice, line 69: please use ButtonVaritaions from pf

PR #831, opened 2022-08-11
Bug 2096780: Wizard with ssh key in additionalResources

upalatucci:
## 📝 Description

Have the wizard play nice with multiple object templates.

SSH auth key section in scripts tab now will add and change secret in the `additionalResources` tabsData. 

Why this pr? 
The template can have in the objects array an already configured secret. With this pr will be detected in the wizard.

PR #833, opened 2022-08-11
Bug 2096780: Wizard with sysprep in additionalResources

upalatucci:
## 📝 Description

same as #831  but for sysprep.

User can edit and detect sysprep config using template objects array

### DIFF
--- a/src/views/catalog/wizard/tabs/scripts/components/Sysprep.tsx
+++ b/src/views/catalog/wizard/tabs/scripts/components/Sysprep.tsx
@@ -1,0 +1,111 @@
+import * as React from 'react';
+
+import { produceVMSysprep, useWizardVMContext } from '@catalog/utils/WizardVMContext';
+import { WizardDescriptionItem } from '@catalog/wizard/components/WizardDescriptionItem';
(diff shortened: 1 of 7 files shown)


src/views/catalog/wizard/tabs/scripts/components/Sysprep.tsx
```diff
@@ -0,0 +1,112 @@
...
+  const currentVMSysprepName = currentSysprepVolume?.sysprep?.configMap?.name;
+
+  const filterSysprepByName = isSysprepConfig(currentVMSysprepName);
+
+  const sysPrepObject = tabsData?.additionalObjects?.find(
+    filterSysprepByName,
+  ) as IoK8sApiCoreV1ConfigMap;
+
+  const unattend = sysPrepObject?.data?.[UNATTEND];
+  const autoUnattend = sysPrepObject?.data?.[AUTOUNATTEND];
```
metalice, line 41: can be in a single line

metalice, line 41: can be in a single line

upalatucci: Okay great

PR #838, opened 2022-08-18
Create resources inventory card

pcbailey:
## 📝 Description

This PR adds the new resources inventory card. The original inventory card will be removed in a future PR.

### Admin version
Selection_161

### Non-admin version
Selection_142


src/views/clusteroverview/OverviewTab/resource-counts-card/ResourceCountsCard.scss
```diff
@@ -0,0 +1,30 @@
...
+    color: var(--co-global--palette--blue-400);
+    font-size: var(--pf-global--FontSize--4xl);
+    text-align: center;
+    margin-bottom: -10px;
+  }
+
+  &__count-name {
+    text-align: center;
+  }
+}
```
metalice, line 30: can we use var values here? instead of absolute numbers?

src/views/clusteroverview/OverviewTab/resource-counts-card/ResourceCountsCard.tsx
```diff
@@ -0,0 +1,67 @@
+import * as React from 'react';
```
metalice, line 1: we can now import simply React no need for * as React anymore, and then name import {useState}, etc..

src/views/clusteroverview/OverviewTab/resource-counts-card/ResourceCountsCard.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+import useInventoryCounts from './hooks/useInventoryCounts';
+import ResourceCountItem from './ResourceCountItem';
+
+import './ResourceCountsCard.scss';
+
+const ResourceCountsCard: React.FC = () => {
+  const { t } = useKubevirtTranslation();
+  const isAdmin = useIsAdmin();
+  const countsData = useInventoryCounts();
```
metalice, line 19: i think u can simply deconstruct here vms, temples, etc.. instead of countsData?....

src/views/clusteroverview/OverviewTab/resource-counts-card/hooks/useInventoryCounts.ts
```diff
@@ -0,0 +1,56 @@
+import * as React from 'react';
```
metalice, line 1: as previous comment :)

src/views/clusteroverview/OverviewTab/resource-counts-card/hooks/useInventoryCounts.ts
```diff
@@ -0,0 +1,56 @@
...
+    [resources, isAdmin],
+  );
+
+  const nads = React.useMemo(
+    () =>
+      isAdmin
+        ? resources?.nads?.data
+        : getAllowedResourceData(resources, NetworkAttachmentDefinitionModel)?.data,
+    [resources, isAdmin],
+  );
```
metalice, line 43: it looks like the same function four times when only the field [vms, templates, nads, nodes] and model are different.
maybe let's create a function that receives field and model?
or something like that? - by returning only the length, you can save the checks of ?.length || 0 on the return statement and simply spread this object with a loaded prop.

```suggestion
const resourcesData = {vms: VirtualMachineModel, templates: TemplateModel, nads: NetworkAttachmentDefinitionModel, nodes: NodeModel};

  Object.entries(resourcesData).reduce((acc, [field, model])=>{
  const data = isAdmin ? resources?.[field]?.data : getAllowedResourceData(resources, model)?.data
  acc[field] = data?.length || 0;
  return acc;
  },{})
```

pcbailey:
/hold

Holding until https://github.com/openshift/console/pull/12001 passes and the namespace bar can be added to the overview page.

src/views/clusteroverview/OverviewTab/resource-counts-card/hooks/useInventoryCounts.ts
pcbailey: I had to update that logic, anyway. I still need to do a last look through it to see if I missed anything, but would you mind giving it another look?

src/views/clusteroverview/OverviewTab/resource-counts-card/hooks/useInventoryCounts.ts
pcbailey: Fixed!

src/views/clusteroverview/OverviewTab/resource-counts-card/ResourceCountsCard.tsx
pcbailey: Fixed!

src/views/clusteroverview/OverviewTab/resource-counts-card/ResourceCountsCard.tsx
pcbailey: I'm good with it. =) When did we start doing that? Fixed!

src/views/clusteroverview/OverviewTab/resource-counts-card/ResourceCountsCard.scss
pcbailey: Fixed!

src/views/clusteroverview/OverviewTab/resources-inventory-card/ResourcesInventoryCard.tsx
```diff
@@ -0,0 +1,75 @@
...
+const ResourcesInventoryCard: React.FC = () => {
+  const { t } = useKubevirtTranslation();
+  const isAdmin = useIsAdmin();
+
+  const [activeNamespace] = useActiveNamespace();
+  const namespace = React.useMemo(
+    () => (activeNamespace === ALL_NAMESPACES_SESSION_KEY ? null : activeNamespace),
+    [activeNamespace],
+  );
+  const { networks, nodes, vms, vmTemplates } = useResourcesQuantities(namespace);
```
metalice, line 27: im guessing this hook can also consume useactivename space hook and no need to transfer namespace

src/views/clusteroverview/OverviewTab/resources-inventory-card/hooks/useResourcesQuantities.ts
```diff
@@ -0,0 +1,74 @@
...
+        return acc;
+      },
+      { loaded: false, networks: 0, nodes: 0, vms: 0, vmTemplates: 0 },
+    );
+
+    resourceQuantities.loaded = Object.values(resources).every((resource) => resource?.loaded);
+
+    return resourceQuantities;
+  }, [resources]);
+};
```
metalice, line 72:
```suggestion
  return useMemo(() => {
    return Object.entries([key, values]).reduce(
      (acc, key) => {
        acc[key] = values?.data?.length;
        acc["loaded"] = acc["loaded"] ? acc["loaded"] && values?.loaded : values?.loaded
        return acc;
      },
      { loaded: false, networks: 0, nodes: 0, vms: 0, vmTemplates: 0 },
    );
  }, [resources]);
};
```

something like that?

src/views/clusteroverview/OverviewTab/resources-inventory-card/ResourcesInventoryCard.tsx
pcbailey: Good point. Somehow, I didn't notice that I wasn't using `namespace` anywhere else in this file. Thanks for catching it!

src/views/clusteroverview/OverviewTab/resources-inventory-card/hooks/useResourcesQuantities.ts
pcbailey: Much cleaner, thanks. I short-circuited the logic to:
`acc.loaded = acc.loaded && value.loaded`
since the only combination that would make `acc.loaded` true is if both values are true.

metalice: can it be possible resources is only 1 field object? if so, I'm guessing only acc.loaded && value.loaded isn't enough, but if guaranteed that resources have at least 2 fields, it will be ok..

metalice:
nice code clean! thumbs up!
/lgtm

PR #842, opened 2022-08-22
Remove reselect dev dep

upalatucci:
## 📝 Description

Remove reselect dev dep added in the past as a workaround for one sdk bug


package.json
```diff
@@ -153,7 +152,7 @@
     "tsconfig-paths-webpack-plugin": "^3.5.1",
     "typescript": "^4.5.5",
     "webpack": "^5.68.0",
-    "webpack-cli": "4.9.x",
+    "webpack-cli": "4.10.x",
```
upalatucci, line 155: After the upgrade i got a webpack error ```[webpack-cli] TypeError: cli.isMultipleCompiler is not a function```

Solved upgrading this

webpack.config.ts
```diff
@@ -73,7 +73,8 @@ const config: Configuration = {
       },
       {
         test: /\.scss$/,
-        exclude: /node_modules\/(?!(@patternfly|@openshift-console\/plugin-shared)\/).*/,
+        exclude:
```
upalatucci, line 76: now also `dynamic-plugin-sdk` has some scss files

upalatucci (commented):
for reviewers

webpack.config.ts
```diff
@@ -73,7 +73,8 @@ const config: Configuration = {
...
+          /node_modules\/(?!(@patternfly|@openshift-console\/plugin-shared|@openshift-console\/dynamic-plugin-sdk)\/).*/,
```
upalatucci, line 77: now also `dynamic-plugin-sdk` has some scss files

avivtur (commented):
Hi looks good :) 
we also have `@openshift-console/dynamic-plugin-sdk-webpack` version 0.0.8 and `@openshift-console/dynamic-plugin-sdk-internal` version 0.11 that we can upgrade to, can you please add them as well?

upalatucci:
Sure

PR #845, opened 2022-08-23
CNV-19687: Sidebar editor

upalatucci:
## 📝 Description

Form - yaml editor side by side.
The user should be able to see how the changes in the forms affect the yaml

We have here 3 new components.

Context, Switch, and Editor

Follow-up pr would add this feature also to the wizard and to the templates

## 🎥 Demo

Screenshot from 2022-09-02 15-31-26
Screenshot from 2022-09-02 15-31-20
Screenshot from 2022-09-02 15-31-13

### DIFF
--- a/src/utils/components/SidebarEditor/SidebarEditor.tsx
+++ b/src/utils/components/SidebarEditor/SidebarEditor.tsx
@@ -1,0 +1,97 @@
+import React, { ReactNode, Suspense, useContext, useMemo, useState } from 'react';
+import { dump } from 'js-yaml';
+
+import { K8sResourceCommon, YAMLEditor } from '@openshift-console/dynamic-plugin-sdk';
+import {
+  Button,
+  ButtonVariant,
+  Flex,
+  FlexItem,
+  Sidebar,
+  SidebarContent,
+  SidebarPanel,
+  Stack,
+  StackItem,
+} from '@patternfly/react-core';
+
+import Loading from '../Loading/Loading';
+
+import { SidebarEditorContext } from './SidebarEditorContext';
+import { safeLoad } from './utils';
+
+import './sidebar-editor.scss';
+
+type SidebarEditorProps<Resource> = {
+  resource: Resource;
+  onResourceUpdate?: (newResource: Resource) => Promise<Resource | void>;
+  children: (resource: Resource) => ReactNode;
+};
+
+const SidebarEditor = <Resource extends K8sResourceCommon>({
+  children,
+  resource,
+  onResourceUpdate,
+}: SidebarEditorProps<Resource>): JSX.Element => {
+  const [editableYAML, setEditableYAML] = useState('');
+  const resourceYAML = useMemo(() => {
+    const yaml = dump(resource, { skipInvalid: true });
+    setEditableYAML(yaml);
+    return yaml;
+  }, [resource]);
+
+  const { showEditor } = useContext(SidebarEditorContext);
+  const editedResource = safeLoad<Resource>(editableYAML);
+
+  const changeResource = (newValue: string) => {
+    setEditableYAML(newValue);
+
+    return {};
+  };
+
+  return (
+    <Sidebar isPanelRight hasGutter hasNoBackground className="sidebar-editor">
+      <SidebarContent>{children && children(editedResource ?? resource)}</SidebarContent>
+      {showEditor && (
+        <SidebarPanel
+          width={{ default: 'width_33', lg: 'width_50', xl: 'width_50' }}
+          className="sidebar-editor"
+        >
+          <Stack hasGutter>
+            <StackItem isFilled>
+              <Suspense fallback={<Loading />}>
+                <YAMLEditor
+                  value={editableYAML}
+                  minHeight="300px"
+                  onChange={changeResource}
+                  onSave={() => onResourceUpdate(editedResource)}
+                />
+              </Suspense>
+            </StackItem>
+            <StackItem>
+              <Flex>
+                <FlexItem>
+                  <Button
+                    variant={ButtonVariant.primary}
+                    onClick={() => onResourceUpdate(editedResource)}
+                  >
+                    Save
+                  </Button>
+                </FlexItem>
+                <FlexItem>
+                  <Button
+                    variant={ButtonVariant.secondary}
+                    onClick={() => setEditableYAML(resourceYAML)}
+                  >
+                    Reload
+                  </Button>
+                </FlexItem>
+              </Flex>
+            </StackItem>
+          </Stack>
+        </SidebarPanel>
+      )}
+    </Sidebar>
+  );
+};
+
+export default SidebarEditor;
--- a/src/utils/components/SidebarEditor/SidebarEditorContext.tsx
+++ b/src/utils/components/SidebarEditor/SidebarEditorContext.tsx
@@ -1,0 +1,20 @@
+import React, { createContext, FC, useState } from 'react';
+
+export type SidebarEditorContextType = {
+  showEditor: boolean;
+  setEditorVisible?: (editorVisible: boolean) => void;
+};
+
+export const SidebarEditorContext = createContext<SidebarEditorContextType>({
+  showEditor: false,
+});
+
+export const SidebarEditorProvider: FC = ({ children }) => {
+  const [showEditor, setShowEditor] = useState(false);
+
+  return (
+    <SidebarEditorContext.Provider value={{ showEditor, setEditorVisible: setShowEditor }}>
+      {children}
+    </SidebarEditorContext.Provider>
+  );
+};
--- a/src/utils/components/SidebarEditor/SidebarEditorSwitch.tsx
+++ b/src/utils/components/SidebarEditor/SidebarEditorSwitch.tsx
@@ -1,0 +1,23 @@
+import React, { FC, memo, useContext } from 'react';
+
+import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
+import { Switch } from '@patternfly/react-core';
+
+import { SidebarEditorContext } from './SidebarEditorContext';
+
+const SidebarEditorSwitch: FC = memo(() => {
+  const { t } = useKubevirtTranslation();
+  const { showEditor, setEditorVisible } = useContext(SidebarEditorContext);
+
+  return (
+    <Switch
+      id="sidebar-editor-switch"
+      label={t('YAML: ON')}
+      labelOff={t('YAML: OFF')}
+      isChecked={showEditor}
+      onChange={setEditorVisible}
+    />
+  );
+});
+
+export default SidebarEditorSwitch;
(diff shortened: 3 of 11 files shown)


src/utils/components/SidebarEditor/SidebarEditor.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+const SidebarEditor: React.FC<SidebarEditorProps> = ({ children, resource }) => {
+  const { showEditor } = React.useContext(SidebarEditorContext);
+  return (
+    <Sidebar isPanelRight hasGutter hasNoBackground className="sidebar-editor">
+      <SidebarContent>{children}</SidebarContent>
+      {showEditor && (
+        <SidebarPanel
+          width={{ default: 'width_33', lg: 'width_50', xl: 'width_50' }}
+          style={{ minHeight: '100%' }}
```
metalice, line 29: cant be in the classname ?

src/utils/components/SidebarEditor/SidebarEditor.tsx
```diff
@@ -0,0 +1,41 @@
...
+          className="sidebar-editor"
+        >
+          <React.Suspense fallback={<Loading />}>
+            <YAMLEditor value={dump(resource)} minHeight="300px" />
```
metalice, line 33: dump may result in exception, should be called outside with try catch block

src/utils/components/SidebarEditor/SidebarEditorContext.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+
+export type SidebarEditorContextType = {
+  showEditor: boolean;
+  setEditorVisible?: (editorVisible: boolean) => void;
+};
+
+export const SidebarEditorContext = React.createContext<SidebarEditorContextType>({
```
metalice, line 8: why do we need context for this component?

metalice (commented):
a few months ago we updated tsconfig, so we can import react normally and not * as react , so no need anymore for Reac.useMemo or React.useCallback, you can simply import useMemo or useCallback

metalice:
is the editor it not editable? what happens if a user change values in YAML? will it reflect in the form?

upalatucci:
The editor is editable, unfortunately. I would like not to be like that. As it is the values would not be reflected in the form.

upalatucci:
aaah but there is a readonly option that I can use!

src/utils/components/SidebarEditor/SidebarEditor.tsx
upalatucci: we'll use dump with `skipInvalid` option

src/utils/components/SidebarEditor/utils.ts
```diff
@@ -0,0 +1,10 @@
+import { load } from 'js-yaml';
+
+export const safeLoad = <T>(value: string): T | undefined => {
```
metalice, line 3: what about safeDump?
if it's a generic type, I don't think nested util file under local utils is the correct place, as here it is always Resource type

src/utils/components/SidebarEditor/sidebar-editor.scss
```diff
@@ -0,0 +1,21 @@
...
+  .pf-c-sidebar__main {
+    min-height: 100%;
+  }
+
+  .pf-c-sidebar__main {
+    align-items: stretch;
+  }
+
+  .pf-c-sidebar__panel {
+    padding: 4px;
```
metalice, line 14: can be changed to a var?

src/utils/components/SidebarEditor/SidebarEditorSwitch.tsx
```diff
@@ -0,0 +1,23 @@
...
+  const { t } = useKubevirtTranslation();
+  const { showEditor, setEditorVisible } = React.useContext(SidebarEditorContext);
+
+  return (
+    <Switch
+      id="sidebar-editor-switch"
+      label={t('YAML: ON')}
+      labelOff={t('YAML: OFF')}
+      isChecked={showEditor}
+      onChange={(checked) => setEditorVisible(checked)}
```
metalice, line 18: onChange={setEditorVisible}

src/utils/components/SidebarEditor/SidebarEditor.tsx
```diff
@@ -0,0 +1,95 @@
...
+    reloadResource();
+  }, [reloadResource]);
+
+  const changeResource = (newValue: string) => {
+    setEditableYAML(newValue);
+
+    return {};
+  };
+
+  const edited = editableYAML !== resourceYAML;
```
metalice, line 54: will calculate on each render.. also if used only once you can simply use it at line 76

src/utils/components/SidebarEditor/SidebarEditor.tsx
```diff
@@ -0,0 +1,95 @@
...
+  resource,
+  onResourceUpdate,
+}: SidebarEditorProps<Resource>): JSX.Element => {
+  const resourceYAML = React.useMemo(() => dump(resource, { skipInvalid: true }), [resource]);
+  const { showEditor } = useContext(SidebarEditorContext);
+  const [editableYAML, setEditableYAML] = useState('');
+  const editedResource = safeLoad<Resource>(editableYAML);
+
+  const reloadResource = useCallback(() => {
+    setEditableYAML(resourceYAML);
```
metalice, line 41: can be added to line 35

src/utils/components/SidebarEditor/SidebarEditor.tsx
```diff
@@ -0,0 +1,95 @@
...
+  }, [resourceYAML]);
+
+  useEffect(() => {
```
metalice, line 44: can be removed if reloadResource callback is added at line 35

src/utils/components/SidebarEditor/SidebarEditor.tsx
```diff
@@ -0,0 +1,95 @@
...
+    reloadResource();
+  }, [reloadResource]);
+
+  const changeResource = (newValue: string) => {
+    setEditableYAML(newValue);
+
+    return {};
+  };
```
metalice, line 52: no need, can simply declare at line 67

src/utils/components/SidebarEditor/SidebarEditor.tsx
```diff
@@ -0,0 +1,95 @@
...
+      <SidebarContent>{children && children(editedResource ?? resource)}</SidebarContent>
+      {showEditor && (
+        <SidebarPanel
+          width={{ default: 'width_33', lg: 'width_50', xl: 'width_50' }}
+          className="sidebar-editor"
+        >
+          <Stack hasGutter>
+            <StackItem isFilled>
+              <React.Suspense fallback={<Loading />}>
+                <YAMLEditor value={editableYAML} minHeight="300px" onChange={changeResource} />
```
metalice, line 67: onChange={setEditableYAML}

metalice, line 67: can we set min height via classname?

src/utils/components/SidebarEditor/SidebarEditor.tsx
```diff
@@ -0,0 +1,95 @@
...
+    setEditableYAML(newValue);
+
+    return {};
+  };
+
+  const edited = editableYAML !== resourceYAML;
+
+  return (
+    <Sidebar isPanelRight hasGutter hasNoBackground className="sidebar-editor">
+      <SidebarContent>{children && children(editedResource ?? resource)}</SidebarContent>
```
metalice, line 58: there is no check for children function?

upalatucci: Check is in the type props. Typescript will raise an error if SidebarEditor will be used wrongly

src/utils/components/SidebarEditor/SidebarEditor.tsx
upalatucci: unfortunately, the yaml editor has wrong props. onChange and onSave want an object as return value that in the code is not used, for this reason i return an empty object.
I have a pr for this.

src/utils/components/SidebarEditor/SidebarEditor.tsx
upalatucci: same https://github.com/kubevirt-ui/kubevirt-plugin/pull/845#discussion_r968098081

src/utils/components/SidebarEditor/SidebarEditor.tsx
upalatucci: okay. Removed. the save button will be always enabled. Not too bad

src/utils/components/SidebarEditor/SidebarEditor.tsx
upalatucci: no className props. minHeight is used for inline style in the source code combined with react-measure. But I can do it with the parent

src/utils/components/SidebarEditor/utils.ts
upalatucci: the doc about `skipInvalid` says `do not throw on invalid types (like function in the safe schema) and skip pairs and single values with such types.`. Can we trust it?

src/utils/components/SidebarEditor/SidebarEditor.tsx
upalatucci: good catch

src/utils/components/SidebarEditor/utils.ts
upalatucci: `Resource` in SidebarEditor is a generic type. Can be VM, VMI, Template, etc.

src/utils/components/SidebarEditor/SidebarEditor.tsx
```diff
@@ -0,0 +1,97 @@
...
+  children: (resource: Resource) => React.ReactNode;
+};
+
+const SidebarEditor = <Resource extends K8sResourceCommon>({
+  children,
+  resource,
+  onResourceUpdate,
+}: SidebarEditorProps<Resource>): JSX.Element => {
+  const [editableYAML, setEditableYAML] = useState('');
+  const resourceYAML = React.useMemo(() => {
```
metalice, line 36: can use simply useMemo

src/utils/components/SidebarEditor/SidebarEditorContext.tsx
```diff
@@ -0,0 +1,24 @@
...
+  showEditor: boolean;
+  setEditorVisible?: (editorVisible: boolean) => void;
+};
+
+export const SidebarEditorContext = createContext<SidebarEditorContextType>({
+  showEditor: false,
+});
+
+export const SidebarEditorProvider: FC = ({ children }) => {
+  const [showEditor, setEditor] = useState(false);
```
metalice, line 13: usually the set funciton is exact name of the first variable, setShowEditor , also, do u really need the useCallback on line 15? i think you can simply transfer the setShowEditor itself

upalatucci: Fixed

PR #846, opened 2022-08-24
CNV-19298: SSH access with virtctl command

upalatucci:
## 📝 Description

SSH access (VM details page section) with virtctl and ssh commands

Screenshot from 2022-08-24 16-43-49
Screenshot from 2022-08-24 16-43-47


src/utils/components/SSHAccess/utils.ts
```diff
@@ -68,9 +71,21 @@ export const createSSHService = async (vm: V1VirtualMachine, vmi: V1VirtualMachi
...
 };
+
+export const getConsoleVirtctlCommand = (vmName: string, vmNamespace?: string) => {
+  let command = 'virtctl ';
+
+  if (vmNamespace) {
+    command += `-n ${vmNamespace} `;
+  }
+
+  command += vmName;
```
hstastna, line 88: Are you sure about the command `virtctl -n <vmNamespace> <vmName>` ? According to the doc, we are missing ` ssh ` there...

hstastna (commented):
Just one comment below, otherwise great work! 👍🏽

src/utils/components/SSHAccess/utils.ts
upalatucci: Yes! I've installed an old version and there was no ssh command. Now I installed the 0.53 and there is!

src/utils/components/SSHAccess/utils.ts
```diff
@@ -68,9 +71,21 @@ export const createSSHService = async (vm: V1VirtualMachine, vmi: V1VirtualMachi
...
+export const getConsoleVirtctlCommand = (vmName: string, vmNamespace?: string) => {
+  let command = 'virtctl ';
+
+  if (vmNamespace) {
+    command += `-n ${vmNamespace} `;
+  }
+
+  command += `ssh ${vmName}`;
+
+  return command;
```
metalice, line 90: `virtctl {$vmNamespace ? -n ${vmNamespace} : ""} ssh ${vmName}` - not sure if this function is needed

src/utils/components/SSHAccess/utils.ts
```diff
@@ -68,9 +71,21 @@ export const createSSHService = async (vm: V1VirtualMachine, vmi: V1VirtualMachi
 
   const serviceResource = buildSSHServiceFromVM(vm, labelSelector);
 
-  await k8sCreate({
+  return await k8sCreate({
```
metalice, line 74: return await is same as return of regular promise

src/utils/components/SSHAccess/components/SSHCheckbox.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const [sshService, setSSHService] = React.useState(initialSSHService);
+  const [loading, setLoading] = React.useState(false);
+  const [error, setError] = React.useState<Error>();
+
+  const onSSHChange = React.useCallback(
+    async (enableSSH: boolean) => {
+      setLoading(true);
+      try {
+        if (enableSSH) {
+          setSSHService((await createSSHService(vm, vmi)) as IoK8sApiCoreV1Service);
```
metalice, line 37: u are saving a promise.. no point in await here..

metalice: if you arent dealing with the result of await in the same scope it just the same as regular promise

metalice: maybe createahSSHService.then(()=>setSSHService()) ? can work?

metalice: just remove the finally and move it also to .then

src/utils/components/SSHAccess/utils.ts
upalatucci: Surely Hilda will use it in another place for the dropdown action so I thought a function would be useful

hstastna (approved):
LGTM!

PR #852, opened 2022-08-30
Improving virtual machine list performance

metalice:

## 📝 Description

- Added pagination 
- VMI, VMIM mappers - instead of using find on each for every vm
- Removing VMIM request for each VM
- Improve filters results and logic include sorting

## 🎥 Demo


avivtur (commented):
+1 to pagination :rocket:

PR #856, opened 2022-09-01
CNV19691: Add table column management to VM list

avivtur:
## 📝 Description

Adding column management to VM list view
NOTES: 
- The kebab actions columns have no title and it appears as empty in the column management component, we should consider giving this column the `Actions` title so it will look better.
- When moving to a specific namespace and using the management column, check and uncheck to the namespace column will not make any effect, we should consider removing that column when user is under any namespace scope and not the `all-projects` scope.

@metalice @vojtechszocs @hstastna @pcbailey @upalatucci please review
## 🎥 Demo

https://user-images.githubusercontent.com/67270715/187917424-dcbbcf4a-ff3f-4408-b7f0-c99032ae832e.mp4

### DIFF
--- a/src/views/virtualmachines/list/VirtualMachinesList.tsx
+++ b/src/views/virtualmachines/list/VirtualMachinesList.tsx
@@ -69,7 +69,8 @@ const VirtualMachinesList: React.FC<VirtualMachinesListProps> = ({ kind, namespa
       ? history.push(catalogURL)
       : history.push(`/k8s/ns/${namespace || 'default'}/${VirtualMachineModelRef}/~new`);
 
-  const columns = useVirtualMachineColumns(namespace);
+  const [columns, activeColumns] = useVirtualMachineColumns(namespace);
+
   return (
     <>
       <ListPageHeader title={t('VirtualMachines')}>
@@ -83,12 +84,22 @@ const VirtualMachinesList: React.FC<VirtualMachinesListProps> = ({ kind, namespa
           loaded={loaded}
           rowFilters={filters}
           onFilterChange={onFilterChange}
+          columnLayout={{
+            columns: columns?.map(({ id, title, additional }) => ({
+              id,
+              title,
+              additional,
+            })),
+            id: VirtualMachineModelRef,
+            selectedColumns: new Set(activeColumns?.map((col) => col?.id)),
+            type: t('VirtualMachine'),
+          }}
         />
         <VirtualizedTable<K8sResourceCommon>
           data={data}
           unfilteredData={unfilteredData}
           loaded={loaded}
-          columns={columns}
+          columns={activeColumns}
           loadError={loadError}
           Row={VirtualMachineRow}
           rowData={{ kind, vmis }}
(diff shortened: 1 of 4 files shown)


upalatucci (approved):

src/views/virtualmachines/list/hooks/useVirtualMachineColumns.ts
```diff
@@ -66,7 +68,7 @@ const useVirtualMachineColumns = (namespace: string) => {
         props: { className: 'pf-m-width-10' },
       },
       {
-        title: '',
+        title: ' ',
```
hstastna, line 71: I don't understand the reason of this change. IMO it is unnecessary.

hstastna (commented):
> The kebab actions columns have no title and it appears as empty in the column management component, we should consider giving this column the Actions title so it will look better.

I totally agree.
actions

Could we fix this within this PR? Because it looks bad and very confusing in the modal, from the UX perspective, and it shouldn't be a lot of work (to add "Actions" title to the modal).

Maybe we should consider adding a normal title "Actions" for the Actions column, like other columns have. That would make all the related things much easier. And the user would not have to wonder what is there - under those three dots a.k. a. kebab menu, in each of the columns. @yfrimanm WDYT? Thanks!

Otherwise great work, Aviv! 🎇

metalice:
> 

Why give the option to remove the action column? I think it is always necessary.. IMHO :)

src/views/virtualmachines/list/VirtualMachinesList.tsx
```diff
@@ -83,12 +84,22 @@ const VirtualMachinesList: React.FC<VirtualMachinesListProps> = ({ kind, namespa
...
           rowFilters={filters}
           onFilterChange={onFilterChange}
+          columnLayout={{
+            columns: columns?.map(({ id, title, additional }) => ({
+              id,
+              title,
+              additional,
+            })),
+            id: VirtualMachineModelRef,
+            selectedColumns: new Set(activeColumns?.map((col) => col?.id)),
```
metalice, line 94: Do we need set here? It is possible to have same ids ?

yfrimanm:
@hstastna @metalice @avivtur TMO the actions should be excluded from the option to be removed and should always be shown for the user. Other columns in the table should be included in the column management selector.

avivtur:
@hstastna @metalice @upalatucci @yfrimanm thanks for looking and for your inputs :)
+1 to block editing to the actions column, changed

there's no option to not render the additional columns header, do we want to change some of our current fields to be additional fields? or do we plan on adding more columns in the future? may I suggest moving the `Created` column to an additional column 
Screenshot - 2022-09-02T094414 922

which pages do we also want to an option for managing the columns?

src/views/virtualmachines/list/VirtualMachinesList.tsx
avivtur: we need this because the selectedColumns is of type Set<string>, this avoids TS type error

src/views/virtualmachines/list/hooks/useVirtualMachineColumns.ts
avivtur: reverted

PR #860, opened 2022-09-06
Bug 2119615: The unit field doesn't change in Disk modal

DanaOrr:

## 📝 Description
The unit field doesn't change when the size of the disk changes in Disk modal.
## 🎥 Demo
Before: 
DiskSIzeBefore

After:
DiskSIzeAfter


src/views/catalog/utils/quantity.ts
```diff
@@ -14,7 +14,7 @@ export const bytesFromQuantity = (
   quantity: string | number,
   precision = 0,
 ): [value: number, unit: string] => {
-  let byteSizeResult: ByteSizeResult = undefined;
+  let byteSizeResult: ByteSizeResult = { value: undefined, unit: undefined, long: undefined };
```
pcbailey, line 17: Why have you added the `long` property? It doesn't look like it's used anywhere.

FYI, it's better to use `null` when initializing a variable yourself. I didn't see the original when I was working on it or I would have changed it.

Here's a good explanation of the difference:

Null: It is the intentional absence of the value. It is one of the primitive values of JavaScript.
Undefined: It means the value does not exist in the compiler. It is the global object.

Source: https://www.geeksforgeeks.org/undefined-vs-null-in-javascript/

pcbailey (commented):
Great job! Just one small question and suggestion.

src/utils/components/CapacityInput/CapacityInput.tsx
```diff
@@ -24,37 +23,27 @@ type CapacityInputProps = {
...
 
-  const [value, quantityUnit] = bytesFromQuantity(size);
+  useEffect(() => {
+    if (size) {
+      const [unitValue] = size.match(/[a-zA-Z]+/g);
+      const [sizeValue = 0] = size.match(/[0-9]+/g) || [];
+      setUnit(!unitValue?.endsWith('B') ? unitValue + 'B' : unitValue);
+      setValue(+sizeValue);
+    }
+  }, [size]);
```
upalatucci, line 37:
```suggestion
  const [unitValue] = size?.match(/[a-zA-Z]+/g);
  const [sizeValue = 0] = size?.match(/[0-9]+/g) || [];
  const unit = !unitValue?.endsWith('B') ? unitValue + 'B' : unitValue;
  const value = +sizeValue;
```

What do you think about?

DanaOrr: It is better! Thank you :)

upalatucci (approved):

PR #869, opened 2022-09-13
Bug 2117549: Cannot edit cloud-init data after add ssh key

upalatucci:
## 📝 Description

**Cause**
With ssh key the cloud-init volume does not use `cloudInitNoCloud` but `cloudInitConfigDrive`.

### DIFF
--- a/src/utils/components/CloudinitModal/utils/useCloudInit.ts
+++ b/src/utils/components/CloudinitModal/utils/useCloudInit.ts
@@ -2,11 +2,7 @@ import React from 'react';
 import { load } from 'js-yaml';
 
 import { produceVMDisks } from '@catalog/utils/WizardVMContext';
-import {
-  V1CloudInitNoCloudSource,
-  V1VirtualMachine,
-  V1Volume,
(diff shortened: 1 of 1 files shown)


src/utils/components/CloudinitModal/utils/useCloudInit.ts
```diff
@@ -63,16 +59,25 @@ export const useCloudInit = (vm: V1VirtualMachine): UseCloudInitValues => {
...
       userData: convertUserDataObjectToYAML(userData, shouldAddHeader),
       networkData: convertNetworkDataObjectToYAML(networkData),
     });
 
-    return {
-      name: 'cloudinitdisk',
-      cloudInitNoCloud,
-    };
-  }, [userData, shouldAddHeader, networkData]);
+    if (cloudInitVol.cloudInitConfigDrive) {
```
metalice, line 67: cloudInitVol can be undefined? if so, please add cloudInitVol?.cloudInitConfigDrive (same goes to line 74)

upalatucci: yes! Solved

PR #873, opened 2022-09-13
[release-4.11] Bug 2119607: Env save button fix

upalatucci:
## 📝 Description

**Cause**
Save button enabled and disabled at the wrong moment. Don't use index to compare new and old envs array. The update can change the order

### DIFF
--- a/src/views/catalog/wizard/tabs/environment/hook/useEnvironments.ts
+++ b/src/views/catalog/wizard/tabs/environment/hook/useEnvironments.ts
@@ -66,11 +66,10 @@ const useEnvironments = (
   }, [environments, updateVM, vm]);
 
   React.useEffect(() => {
-    const unchanged = environments.every(
-      ({ name, serial }, index) =>
-        initialEnvironments?.[index]?.name === name &&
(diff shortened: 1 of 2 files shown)


src/views/catalog/wizard/tabs/environment/hook/useEnvironments.ts
```diff
@@ -66,11 +66,10 @@ const useEnvironments = (
   }, [environments, updateVM, vm]);
 
   React.useEffect(() => {
-    const unchanged = environments.every(
-      ({ name, serial }, index) =>
-        initialEnvironments?.[index]?.name === name &&
-        initialEnvironments?.[index]?.serial === serial,
+    const unchanged = environments.every(({ name, serial }) =>
+      initialEnvironments.find((e) => e.name === name && e.serial === serial),
```
hstastna, line 70: I'd better use `e?.name` here, same for the `e.serial`. Same for the other code in this PR.

hstastna (commented):
I've left a small comment, otherwise looks good to me! :)

PR #879, opened 2022-09-16
Bug 2124558: Allow MigrationPolicy creation only to privileged user

hstastna:
## 📝 Description

**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2124558

Allow _MigrationPolicy_ creation only to privileged user by displaying _Create MigrationPolicy_ button only for such users.

## Screenshots
**Before**:
The unprivileged user can click on _Create MigrationPolicy_ button:
beforee
**After**:
No any  _Create MigrationPolicy_ button present for unprivileged user:
afterr

### DIFF
--- a/src/views/migrationpolicies/details/tabs/details/components/MigrationPolicyDetailsSection/MigrationPolicyDetailsSection.tsx
+++ b/src/views/migrationpolicies/details/tabs/details/components/MigrationPolicyDetailsSection/MigrationPolicyDetailsSection.tsx
@@ -42,6 +42,7 @@ const MigrationPolicyDetailsSection: React.FC<MigrationPolicyDetailsSectionProps
   const { t } = useKubevirtTranslation();
   const { createModal } = useModal();
   const hasOwnPropertySpec = (key: string) => key in (mp?.spec || {});
+
   return (
     <div>
       <a href={`${pathname}#details`} className="link-icon">
--- a/src/views/migrationpolicies/list/MigrationPoliciesList.tsx
+++ b/src/views/migrationpolicies/list/MigrationPoliciesList.tsx
@@ -49,7 +49,11 @@ const MigrationPoliciesList: React.FC<MigrationPoliciesListProps> = ({ kind }) =
   return (
     <>
       <ListPageHeader title={t('MigrationPolicies')}>
-        <ListPageCreateDropdown items={createItems} onClick={onCreate}>
+        <ListPageCreateDropdown
+          items={createItems}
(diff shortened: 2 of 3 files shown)


hstastna:

upalatucci:
Warning in advance: don't know precisely how roles are managed in general in the console 

Shouldn't we use `useAccessReview` with `verb: 'create'` and/or `ver: 'patch`? Maybe there can be non-admin roles with creation and editing permission?

pcbailey:
The code itself looks good. Have you asked anyone about Ugo's question about `useAccessReview` vs `useIsAdmin`? I'm curious about that as well. So far, we've been operating under the assumption that there are only cluster admins and non-admins/project admins. It would be nice to have some formal guidance that this is the case.

Also, shouldn't we provide a message when there are no migration policies telling the user that they don't have permissions to create them?

hstastna:
> The code itself looks good. Have you asked anyone about Ugo's question about `useAccessReview` vs `useIsAdmin`? I'm curious about that as well. So far, we've been operating under the assumption that there are only cluster admins and non-admins/project admins. It would be nice to have some formal guidance that this is the case.

Not yet. Figuring out who's the best person to ask about this...

> Also, shouldn't we provide a message when there are no migration policies telling the user that they don't have permissions to create them?

IMO that's the question for the UX. I'm not sure if there is already some design for this case. I will continue investigating.

Anyway, thanks very much for the review! ❇️

avivtur:
Hi @hstastna 
I got confused about something something,

avivtur:
Hi @hstastna 
I must say that I'm confused, for some reason your "before" screenshots show the list of MigrationPolicies resources for a non-privileged user, when I get this error:
non-priviliged-can-create-mp

can you explain how that error does not occur for you? (this error is also it is shown in the BZ screenshot)

hstastna:
> Hi @hstastna I must say that I'm confused, for some reason your "before" screenshots show the list of MigrationPolicies resources for a non-privileged user, when I get this error: non-priviliged-can-create-mp
> 
> can you explain how that error does not occur for you? (this error is also it is shown in the BZ screenshot)

There was a bug, so I better moved the PR to WIP. I need to update it, to fix problems and to make sure about some implementation details. Thanks very much, Aviv, for pointing on this!

avivtur:
Hi @hstastna :) 
I was fixing a similar bug for the DataSources page, I believe you can follow the same pattern here as well
please have a look: https://github.com/kubevirt-ui/kubevirt-plugin/pull/900

PR #886, opened 2022-09-19
Bug 2128002: Display VM Templates list after template deletion

hstastna:
## 📝 Description

**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2128002

## 🎥 Demo
**Before:**
temp_before
**After:**
temp_after

### DIFF
--- a/src/views/templates/actions/hooks/useVirtualMachineTemplatesActions.tsx
+++ b/src/views/templates/actions/hooks/useVirtualMachineTemplatesActions.tsx
@@ -79,8 +79,7 @@ const useVirtualMachineTemplatesActions: useVirtualMachineTemplatesActionsProps 
     await k8sDelete({
       model: TemplateModel,
       resource: template,
-    });
-    history.push(`/k8s/${lastNamespacePath}/templates`);
+    }).then(() => history.push(`/k8s/${lastNamespacePath}/templates`));
   };
 
   const actions = [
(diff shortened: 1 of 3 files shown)


hstastna:

src/views/templates/actions/hooks/useVirtualMachineTemplatesActions.tsx
```diff
@@ -79,8 +79,9 @@ const useVirtualMachineTemplatesActions: useVirtualMachineTemplatesActionsProps
     await k8sDelete({
       model: TemplateModel,
       resource: template,
-    });
-    history.push(`/k8s/${lastNamespacePath}/templates`);
+    })
+      .then(() => history.push(`/k8s/${lastNamespacePath}/templates`))
+      .catch((e) => console.error(e));
```
upalatucci, line 84: If you remove this, the modal will handle the error showing the message on top of submit button

hstastna: The catch block is unnecessary here, of course. But it's a good practice to use it whenever working with promises.

upalatucci: Yes. The modal will not handle the expection because it's catched here. So no error message will be prompted

pcbailey: @hstastna Can you confirm that you don't want the error message to be shown? I don't feel comfortable lgtm'ing this without resolving this conversation.

PR #888, opened 2022-09-20
CNV 19687: SidebarEditor for wizard overview & envs tabs

upalatucci:
## 📝 Description

Sidebar Editor with buttons secondary editor style (like yaml tab editor)

## 🎥 Demo

Screenshot from 2022-09-20 16-42-23

Screenshot from 2022-09-20 16-42-13


src/utils/styles/_secondary-editor.scss
```diff
@@ -1,37 +1,58 @@
-.yaml-editor__buttons {
```
metalice, line 1: why does the styling name start with _ ? any benefits? , I can see the file is used for the @import, but why the _ ? :) 
we need to try to avoid !important as much as possible. Most of the time, if components don't get inline styling, it's possible due to better selectors... btw I like the mixin and includes

src/views/catalog/wizard/tabs/environment/wizard-environment-tab.scss
```diff
@@ -0,0 +1,7 @@
+.wizard-environment-tab {
+    height: 100%;
+
+    .sidebar-editor .pf-c-sidebar__panel {
+        margin-bottom: 1rem;
```
metalice, line 5: var?

src/views/catalog/wizard/tabs/overview/WizardOverviewTab.scss
```diff
@@ -1,3 +1,9 @@
 .wizard-overview-description-left-column {
   grid-column: 1;
 }
+
+.wizard-overview-tab {
+  &__grid {
+    margin-top: 1rem;
```
metalice, line 7: var?

src/utils/styles/_secondary-editor.scss
upalatucci: Don't know. In my defence I can say: it was already like that :grin:

upalatucci:
fixed @metalice

PR #889, opened 2022-09-21
CNV 19687: Sidebar editor in template parameters tab

upalatucci:
## 📝 Description

Sidebar editor for parameters tab with only one save button

## 🎥 Demo

image

### DIFF
--- a/src/views/templates/details/tabs/parameters/TemplateParametersPage.tsx
+++ b/src/views/templates/details/tabs/parameters/TemplateParametersPage.tsx
@@ -1,11 +1,22 @@
-import * as React from 'react';
+import React, { useEffect } from 'react';
 import { RouteComponentProps, useHistory } from 'react-router-dom';
 import { useImmer } from 'use-immer';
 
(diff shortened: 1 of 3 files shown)


src/views/templates/details/tabs/parameters/TemplateParametersPage.tsx
```diff
@@ -21,41 +31,39 @@ type TemplateParametersPageProps = RouteComponentProps<{
...
+    setEditableTemplate(({ parameters: draftParameters }) => {
       const parameterIndex = draftParameters.findIndex((p) => p.name === parameter.name);
       draftParameters[parameterIndex] = parameter;
     });
   };
 
+  const parameters = editableTemplate.parameters;
+
+  const isSaveDisabled =
+    JSON.stringify(template.parameters) === JSON.stringify(parameters) || isEditDisabled;
```
metalice, line 55: stringify can be sometimes a bit heavy, consider putting those lines of params and stringify in a usememo, as this component has a lot of states which will cause a lot or recalculate that arent needed if parmas isn't updated

PR #901, opened 2022-09-30
Fix type of the t function as the argument of getDataImportCronFilter

hstastna:
## 📝 Description

Fix the incorrect type of the `t` function, used as the argument of `getDataImportCronFilter`,
also fix some antipattern in `getDataImportCronFilter`.

## 🎥 Demo
**Before:**
t_before
**After:**
t_after

### DIFF
--- a/src/views/datasources/list/DataSourcesListFilters.ts
+++ b/src/views/datasources/list/DataSourcesListFilters.ts
@@ -1,16 +1,15 @@
-import { TFunction } from 'react-i18next';
-
 import { V1beta1DataSource } from '@kubevirt-ui/kubevirt-api/containerized-data-importer/models';
+import { t } from '@kubevirt-utils/hooks/useKubevirtTranslation';
 import { RowFilter } from '@openshift-console/dynamic-plugin-sdk';
 
 import { getDataSourceCronJob } from '../utils';
 
-export const getDataImportCronFilter = (t: TFunction): RowFilter<V1beta1DataSource>[] => [
+export const getDataImportCronFilter = (): RowFilter<V1beta1DataSource>[] => [
   {
     filterGroupName: t('Source'),
     type: 'data-cron-available',
     reducer: (obj) => (getDataSourceCronJob(obj) ? 'available' : ''),
-    filter: ({ selected }, obj) => selected?.length === 0 || !!getDataSourceCronJob(obj),
+    filter: ({ selected }, obj) => selected?.length === 0 || Boolean(getDataSourceCronJob(obj)),
     items: [
       {
         id: 'available',
--- a/src/views/datasources/list/DataSourcesList.tsx
+++ b/src/views/datasources/list/DataSourcesList.tsx
@@ -39,7 +39,7 @@ const DataSourcesList: React.FC<DataSourcesListProps> = ({ kind, namespace }) =>
     namespace,
   });
   const [columns, activeColumns] = useDataSourcesColumns(namespace);
-  const filters = getDataImportCronFilter(t);
+  const filters = getDataImportCronFilter();
   const [unfilteredData, data, onFilterChange] = useListPageFilter(dataSources, filters);
 
   const createItems = {


hstastna:

src/views/datasources/list/DataSourcesListFilters.ts
```diff
@@ -5,12 +5,14 @@ import { RowFilter } from '@openshift-console/dynamic-plugin-sdk';
 
 import { getDataSourceCronJob } from '../utils';
 
-export const getDataImportCronFilter = (t: TFunction): RowFilter<V1beta1DataSource>[] => [
+export const getDataImportCronFilter = (
+  t: TFunction<'plugin__kubevirt-plugin', undefined>,
```
vojtechszocs, line 9: Maybe it's worth adding
```ts
export type KubevirtTFunction = TFunction<'plugin__kubevirt-plugin', undefined>;
```
(or omit the second type argument, unless it's strictly required)

to `src/utils/hooks/useKubevirtTranslation.ts` and update the code to use `KubevirtTFunction` consistently throughout the codebase.

metalice:
Please check if we can remove t function completely and consume it directly in this file. There are example in our code base.

src/views/datasources/list/DataSourcesListFilters.ts
hstastna: That sounds great to me! I'd like do this in a followup PR.

hstastna:
> Please check if we can remove t function completely and consume it directly in this file. There are example in our code base.

i think it is easily, but we would have to rename `getDataImportCronFilter`  this to `use..` and to make a hook from it. Do you agree on that, @metalice ?

metalice:
> > Please check if we can remove t function completely and consume it directly in this file. There are example in our code base.
> 
> I think it is easily doable, but we would have to rename `getDataImportCronFilter` to `use...` and to make a hook from it. Do you agree on that, @metalice ?

you can simply import {t} from "......useKubevirtTranslation" , no need to custom hook, we exported solo t for cases like this when its a util file and there is a function that needs t. so t will not be transferred to the function, but t will be imported directly where needed, you can have a look inside useKubevirtTranslation file

hstastna:
> > > Please check if we can remove t function completely and consume it directly in this file. There are example in our code base.
> > 
> > 
> > I think it is easily doable, but we would have to rename `getDataImportCronFilter` to `use...` and to make a hook from it. Do you agree on that, @metalice ?
> 
> you can simply import {t} from "......useKubevirtTranslation" , no need to custom hook, we exported solo t for cases like this when its a util file and there is a function that needs t. so t will not be transferred to the function, but t will be imported directly where needed, you can have a look inside useKubevirtTranslation file

Okay, so I'll move the `t` function inside the _getDataImportCronFilter_. Thanks! 👍🏽

hstastna:
> > > Please check if we can remove t function completely and consume it directly in this file. There are example in our code base.
> > 
> > 
> > I think it is easily doable, but we would have to rename `getDataImportCronFilter` to `use...` and to make a hook from it. Do you agree on that, @metalice ?
> 
> you can simply import {t} from "......useKubevirtTranslation" , no need to custom hook, we exported solo t for cases like this when its a util file and there is a function that needs t. so t will not be transferred to the function, but t will be imported directly where needed, you can have a look inside useKubevirtTranslation file

@metalice so I've made a change and this has happened:
uncheck_and_fix_not_working_clusters_ maybe
This is what I was talking about when talking about renaming `getDataImportCronFilter` to use the `use` prefix and to make it a hook from it...

metalice:
> > > > Please check if we can remove t function completely and consume it directly in this file. There are example in our code base.
> > > 
> > > 
> > > I think it is easily doable, but we would have to rename `getDataImportCronFilter` to `use...` and to make a hook from it. Do you agree on that, @metalice ?
> > 
> > 
> > you can simply import {t} from "......useKubevirtTranslation" , no need to custom hook, we exported solo t for cases like this when its a util file and there is a function that needs t. so t will not be transferred to the function, but t will be imported directly where needed, you can have a look inside useKubevirtTranslation file
> 
> @metalice so I've made a change and this has happened: use This is what I was talking about when talking about renaming `getDataImportCronFilter` to use the `use` prefix and to make it a hook from it...

that's because u tried to import the hook and not the exported t function, try like this:
image

hstastna:
> that's because u tried to import the hook and not the exported t function, try like this: image

I did it because `import { t } `didn't work for me, not sure why, I'm still investigating..

PR #906, opened 2022-10-06
Adjusting ds example yaml

metalice:

## 📝 Description
Adjusting ds example yaml - ds name from example to ds-example :)

## 🎥 Demo


hstastna (approved):

avivtur:
/hold
the convention should be `example` as the name of the resource and we should fix the VM YAML example and the VM Template example to follow that convention, we have this BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2132682

PR #909, opened 2022-10-07
Fix storage throughput value in top consumers card

pcbailey:
## 📝 Description

The values for storage throughput in the top consumers card were being reported incorrectly when compared with the same metric in the top consumers dashboard in Observe > Dashboards. This PR fixes the issue by using the correct function to convert from the bytes returned from Prometheus to the unit to be displayed.


pcbailey:
Fixed in auto-backport: https://github.com/kubevirt-ui/kubevirt-plugin/pull/921

PR #910, opened 2022-10-07
Bug 2071491: Fix storage throughput value in Top Consumers card

pcbailey:
## 📝 Description

The values for storage throughput in the top consumers card were being reported incorrectly when compared with the same metric in the top consumers dashboard in Observe > Dashboards. This PR fixes the issue by using the correct function to convert from the bytes returned from Prometheus to the unit to be displayed.

## 🎥 Screenshot
Before:
Selection_169

After:
Selection_167

### DIFF
--- a/src/views/clusteroverview/MonitoringTab/top-consumers-card/utils/utils.ts
+++ b/src/views/clusteroverview/MonitoringTab/top-consumers-card/utils/utils.ts
@@ -3,7 +3,6 @@ import { TFunction } from 'i18next';
 import {
   humanizeBinaryBytes,
   humanizeDecimalBytes,
-  humanizeDecimalBytesPerSec,
   humanizeSeconds,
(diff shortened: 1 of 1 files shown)


upalatucci:
Should the `storage throughput` be Bytes per second?

pcbailey:
> Shouldn't the storage throughput be bytes per second instead of bytes? I think the issue is that numbers were interpreted as MBs and not Bs

I originally thought that as well, which is why the bug existed in the first place. If you look at the y-axis of the graph in the left side of the screenshot, they're using the same units now. The queries used for the left and right sides are identical.

I believe the metric is supposed to indicate the total amount of data written and read at a given point in time.

pcbailey:
@yaacov I'm in the kubevirt-ui org. Do you have any idea why the cherrypick command failed for me?

PR #913, opened 2022-10-11
Bug 2132793: Load and show boot source reference

upalatucci:
## 📝 Description

**Cause**
The EditBootSourceModal does not fetch the dataVolume and does not show the dataVolume source.

**Solution**
Fetch DV from DataSource and get from it the specs. Populate the form with specs data and be able to edit them. On submission, edit the DV

Most of the changes were made in the [SelectSource](https://github.com/kubevirt-ui/kubevirt-plugin/pull/913/files#diff-67651d5ff71151c93431684b8f429ba4985aef31fb688bbb6ee994ebb0b09daf).
The previous select source was not accepting a state or an initial state so was impossible to solve the bug without touching it.
Also in the component, there were a giant `useEffect` runner at every update. 
Now the SelectSource accepts a state (data volume spec) and all the component states were transformed into constants dependent on that state. There is just one callback when the volume or source type change
 
## 🎥 Demo

[Screencast from 11-10-2022 10:46:52.webm](https://user-images.githubusercontent.com/29160323/195043359-5096aedd-c821-45ac-93e6-e3e96d320f42.webm)

### DIFF
--- a/src/views/templates/actions/components/PersistentVolumeClaimSelect/PersistentVolumeClaimSelect.tsx
+++ b/src/views/templates/actions/components/PersistentVolumeClaimSelect/PersistentVolumeClaimSelect.tsx
@@ -10,34 +10,32 @@ import './PersistentVolumeClaimSelect.scss';
 type PersistentVolumeClaimSelectProps = {
   pvcNameSelected: string;
   projectSelected: string;
-  selectNamespace: (namespace: string) => void;
-  selectPVCName: (pvcName: string) => void;
+  selectPVC: (pvcNamespace: string, pvcName?: string) => void;
 };
 
 export const PersistentVolumeClaimSelect: React.FC<PersistentVolumeClaimSelectProps> = ({
   pvcNameSelected,
   projectSelected,
-  selectPVCName,
-  selectNamespace,
+  selectPVC,
 }) => {
-  const { projectsNames, filteredPVCNames, loaded } = useProjectsAndPVCs(projectSelected);
+  const { projectsNames, filteredPVCNames, projectsLoaded, pvcsLoaded } =
+    useProjectsAndPVCs(projectSelected);
 
   const onSelectProject = React.useCallback(
     (newProject) => {
-      selectNamespace(newProject);
-      selectPVCName(undefined);
+      selectPVC(newProject);
     },
-    [selectNamespace, selectPVCName],
+    [selectPVC],
   );
 
   const onPVCSelected = React.useCallback(
     (selection) => {
-      selectPVCName(selection);
+      selectPVC(projectSelected, selection);
     },
-    [selectPVCName],
+    [selectPVC, projectSelected],
   );
 
-  if (!loaded) return <PersistentVolumeClainSelectSkeleton />;
+  if (!projectsLoaded) return <PersistentVolumeClainSelectSkeleton />;
 
   return (
     <div>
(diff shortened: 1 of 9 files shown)


src/views/templates/actions/editBootSource.ts
```diff
@@ -79,14 +81,29 @@ export const getBootDataSource = async (
...
+export const getDataSourceDataVolume = async (
+  dataSourcePVCName: string,
+  dataSourcePVCNamespace: string,
+): Promise<V1beta1DataVolume | undefined> => {
+  let dataVolume: V1beta1DataVolume = null;
+
+  try {
+    dataVolume = await getDataVolume(dataSourcePVCName, dataSourcePVCNamespace);
+  } catch (error) {
+    console.warn(error);
```
hstastna, line 93: Not sure if we really want `console.warn` here, if it's about catching error. 
Also maybe we want to display the error/warning the different way - visibly for the user - as in other places in the code... Can you, please, explain the reason of using `console.warn` here? Just for better understanding of this change. Thanks! :)

hstastna (commented):
Left a few comments to make sure about some code changes. Anyway, great work, Ugo! ❇️

src/views/templates/actions/components/PersistentVolumeClaimSelect/PersistentVolumeClaimSelect.tsx
```diff
@@ -10,34 +10,32 @@ import './PersistentVolumeClaimSelect.scss';
 type PersistentVolumeClaimSelectProps = {
   pvcNameSelected: string;
   projectSelected: string;
-  selectNamespace: (namespace: string) => void;
-  selectPVCName: (pvcName: string) => void;
+  selectPVC: (pvcNamespace: string, pvcName?: string) => void;
```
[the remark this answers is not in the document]
upalatucci: Because when the user change namespace, the name needs to be undefined

src/views/templates/actions/editBootSource.ts
upalatucci: I used warn because deep source wants something in the catch. If there is no dataVolume it's not an error. Just means that there is no DV

hstastna: Hmm, maybe `console.warn` + some comment in the code about why we use `warn` and not `error`, could be sufficient :)

upalatucci: Maybe if the client lost connection? So also the creation will fail.

hstastna (approved):

src/views/templates/actions/components/PersistentVolumeClaimSelect/PersistentVolumeClaimSelect.tsx
```diff
@@ -10,34 +10,32 @@ import './PersistentVolumeClaimSelect.scss';
...
     (selection) => {
-      selectPVCName(selection);
+      selectPVC(projectSelected, selection);
     },
-    [selectPVCName],
+    [selectPVC, projectSelected],
   );
 
-  if (!loaded) return <PersistentVolumeClainSelectSkeleton />;
+  if (!projectsLoaded) return <PersistentVolumeClainSelectSkeleton />;
```
pcbailey, line 38: There's a typo in that component name. It should be 'Claim' not 'Clain'. We should probably fix that in a separate PR.

PR #916, opened 2022-10-18
Bug 2131165: Fixed in virtual machine statuses card the dark mode and removing icons

DanaOrr:

## 📝 Description
In dark mode the additional status in the virtual machine statuses card was not visible enough.
And I removed the icons in the additional status because they are the same. 
## 🎥 Demo

Before:
image
image

After: 
image
image

### DIFF
--- a/src/views/clusteroverview/OverviewTab/vm-statuses-card/VMStatusesCard.scss
+++ b/src/views/clusteroverview/OverviewTab/vm-statuses-card/VMStatusesCard.scss
@@ -52,7 +52,6 @@
       &:active {
         background-color: transparent;
       }
-      color: #151515;
     }
   }
 }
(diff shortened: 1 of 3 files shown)


src/views/clusteroverview/OverviewTab/vm-statuses-card/VMStatusItem.tsx
```diff
@@ -14,21 +14,24 @@ type VMStatusItemProps = {
...
   namespace?: string;
+  showIcon?: boolean;
 };
 
-const VMStatusItem: React.FC<VMStatusItemProps> = ({ status, count, namespace }) => {
+const VMStatusItem: React.FC<VMStatusItemProps> = ({
+  status,
+  count,
+  namespace,
+  showIcon = true,
```
avivtur, line 24: I think we can go with a different approach here to avoid adding the `showIcon` prop
`vmStatusIcon` on line 26 is an object to map a string value to an icon, but if we don't use some of the icons then maybe we should change the object to return null on the additional statuses so it won't show the icons and we avoid adding another prop

DanaOrr: Thank you Aviv! :)

avivtur (approved):
/lgtm

PR #926, opened 2022-10-24
Bug 2132721: Add missing texts to Delete dialogs

hstastna:
## 📝 Description
**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2132721

Add missing "?" or some more text to the following dialogs:

- Delete VM
- Detach VM disk
- Delete VM NIC
- Delete VM template
- Delete VM template disk
- Delete VM Template NIC
- Delete VirtualMachineSnapshot
- Delete MigrationPolicy
- Edit Boot source reference
- Delete NIC dialog while customizing VM
- Delete VM Snapshot from the Snapshots card in the Overview tab

Also delete unnecessary words in some of the related kebab menus for consistency reasons.

## 🎥 Screenshots (not all the changes included)
**Before:**
Delete VirtualMachine dialog:
vm_before
Delete template NIC dialog:
template_nic_before
VM Snapshot kebab menu:
snap_before1
Delete MigrationPolicy:
migr_before
Edit Boot source reference:
bsr_before
**After:**
vm_after
template_nic_after
snap_after1
migr_after
bsr_after


hstastna:

src/views/templates/actions/components/EditBootSourceModal.tsx
```diff
@@ -62,29 +64,28 @@ const EditBootSourceModal: FC<EditBootSourceModalProps> = ({
...
-            </Popover>{' '}
+            <Button variant="link" isInline ref={popoverRef}>
+              all templates
+            </Button>{' '}
             that are currently using this DataSource.
-          </Trans>
-        </Alert>
+          </Alert>
+        </Trans>
+        <Popover
```
pcbailey, line 76: Is there a benefit to using a ref for the Popover instead of wrapping the button with it?

src/views/templates/details/tabs/disks/components/DiskRowActions.tsx
```diff
@@ -62,12 +62,13 @@ const DiskRowActions: React.FC<DiskRowActionsProps> = ({ diskName, vm, onUpdate,
...
         obj={vm}
         onSubmit={onDelete}
-        headerText={t('Detach {{diskName}} disk', { diskName })}
+        headerText={t('Detach {{diskName}} disk?', { diskName })}
         submitBtnText={deleteBtnText}
         submitBtnVariant={ButtonVariant.danger}
       >
         <Trans t={t}>
-          Are you sure you want to detach <strong>{diskName} </strong>
+          Are you sure you want to detach <strong>{diskName} </strong>in namespace{' '}
```
pcbailey, line 70: Why not use `DeleteResourceMessage` here?

src/views/virtualmachines/details/tabs/disk/modal/DeleteDiskModal.tsx
```diff
@@ -98,15 +98,15 @@ const DeleteDiskModal: React.FC<DeleteDiskModalProps> = ({ vm, volume, isOpen, o
...
       onSubmit={onSubmit}
-      headerText={t('Detach {{diskName}} disk', { diskName })}
+      headerText={t('Detach {{diskName}} disk?', { diskName })}
       submitBtnText={t('Detach')}
       submitBtnVariant={ButtonVariant.danger}
     >
       <Stack hasGutter>
         <StackItem>
           <Trans t={t}>
             Are you sure you want to detach <strong>{diskName} </strong>in namespace{' '}
```
pcbailey, line 108: You could add an `action` prop to `DeleteResourceMessage`, rename it to something like `ConfirmResourceActionMsg`, and use it for the detach messages as well. Just a suggestion.

src/views/templates/details/tabs/disks/components/DiskRowActions.tsx
hstastna: Because here we use the verb "detach", not "delete" as it is in `DeleteResourceMessage`, so I better left this as it is.

src/views/templates/actions/components/EditBootSourceModal.tsx
hstastna: Yes. Wrapping the button with it - as it was originally here, didn't work to display the popover correctly.

src/views/catalog/wizard/tabs/disks/components/DiskRowActions.tsx
```diff
@@ -74,12 +74,13 @@ const DiskRowActions: React.FC<DiskRowActionsProps> = ({ diskName }) => {
...
         obj={vm}
         onSubmit={onDelete}
-        headerText={t('Detach {{diskName}} disk', { diskName })}
+        headerText={t('Detach {{diskName}} disk?', { diskName })}
         submitBtnText={deleteBtnText}
         submitBtnVariant={ButtonVariant.danger}
       >
         <Trans t={t}>
-          Are you sure you want to detach <strong>{diskName} </strong>
+          Are you sure you want to detach <strong>{diskName} </strong>in namespace{' '}
```
metalice, line 82: in file above [DeleteResourceMessage.tsx](https://github.com/kubevirt-ui/kubevirt-plugin/pull/926/files#diff-5170b9ba43e92451f91fe0e0a71393282c4df1b33244ab9ebbf9b9d0768cc1a7) you add a check for namespace but here no? i think better keep on consistency, not sure namespace check is required, as you don't check for name also, they both comes together.. or I'm missing something ?

src/views/templates/actions/components/EditBootSourceModal.tsx
```diff
@@ -62,29 +64,28 @@ const EditBootSourceModal: FC<EditBootSourceModalProps> = ({
         isOpen={isOpen}
         onClose={onClose}
       >
-        <Alert isInline className="margin-bottom-md" variant="warning" title={t('Warning')}>
-          <Trans ns="plugin__kubevirt-plugin">
+        <Trans ns="plugin__kubevirt-plugin">
```
metalice, line 67: im not sure Trans is needed here

src/views/templates/actions/components/EditBootSourceModal.tsx
metalice: @hstastna I think we can do it without ref. If we need a ref it feels like the usage of the popover is missing something.. please check

hstastna: Yes. `Popover` component seems to not working 100% correctly. This seems to be related to Patternfly and changes of their components we use. And any other component around could influence that the original solution stopped working, not only Popover.. 
IMO fixing this can take some time, so I'd suggest to better go for the actual proposed and working solution, until we find some better way. IMO using ref is nothing bad here, just a different way that could be used.

src/views/catalog/wizard/tabs/disks/components/DiskRowActions.tsx
hstastna: Here we use "detach", so I haven't used `DeleteResourceMessage` component, and it is very highly expected that we display the namespace in this modal, too. 
In `DeleteResourceMessage`, there's check for namespace added, because displaying namespace is redundant for MigrationPolicy deletion. In all other cases, we do want to display the namespace. Also using `vm?.metadata?.namespace` syntax makes sure the app doesn't break in (very rare and unexpected) case the namespace is not present.

src/views/templates/actions/components/EditBootSourceModal.tsx
hstastna: You are right, this line of the code can be in its original place.

src/utils/components/ConfirmActionMessage/ConfirmActionMessage.tsx
```diff
@@ -0,0 +1,28 @@
+import * as React from 'react';
+import { Trans } from 'react-i18next';
+
+import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
+import { K8sResourceCommon } from '@openshift-console/dynamic-plugin-sdk';
+
+const ConfirmActionMessage: React.FC<{ obj: K8sResourceCommon; action?: string }> = ({
+  obj,
```
pcbailey, line 8: I think it might be a little cleaner to accept `name` and `namespace` props instead of building custom objects when you use the components. What do you think?

hstastna: Well, this depends on where the component is used. I'd leave it as it is, as we need a custom object only in a few cases.

pcbailey: The component doesn't need a `K8sResourceCommon` object; it needs a `name` and a `namespace`. Passing only what is needed would make both the component and the places it's used cleaner. Also, the `namespace` prop would be optional, which would help in making the component's interface even clearer.

hstastna: From my perspective, it's better to send `K8sResourceCommon` object as a prop to the component - as it was originally done. I believe this was done with a good purpose. Also the component takes care by itself about what is needed from the object. If we did it the different way, we would have to extract the name and the namespace - to do the same stuff in all the places where the component is used - I consider this unnecessary.

pcbailey: Having `K8sResourceCommon` as a prop falsely implies that the component only works for `K8sResourceCommon` objects, which isn't true. Maybe it would have been better to just create a `ConfirmDetachMessage` component and leave this one as it was originally.

src/utils/components/ConfirmActionMessage/ConfirmActionMessage.tsx
```diff
@@ -0,0 +1,28 @@
...
+  action,
+}) => {
+  const { t } = useKubevirtTranslation();
+  const objNamespace = obj?.metadata?.namespace;
+
+  return (
+    <Trans t={t}>
+      Are you sure you want to {action || 'delete'} <strong>{obj?.metadata?.name}</strong>
```
pcbailey, line 16: Sorry, I missed this before. You can set 'delete' as the default value in the function arguments list. That would get rid of the `|| 'delete'` on this line.

hstastna: Yeah, thanks, sometimes there are more options how to implement something :)

src/utils/components/ConfirmActionMessage/ConfirmActionMessage.tsx
hstastna: Thank you, Phillip, this explanation did persuade me to think that we don't want `K8sResourceCommon` as a prop type :)
 
Yes, I was also thinking about creating a separate component for "detach" message, not to mix different types of message components, but I'm not sure if it worth, as the messages are almost the same. I need to re-think this. If you have any other ideas about this or opinions what's a better solution, let me know. Thank you very much! :)

pcbailey: I think the right solution hinges on whether or not there is a good chance that other data from the object will be needed in the future. If yes, then it's cleaner IMO to create a separate component for detach, since there is no `K8sResourceCommon` object used in those cases. If there's not, then I think it's cleaner to use a single component that takes only the props needed to function.

That said, if you decide you still want to keep it the way it is, I'll lgtm it. Just let me know.

src/utils/components/ConfirmActionMessage/ConfirmActionMessage.tsx
hstastna: Agreed :)

PR #928, opened 2022-10-25
Bug 2023393: Adding info message when boot source SC is different from default SC

metalice:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2023393

## 📝 Description

When cluster default SC is different from boot source SC it can cause slow imports.
Adding an info message regarding this issue on disk modal.

## 🎥 Demo

Disk modal:
image


src/utils/hooks/useBootSourceDefaultStorageClass.ts
```diff
@@ -0,0 +1,21 @@
+import { IoK8sApiStorageV1StorageClass } from '@kubevirt-ui/kubevirt-api/kubernetes/models';
+import useDefaultStorageClass from '@kubevirt-utils/hooks/useDefaultStorage/useDefaultStorageClass';
+import { TemplateBootSource } from '@kubevirt-utils/resources/template/hooks/useVmTemplateSource/utils';
+
+type UseBootSourceDefaultStorageClass = (bootSource?: TemplateBootSource) => {
+  isBootSourceSCisDefaultSC: boolean;
```
avivtur, line 6: this name looks a bit strange to me, maybe `bootSourceDefaultSC` wdyt?

avivtur (commented):
besides a small suggestion, it looks good to me :)

src/utils/hooks/useBootSourceDefaultStorageClass.ts
metalice: changed it a bit :)

PR #931, opened 2022-10-26
Bug 2137243: Fixing registry input in Edit boot source reference

DanaOrr:

## 📝 Description
Fixing in the Edit boot source reference the added docker prefix once more if a letter or sign is deleted from- "docker://".

## 🎥 Demo
Before:
RegistryInput1

After:
RegistryInput


src/views/templates/actions/components/SelectSource.tsx
```diff
@@ -119,14 +116,13 @@ export const SelectSource: React.FC<SelectSourceProps> = ({
           fieldId={`disk-source-required-${selectedSourceType}`}
           isRequired
           className="disk-source-form-group"
-          helperText={registrySourceHelperText}
+          helperText={<>{t('Example: quay.io/containerdisks/fedora:latest')}</>}
```
avivtur, line 119: you should translate only the Example text and not the URL itself

avivtur (approved):
looks good

PR #932, opened 2022-10-26
Bug 2136425: Windows 11 is detected as Windows 10

avivtur:

## 📝 Description

> Guest agent is showing wrong pretty name data for win 11 so we change the way we evaluate the OS for the single VM overview details card


dominikholler:
@avivtur thanks for addressing the issue so quickly! Do you have a screenshot how this looks for RHEL and Windows 11?

PR #941, opened 2022-11-02
Bug 2139306: Process template on selected project

upalatucci:
## 📝 Description

non-priv user does not have access to other namespaces so we have to process template in the selected namespace

This fix targets also [this bug](https://bugzilla.redhat.com/show_bug.cgi?id=2139299)


avivtur:
/retitle Bug 2139299: Process template on selected project

PR #944, opened 2022-11-02
Bug 2139296: Non-admin cannot load MigrationPolicies page

avivtur:
## 📝 Description

> requiring `CAN_LIST_NS` flag for MigrationPolicies page to show this page only for privileged users
### Before
mp-page-error

### After
mp-page-error-after

### DIFF
--- a/console-extensions.json
+++ b/console-extensions.json
@@ -384,6 +384,9 @@
         "data-quickstart-id": "qs-nav-migrationpolicies",
         "data-test-id": "migrationpolicies-nav-item"
       }
+    },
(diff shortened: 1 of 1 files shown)


avivtur:
@hstastna AFAIK yes, we should show navigation items for users if they have access to related resources on the same navigation item on any scope (cluster or project), but in this case, they don't have permission to list on any scope (migration policies are on cluster scope only for now), we should not present this navigation item as it's not really functional for non-priv user as other navigation items are hidden for non-priv user in case they are completely not functional.

PR #947, opened 2022-11-03
Bug 2139687: catalog shows template list for non-priv users

avivtur:
## 📝 Description

> Non-privileged user could see the template's list in the catalog and under a namespace with no permissions can also click the template and crash when clicking on customize vm

### Before
https://user-images.githubusercontent.com/67270715/199687089-23236aeb-e49a-4dfc-be90-abb921635e50.mp4

catalog-shows-templates-when-cant-process-template-b4

### After
catalog-shows-templates-when-cant-process-template-with-ns
catalog-shows-templates-when-cant-process-template-with-ns-permissions
catalog-shows-templates-when-cant-process-template


upalatucci (approved):
Great job!

PR #950, opened 2022-11-07
Bug 2027959: Delete with Grace Period

upalatucci:
## 📝 Description

Delete with gracePeriodSeconds

## 🎥 Demo

Screenshot from 2022-11-07 11-56-05
Screenshot from 2022-11-07 11-56-01


src/views/virtualmachines/actions/components/DeleteVMModal/components/GracePeriodInput.tsx
```diff
@@ -0,0 +1,89 @@
...
+  InputGroupText,
+  Popover,
+  StackItem,
+  TextInput,
+} from '@patternfly/react-core';
+import { HelpIcon } from '@patternfly/react-icons';
+
+import './grace-period-input.scss';
+
+const DEFAULT_GRACE_PERIOD = 180;
```
avivtur, line 18: it is a bit strange to me to hard-code the default for the seconds, isn't this default derived from somewhere?

src/views/virtualmachines/actions/components/DeleteVMModal/components/GracePeriodInput.tsx
```diff
@@ -0,0 +1,89 @@
...
+  setGracePeriodSeconds: (number) => void;
+}
+
+export const GracePeriodInput: FC<GracePeriodInputProps> = ({
+  gracePeriodSeconds,
+  setGracePeriodSeconds,
+}) => {
+  const { t } = useTranslation();
+
+  const onGracePeriodCheckChange = (checked: boolean) => {
```
avivtur, line 31: I think resetting the seconds on check click is not a good practice, since in case the user clicked the checkbox by accident it will delete his previous input which can be annoying.
I would have used with another <boolean>useState for the checkbox, unless it's a requirement to reset on every check/uncheck

src/views/virtualmachines/actions/components/DeleteVMModal/components/GracePeriodInput.tsx
```diff
@@ -0,0 +1,89 @@
...
+    if (checked) setGracePeriodSeconds(DEFAULT_GRACE_PERIOD);
+    else setGracePeriodSeconds(null);
+  };
+
+  const onGracePeriodSecondsChange = (value: string) => {
```
avivtur, line 36: this function might be having an issue when you try to delete the input, but the 0 keeps poping back.
also `parseInt` has a default of base 10, so you can omit the 10 at line 40.

src/views/virtualmachines/actions/components/DeleteVMModal/components/GracePeriodInput.tsx
upalatucci: It's something defined in the backend. By default, the grace period is 3 minutes

src/views/virtualmachines/actions/components/DeleteVMModal/components/GracePeriodInput.tsx
upalatucci: Yes. Otherwise, I would leave the input blank and handle the case of submission with the checkbox checked and input blank.

src/views/virtualmachines/actions/components/DeleteVMModal/DeleteVMModal.tsx
```diff
@@ -24,10 +25,14 @@ type DeleteVMModalProps = {
   onClose: () => void;
 };
 
-const DeleteVMModal: React.FC<DeleteVMModalProps> = ({ vm, isOpen, onClose }) => {
+const DEFAULT_GRACE_PERIOD = 180;
```
metalice, line 28: extract to constants?

src/views/virtualmachines/actions/components/DeleteVMModal/DeleteVMModal.tsx
```diff
@@ -24,10 +25,14 @@ type DeleteVMModalProps = {
...
+
+const DeleteVMModal: FC<DeleteVMModalProps> = ({ vm, isOpen, onClose }) => {
   const { t } = useKubevirtTranslation();
   const history = useHistory();
-  const [deleteOwnedResource, setDeleteOwnedResource] = React.useState(true);
+  const [deleteOwnedResource, setDeleteOwnedResource] = useState(true);
```
metalice, line 33: <boolean>

src/views/virtualmachines/actions/components/DeleteVMModal/components/GracePeriodInput.tsx
```diff
@@ -0,0 +1,84 @@
...
+  setGracePeriodSeconds,
+}) => {
+  const { t } = useTranslation();
+
+  const onGracePeriodSecondsChange = (value: string) => {
+    if (value === '') {
+      setGracePeriodSeconds(null);
+    } else {
+      setGracePeriodSeconds(parseInt(value));
+    }
```
metalice, line 38: setGracePeriodSeconds(isEmpty(value) ? null : value)

src/views/virtualmachines/actions/components/DeleteVMModal/DeleteVMModal.tsx
```diff
@@ -86,6 +91,9 @@ const DeleteVMModal: React.FC<DeleteVMModalProps> = ({ vm, isOpen, onClose }) =>
     await k8sDelete({
       model: VirtualMachineModel,
       resource: updatedVM,
+      json: withGracePeriod
+        ? { kind: 'DeleteOptions', apiVersion: 'v1', gracePeriodSeconds }
+        : null,
```
metalice, line 96: please try seems to be working here withGracePeriod ?? { kind: 'DeleteOptions', apiVersion: 'v1', gracePeriodSeconds }

src/views/virtualmachines/actions/components/DeleteVMModal/DeleteVMModal.tsx
```diff
@@ -24,10 +25,14 @@ type DeleteVMModalProps = {
...
 
-const DeleteVMModal: React.FC<DeleteVMModalProps> = ({ vm, isOpen, onClose }) => {
+const DEFAULT_GRACE_PERIOD = 180;
+
+const DeleteVMModal: FC<DeleteVMModalProps> = ({ vm, isOpen, onClose }) => {
   const { t } = useKubevirtTranslation();
   const history = useHistory();
-  const [deleteOwnedResource, setDeleteOwnedResource] = React.useState(true);
+  const [deleteOwnedResource, setDeleteOwnedResource] = useState(true);
+  const [withGracePeriod, setWithGracePeriod] = useState<boolean>(false);
```
metalice, line 34: consider naming gracePeriodCheckbox

src/views/virtualmachines/actions/components/DeleteVMModal/components/GracePeriodInput.tsx
metalice: or ??

src/views/virtualmachines/actions/components/DeleteVMModal/DeleteVMModal.tsx
upalatucci: the operator `??` would return deleteoptions only if withGracePeriod is `null` or `undefined`

src/views/virtualmachines/actions/components/DeleteVMModal/components/GracePeriodInput.tsx
upalatucci: In the backend this value depends on the VM os. I think 3 minutes for linux distro and 1 hour for windows. Don't know if the backend exposes this kind of data but I don't think is critical here. Just to have something in the input at the beginning and 180 seconds seems reasonable. 
Having more logic here for this thing would not bring any value from my point of view.

avivtur:
LGTM with me :) 
please let @metalice to have a final look at his comments

src/views/virtualmachines/actions/components/DeleteVMModal/components/GracePeriodInput.tsx
avivtur: cant you take the default from the vm YAML (`terminationGracePeriodSeconds`)? - http://kubevirt.io/api-reference/main/definitions.html#_v1_virtualmachineinstancespec

upalatucci: Great idea! I can get this value from `vm.spec.template.spec.terminationGracePeriodSeconds` and use `DEFAULT_GRACE_PERIOD` if this value is not set (cause it's optional)

PR #952, opened 2022-11-07
Bug 2139574: Fix catalog for template no bootsource

upalatucci:
## 📝 Description

Fix template catalog for template with no bootsource

## 🎥 Demo

Screenshot from 2022-11-07 12-51-53


hstastna:
@upalatucci Can you, please, add _Before_ and _After_ screenshots or a video of what exactly has been visually changed? It's not obvious from the current screenshot (and also not from the BZ description). Thanks! :)

avivtur:
@hstastna by the look of the change is setting the loader to stop with no boot source available as a default so the skeleton would stop load without stopping.
/lgtm

upalatucci:
Yes, sorry guys. 
The issue was that for templates with no boot source, the drawer footer was stuck on an infinite loading.

PR #955, opened 2022-11-07
Bug 2140627: Allow to select storageClass in Add disk modal

hstastna:
## 📝 Description

**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2140627

Allow to select StorageClass when adding a new Disk to a VM if there is no default storageclass defined.

Note that if you want to reproduce this bug, you need to:
- set the annotation _storageclass.kubernetes.io/is-default-class_ to `'false'` for all storage classes
  (in _Storage_ > _StorageClasses_)

## 🎥 Screenshots
**Before:**
Not being able to select a different StorageClass than already selected (nothing happens when clicking on it):
sc_before
Error in the browser console:
sc_err
**After:**
Successfully selected a different StorageClass, incl. displaying the correct info:
sc_after


hstastna:

PR #958, opened 2022-11-09
Bug 2140730: Make links to resources on Overview page work correctly

hstastna:
## 📝 Description

**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2140730

Make links to the lists of VMs, Templates, Networks work correctly - to lead to the correct list of the resources with the correct project/namespace name.

## 🎥 Screenshots
**Before:** 
After clicking on the number of _VirtualMachines_ in the _Virtualization > Overview_ page, VMs for _All Projects_ are displayed, all the VMs incl. those that don't belong to the namespace or project originally set in the _Overview_ page (same for the Templates, Networks):
proj_before
**After:**
proj_after

### DIFF
--- a/src/utils/resources/shared.ts
+++ b/src/utils/resources/shared.ts
@@ -1,5 +1,6 @@
 import { modelToGroupVersionKind } from '@kubevirt-ui/kubevirt-api/console';
 import { V1alpha1Condition, V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import { ALL_NAMESPACES_SESSION_KEY } from '@kubevirt-utils/hooks/constants';
 import { TemplateModel } from '@kubevirt-utils/models';
 import {
   AccessReviewResourceAttributes,
@@ -41,20 +42,27 @@ export const getLabel = (entity: K8sResourceCommon, label: string, defaultValue?
  * function for getting a resource URL
  * @param {k8sModel} model - model to get the URL from
  * @param {K8sResourceCommon} resource - resource to get the URL from
+ * @param {string} activeNamespace - name of the actual project (the active namespace)
+
  * @returns the URL for the resource
  */
-export const getResourceUrl = (model: K8sModel, resource?: K8sResourceCommon): string => {
+export const getResourceUrl = (
+  model: K8sModel,
+  resource?: K8sResourceCommon,
+  activeNamespace?: string,
+): string => {
   if (!model) return null;
   const { crd, namespaced, plural } = model;
 
-  const namespace = resource?.metadata?.namespace
-    ? `ns/${resource.metadata.namespace}`
-    : 'all-namespaces';
+  const namespace =
+    resource?.metadata?.namespace ||
+    (activeNamespace !== ALL_NAMESPACES_SESSION_KEY && activeNamespace);
+  const namespaceUrl = namespace ? `ns/${namespace}` : 'all-namespaces';
 
   const ref = crd ? `${model.apiGroup || 'core'}~${model.apiVersion}~${model.kind}` : plural || '';
   const name = resource?.metadata?.name || '';
 
-  return `/k8s/${namespaced ? namespace : 'cluster'}/${ref}/${name}`;
+  return `/k8s/${namespaced ? namespaceUrl : 'cluster'}/${ref}/${name}`;
 };
 
 /**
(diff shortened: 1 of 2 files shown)


hstastna:

src/views/clusteroverview/OverviewTab/resources-inventory-card/ResourcesInventoryCard.tsx
```diff
@@ -26,7 +28,7 @@ const ResourcesInventoryCard: React.FC = () => {
             <ResourceInventoryItem
               quantity={vms}
               label={t('VirtualMachines')}
-              path={getResourceUrl(VirtualMachineModel)}
+              path={getResourceUrl(VirtualMachineModel, undefined, activeNamespace)}
```
pcbailey, line 31: I recommend changing `undefined` to `null` here and in the other places you've used it in this file. This [page](https://www.geeksforgeeks.org/undefined-vs-null-in-javascript/) gives a good explanation of the reasoning. The short version is: 

> **Null**: It is the intentional absence of the value. It is one of the primitive values of JavaScript.
> **Undefined**: It means the value does not exist in the compiler. It is the global object.

hstastna: Well, in some other resources I've found, undefined is recommended to omit a specific optional parameter if there are more of them. But I will reconsider this. Thanks!

pcbailey: [This](https://stackoverflow.com/questions/33232066/do-you-have-to-give-all-parameters-when-firing-a-function-in-javascript) supports that. I had never heard that, so thanks for teaching me something new. =)

src/utils/resources/shared.ts
```diff
@@ -41,20 +42,27 @@ export const getLabel = (entity: K8sResourceCommon, label: string, defaultValue?
...
+): string => {
   if (!model) return null;
   const { crd, namespaced, plural } = model;
 
-  const namespace = resource?.metadata?.namespace
-    ? `ns/${resource.metadata.namespace}`
-    : 'all-namespaces';
+  const namespace =
+    resource?.metadata?.namespace ||
+    (activeNamespace === ALL_NAMESPACES_SESSION_KEY ? null : activeNamespace);
```
metalice, line 59: can this also work?
const namespace =
    resource?.metadata?.namespace ||
    (activeNamespace !== ALL_NAMESPACES_SESSION_KEY && activeNamespace); ?

src/views/clusteroverview/OverviewTab/resources-inventory-card/ResourcesInventoryCard.tsx
metalice: To avoid null or undefined we can change getResourceUrl , to receive and obj instead of separate args.

src/utils/resources/shared.ts
hstastna: Yeah, that should work, too. Just more nicely written. Thanks, Matan! 😊

src/views/clusteroverview/OverviewTab/resources-inventory-card/ResourcesInventoryCard.tsx
hstastna: Yes. I was thinking about creating an interface for `getResourceUrl` to accept an object instead, but it does not seem to be very useful for other cases than this one in here..

src/utils/resources/shared.ts
hstastna: Changed!

src/views/clusteroverview/OverviewTab/resources-inventory-card/ResourcesInventoryCard.tsx
metalice: > Yes. I was thinking about creating an interface for `getResourceUrl` to accept an object instead, but it does not seem to be very useful for other cases than this one in here..

other cases will need to be edited into an obj, but this makes the function much nicer and also if in the future it will need to receive another arg make it much faster and easier.

metalice: if we can make our code base better we should try :)

hstastna: See the followup PR: https://github.com/kubevirt-ui/kubevirt-plugin/pull/962

PR #961, opened 2022-11-09
Bug 2140982: The base template of cloned template is "Not available"

avivtur:
## 📝 Description

when cloning a template we need to add labels to the template object so we can identify the base template

### After
https://user-images.githubusercontent.com/67270715/200864410-f6a3ab64-c2fc-48c0-823f-b839605af3b8.mp4

### DIFF
--- a/src/utils/components/CloneTemplateModal/CloneTemplateModal.tsx
+++ b/src/utils/components/CloneTemplateModal/CloneTemplateModal.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import produce from 'immer';
+import { LABELS } from 'src/views/templates/utils/constants';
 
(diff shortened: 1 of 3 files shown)


src/utils/components/CloneTemplateModal/CloneTemplateModal.tsx
```diff
@@ -64,6 +65,8 @@ const CloneTemplateModal: React.FC<CloneTemplateModalProps> = ({
           ...draftTemplate?.metadata?.labels,
           'template.kubevirt.io/type': TEMPLATE_TYPE_VM,
```
metalice, line 66: i know its not related to your PR, but maybe change this also to a const? add to LABELS?

PR #968, opened 2022-11-15
Bug 2142647: Delete modal lead to wrong URL

avivtur:
## 📝 Description

> Delete modal was building the return URL in the wrong way which caused to 404 error when deleting a resource with a short name like `a`
basing now the return URL on the deleted resource's model to create the URL

### Before

https://user-images.githubusercontent.com/67270715/201921035-2b98ac3d-3cf0-49ed-a201-995e75766750.mp4

### After

https://user-images.githubusercontent.com/67270715/201921026-5ffb0076-3679-4dc8-a900-6e79c2fb66c6.mp4


upalatucci (approved):
Nice

hstastna (approved):
Thank you very much for fixing this! :)

PR #970, opened 2022-11-16
Bug 2143216: Button with permission tool tip

upalatucci:
## 📝 Description

Just make a component for common logic. 
We can use this button whenever there is an action to edit the template somehow. 
Following UX guidelines.

The same type of component can also be done for VM and other resources. 
For templates, we have a different logic (we also have isCommonTemplate logic that cannot be edited) but for other resources maybe we can build a generic edit button that handle permissions for different resources


src/views/templates/details/TooltipNoEditPermissions.tsx
```diff
@@ -1,29 +0,0 @@
...
-        {/* Span here as disabled buttons do not fire any event. We need the 'hover' event. */}
-        <span>{children}</span>
-      </Tooltip>
-    );
-  }
-
-  return <>{children}</>;
-};
-
-export default TooltipNoEditPermissions;
```
avivtur, line 29: I think this component is not bad, and can be expanded for other cases like VM, MigrationPolicies, DataSources etc. adding a prop for the tooltip content pretty much settels it, or am I missing something

upalatucci: yes. I've just noticed that this component is used in combination with a disabled button so i created `ButtonWithPermissionTooltip](https://github.com/kubevirt-ui/kubevirt-plugin/pull/970/files#diff-fbc8d130302d440edfcbeac587df7101fb481eedc25efd0504f8faaef3cc7082)`

PR #972, opened 2022-11-17
Bug 2141089: Ensure vm template object has namespace

upalatucci:
## 📝 Description

**Cause**
In CDI upload, dataVolume has no namespace as the template VM object has no namespace. 

**Solution**
when VM object in template YAML has no namespace, the template namespace is implied

NOTE: change the `dvName` default name as `VM.metadata.name` in case of a template VM can be a parameter


avivtur (commented):
please add screenshots to PR's description
why implying the template's namespace if VM has no namespace? 
using the active namespace might be a better approach, wdyt?

upalatucci:
What's the benefit of using the active namespace instead of the template namespace? Are there conditions where the template has no namespace?

src/utils/components/DiskModal/utils/helpers.ts
```diff
@@ -131,7 +131,7 @@ export const getDataVolumeFromState = ({
   createOwnerReference?: boolean;
 }): V1beta1DataVolume => {
   const dataVolume = getEmptyVMDataVolumeResource(vm, createOwnerReference);
-  const dvName = resultVolume?.dataVolume?.name || `${vm?.metadata?.name}-${diskState.diskName}`;
```
upalatucci, line 134: vm name can be a template parameter so just use another thing

avivtur:
@upalatucci that's great that you tested what is happening on cli, may I assume you were logged as admin user?
if so I wonder what happens for non-admin users if the VM's namespace is not specified

> Anyway, in the template details page, the `activeNamespace` is the template namespace so I don't know the difference here. We have just less code using template namespace

than I guess using the template's namespace is really the same as using activeNamespace

src/views/templates/details/tabs/disks/utils.ts
```diff
@@ -0,0 +1,17 @@
+import { V1Template } from '@kubevirt-ui/kubevirt-api/console';
+import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import { getTemplateVirtualMachineObject } from '@kubevirt-utils/resources/template';
+
+export const getTemplateVMWithNamespace = (template: V1Template): V1VirtualMachine => {
+  const vm = getTemplateVirtualMachineObject(template);
+
+  if (vm) {
```
avivtur, line 8: maybe using produce from `immer` in here would be nicer, wdyt?

upalatucci: okay. Done

PR #977, opened 2022-11-21
Bug 2143498: Could not load template while creating VM from catalog

avivtur:
## 📝 Description

falling back to `default` namespace if the namespace is not selected (non-admin user can not click create for all projects so the `default` namespace won't apply as he must select a namespace to create)

### After
https://user-images.githubusercontent.com/67270715/203059012-9ebbd78b-fd5d-4c82-b49e-30d65b2a6473.mp4

### DIFF
--- a/src/utils/resources/template/hooks/useProcessedTemplate.ts
+++ b/src/utils/resources/template/hooks/useProcessedTemplate.ts
@@ -1,6 +1,7 @@
 import * as React from 'react';
 
 import { ProcessedTemplatesModel, V1Template } from '@kubevirt-ui/kubevirt-api/console';
+import { DEFAULT_NAMESPACE } from '@kubevirt-utils/constants/constants';
 import { k8sCreate } from '@openshift-console/dynamic-plugin-sdk';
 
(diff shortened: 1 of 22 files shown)


src/utils/resources/template/hooks/useProcessedTemplate.ts
```diff
@@ -11,7 +11,7 @@ import { k8sCreate } from '@openshift-console/dynamic-plugin-sdk';
  */
 export const useProcessedTemplate = (
   template: V1Template,
-  namespace: string,
+  namespace = 'default',
```
upalatucci, line 14: WDYT of creating a constant file under `/src/utils` for the `default` namespace? Also for `getTemplateVMWithNamespace`

avivtur: I saw we do have one under catalog directory, I'll move it

orenc1:
/cherry-pick release-4.12
https://bugzilla.redhat.com/show_bug.cgi?id=2143964

PR #978, opened 2022-11-21
Bug 2143216: Template with no edit permission

upalatucci:
## 📝 Description

Template view-mode message updated.

Do not use buttons with tooltips. 
Just use the header alert

## 🎥 Demo

image


src/views/templates/utils/constants.ts
```diff
@@ -30,10 +30,8 @@ export type SOURCE_OPTIONS_IDS =
...
-export const NO_EDIT_TEMPLATE_PERMISSIONS =
-  'You do not have permission to edit. To make changes to this template, contact your administrator.';
+// t('Can not edit in view-only mode')
+export const NO_EDIT_TEMPLATE_PERMISSIONS = 'Can not edit in view-only mode';
 
-// t('You do not have permission to delete. To make changes to this template, contact your administrator.')
-export const NO_DELETE_TEMPLATE_PERMISSIONS =
-  'You do not have permission to delete. To make changes to this template, contact your administrator.';
+// t('Can not delete in view-only mode')
+export const NO_DELETE_TEMPLATE_PERMISSIONS = 'Can not delete in view-only mode';
```
avivtur, line 37: it doesn't seem these constants are being used, if not can they be used, or if not needed then removed?

PR #986, opened 2022-11-25
CNV-21974: Part I - watch boot source in template list

upalatucci:
## 📝 Description

Make the template list responsive on boot source change. 
No screenshots as I just changed the code. 
When the user is on the template list, the 'Source available' label will appear or disappear if the data source will become available or not. 

Next pr will add the label 'Clone in progress' when data sources would be in other states.
When the user edits the boot source reference, will see the label 'Clone in progress' 

Bonus: we cache those requests in redux using `useK8sWatchResources`

Video (I've waited for 40 s for the DS import)

[Screencast from 28-11-2022 09:13:36.webm](https://user-images.githubusercontent.com/29160323/204227131-78a6de8f-ebb8-44be-89e9-898819d8e22d.webm)

### DIFF
--- a/src/views/catalog/templatescatalog/hooks/useAvailableDataSourcesAndPVCs.ts
+++ b/src/views/catalog/templatescatalog/hooks/useAvailableDataSourcesAndPVCs.ts
@@ -1,19 +1,22 @@
 import React from 'react';
 
-import { V1Template } from '@kubevirt-ui/kubevirt-api/console';
+import { PersistentVolumeClaimModel, V1Template } from '@kubevirt-ui/kubevirt-api/console';
+import DataSourceModel from '@kubevirt-ui/kubevirt-api/console/models/DataSourceModel';
 import { V1beta1DataSource } from '@kubevirt-ui/kubevirt-api/containerized-data-importer/models';
+import { IoK8sApiCoreV1PersistentVolumeClaim } from '@kubevirt-ui/kubevirt-api/kubernetes';
 import { BOOT_SOURCE } from '@kubevirt-utils/resources/template';
 import {
-  getDataSource,
-  getPVC,
   getTemplateBootSourceType,
   isDataSourceReady,
 } from '@kubevirt-utils/resources/template/hooks/useVmTemplateSource/utils';
-
-import { assertFulfilled, getSourcePromises } from './utils';
+import {
+  getGroupVersionKindForModel,
+  useK8sWatchResources,
+  WatchK8sResource,
+} from '@openshift-console/dynamic-plugin-sdk';
 
 type UniqueSourceType = {
-  [namespace: string]: Set<string>;
+  [key in string]: WatchK8sResource;
 };
 
 /**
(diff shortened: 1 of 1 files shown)


src/views/catalog/templatescatalog/hooks/useAvailableDataSourcesAndPVCs.ts
```diff
@@ -1,91 +1,100 @@
...
-        }
+    templates.forEach((t) => {
+      const bootSource = getTemplateBootSourceType(t);
 
-        if (bootSource.type === BOOT_SOURCE.PVC) {
-          const pvc = bootSource?.source?.pvc;
-          pvcs[pvc?.namespace] = new Set([...(pvcs?.[pvc?.namespace] || []), pvc?.name]);
-        }
-      });
+      if (bootSource.type === BOOT_SOURCE.DATA_SOURCE) {
```
hstastna, line 45: Do we have guaranteed that `bootSource` will always have `type`? Not sure if this won't crash, even if it's part of the if-block...

hstastna (commented):
Looks good to me! Anyway, Ugo, I'd really like to see some example of a screenshots how it visually changes in the UI (vs how it would look like without this code), as this PR does not seem to be refactoring ;)

src/views/catalog/templatescatalog/hooks/useAvailableDataSourcesAndPVCs.ts
upalatucci: there will always be a type. The worst case would be  `NONE`

upalatucci: video attached

metalice: I think u can change to reduce

metalice, line 45: can be used with reduce

upalatucci: right

src/views/catalog/templatescatalog/hooks/useAvailableDataSourcesAndPVCs.ts
```diff
@@ -26,66 +29,73 @@ export const useAvailableDataSourcesAndPVCs = (
...
   }, [templates, templatesLoaded]);
 
-  React.useEffect(() => {
-    if (uniqueSources) {
-      setLoaded(false);
-      const { uniqueDataSources, uniquePVCs } = uniqueSources;
-      const dataSourcesPromises = getSourcePromises(getDataSource, uniqueDataSources);
-      const pvcPromises = getSourcePromises(getPVC, uniquePVCs);
+  const watchDataSources = useK8sWatchResources<{
+    string?: V1beta1DataSource;
```
avivtur, line 73: can you rename `string?` to something else?

src/views/catalog/templatescatalog/hooks/useAvailableDataSourcesAndPVCs.ts
```diff
@@ -26,66 +29,73 @@ export const useAvailableDataSourcesAndPVCs = (
...
+  }>(uniqueDataSources);
+
+  const watchPVCs = useK8sWatchResources<{
+    string?: IoK8sApiCoreV1PersistentVolumeClaim;
```
avivtur, line 77: here too can you rename `string?` to something else?

src/views/catalog/templatescatalog/hooks/useAvailableDataSourcesAndPVCs.ts
```diff
@@ -1,19 +1,22 @@
...
-import { assertFulfilled, getSourcePromises } from './utils';
+import {
+  getGroupVersionKindForModel,
+  useK8sWatchResources,
+  WatchK8sResource,
+} from '@openshift-console/dynamic-plugin-sdk';
 
 type UniqueSourceType = {
-  [namespace: string]: Set<string>;
+  [key in string]: WatchK8sResource;
```
avivtur, line 19: why did you changed to `key in string`?

avivtur (commented):
looks good :) 
did you also check for non-admin users?

src/views/catalog/templatescatalog/hooks/useAvailableDataSourcesAndPVCs.ts
upalatucci: because now the object key is not the namespace but is `NAMESPACE-NAME`

avivtur:
LGTM
@metalice @hstastna can you also take another look please?

PR #987, opened 2022-11-28
Bug 2139574: blank vm name in customize

upalatucci:
## 📝 Description

Don't re-process the template just for the VM name. Use the already processed template in the parent component.
If the template is not processable due to required parameters, `canQuickCreate` is false and the user would not have the name input field. The Customize page would generate the name instead (if generatable). 

## 🎥 Demo
**Before**

https://user-images.githubusercontent.com/29160323/207339710-cba7c5e0-cb65-4c7f-bc7c-d954b49523c0.mp4

**After**

https://user-images.githubusercontent.com/29160323/207340550-8d30c402-e6e2-4531-b6b6-d2fc851b7847.mp4


hstastna (changes requested):
Using the already processed template in the parent component is a great way to get the VM name, thanks for this, Ugo! Looks like this is the improvement PR of https://github.com/kubevirt-ui/kubevirt-plugin/pull/787 :)

However, the code changes don't work in one specific case  - when WE need to generate the VM name (I mean the name with those 3 human readable words - as we used to do earlier by default no matter the template NAME parameter). For this case, the empty VM name field occurs (and it shouldn't, it worked with ):
no_name
See how the yaml looks like in this case (we need to support this case, too):
no_name_yaml
This happens also when the VM cannot be quick created  (and has the yaml as above - orange rectangle):
nono

Note that the buggy case mentioned above worked well without code changes in this PR.

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
```diff
@@ -29,173 +29,170 @@ import {
...
+            {quickCreateError && (
               <StackItem>
-                <Checkbox
-                  id="start-after-create-checkbox"
-                  isChecked={startVM}
-                  onChange={(v) => setStartVM(v)}
-                  label={t('Start this VirtualMachine after creation')}
-                />
+                <Alert variant="danger" title={t('Quick create error')} isInline>
+                  {quickCreateError.message}
```
metalice, line 152: please add ?.

upalatucci: okay

upalatucci:
thanks @hstastna i've just fixed that case

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
```diff
@@ -29,173 +29,170 @@ import {
...
 
-    const onQuickCreate = () => {
-      setIsQuickCreating(true);
-      setQuickCreateError(undefined);
+      const onQuickCreate = () => {
+        setIsQuickCreating(true);
+        setQuickCreateError(undefined);
 
-      const parameterForName = extractParameterNameFromMetadataName(template);
+        const parameterForName = extractParameterNameFromMetadataName(template);
```
hstastna, line 63: Maybe better 
`const parameterForName = extractParameterNameFromMetadataName(template) || 'NAME';` ? WDYT?

hstastna (commented):
The issue I've mentioned earlier is fixed! 👍🏽 
Now just one little thing and then I think we are good to go, Ugo! 😊

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
upalatucci: okay

hstastna:
@avivtur @metalice @pcbailey @vojtechszocs please review

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
```diff
@@ -29,173 +29,170 @@ import {
...
+            <StackItem />
+            {quickCreateError && (
               <StackItem>
-                <Checkbox
-                  id="start-after-create-checkbox"
-                  isChecked={startVM}
-                  onChange={(v) => setStartVM(v)}
-                  label={t('Start this VirtualMachine after creation')}
-                />
+                <Alert variant="danger" title={t('Quick create error')} isInline>
```
metalice, line 151: U can use AlertVariant from patternfly

src/views/catalog/customize/utils.ts
```diff
@@ -82,6 +82,12 @@ export const extractParameterNameFromMetadataName = (template: V1Template): stri
   return virtualMachineObject?.metadata.name?.replace(/[${}"]+/g, '');
 };
 
+export const getTemplateNameParameterValue = (template: V1Template): string => {
+  const nameParameter = extractParameterNameFromMetadataName(template) || 'NAME';
```
avivtur, line 86: `extractParameterNameFromMetadataName(template) || 'NAME'` is repeating here and in another file in this PR,
I would consider moving the `|| 'NAME'` part to the return value of the `extractParameterNameFromMetadataName` function

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
```diff
@@ -29,173 +29,170 @@ import {
...
+              <Split hasGutter>
+                {canQuickCreate && (
+                  <SplitItem>
+                    <Button
+                      data-test-id="quick-create-vm-btn"
+                      type="submit"
+                      form="quick-create-form"
+                      isLoading={isQuickCreating || modelsLoading}
+                      isDisabled={
+                        !isBootSourceAvailable || isQuickCreating || !vmName || modelsLoading
```
avivtur, line 167: I think it would be better to use isEmpty(models) instead of modelsLoading, wdyt?

avivtur (commented):
according to your videos I can't understand if you really fixed the bug, it still seems stuck at a loading skeleton after you click the customize, can you please provide a clip including the vm being created?

upalatucci:
OMG my videos are so bad ahahahah! 
I thought they were good but they don't even show the real issue.
The real issue is that when you click really fast the 'Customize' button on the catalog, the vmName does not shows up in the next page.

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
hstastna: Hmm, what is the preferred component we should use and why? I don't see much difference, at least from the functional point of view, but maybe there is some difference...

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
upalatucci: Yes, we can do that.

hstastna:
> according to your videos I can't understand if you really fixed the bug, it still seems stuck at a loading skeleton after you click on customize, can you please provide a clip including the vm being created?

I've tested this PR and I was not able to reproduce the issue anymore. The issue is presented in the bugzilla (video).

upalatucci:
I could reproduce it. You have to be a little bit fast or set up a throttling in the network dev tools

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
upalatucci: done thanks

avivtur (approved):

PR #994, opened 2022-12-05
Bug 2148322: Add help text to DataImportCron

hstastna:
## 📝 Description

**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2148322

Add the missing help text to _DataImportCron_ field, in the _Details_ page/tab of a _DataSource_.

## 🎥 Screenshots
**Before:**
d_before
**After:**
d_after


hstastna:

hstastna:

locales/en/plugin__kubevirt-plugin.json
```diff
@@ -876,6 +876,7 @@
   "Terms": "Terms",
   "The changes you have made require this VirtualMachine to be restarted.": "The changes you have made require this VirtualMachine to be restarted.",
   "The channel to track and receive the updates from.": "The channel to track and receive the updates from.",
+  "The DataImportCron polls disk images and imports the images as persistent volume claims. You can configure the image source and other settings on the DataImportCron details page.": "The DataImportCron polls disk images and imports the images as persistent volume claims. You can configure the image source and other settings on the DataImportCron details page.",
```
pcbailey, line 879: 'The DataImportCron...' sounds strange here. I think it should either be 'DataImportCron...' or 'The DataImportCron job...'.

'polls disk images and imports the images as...' sounds redundant. I think 'polls disk images and imports them as...' sounds cleaner.

I believe the convention for CRs is to spell them in camel case, such as `PersistentVolumeClaims`.

hstastna: The message comes from the suggestion of Avital. IMHO he is not wrong, maybe just not 100% correct. And there was another suggestion of Ronen: https://bugzilla.redhat.com/show_bug.cgi?id=2148322#c9
The help text was unfortunately not part of the design and it seems there is no one except Ronen and Avital who could help to formulate the message. It's totally up to us 🤷🏼‍♀️
Maybe we could discuss that also with someone from the doc team? Or just to copy something from the doc? Or you can add also your own suggestion ;) I'd be very happy for that! :)

pcbailey:
/lgtm

We can address the wording later, if needed.

locales/en/plugin__kubevirt-plugin.json
apinnick: For the first, I would go with 'DataImportCron'. Although "job" makes logical sense, it looks odd after an object.

+1 "imports them"

Yes, CRs are spelled in camel case.

locales/en/plugin__kubevirt-plugin.json
```diff
@@ -876,6 +876,7 @@
   "Terms": "Terms",
   "The changes you have made require this VirtualMachine to be restarted.": "The changes you have made require this VirtualMachine to be restarted.",
```
apinnick, line 877: Suggest: This VirtualMachine must be restarted to apply changes.

locales/en/plugin__kubevirt-plugin.json
```diff
@@ -876,6 +876,7 @@
...
   "The channel to track and receive the updates from.": "The channel to track and receive the updates from.",
```
apinnick, line 878: Maybe "The channel used to track and receive updates." I'm not sure what the meaning is supposed to be.

locales/en/plugin__kubevirt-plugin.json
```diff
@@ -876,6 +876,7 @@
...
+  "The DataImportCron polls disk images and imports the images as persistent volume claims. You can configure the image source and other settings on the DataImportCron details page.": "The DataImportCron polls disk images and imports the images as persistent volume claims. You can configure the image source and other settings on the DataImportCron details page.",
   "The Descheduler can be used to evict a running VirtualMachine so that the VirtualMachine can be rescheduled onto a more suitable Node via a live migration.": "The Descheduler can be used to evict a running VirtualMachine so that the VirtualMachine can be rescheduled onto a more suitable Node via a live migration.",
```
apinnick, line 880: I'm not sure what "can be used" means in this context. Are you implying that the Descheduler has other uses? 

Suggestion: "The Descheduler evicts a running VirtualMachine so that the VirtualMachine can be rescheduled on a more suitable Node by using a live migration."

I know that "by using" is more verbose than "via" but our doc guidelines do not allow Latin words. Guidelines also say that "by using" is better than "using" because it creates fewer problems for localization.

apinnick: I will try to review the other lines when I have time.

locales/en/plugin__kubevirt-plugin.json
hstastna: Thanks for the suggestion, Avital. However, this line of the code is unchanged and not related to adding help text to DataImportCron field. I'll go back to this later if needed (and fix that in a separate PR). Same for the next line and most of the other lines you've commented. The only related/changed code is the code with the green/red background in here, when looking at the code changes.

locales/en/plugin__kubevirt-plugin.json
```diff
@@ -876,6 +876,7 @@
...
   "The duration in seconds before the object should be deleted. Value must be non-negative integer. The value zero indicates delete immediately. If this value is nil, the default grace period for the specified type will be used. Defaults to a per object value if not specified. zero means delete immediately.": "The duration in seconds before the object should be deleted. Value must be non-negative integer. The value zero indicates delete immediately. If this value is nil, the default grace period for the specified type will be used. Defaults to a per object value if not specified. zero means delete immediately.",
```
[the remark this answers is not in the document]
hstastna: There's no any issue in here, don't worry. That's how this file is supposed to look like and is related to internationalization, supporting different languages in our project ;)

locales/en/plugin__kubevirt-plugin.json
apinnick: Understood. Thanks for the explanation!

PR #995, opened 2022-12-05
CNV-21086: Add new Live migratable column to VMs list

hstastna:
## 📝 Description

This is the first PR for the feature: https://issues.redhat.com/browse/CNV-21086

Add the new column + filtering VMs according to if they can be live migratable or not.

Also move the content of the file _src/views/virtualmachines/utils/utils.ts_ to _constants.ts_ (in the same dir) as it contained only constants, and add a new function `isLiveMigratable` to the original file.

## 🎥 Demo
**Before:**
add_before
**After:**
add_live_migratable_column


hstastna:

src/views/virtualmachines/list/components/VirtualMachineRow/VirtualMachineRowLayout.tsx
```diff
@@ -13,6 +14,13 @@ import { VMStatusConditionLabelList } from '../VMStatusConditionLabel';
 
 import './virtual-machine-row-layout.scss';
 
+export const isLiveMigratable = (vm: V1VirtualMachine, isSingleNodeCluster: boolean): boolean =>
+  !isSingleNodeCluster &&
+  vm?.status?.printableStatus === 'Running' &&
```
avivtur, line 19: please use printableVMStatus.Running instead of hard-coded 'Running'

src/views/virtualmachines/list/components/VirtualMachineRow/VirtualMachineRowLayout.tsx
```diff
@@ -13,6 +14,13 @@ import { VMStatusConditionLabelList } from '../VMStatusConditionLabel';
 
 import './virtual-machine-row-layout.scss';
 
+export const isLiveMigratable = (vm: V1VirtualMachine, isSingleNodeCluster: boolean): boolean =>
```
avivtur, line 17: can you please move this function to a util file? 
this function can also be used in VirtualMachineActionFactory.tsx file it will be nice if you use it there too :)

src/views/virtualmachines/utils/virtualMachineRowFilter.ts
```diff
@@ -55,6 +57,28 @@ const useStatusFilter = (): RowFilter => ({
...
 });
 
+const useLiveMigratableFilter = (): RowFilter => {
+  const [isSingleNodeCluster] = useSingleNodeCluster();
+
+  return {
+    filterGroupName: t('Live migratable'),
+    type: 'live-migratable',
+    items: [
+      { id: 'yes', title: t('Yes') },
```
avivtur, line 67: can you please add 'yes' and 'no' to a constants file? for the id's only

avivtur (commented):
some small suggestions :) 
looks good

src/views/virtualmachines/list/components/VirtualMachineRow/VirtualMachineRowLayout.tsx
hstastna: Yeah, I was not sure where exactly to put this function. Thanks! :)

src/views/virtualmachines/utils/virtualMachineRowFilter.ts
hstastna: That's a good suggestion! I totally agree! Thanks! :)

src/views/virtualmachines/details/tabs/scheduling/components/VirtualMachineSchedulingLeftGrid.tsx
```diff
@@ -40,13 +40,14 @@ const VirtualMachineSchedulingLeftGrid: React.FC<VirtualMachineSchedulingLeftGri
   const { t } = useKubevirtTranslation();
   const { createModal } = useModal();
 
-  const isMachineNotLiveMirgation = !!vm?.status?.conditions?.find(
-    ({ type, status }) => type === 'LiveMigratable' && status === 'False',
+  // TODO: Can "LiveMigratable" condition have any more values than "True" and "False"?
```
avivtur, line 43: I think it might be better to use here the `useSingleNodeCluster` hook to be able to use the function you wrote `isLiveMigratable` which is a very good solution to your comment IMO, since when you check if 
`status === 'True'` you don't really care if there's other values which is exactly what we want :)

src/views/virtualmachines/utils/virtualMachineRowFilter.ts
```diff
@@ -55,6 +58,28 @@ const useStatusFilter = (): RowFilter => ({
...
+  const [isSingleNodeCluster] = useSingleNodeCluster();
+
+  return {
+    filterGroupName: t('Live migratable'),
+    type: 'live-migratable',
+    items: [
+      { id: booleanTextIds.yes, title: t('Yes') },
+      { id: booleanTextIds.no, title: t('No') },
+    ],
+    reducer: (obj) => (isLiveMigratable(obj, isSingleNodeCluster) ? 'yes' : 'no'),
```
avivtur, line 71: can use booleanTextIds here too :)

src/views/virtualmachines/utils/virtualMachineRowFilter.ts
```diff
@@ -55,6 +58,28 @@ const useStatusFilter = (): RowFilter => ({
...
+    filter: (selectedItems, obj) => {
+      const isMigratable = isLiveMigratable(obj, isSingleNodeCluster);
+      return (
+        selectedItems?.selected?.length === 0 ||
+        (selectedItems?.selected?.includes('yes') && isMigratable) ||
+        (selectedItems?.selected?.includes('no') && !isMigratable)
```
avivtur, line 77: can use booleanTextIds here too :)

avivtur (commented):
looking good :) 
@hstastna I looked again and saw only one comment you left for me (I hope I didn't missed anything else) and responded
left some more suggestions :)

src/views/virtualmachines/details/tabs/scheduling/components/VirtualMachineSchedulingLeftGrid.tsx
hstastna: @avivtur I had more dilemmas in this place of the code, and not because I wouldn't know how to use `isLiveMigratable` here (at first I wanted to use it here), but because here it is about checking if we can edit Descheduler field. And if you look into the popover text, there's:
_Note: if VirtualMachine have LiveMigration=False condition, edit is disabled._ 
So I assumed if it is not `"False"`, then Descheduler should be editable, no matter the VM status or if there's just a one node in the cluster. And as we know, this all is checked within `isLiveMigratable`. So I assumed I better won't use `isLiveMigratable` in here + I will keep checking for "False", not "True". But if there are more options than True and False, the check for editability of the Descheduler may not be quite correct, IMO.

avivtur: I understand what you're saying that's some good thinking, I'm not sure if there's any more values other than 'True' or 'False' I'm afraid but I think that the actual descheduler is also making the proper checks in the backend so that's why checking for the LiveMigration=False condition as the helper text suggest will be enough

avivtur (approved):

PR #1006, opened 2023-01-02
[release-4.11] Bug 2119724: Customizing VM creation via template sets storageclass configuration to default

avivtur:
## 📝 Description

we apply the boot disk (source PVC) storage class to the template before proccessing it which causing the VM to be created with wrong SC.

priorities to SC should go as follows:
1. SC defined in customize VM wizard (if changed)
2. SC defined in the VM Template (can be seen in the disks tab or in the YAML under the VM's object dataVolumeTemplate)
3. Cluster's default SC (if multiple default storage classes exists, should take the first)
4. Apply with the source PVC's storage class (if boot disk is created with PVC)
5. Assign the first storage class existing

NOTE: 4 and 5 needs to be implemented.

### Before

https://user-images.githubusercontent.com/67270715/210219711-a2523912-c428-4f94-a710-a0be32240a97.mp4

### After

https://user-images.githubusercontent.com/67270715/210219719-0388713a-5a10-4cb8-9c12-2848c34bccf7.mp4


avivtur:
@tnisan the command is `/approve` without a d at the end :)

PR #1014, opened 2023-01-10
CNV-21086: Add 'Not migratable' badge to VM Overview tab

hstastna:
## 📝 Description

Add 'Not migratable' badge to _Status_ in the _Details_ card of the VM _Overview_ tab, instead of 'Live migratable' field (removed in this PR), to clearly mark the VM that isn't live migratable (only if it isn't, of course).

This change follows the very recent design changes. With this change, we achieve the UI to be more "airy" and the non live migratable VM is very easily recognized, and we also save some space in the UI.

## 🎥 Demo
**Before:**
_Live migratable_ field always present in the VM _Overview_:
over_before
**After:**
'Not migratable' badge displayed only if the VM is not live migratable:
over_after

### DIFF
--- a/src/views/virtualmachines/details/tabs/details/components/grid/rightGrid/VirtualMachineDetailsRightGridLayout.tsx
+++ b/src/views/virtualmachines/details/tabs/details/components/grid/rightGrid/VirtualMachineDetailsRightGridLayout.tsx
@@ -82,18 +82,11 @@ const VirtualMachineDetailsRightGridLayout: React.FC<VirtualMachineDetailsRightG
     });
   };
 
-  const isMigratable = isLiveMigratable(vm, isSingleNodeCluster);
-
   return (
     <GridItem span={5}>
       <DescriptionList>
         <VirtualMachineDescriptionItem
-          descriptionData={
-            <VirtualMachineStatus
-              printableStatus={vm?.status?.printableStatus}
-              isMigratable={isMigratable}
-            />
-          }
+          descriptionData={<VirtualMachineStatus vm={vm} />}
           descriptionHeader={t('Status')}
           data-test-id={`${vm?.metadata?.name}-status`}
         />
--- a/src/views/virtualmachines/details/tabs/overview/components/VirtualMachinesOverviewTabDetails/VirtualMachinesOverviewTabDetails.tsx
+++ b/src/views/virtualmachines/details/tabs/overview/components/VirtualMachinesOverviewTabDetails/VirtualMachinesOverviewTabDetails.tsx
@@ -1,5 +1,4 @@
 import React, { FC, useMemo } from 'react';
-import { getBooleanText } from 'src/views/migrationpolicies/utils/utils';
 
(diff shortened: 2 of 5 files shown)


hstastna:

avivtur (commented):
nice work :) 
nit suggestion, make it an external component under utils directory that will also encapsulate the `useSingleNodeCluster` hook and get the `vm` as a prop, wdyt?

hstastna:
> nice work :) nit suggestion, make it an external component under utils directory that will also encapsulate the `useSingleNodeCluster` hook and get the `vm` as a prop, wdyt?

I was thinking about this already for some time, but I'd have to rename isLiveMigratable to "use..." to be able to use the hook useSingleNodeCluster in there, and I didn't like it as it wouldn't be so explanatory... So I'd better left it as it is for now..

avivtur:
@hstastna I was thinking about something like this:
```
import React, { FC } from 'react';

type NotMigratableBadgeProps = {
  vm: V1VirtualMachine;
}

const NotMigratableBadge: FC<NotMigratableBadgeProps> = ({ vm }) => {
  const [isSingleNodeCluster] = useSingleNodeCluster();
  const isMigratable = isLiveMigratable(vm, isSingleNodeCluster);

  return !isMigratable ? 
                      <>
                        {' '}
                        <Badge key="available-boot">{t('Not migratable')}</Badge>
                      </> 
: null
}

export default NotMigratableBadge;
```
and than you can simply call the component wherever need `<NotMigratableBadge vm={vm} />` :)

hstastna:
> @hstastna I was thinking about something like this:
> 
> ```
> import React, { FC } from 'react';
> 
> type NotMigratableBadgeProps = {
>   vm: V1VirtualMachine;
> }
> 
> const NotMigratableBadge: FC<NotMigratableBadgeProps> = ({ vm }) => {
>   const { t } = useKubevirtTranslation();
>   const [isSingleNodeCluster] = useSingleNodeCluster();
>   const isMigratable = isLiveMigratable(vm, isSingleNodeCluster);
> 
>   return !isMigratable ? 
>                       <>
>                         {' '}
>                         <Badge key="available-boot">{t('Not migratable')}</Badge>
>                       </> 
> : null
> }
> 
> export default NotMigratableBadge;
> ```
> 
> and than you can simply call the component wherever need `<NotMigratableBadge vm={vm} />` :)

Ah, hmm, I am not sure if it worth if using this in 2 places, and it's not a lot of code...

avivtur:
> Ah, hmm, I am not sure if it worth if using this in 2 places, and it's not a lot of code...

You're right it's not a lot of code, and we currently planning to use it in only 2 places, but we should always think this might change in the future, and it would be nicer to fix in one place instead of 2 or more

hstastna:
/hold
@avivtur  Let me check the usage of SplitItem component in here, instead of using {' '}, then we could refactor the code on 3 places, not 2.

avivtur (approved):
Great work!
:trophy: :trophy:

PR #1016, opened 2023-01-11
Add StorageClass input to 'Add bootable Volume' modal

avivtur:
## 📝 Description

Main goal: 
- Adding Storage class dropdown to `Add bootable Volume` modal under upload image form only
After:
sc-input

- Making the form submit to create a DS on point at existing PVC or at the cloned PVC or uploaded disk
Side goals:
- Add disk modal was not working
Before: 
cant-create-disk

- Storage class dropdown taking the default SC as value automatically on the first load, either if we already have a previous value or not, which is incorrect (should simply take an existing value from YAML file if exists or stay empty if not selected)
Before: 
https://user-images.githubusercontent.com/67270715/211790070-f496c6f4-c339-48a3-9f05-b31d90fdcffa.mp4

## 🎥 Demo
Disk modal is now able to create a disk without enforcing any value from the UI and still get to proper default SC by the backend. (not included in video/screenshot but hotplug disk creation was having same issues which now are resolved)

https://user-images.githubusercontent.com/67270715/211790997-48488b41-995d-492e-a644-48b4c4745b18.mp4


pcbailey:
Looks good to me, but I'd like to have someone else review it as well since it's used in multiple places and I don't have that much direct experience with this area of the code.

src/views/catalog/CreateVMHorizontalNav/components/CreateFromInstanceTypes/components/AddBootableVolumeButton/components/AddBootableVolumeModal/AddBootableVolumeModal.tsx
```diff
@@ -16,12 +16,14 @@ import CapacityInput from '@kubevirt-utils/components/CapacityInput/CapacityInpu
 import { removeByteSuffix } from '@kubevirt-utils/components/CapacityInput/utils';
```
metalice, line 16: i think that some functions can be extracted to utiil? what do u think?

avivtur: I agree with that, I think this PR is already quite big so maybe can you name which functions other than this one you had in mind and I'll create a separate PR for moving them to utils, wdyt?

src/views/catalog/CreateVMHorizontalNav/components/CreateFromInstanceTypes/components/AddBootableVolumeButton/components/AddBootableVolumeModal/AddBootableVolumeModal.tsx
```diff
@@ -179,21 +189,34 @@ const AddBootableVolumeModal: FC<AddBootableVolumeModalProps> = ({
...
+        <FormGroup label={t('Preferences')} isRequired>
           <FilterSelect
             selected={labels?.[DEFAULT_PREFERENCE_LABEL]}
             setSelected={setBootableVolumeField('labels', DEFAULT_PREFERENCE_LABEL)}
             options={preferencesNames}
             optionLabelText={t('Preference')}
           />
         </FormGroup>
-        <FormGroup label={t('Preferred Instancetype')}>
+        <FormGroup label={t('Instancetypes')} isRequired>
```
pcbailey, line 214: Shouldn't this be 'InstanceTypes'?

PR #1020, opened 2023-01-16
Bug 2151056: Improve descriptive text of cloud-init and ssh-key

hstastna:
## 📝 Description

**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2151056

Improve the descriptive text of the _Cloud-init_ field, to be more precise, in th VM's _Scripts_ tab, for both - already existing VM and also for the VM that is going to be created.

For _Authorized SSH Key_ field, remove "The key will be stored after the machine is created" text - for the VM that is already created.

## 🎥 Screenshots
**Before:**
Already existing VM:
cl_before1
Customizing VM that's going to be created:
cl_before2
**After:**
Already existing VM:
cl_after1
Customizing VM that's going to be created:
cl_after2


hstastna:

PR #1026, opened 2023-01-23
CNV-23381: Make necessary changes to VM details tabs

avivtur:
## 📝 Description

> This PR focuses on making sure the VM details tabs view is still functional and nothing crashes

** Fixing a crash of VM disks tab, and cancelling the pending changes for VMs with instancetype or preference section included in the VM YAML

## 🎥 Demo

> https://user-images.githubusercontent.com/67270715/214373191-b4cef08b-e4ac-4f78-98ed-0b9ae9a4a5b9.mp4

### DIFF
--- a/src/utils/resources/vm/utils/disk/rowData.ts
+++ b/src/utils/resources/vm/utils/disk/rowData.ts
@@ -13,6 +13,7 @@ import {
   getPrintableDiskInterface,
 } from '@kubevirt-utils/resources/vm/utils/disk/selectors';
 import { formatBytes } from '@kubevirt-utils/resources/vm/utils/disk/size';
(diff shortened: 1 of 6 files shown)


src/views/virtualmachines/actions/actions.ts
```diff
@@ -108,3 +109,6 @@ export const deleteVM = async (vm: V1VirtualMachine) => {
     resource: vm,
   });
 };
+
+export const expandVirtualMachineSpec = async (vm: V1VirtualMachine) =>
```
pcbailey, line 113: I think we should use `expandVMSpec` to stay consistent with the other commands here.

avivtur: fixed thanks :)

PR #1034, opened 2023-01-31
CNV-21974: clone in progress label

upalatucci:
## 📝 Description

Forward also cloning DS and add label to them in the template list

Do you think other states will be added in the future? Maybe better to consider forwarding all the DS and filtering based on states later?  

## 🎥 Demo

**Before**
Screenshot from 2023-01-31 17-48-32

**After**
Screenshot from 2023-01-31 17-48-41

### DIFF
--- a/src/utils/resources/template/hooks/useVmTemplateSource/utils.ts
+++ b/src/utils/resources/template/hooks/useVmTemplateSource/utils.ts
@@ -109,6 +109,19 @@ export const isDataSourceReady = (dataSource: V1beta1DataSource): boolean =>
   dataSource?.status?.conditions?.some((c) => c.type === 'Ready' && c.status === 'True');
 
 /**
+ * a function that returns true if the data source is cloning in progress
+ * @param dataSource the data source to check if cloning
+ * @returns true if the data source is in cloning state, false otherwise
+ */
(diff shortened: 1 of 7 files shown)


src/utils/resources/template/hooks/useVmTemplateSource/utils.ts
```diff
@@ -108,6 +108,19 @@ export const getDataSourcePVC = (name: string, ns: string) =>
...
 
+/**
+ * a function that returns true if the data source is cloning in progress
+ * @param dataSource the data source to check if cloning
+ * @returns true if the data source is in cloning state, false otherwise
+ */
+export const isDataSourceCloning = (dataSource: V1beta1DataSource): boolean =>
+  dataSource?.status?.conditions?.some(
+    (c) =>
+      c.type === 'Ready' &&
```
hstastna, line 119: Do the conditions always have `type`, `status` and `reason` present?

hstastna (commented):
Just had one question, otherwise looks good to me! 💪🏽

src/utils/resources/template/hooks/useVmTemplateSource/utils.ts
upalatucci: `reason` is optional. I'll fix that. Thanks for the good catch

hstastna (approved):

PR #1036, opened 2023-02-03
CNV-23494: Select loadbalancer as ssh service type

upalatucci:
## 📝 Description

Add the ability to choose which type of service to create.
Load Balancer (only if metalLB is installed)  or NodePort

## 🎥 Demo

Screenshot from 2023-02-03 13-55-53
Screenshot from 2023-02-03 13-55-49


hstastna (commented):
I've checked the code and also tested in the UI, works nice :) 
Please, fix the typo. Maybe it worth also to consider to implement some check if there really was a change, regarding the selected SSH service type. I've tried to click on the same type as already chosen, and it triggered `onSSHChange` function unnecessary. Maybe it worth to avoid creating/deleting promises etc, I mean executing the whole code of `onSSHChange` if it's unnecessary. WDYT? 🤔 
Otherwise, great work, Ugo! ❇️

src/utils/components/SSHAccess/components/SSHCommand.tsx
```diff
@@ -38,22 +35,37 @@ const SSHCommand: React.FC<SSHCommandProps> = ({
...
 
-  const onSSHChange = (enableSSH: boolean) => {
+  const onSSHChange = async (newServiceType: SERVICE_TYPES) => {
     setLoading(true);
 
-    const request = enableSSH
-      ? createSSHService(vm, vmi).then(setSSHService)
-      : deleteSSHService(sshService).then(() => setSSHService(undefined));
+    try {
+      const deletePromise = sshService
```
metalice, line 47: why u need to use empty promise? what happen if u return undefined Promise.allSettled don't like it?

src/utils/components/SSHAccess/components/SSHCommand.tsx
```diff
@@ -38,22 +35,37 @@ const SSHCommand: React.FC<SSHCommandProps> = ({
...
+        ? deleteSSHService(sshService)
+        : new Promise<void>((resolve) => resolve());
 
-    request.catch(setError).finally(() => setLoading(false));
-  };
+      const createPromise =
```
metalice, line 51: same as above

src/utils/components/SSHAccess/components/SSHCommand.tsx
```diff
@@ -38,22 +35,37 @@ const SSHCommand: React.FC<SSHCommandProps> = ({
...
+        !newServiceType || newServiceType === SERVICE_TYPES.NONE
+          ? new Promise((resolve) => resolve(null))
+          : createSSHService(vm, vmi, newServiceType);
 
-  const hasNoUsername = user === CLOUD_INIT_MISSING_USERNAME;
+      await Promise.allSettled([deletePromise, createPromise]).then(
+        ([deleteResult, createResult]) => {
```
metalice, line 57: why do u combine them? create and delete and the same logic? its 2 different actions.

src/utils/components/SSHAccess/components/SSHServiceSelect.tsx
```diff
@@ -0,0 +1,68 @@
...
+const SSHServiceSelect: FC<SSHServiceSelectProps> = ({
+  sshService,
+  sshServiceLoaded,
+  onSSHChange,
+}) => {
+  const { t } = useKubevirtTranslation();
+  const [isOpen, setIsOpen] = React.useState(false);
+  const [models] = useK8sModels();
+
+  const hasSomeMetalCrd = Object.keys(models).some((modelGroupVersionKind) =>
```
metalice, line 25: maye add usememo with [models]?

src/utils/components/SSHAccess/components/SSHCommand.tsx
upalatucci: Yes, i would not wait for delete to start the creation request. But okay, I'll wait first for the delete and then create

src/utils/components/SSHAccess/components/SSHServiceSelect.tsx
```diff
@@ -0,0 +1,68 @@
...
+      onSelect={handleChange}
+      variant={SelectVariant.single}
+      selections={sshService?.spec?.type ?? SERVICE_TYPES.NONE}
+      isDisabled={!sshServiceLoaded}
+    >
+      <SelectOption value={SERVICE_TYPES.NONE}>{t('None')}</SelectOption>
+      <SelectOption
+        value={SERVICE_TYPES.NODE_PORT}
+        description={t(
+          'Opens a specific port on all Nodes in the cluster. If the Node is publicly accessible, any trtaffic that is sent to this port is forwarded to the Service',
```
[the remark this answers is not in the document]
upalatucci: the logic to prevent onSSHChange from firing unnecessarily is [here](https://github.com/kubevirt-ui/kubevirt-plugin/pull/1036/files#diff-4733b7d6e33fb5a1cfc1fb47ee70a01dede35dbf7794de437c6948fddfb69703R38). There was a bug. Fixed I think :D

PR #1039, opened 2023-02-08
Display only available DataSources for VM creation

hstastna:
## 📝 Description

Display **only 'available' DataSources** when creating a new VM from a bootable volume, instead of displaying all of them, and so fix the bug leading to not being able to create a VM or not being able to start a created VM - a VM created by selection of a volume (for now a `DataSource` resource) that is not ready/available (a.ka. `DataSource` with status condition type `'Ready'` and status `'False'`, for example when a PVC does not exist).

Sort the Instancetypes and preferences names in the appropriate drop downs alphabetically, for better search experience.

Change 'Upload Data' to 'Upload data', also 'Preference' to 'preference' in the _Add bootable volume_ modal, according to to design suggestions and UI consistency.

## 🎥 Demo
**Before:**
Displaying all the DataSources, not only those being 'ready':
list_before
After choosing such a DataSource (or a 'bootable volume', when creating a VM) and checking the checkbox to start a VM right after creation, nothing happens, the VM remains 'stopped' + error in the browser console:
stopped_vm
_Add bootable volume_ modal with unsorted preferences' (and instance types') names:
modal_before
**After:**
Only the available/ready DataSources being displayed:
list_after
VM not being stuck in 'Stopped' status, after its creation:
after_vm_creation
_Add bootable volume_ modal providing alphabetically sorted preferences' (and instance types') names:
modal_after


hstastna:

PR #1043, opened 2023-02-09
fix: create ssh service with no labels obj

upalatucci:
## 📝 Description

Error during ssh service creation.
VMs without `spec.template.metadata.labels` throw errors on k8sPatch.

Switch to `k8sUpdate` creating first the label object and after adding it

### DIFF
--- a/src/utils/components/SSHAccess/utils.ts
+++ b/src/utils/components/SSHAccess/utils.ts
@@ -1,14 +1,16 @@
+import produce from 'immer';
+
 import { ServiceModel } from '@kubevirt-ui/kubevirt-api/console';
 import VirtualMachineInstanceModel from '@kubevirt-ui/kubevirt-api/console/models/VirtualMachineInstanceModel';
 import VirtualMachineModel from '@kubevirt-ui/kubevirt-api/console/models/VirtualMachineModel';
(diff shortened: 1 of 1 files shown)


src/utils/components/SSHAccess/utils.ts
```diff
@@ -44,44 +46,60 @@ export const deleteSSHService = (sshService: IoK8sApiCoreV1Service) =>
...
 
-export const createSSHService = (
+export const addSSHSelectorLabelToVM = async (
+  vm: V1VirtualMachine,
+  vmi: V1VirtualMachineInstance,
+  labelValue,
+) => {
+  const vmWithLabel = produce(vm, (draftVM) => {
+    if (!draftVM?.spec?.template?.metadata) {
+      draftVM.spec.template.metadata = {};
```
hstastna, line 56: Maybe you could use `ensurePath` here, the one from the other PR? :)

hstastna (commented):
I've Just left one comment , otherwise looks good to me!

src/utils/components/SSHAccess/utils.ts
upalatucci: fixed

hstastna (approved):

PR #1054, opened 2023-02-13
CNV-18014: Add Size drop down to Add volume modal

hstastna:
## 📝 Description

Add missing _Size_ dropdown to _Add volume to boot from_ modal when creating a VM, also to _Edit_ modal in the _Bootable volumes_ list, to separate workload type and size, and so to set an appropriate Instancetype label to the selected `DataSource`.

Fixes:

## 🎥 Demo
**Before:**
 _Edit_ modal in the _Bootable volumes_ list:
bootable_before
 No _Size_ drop down in _Add volume to boot from_ modal:
size_before

**After:**
 _Edit_ modal in the _Bootable volumes_ list:
bootable_after
_Size_ drop down in _Add volume to boot from_ modal:
size_after
After _Instancetype_ selection, user can select an appropriate size:
size2_after


hstastna:

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
```diff
@@ -41,51 +41,106 @@ const EditBootableVolumesModal: FC<EditBootableVolumesModalProps> = ({
     () => Object.keys(convertResourceArrayToMap(preferences)).sort((a, b) => a.localeCompare(b)),
     [preferences],
   );
-  const initialMetadata: BootableVolumeMetadata = useMemo(() => {
+
+  const instanceTypesAndSizesObject = useMemo(
```
pcbailey, line 45: How would you feel about changing the name to `instanceTypesToSizesMap`? I just looked and it appears we're passing instanceTypesNames down through a lot of components, but I'm not sure it's entirely necessary. We can look at that outside of this PR, though.

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
```diff
@@ -41,51 +41,106 @@ const EditBootableVolumesModal: FC<EditBootableVolumesModalProps> = ({
...
+    () => getInstanceTypesWithSizes(instanceTypesNames),
+    [instanceTypesNames],
+  );
+
+  const initialParams = useMemo(() => {
+    const instanceTypeLabel =
```
pcbailey, line 51: How would you feel about changing the name to `instanceTypeCode`? 'Label' makes me think of the display name.

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
```diff
@@ -41,51 +41,106 @@ const EditBootableVolumesModal: FC<EditBootableVolumesModalProps> = ({
...
 
-  const setBootableVolumeMetadata = (key: string, fieldKey: string) => (value: string) =>
-    setMetadata((prevState) => ({
-      ...prevState,
-      [key]: { ...prevState[key], [fieldKey]: value },
-    }));
+  const [preference, setPreference] = useState<string>(initialParams.preference);
+  const [instanceType, setInstanceType] = useState<string>(initialParams.instanceType);
+  const [size, setSize] = useState<string>(initialParams.size);
+  const [sizeOptions, setSizeOptions] = useState<string[]>(
```
pcbailey, line 65: nitpick: Just for organization purposes, I would move this below the description state to keep all of the state coming from the initialParams object together. You could also use `useMemo` to have it automatically recalculate whenever `instanceType` changes instead of setting it manually.

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
```diff
@@ -41,51 +41,106 @@ const EditBootableVolumesModal: FC<EditBootableVolumesModalProps> = ({
...
+    instanceTypesAndSizesObject[instanceType] || [],
+  );
+  const [description, setDescription] = useState<string>(initialParams.description);
+
+  const onInstanceTypeChange = (value: string) => {
```
pcbailey, line 70: I recommend using a name more specific than `value` to indicate what information it carries. I assume these are `InstanceTypeCategory` values? If so, you can also try using that for the type to provide more information.

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
```diff
@@ -41,51 +41,106 @@ const EditBootableVolumesModal: FC<EditBootableVolumesModalProps> = ({
...
-            selected={metadata?.labels?.[DEFAULT_INSTANCETYPE_LABEL]}
-            setSelected={setBootableVolumeMetadata('labels', DEFAULT_INSTANCETYPE_LABEL)}
-            options={instanceTypesNames}
-            groupVersionKind={VirtualMachineClusterInstancetypeModelGroupVersionKind}
-            optionLabelText={t('Instancetype')}
-          />
-        </FormGroup>
+        <Grid hasGutter>
+          <GridItem span={6}>
+            <FormGroup label={t('Default Instancetype')}>
```
pcbailey, line 118: I think this should be 'InstanceType' based on what you told me today.

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
```diff
@@ -41,51 +41,106 @@ const EditBootableVolumesModal: FC<EditBootableVolumesModalProps> = ({
...
+              <FilterSelect
+                selected={instanceType}
+                setSelected={onInstanceTypeChange}
+                options={Object.keys(instanceTypesAndSizesObject) || []}
+                groupVersionKind={VirtualMachineClusterInstancetypeModelGroupVersionKind}
+                optionLabelText={t('Instancetype')}
```
pcbailey, line 124: I think this should be 'InstanceType' based on what you told me today.

src/views/bootablevolumes/utils/utils.ts
```diff
@@ -81,3 +81,24 @@ export const changeBootableVolumeMetadata =
         ],
       }));
   };
+
+interface InstanceTypesAndSizesObject {
```
pcbailey, line 85: I suggest renaming the type to `InstanceTypesToSizesMap`, too. I think the type should be moved to bootablevolumes/utils/types.

src/views/bootablevolumes/utils/utils.ts
```diff
@@ -81,3 +81,24 @@ export const changeBootableVolumeMetadata =
...
+  [key: string]: string[];
+}
+
+export const getInstanceTypesWithSizes = (
+  instanceTypesNames: string[] = [],
+): InstanceTypesAndSizesObject => {
+  // eslint-disable-next-line prefer-const
+  let instanceTypesAndSizes: InstanceTypesAndSizesObject = {};
+
+  instanceTypesNames.forEach((instanceType) => {
```
pcbailey, line 95: You could use `reduce` here to create the object. As is, I believe you can use `const` for the object variable since adding properties to the object doesn't change its reference.

src/views/catalog/CreateFromInstanceTypes/components/AddBootableVolumeModal/AddBootableVolumeModal.tsx
```diff
@@ -71,6 +72,46 @@ const AddBootableVolumeModal: FC<AddBootableVolumeModalProps> = ({
 
   const isUploadForm = formSelection === RADIO_FORM_SELECTION.UPLOAD_IMAGE;
 
+  const instanceTypesAndSizesObject = useMemo(
```
pcbailey, line 75: Same rename suggestion: `instanceTypesToSizesMap`

src/views/catalog/CreateFromInstanceTypes/components/AddBootableVolumeModal/AddBootableVolumeModal.tsx
```diff
@@ -71,6 +72,46 @@ const AddBootableVolumeModal: FC<AddBootableVolumeModalProps> = ({
...
+    setBootableVolume((prevState) => ({
+      ...prevState,
+      labels: {
+        ...prevState?.labels,
+        [DEFAULT_INSTANCETYPE_LABEL]: `${value}.${newSizesArray[0]}`,
+      },
+    }));
+  };
+
+  const onInstanceSizeChange = (value: string) => {
```
pcbailey, line 104: Recommend a more descriptive parameter name.

src/views/catalog/CreateFromInstanceTypes/components/AddBootableVolumeModal/AddBootableVolumeModal.tsx
```diff
@@ -71,6 +72,46 @@ const AddBootableVolumeModal: FC<AddBootableVolumeModalProps> = ({
...
+    [labels],
+  );
+
+  const [instanceType, setInstanceType] = useState<string>(instanceTypeAndSize[0]);
+  const [instanceSize, setInstanceSize] = useState<string>(instanceTypeAndSize[1]);
+  const [sizeOptions, setSizeOptions] = useState<string[]>(
+    instanceTypesAndSizesObject[instanceType] || [],
+  );
+
+  const onInstanceTypeChange = (value: string) => {
```
pcbailey, line 90: Recommend a more descriptive parameter name.

src/views/catalog/CreateFromInstanceTypes/components/AddBootableVolumeModal/AddBootableVolumeModal.tsx
```diff
@@ -190,15 +231,32 @@ const AddBootableVolumeModal: FC<AddBootableVolumeModalProps> = ({
...
-            setSelected={setBootableVolumeField('labels', DEFAULT_INSTANCETYPE_LABEL)}
-            options={instanceTypesNames}
-            groupVersionKind={VirtualMachineClusterInstancetypeModelGroupVersionKind}
-            optionLabelText={t('Instancetype')}
-          />
-        </FormGroup>
+
+        <Grid hasGutter>
+          <GridItem span={6}>
+            <FormGroup label={t('Default Instancetype')}>
```
pcbailey, line 237: I think this should be 'InstanceType'.

src/views/catalog/CreateFromInstanceTypes/components/AddBootableVolumeModal/AddBootableVolumeModal.tsx
```diff
@@ -190,15 +231,32 @@ const AddBootableVolumeModal: FC<AddBootableVolumeModalProps> = ({
...
+              <FilterSelect
+                selected={instanceType}
+                setSelected={onInstanceTypeChange}
+                options={Object.keys(instanceTypesAndSizesObject) || []}
+                groupVersionKind={VirtualMachineClusterInstancetypeModelGroupVersionKind}
+                optionLabelText={t('Instancetype')}
```
pcbailey, line 243: I think this should be 'InstanceType'.

pcbailey (commented):
It's looking really good! I just had some minor suggestions.

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
hstastna: Okay, I'll move it below the description state. But I am not sure how to combine useState with useMemo for sizeOptions (and if it ever is a good idea)... Or how did you mean that?

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
hstastna: Yes, and that is fixed in a different PR: https://github.com/kubevirt-ui/kubevirt-plugin/pull/1056

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
hstastna: Yes, and that is fixed in a different PR: https://github.com/kubevirt-ui/kubevirt-plugin/pull/1056

src/views/bootablevolumes/utils/utils.ts
hstastna: Yes, after I created this function, I realized I could achieve the same using `reduce`, but this code is much more readable as it is now. Also I haven't found any significant advantage of using reduce, from the performance point of view. I consider readability of this part of the code to be important, because it may not be clear to other programmers what we are trying to achieve in here. We are doing uncommon thing with the specific label - dividing the string into 2 and creating a special object for the data to be easily used in "Default InstanceType" and "Size" dropdowns. Because we are dealing with 2 things but those are somehow connected via one resource (`VirtualMachineClusterInstancetype`) and one label (_instancetype.kubevirt.io/default-instancetype)_, which is IMO not the best idea. But that's how the API for our feature was designed, unfortunately...
Regarding using `const` - I'll try (again), but I think it didn't work when not using `reduce` (not sure why).

src/views/catalog/CreateFromInstanceTypes/components/AddBootableVolumeModal/AddBootableVolumeModal.tsx
hstastna: Already fixed in another PR ;)

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
pcbailey: Ah, ok. That makes sense. =)

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
hstastna: > I just looked and it appears we're passing instanceTypesNames down through a lot of components

I looked into that and there isn't any case where it wouldn't be useless to pass it down thru lots of components. We need it in many places. Also note that we are doing it to minimize the amount of calls to the server for this resource, so we call it once in the parent component and pass it down to the other components that need the same info.

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
pcbailey: `useMemo` caches the result of a calculation and returns it on each re-render unless one of its dependencies changes. It's made specifically to watch props and state for changes.

In this case, it would look something like: 

`const sizeOptions = useMemo(() => instanceTypesToSizeMap[instanceType], [instanceType, instanceTypesToSizeMap]);`

You could then remove the call to `setSizeOptions` as it would no longer be needed.

src/views/bootablevolumes/utils/utils.ts
pcbailey: I'm a huge advocate for readability and I can't see an impact to readability here unless the programmer doesn't understand what `reduce` does, which would be odd for a JavaScript programmer.

In this case, the logic in the body wouldn't really change at all, but it would prevent the requirement for creating an object externally and using side effects from the loop to populate it. 

```
const sizesByInstanceType: SizesByInstanceType = instanceTypeNames.reduce((instanceTypesToSizesMap, instanceType) => {
  const [instanceTypeName, size] = instanceType.split('.');

  instanceTypesToSizesMap[instanceTypeName] !== undefined
    ? instanceTypesToSizesMap[instanceTypeName].push(size)
    : (instanceTypesToSizesMap[instanceTypeName] = [size]);
  });
  
  return instanceTypesToSizesMap;
}, {});
```

src/views/catalog/CreateFromInstanceTypes/components/AddBootableVolumeModal/AddBootableVolumeModal.tsx
```diff
@@ -71,6 +72,46 @@ const AddBootableVolumeModal: FC<AddBootableVolumeModalProps> = ({
...
+    setBootableVolume((prevState) => ({
+      ...prevState,
+      labels: {
+        ...prevState?.labels,
+        [DEFAULT_INSTANCETYPE_LABEL]: `${instanceTypeName}.${newSizesArray[0]}`,
+      },
+    }));
+  };
+
+  const onInstanceSizeChange = (sizeValue: string) => {
```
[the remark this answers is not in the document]
hstastna: Because it is already declared in the file earlier...

src/views/bootablevolumes/utils/types.ts
```diff
@@ -2,3 +2,7 @@ export type BootableVolumeMetadata = {
   labels: { [key: string]: string };
   annotations: { [key: string]: string };
 };
+
+export type InstanceTypesAndSizesMapType = {
```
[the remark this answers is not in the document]
hstastna: I would not use "To" substring in the naming because IMO it is misleading. We don't convert/change anything to size, size is part of the string/whole information we are dealing with. We just divide the whole label's string in here. 

Another note from me is that I don't think it's always good idea to follow some conventions/general recommendations. IMO readability of the code for a programmer as a higher priority to me, and in this case the suggested naming is less clear or self explaining than the original one.

src/views/bootablevolumes/utils/utils.ts
hstastna: however I don't quite agree with everything mentioned above and even when the code above won't work, I'll update this part asap ;) 👍🏽  Thanks!

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
hstastna: I understand what you mean, seems good to me, except that I am not sure if the Size drop down will be re-rendered accordingly, when value of `instanceType` changes...

src/views/bootablevolumes/utils/types.ts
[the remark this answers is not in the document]
pcbailey: From what I can tell here, you're taking InstanceType codes such as `c4.xlarge` and breaking them down into the InstanceType name shorthand (c4) and the size (xlarge). You're then creating a mapping, like below. Correct me if I'm wrong. Perhaps I'm simply misunderstanding what the code is doing here.

```
const map = {
  c4: [xlarge,...],
  ...
}
``` 

If the above example is correct, "To" is a much more accurate term than "and" for this use case. A map connects the values to the keys by mapping the key TO the value. Using "and" implies a group of InstanceTypes and sizes together in one group/array, which is incorrect. Once you access the map using an InstanceType shorthand name, you only get the sizes for that InstanceType category, not a list of sizes AND InstanceType names.

I completely agree that readability is paramount. Conventions typically exist to improve readability not hinder it. I don't see how "XWithY" is clearer than "XtoYMap"/"XToY" or "YByX". I have no idea what "with" is supposed to mean in this instance.

In any case, "Type" shouldn't be used in the name.

src/views/catalog/CreateFromInstanceTypes/components/AddBootableVolumeModal/AddBootableVolumeModal.tsx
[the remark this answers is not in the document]
pcbailey: Ah, gotcha.

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
pcbailey: When `instanceType` changes, the component will re-render. When it re-renders, `useMemo` will know that it has changed and recompute `sizeOptions`, which will then be passed to the dropdown causing it to re-render since its props will have changed.

src/views/bootablevolumes/utils/types.ts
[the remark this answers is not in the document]
hstastna: > In any case, "Type" shouldn't be used in the name.

I mostly agree, but not 100%. IMO, it depends on a situation...

Anyway, we don't have much time, so can you please check the actual version of the code? I think you are still looking at some old version of the code. It was updated :)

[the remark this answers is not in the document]
pcbailey: Weird. I refreshed last night before commenting.

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
```diff
@@ -41,51 +41,107 @@ const EditBootableVolumesModal: FC<EditBootableVolumesModalProps> = ({
...
+    () => instanceTypesToSizesMap[instanceType] || [],
+    [instanceType, instanceTypesToSizesMap],
+  );
+
+  const onInstanceTypeChange = (instanceTypeName: string) => {
+    setInstanceType(instanceTypeName);
+    setSize(instanceTypesToSizesMap[instanceTypeName][0]);
+  };
+
+  const onChangeVolumeParams = () => {
```
vojtechszocs, line 77: Just a note, any functions passed to React components as props (e.g. callbacks) should be referentially stable across re-renders. Otherwise, you're creating a new function instance on every render here, passing it to other React components and causing re-render of these components.

This can be done via `React.useCallback` hook https://reactjs.org/docs/hooks-reference.html#usecallback

On the other hand, the extra re-renders aren't normally a big issue, so developers may opt into not doing this.

hstastna: `React.useCallback` was added in some of the followup PRs.

PR #1055, opened 2023-02-13
CNV-15501: Update create VM from volume pages

hstastna:
## 📝 Description

This PR is related to the feature: https://issues.redhat.com/browse/CNV-15501

Update the pages according to the design changes, fix various issues, such as:
- use 'Advanced settings' instead of 'Advanced'
- change title of the dialog window containing list of bootable volumes from 'Bootable volumes' to 'Volumes to boot from'
- change title of the button to create a VM from 'Quick create VirtualMachine' to 'Create VirtualMachine', as there's not gonna be any customization, so 'quick' creation doesn't make sense
- change 'Gi' units to more readable 'GiB'
- display the bootable volumes names as black texts in the list (and in the dialog displaying list of all the volumes), as there's not gonna be any redirection to the details page, when creating a VM from a volume

## 🎥 Demo
**Before:**
'Advanced' what?
adv_before
Older 'Bootable volumes' dialog window (with the blue text color for the volumes' names):
all_before
'Quick create VirtualMachine' (while there's not gonna be any VM customization, in these pages):
create-before
'Gi' + no any space between the value and the unit:
mem-before
The blue text color for the names in the list (when clicking, it leads to nowhere):
text_before

**After:**
It's clear what this expandable section is about:
adv_after
'Volumes to boot from' dialog window (with nice, black and clearly unclickable volumes' names):
all_after
Simply 'Create VirtualMachine':
create-after
'GiB' + a space between the value and the unit, for better readability:
mem-after
The black text color for the names in the list, so the user does not expect to go to some details pages, when clicking:
text_after


hstastna:
@metalice @upalatucci @vojtechszocs please review

PR #1056, opened 2023-02-15
CNV-18014: Update text in Add volume modal

hstastna:
## 📝 Description

Update the strings in the _Add volume to boot from_ modal, same ones in the _Edit_ modal in the _Bootable volumes_ list, according to the design changes. Add the subtitle to _Add volume to boot from_ modal: "You can upload a new volume or use an existing PersistentVolumeClaim (PVC)".

Add missing helper texts to some of the modal's fields - to _Volume name_ and _PVC project_.

Update also some of the strings in the same components (as in _Add volume to boot from_ modal) but used in other places (e.g. _PVC project_ and _PVC name_ drop downs), for consistency reasons.

Update the subtitle for _Create new VirtualMachine_ page, according to the design document to "Select an option to create a VirtualMachine from."

Make the checkbox _Clone existing PVC_ checked and disabled (so the user cannot change it, at least for now), so the user has to clone the PVC by default (for 4.13, may change in the future versions).

Make the `HelpIcon`  next to the checkbox _Clone existing PVC_ smaller, to be displayed as expected.

_This PR is related to features:_

## 🎥 Demo
**Before:**
When uploading a new volume/image:
a1_before
When using an existing PVC:
a2_before
_Edit_ modal for the volumes in the _Bootable volumes_ list:
edit_before
Creating a VM from volume:
select_before

**After:**
When uploading a new volume/image:
a1_afterr
When using an existing PVC:
a2_afterr
_Edit_ modal for the volumes in the _Bootable volumes_ list:
edit_vol_after
Creating a VM from volume:
select_after


hstastna:

PR #1058, opened 2023-02-16
CNV-21086: Display "Not migratable" only for running VMs

hstastna:
## 📝 Description

Display "Not migratable" label only for the VMs that are running.

The purpose of this change is to limit the amount of labels and to display the labels only when necessary.

## 🎥 Demo
**Before:**
live_before

**After:**
live_after


hstastna:

PR #1064, opened 2023-02-20
CNV-18014: Add help text to Add volume and Edit modals

hstastna:
## 📝 Description

Add help icons with popovers to _Add volume to boot from_ and to _Edit volume parameters_ modals, to _Preference_ and also _Default InstanceType_ fields, to provide more info to the user about those new fields.

_This PR is related to features:_

## 🎥 Screenshots
**Before:**
_Add volume to boot from_ dialog window:
1_before
_Edit volume parameters_ dialog window:
2_before

**After:**
_Add volume to boot from_ dialog window:
1_after
2_after
_Edit volume parameters_ dialog window:
3_after
4_after


hstastna:

src/utils/components/PreferenceHelpTextIcon/PreferenceHelpTextIcon.tsx
```diff
@@ -0,0 +1,24 @@
+import * as React from 'react';
```
metalice, line 1: import React from 'react' can be used

src/utils/components/PreferenceHelpTextIcon/PreferenceHelpTextIcon.tsx
```diff
@@ -0,0 +1,24 @@
...
+      bodyContent={t('The default set of preferences to associate with the volume')}
+      className="icon-size-small"
+      position="right"
+    >
+      <HelpIcon />
+    </Popover>
+  );
+};
+
+export default PreferenceHelpTextIcon;
```
metalice, line 24: it is a duplication of [InstanceTypeHelpTextIcon.tsx](https://github.com/kubevirt-ui/kubevirt-plugin/pull/1064/files#diff-e929f0c292177b77ceca3e970f211692c006260c72cf88cf3aa12e464586a67a)

Why not simply make it a more generic component that gets the body content string as a prop?

The CSS class name should be more in BEM convention and start with scoping with component name to prevent any conflicts in the future

hstastna: > Why not simply make it a more generic component that gets the body content string as a prop?

Because it was not so simple ;P... Why? Of course, I was thinking about that, but it was more like a hesitating between creating this component and then making 2 other places of the code (where the component needed) simple and nice vs. not creating this component (but more generic one) but that required using the same body content in more places than 1, which is a bit less nice (but nothing horrible, of course) but.. but then it would be right to refactor even more code places, because we have more components similar to this one across the whole code, just with some little differences. So I chose the 1st option and just followed what we already do in our code base, as this is not a refactoring PR and we don't have much time before our code freeze. So that's why :) 

Regarding the css, the same: because that's what we already do in our codebase ;) Sometimes it's better to follow one way of doing something in one project, even when it's not the best, because it provides consistency and so later easier to refactor, better than to use many different ways of implementing something in one project - but that's we unfortunately already do for some time.... Anyway I do like you strongly suggest to follow BEM, so I can change that, of course.

src/utils/components/PreferenceHelpTextIcon/PreferenceHelpTextIcon.tsx
hstastna: Of course, it can. Just IMO it didn't worth to do that for such a little component.. We could find many places across our codebase where we could do that.. and I wonder why we didn't do that from the beginning everywhere and by default..

metalice: we didn't do it from the beginning because the tsconfig file wasn't updated so we weren't able, since we update the tsconfig file and we can do it, we try to keep the import in this preferred way :)

src/utils/components/PreferenceHelpTextIcon/PreferenceHelpTextIcon.tsx
metalice: You are right that we started the project in some way that isn't the approach we should always follow. We missed some of react essential advantages, such as generic components (without abusing), and some basic coding principles, such as code reusability. IMHO, we should not add new things that still follow those practices (all with deep thought on each component and its need and complexity). So if we look at this specific component, in which there isn't any logic inside, we can easily make it generic by getting the body content as a prop. Then this component can be saved in a generally shared folder and be used in other places as well - it has no logic, just appearance, and it's perfect for sharing.  The same goes for CSS; we should follow the industry standards such as BEM :)

PR #1071, opened 2023-02-22
Bug 2169699: Add ssh select id for testing purposes

upalatucci:
## 📝 Description

Just for test purposes


hstastna (approved):
Looks good!!

PR #1076, opened 2023-02-23
Bug 2155403: Adding labels for Linux only and Windows only next to the "SSH" and "Sysprep"

DanaOrr:
## 📝 Description

> Add a brief description

## 🎥 Demo
Before:
image

image

After:
image

image


src/utils/components/Labels/LinuxLabel.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { LinuxIcon } from '@patternfly/react-icons';
+
+import './labels.scss';
+
+const LinuxLabel: FC = () => {
+  const { t } = useKubevirtTranslation();
+
+  return (
+    <Label variant="outline" className="title-label" icon={<LinuxIcon />}>
+      {t('Linux only')}
```
hstastna, line 14: I am not sure if it worth to create separate components for this. Maybe creating one compoment, maybe LabelOutline or something like that, and to get the text and the icon from props?

src/utils/components/Labels/labels.scss
```diff
@@ -0,0 +1,4 @@
+.title-label {
+    font-size: var(--pf-global--FontSize--xs);
+    font-weight: var(--pf-global--FontWeight--light);
+};
```
hstastna, line 4: Newline missing in this file ;)

src/views/catalog/wizard/components/WizardDescriptionItem/WizardDescriptionItem.tsx
```diff
@@ -74,7 +76,11 @@ export const WizardDescriptionItem: React.FC<WizardDescriptionItemProps> = React
         );
       }
 
-      return <DescriptionListTerm>{titleWithCount}</DescriptionListTerm>;
+      return (
+        <DescriptionListTerm>
+          {titleWithCount} {label}
```
hstastna, line 81: Similarly as above, I'd recommend to add an extra space between the text and the label. Maybe adding it to the label component? Up to you...

hstastna (commented):
Left a few comments, PTAL. Thanks, Dana! :)

src/utils/components/DescriptionItem/DescriptionItemHeader.tsx
```diff
@@ -62,5 +64,9 @@ export const DescriptionItemHeader: React.FC<DescriptionItemHeaderProps> = ({
     );
   }
 
-  return <DescriptionListTerm>{descriptionHeader}</DescriptionListTerm>;
+  return (
+    <DescriptionListTerm>
+      {descriptionHeader} {label}
```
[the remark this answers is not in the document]
upalatucci: i think  she should update the screenshots. Now there should be some padding left on the label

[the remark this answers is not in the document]
hstastna: Yep, adding some padding would be even better.

src/utils/components/Labels/LinuxLabel.tsx
upalatucci: As the label has unic text and icons I'm okay with that.
Having LabelOutline with the same text and icon (with its import) as props all around the repo is annoying for me

upalatucci:
lgtm

src/views/virtualmachines/details/tabs/configuration/scripts/ScriptsTab.tsx
```diff
@@ -101,6 +103,11 @@ const ScriptsTab: React.FC<VirtualMachineScriptPageProps> = ({ obj: vm }) => {
               }
               descriptionHeader={t('Authorized SSH Key')}
               isEdit={canUpdateVM}
+              label={
+                <span className="pf-u-pl-xs pf-u-ml-xs">
```
hstastna, line 107: I can see the` <span>` element around `OSLabel` component everywhere in this PR. What is that span for? Is it really needed? Can't we just add the styling we need to the OSLabel by defining the css for that component? Or if we do need that span, we could move it to the OSLabel, too. WDYT? Thanks!

src/utils/components/Labels/OSLabel.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { Label } from '@patternfly/react-core';
+
+import './os-labels.scss';
+
+type OSLabelProps = {
+  text: string;
+  icon: ReactNode;
+};
+const OSLabel: FC<OSLabelProps> = ({ text, icon }) => {
+  return (
```
hstastna, line 12: You can omit the `return` in here, if there is nothing else in the component, except the return function...

hstastna (commented):
Left a few comments, PTAL :) Thanks a lot!

src/utils/components/Labels/LinuxLabel.tsx
hstastna: Sometimes it is not so obvious where are the boundaries of what worth and what not, especially if we don't know how's gonna be in the future - that could help to make better decisions. Maybe there's gonna be many other similar use cases so we'd save some settings or css styling, maybe not. But for now yes, it probably doesn't worth.

But what to do anyway? This is IMO more like a question of an individual opinion. One or another approach isn't really bad. It will work without any issues or without any significant impact on the readability of the code or app performance. I'd simply go for the easier way, which is not creating a custom component.

But how this component was ever created? Let's understand this. In the earlier stages of this PR, there was a bit more stuff to put in the `OSLabel`. So it really seemed that it did worth to create it. But then I found how to simply add some more style for the Label component in this PR the way that we don't need extra elements to wrap the Label anymore, to achieve the expected look and especially extra space around the component. So those were removed, but the existence of the OSLabel remained. And now, we are simply in another stage, after the previous improvements, when we realize we probably don't even need OSLabel anymore, and can simply use the Label component as it is..

src/utils/components/Labels/WindowsLabel.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { Label } from '@patternfly/react-core';
+import { WindowsIcon } from '@patternfly/react-icons';
+
+import './labels.scss';
+
+const WindowsLabel: FC = () => {
+  const { t } = useTranslation();
+  return (
+    <Label variant="outline" className="os-label" icon={<WindowsIcon />}>
+      {t('Windwos only')}
```
avivtur, line 13: should be `Windows` :)

DanaOrr: thank you :)

hstastna: @avivtur are you sure it should be only "Windows"? IMHO "Windows only" is better than just "Windows". If I was a default normal user, seeing "Windows" would tell me nothing about that it is related to ONLY Windows.

avivtur: I meant she misspelled the word Windows :)

PR #1087, opened 2023-02-28
Create VM from boot source with correct SC

avivtur:
## 📝 Description

Apply the SC to the VM's DataVolumeTemplate section from the underlying PVC

## 🎥 Demo

https://user-images.githubusercontent.com/67270715/221886955-fd368a40-04e0-4820-af5f-476ac1c38a6b.mp4

### DIFF
--- a/src/views/catalog/CreateFromInstanceTypes/components/VMDetailsSection/components/VMDetailsBody/components/DetailsRightGrid.tsx
+++ b/src/views/catalog/CreateFromInstanceTypes/components/VMDetailsSection/components/VMDetailsBody/components/DetailsRightGrid.tsx
@@ -1,34 +1,20 @@
-import React, { useState } from 'react';
+import React from 'react';
 
-import { V1beta1DataSource } from '@kubevirt-ui/kubevirt-api/containerized-data-importer/models';
 import { V1alpha1PersistentVolumeClaim } from '@kubevirt-ui/kubevirt-api/kubevirt';
 import { t } from '@kubevirt-utils/hooks/useKubevirtTranslation';
-import { getPVC } from '@kubevirt-utils/resources/template/hooks/useVmTemplateSource/utils';
 import { DescriptionList } from '@patternfly/react-core';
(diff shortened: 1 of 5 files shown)


src/views/catalog/CreateFromInstanceTypes/components/VMDetailsSection/components/VMDetailsBody/components/DetailsRightGrid.tsx
```diff
@@ -12,23 +12,40 @@ import { humanizeBinaryBytes } from '../../../../../../../../utils/utils/humaniz
...
+      pvcSource?.metadata?.namespace !== pvcMetadata?.namespace
+    ) {
+      getPVC(pvcMetadata?.name, pvcMetadata?.namespace)
+        .then((pvc) => {
+          setPVCSource(pvc);
+          setStorageClassName(pvc.spec.storageClassName);
+        })
+        .catch(console.log);
+    }
+  }, [bootSource, pvcSource, setStorageClassName]);
```
metalice, line 45: why this call is in inner component instead of upper component and then u don't need to do prop drilling of 3 floors?

avivtur: not really sure why to be honest, will try improve this, thanks :)

avivtur: made some changes,
now the pvc call is from the outer component but I still had to drill down the pvc for 3 components.
I also think we might want to switch the 5 useState on CreateFromInstanceType.tsx that is related to VM configuration into one useReducer hook. wdyt?

metalice: yes, please do that, and on the way of doing it, consider better naming of state fields :)

avivtur: I think it will be in a better context to do that in a separated PR

PR #1088, opened 2023-02-28
Enable column sorting for bootable volumes list

avivtur:
## 📝 Description

Enable column sorting for bootable volumes list under instance type VM create flow

## 🎥 Demo

https://user-images.githubusercontent.com/67270715/221902137-1ff2e458-f5bb-4ca4-bbe0-1ffd6eff449e.mp4

### DIFF
--- a/src/views/catalog/CreateFromInstanceTypes/components/BootableVolumeList/BootableVolumeList.tsx
+++ b/src/views/catalog/CreateFromInstanceTypes/components/BootableVolumeList/BootableVolumeList.tsx
@@ -1,4 +1,4 @@
-import React, { FC, useState } from 'react';
+import React, { Dispatch, FC, SetStateAction, useState } from 'react';
 
 import { V1beta1DataSource } from '@kubevirt-ui/kubevirt-api/containerized-data-importer/models';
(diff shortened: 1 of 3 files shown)


src/views/catalog/CreateFromInstanceTypes/components/BootableVolumeList/BootableVolumeList.tsx
```diff
@@ -4,7 +4,8 @@ import { V1beta1DataSource } from '@kubevirt-ui/kubevirt-api/containerized-data-
 import { V1alpha2VirtualMachineClusterPreference } from '@kubevirt-ui/kubevirt-api/kubevirt';
```
metalice, line 4: why not sperate to a hook? instead of recalculate and rerender each time?

src/views/catalog/CreateFromInstanceTypes/components/BootableVolumeList/hooks/useBootVolumeSortColumns.ts
```diff
@@ -0,0 +1,55 @@
+import { useState } from 'react';
```
metalice, line 1: this hook can return the sorted data, instead of doing slice, sort in the component

PR #1090, opened 2023-03-01
Bug 2158424: nads different namespaces

metalice:
## 📝 Description

This PR is fixing a minor bug https://bugzilla.redhat.com/show_bug.cgi?id=2158424

While fixing this bug, I noticed this bug needs to be set in 3 places! I refactored all NetworkInterfaceModal
that is now shareable between wizard/vm details/templates details. Only onSubmit is passed as a prop, as it might differ for each case.

While refactoring, I found another bug in template/wizard that edit a nic isn't working, so I fixed it also. the bug is related to 
https://github.com/kubevirt-ui/kubevirt-plugin/commit/4cd451ed5e9714e305eb0286e9fd254316fad542#diff-df0e15fd36de6c7fd62d708221e2706cee5c5ae5717f8ad1087c73a1ac47d110R72
the filtering isn't making anything, and its duplicates, which is not correct. This bug is also found in the wizard, as this is a duplicate code. Duplicate code means duplicate bugs :)

I open this PR without a squash with two commits so you can see the first commit, the refactor, and the second commit, the bug fix.

## 🎥 Demo


metalice:
@avivtur @pcbailey @hstastna @upalatucci @DanaOrr  PTAL :)

upalatucci (approved):

src/views/catalog/wizard/tabs/network/components/modal/NetworkInterfaceModal.tsx
```diff
@@ -1,17 +1,12 @@
...
-
-import NameFormField from './components/NameFormField';
-import NetworkInterfaceMACAddressInput from './components/NetworkInterfaceMacAddressInput';
-import NetworkInterfaceModelSelect from './components/NetworkInterfaceModelSelect';
-import NetworkInterfaceNetworkSelect from './components/NetworkInterfaceNetworkSelect';
-import NetworkInterfaceTypeSelect from './components/NetworkInterfaceTypeSelect';
-import { interfaceModelType, interfaceTypeTypes } from './utils/constants';
-import { generateNicName, networkNameStartWithPod, podNetworkExists } from './utils/helpers';
+import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import SharedNetworkInterfaceModal from '@kubevirt-utils/components/NetworkInterfaceModal/NetworkInterfaceModal';
```
avivtur, line 5: suggestion: It's a bit confusing to me when here you call the shared component SharedNetworkInterfaceModal and in the Edit just NetworkInterfaceModal, can you just call SharedNetworkInterfaceModal to all of them? also maybe a better name than Shared... ?

avivtur (commented):
NICE, great work! :1st_place_medal: 
other than one small suggestion I don't see anything wrong with this PR

src/views/catalog/wizard/tabs/network/components/modal/NetworkInterfaceModal.tsx
metalice: I thought i would get away with it, lol. I didn't have a better naming... i will think about something :)

metalice: done - i took a different approach instead of renaming the shared one, I renamed the calling components to match its place of rendering name

avivtur (approved):

src/views/catalog/wizard/tabs/network/components/modal/NetworkInterfaceModal.tsx
avivtur: I secretly hoped that's what you'll do :) 
great minds think alike

PR #1092, opened 2023-03-01
Fix OS display name

avivtur:
## 📝 Description

Fixing OS name in the VM details section in VM create with Instance type flow

## 🎥 Demo

Before:
os-display-name-b4

After:
os-display-name


hstastna (commented):
@avivtur Looks good, I'm just not sure about the reasons of needing this change and not simply using getting the readable OS name the same way as we get that info for the _Operating system_ column, where it's already displayed correctly..

PR #1094, opened 2023-03-02
Bug 2174288: No storageClass is selected by default while adding/editing a disk

avivtur:
## 🎥 Demo

After:

https://user-images.githubusercontent.com/67270715/222400730-7428e6f5-8ece-4af8-a994-30ae91a0512d.mp4

### DIFF
--- a/src/utils/components/DiskModal/DiskFormFields/StorageClassSelect.tsx
+++ b/src/utils/components/DiskModal/DiskFormFields/StorageClassSelect.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useMemo, useState } from 'react';
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
 
 import { IoK8sApiStorageV1StorageClass } from '@kubevirt-ui/kubevirt-api/kubernetes/models';
 import Loading from '@kubevirt-utils/components/Loading/Loading';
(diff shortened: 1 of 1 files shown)


src/utils/components/DiskModal/DiskFormFields/StorageClassSelect.tsx
```diff
@@ -45,6 +45,12 @@ const StorageClassSelect: React.FC<StorageClassSelectProps> = ({
     [storageClasses],
   );
 
+  useEffect(() => {
+    if (!storageClass) {
```
upalatucci, line 49:
```suggestion
    if (!storageClass && loaded) {
```
As the continuous change of storageClasses during load can create some ui flickering probably

gouyang:
There is no default storageclass set in 'Add volume' modal, can we apply the fix to it as well?

avivtur:
> As the continuous change of storageClasses during load can create some ui flickering probably

@upalatucci changed, thanks :) 

> LGTM, just a question: shouldn't we also `setStorageClassProvisioner`? If we already set storage class name (when `storageClass` not defined). Thanks!

@hstastna added, thanks :) 

> There is no default storageclass set in 'Add volume' modal, can we apply the fix to it as well?

@gouyang since we are using a general component for storage class across the app, this fix should apply to the `Add volume` modal in the InstanceType VM create flow :)
when this bug is fixed please verify for add/edit disk modal and the add volume modal, thanks :)

PR #1098, opened 2023-03-02
Bug 2168488: add descriptions to workloads

upalatucci:
## 🎥 Demo
Screenshot from 2023-03-02 14-01-43


src/utils/components/WorkloadProfileModal/WorkloadProfileModal.tsx
```diff
@@ -6,6 +6,8 @@ import { WORKLOADS, WORKLOADS_LABELS } from '@kubevirt-utils/resources/template'
 import { K8sResourceCommon } from '@openshift-console/dynamic-plugin-sdk';
 import { Form, FormGroup, Select, SelectOption } from '@patternfly/react-core';
 
+import { WORKLOADS_DESCRIPTIONS } from './constants';
```
metalice, line 9: please consider expanding WORKLOADS_LABELS , with title and description.. i think it will be easier and in a single place. as for now u import WORKLOADS_LABELS from templates but the description is locally...

src/utils/components/WorkloadProfileModal/WorkloadProfileModal.tsx
```diff
@@ -42,7 +44,7 @@ const WorkloadProfileModal: React.FC<WorkloadProfileModalProps> = React.memo(
               selections={workload}
             >
               {Object.entries(WORKLOADS_LABELS).map(([key, value]) => (
-                <SelectOption key={key} value={key}>
+                <SelectOption key={key} value={key} description={t(WORKLOADS_DESCRIPTIONS[key])}>
```
avivtur, line 47: translating a const, does not work well with i18n,
please import t as a function to the `constants.ts` file and translate the keys directly instead of translating on components, then you can also remove the translation comments above each key

upalatucci: okay thanks for the info

src/utils/components/WorkloadProfileModal/WorkloadProfileModal.tsx
upalatucci: fixed. I've added a new object below `WORKLOADS_LABELS`

__mocks__/react-i18next.ts
```diff
@@ -1,4 +1,7 @@
 module.exports = {
   useTranslation: () => ({ t: (key) => key }),
   Trans: ({ children }) => children,
+  getI18n: () => ({
```
metalice, line 4: why we need it?

upalatucci: For tests we mock react-i18next and they fails because the script does not find that function.
Tests started failing when i imported the `t` from the utils in the constants file

PR #1102, opened 2023-03-03
Bug 2174619: render any device as bootable

upalatucci:
## 📝 Description

Guys, is it right that the modal should show all the possible devices? Even if they don't have a bootOrder?
Are all devices in theory bootable?


upalatucci:
@metalice can you please fix the labels?

PR #1105, opened 2023-03-03
Bug 2175274: Display Edit CPU | Memory modal in VM Details

hstastna:
## 📝 Description

**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2175274

Fix the error, prevent blank page to be displayed, after trying to edit CPU/Memory of a VM, created from a template without `spec.template.spec.domain.cpu` defined in its VM object (see the YAML of a VM or a Template).

Fix the error by making sure the path exists (using `ensurePath`), when editing the VM or Template object (the same problem occurs in more places).

## 🎥 Demo
**Before:**
cpu_before

**After:**
cpu_after


hstastna:

PR #1109, opened 2023-03-06
Bug 2170703: Fix template catalog search

metalice:
## 📝 Description

Patternfly probably has changed and updated since this search input has created.
I updated patternfly again to fix a bug in typescript they had.
Because of the update had to change in migration policy files to match the new patternfly typescript and components change.

## 🎥 Demo

After:
search

Before:
search-before


avivtur (approved):

avivtur (approved):
great job  :trophy:

PR #1113, opened 2023-03-06
Bug 2172371: Disable "Restore template settings" for VM without template

hstastna:
## 📝 Description

**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2172371

Disable _Restore template settings_ button in _Edit CPU | Memory_ modal in VM _Details_ tab, because it does not make sense if the VM has no base template specified (check _Template_ field in the same page in the UI).

## 🎥 Screenshots
**Before:**
_Restore template settings_ button was loaded and is now clickable even it makes no sense:
none_before

**After:**
_Restore template settings_ disabled for the VM without base template:
none_after

### DIFF
--- a/src/utils/components/CPUMemoryModal/CpuMemoryModal.tsx
+++ b/src/utils/components/CPUMemoryModal/CpuMemoryModal.tsx
@@ -3,6 +3,8 @@ import produce from 'immer';
 
 import { V1VirtualMachine, V1VirtualMachineInstance } from '@kubevirt-ui/kubevirt-api/kubevirt';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
+import { getLabel } from '@kubevirt-utils/resources/shared';
+import { VM_TEMPLATE_ANNOTATION } from '@kubevirt-utils/resources/vm';
 import { toIECUnit } from '@kubevirt-utils/utils/units';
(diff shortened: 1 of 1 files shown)


hstastna:

src/utils/components/CPUMemoryModal/CpuMemoryModal.tsx
```diff
@@ -115,12 +119,13 @@ const CPUMemoryModal: React.FC<CPUMemoryModalProps> = ({ vm, isOpen, onClose, on
...
           variant={ButtonVariant.secondary}
           isDisabled={
+            !templateName ||
             !defaultsLoaded ||
             !templateDefaultsData?.defaultCpu ||
             !templateDefaultsData?.defaultMemory ||
             defaultLoadError
           }
-          isLoading={!defaultsLoaded}
+          isLoading={templateName && !defaultsLoaded}
```
avivtur, line 128: why do we need to add templateName to isLoading?

avivtur (commented):
Thank you for this PR :)
just one question as I didn't understand something

src/utils/components/CPUMemoryModal/CpuMemoryModal.tsx
hstastna: Because why do we need to show the button loading if the button will be disabled anyway? It would be just misleading to the user.

PR #1116, opened 2023-03-07
Add new SSH section

pcbailey:
## 📝 Description

This PR:

- Replaces the old accordion-style SSH section with a simplified version of the new SSH section design
- Adds the 'SSH key name' field to the VM details section
- Left aligns value fields in the VM details section
- Updates the titles in the Select InstanceType section

### None selected
Selection_223

### Use existing selected
Selection_224

### Add new selected
Selection_225

### Add new selected with non-unique name error
Selection_226

### Updated VM details section
Selection_228


src/utils/components/SSHSecretSection/utils/components/SSHKeyUpload/SSHKeyUpload.tsx
```diff
@@ -0,0 +1,82 @@
...
+  return (
+    <Form isHorizontal>
+      <FileUpload
+        id={'ssh-key-upload'}
+        className="ssh-secret-selection__file-upload"
+        type="text"
+        value={sshSecretKey}
+        onChange={(v: string) => {
+          setIsValidKey(validateSSHPublicKey(v));
+          // setNewSecretKey(v?.trim());
```
avivtur, line 45: intentional?

pcbailey: Nope. Fixed.

upalatucci (approved):
so great this design

pcbailey:
> so great this design

It's so much better than the existing design.

src/utils/components/SSHSecretSection/SSHSecretSection.tsx
```diff
@@ -0,0 +1,78 @@
...
+  return (
+    <Grid className="ssh-secret-section">
+      <GridItem span={12}>
+        <SecretSelectionRadioGroup
+          selectedOption={secretSelectionOption}
+          setSelectedOption={setSecretSelectionOption}
+        />
+      </GridItem>
+      <GridItem span={12} className="ssh-secret-section__body">
+        {secretSelectionOption === SecretSelectionOption.none && null}
```
metalice, line 58: && null ?

src/utils/components/SSHSecretSection/utils/components/SelectSecret.tsx
```diff
@@ -0,0 +1,86 @@
+import * as React from 'react';
```
metalice, line 1: Can remove *

src/utils/components/SSHSecretSection/utils/components/SelectSecret.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  const [secrets, secretsLoaded, secretsError] = secretsResourceData;
+  const sshKeySecrets = secrets
+    ? secrets?.filter((secret) => validateSSHPublicKey(decodeSecret(secret)))
+    : [];
+
+  const filterSecrets = (_, value: string): React.ReactElement[] => {
+    let filteredSecrets = sshKeySecrets;
+
+    if (value) {
```
metalice, line 37: What happen if you remove this if statement and set default value to value of "" ?

src/utils/components/SSHSecretSection/utils/components/SelectSecret.tsx
```diff
@@ -0,0 +1,86 @@
...
+      onSelect={onSelect}
+      variant={SelectVariant.single}
+      onFilter={filterSecrets}
+      hasInlineFilter
+      selections={selectedSecretName}
+      placeholderText={t('--- Select secret ---')}
+      maxHeight={400}
+      id={id || 'select-secret'}
+    >
+      {sshKeySecrets?.map((secret) => (
```
metalice, line 77: Not using filteredSecrets here?

src/utils/components/SSHSecretSection/SSHSecretSection.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+const SSHSecretSection: FC<SSHSecretSectionProps> = ({ sshSecretDetails, setSSHSecretDetails }) => {
+  const [activeNamespace] = useActiveNamespace();
+  const [secretSelectionOption, setSecretSelectionOption] = useState<SecretSelectionOption>(
+    SecretSelectionOption.none,
+  );
+
+  const secretsResourceData = useK8sWatchResource<IoK8sApiCoreV1Secret[]>(
+    activeNamespace
+      ? {
```
metalice, line 31: Does && works as well it it must be null?

metalice:
Nice work!! Any plans for this new design to replace all other places where we have ssh ?

src/utils/components/SSHSecretSection/utils/components/SelectSecret.tsx
pcbailey: What default value?

Ugo wrote this. I just copied it over to this location since the other SSH code will eventually go away. That allowed me to do a little cleanup. I'm not sure what this regex is supposed to be doing, though. @upalatucci Can you explain what this is doing?

Additionally, this is why I think we should avoid using `value` as a parameter name when a more descriptive name can be used. I'm guilty of doing it myself, but we should start holding each other accountable for it, IMO.

src/utils/components/SSHSecretSection/utils/components/SelectSecret.tsx
pcbailey: @upalatucci Why do we need both the filter function and the map here?

src/utils/components/SSHSecretSection/SSHSecretSection.tsx
pcbailey: Yes. Nothing should be displayed here when `None` is selected.

src/utils/components/SSHSecretSection/utils/components/SelectSecret.tsx
pcbailey: Done!

src/utils/components/SSHSecretSection/SSHSecretSection.tsx
pcbailey: Done!

pcbailey:
@metalice Thanks! This is the simplified version of a design that Tal Tobias is working on that as far as I'm aware will replace all of the existing SSH components in 4.14.

src/utils/components/SSHSecretSection/utils/components/SelectSecret.tsx
avivtur: the unfiltered data should be as array of `children`, while the `onFilter` prop is getting a function that returns array of filtered data as ReactElement which eventually calculated as the filtered options of the Select.
bottom line, this should be ok

pcbailey: @avivtur Ah, that makes sense. It was late when I responded and I forgot that this was a filterable select component. 🤦🏼‍♂️ Thanks for clearing it up!

metalice:
> @metalice Thanks! This is the simplified version of a design that Tal Tobias is working on that as far as I'm aware will replace all of the existing SSH components in 4.14.

Good to know, so in this case. i would suggest make those components global shared as it be used in other places as well

src/utils/components/SSHSecretSection/SSHSecretSection.tsx
metalice: so it can be removed

src/utils/components/SSHSecretSection/utils/components/SelectSecret.tsx
upalatucci: The `value` here is just what the user type in the searchbox when the dropdown is open. If the value is undefined or empty, the user should see all the secrets. If the user type something, should see the options filtered.

Sorry for this sad naming. You are right about that.

src/utils/components/SSHSecretSection/utils/components/SelectSecret.tsx
upalatucci: Yes guys, just a little bit confusing but this is how `Select` works

src/utils/components/SSHSecretSection/utils/components/SelectSecret.tsx
```diff
@@ -0,0 +1,86 @@
...
+      const regex = new RegExp(value, 'i');
+      filteredSecrets = sshKeySecrets.filter((secret) => regex.test(secret?.metadata?.name));
+    }
+
+    return filteredSecrets.map((secret) => (
+      <SelectOption key={secret?.metadata?.name} value={secret?.metadata?.name}>
+        <ResourceLink kind={SecretModel.kind} name={secret?.metadata?.name} linkTo={false} />
+      </SelectOption>
+    )) as ReactElement[];
+  };
```
metalice, line 47:
```suggestion
  const filterSecrets = (_, value = '': string): ReactElement[] => {
    const regex = new RegExp(value, 'i');
    filteredSecrets = sshKeySecrets.filter((secret) => regex.test(secret?.metadata?.name));

   return filteredSecrets.map((secret) => (
     <SelectOption key={secret?.metadata?.name} value={secret?.metadata?.name}>
       <ResourceLink kind={SecretModel.kind} name={secret?.metadata?.name} linkTo={false} />
     </SelectOption>
   )) as ReactElement[];
 };
```

What I mean is to check if something like this should work.
Also, I would have split this function, as this function eventually returns elements and is not related to filterSecret meaning. and the actually filter function can be in a util file

src/utils/components/SSHSecretSection/utils/components/SelectSecret.tsx
pcbailey: Thanks, Ugo! That makes sense. I'll update the name. No need to apologize. We're all guilty of doing it from time to time.

src/utils/components/SSHSecretSection/utils/components/SelectSecret.tsx
pcbailey: Ugo added the if statement to avoid iterating over the array if there was no input, but you're correct that it isn't necessary since the filter function isn't called unless there's input. I removed it.

I don't believe I can move it to a util file because it needs `sshKeySecrets` and as far as I can tell there's no way to pass the list of options to the function as a parameter.

src/utils/components/SSHSecretSection/utils/components/SelectSecret.tsx
metalice: When looking on docs i can see onTypeaheadInputChanged prop on Select will give the input search value.
Then you dont need the filterChanged prop and just render instead of sshKeySecrets?.map just the filtered version. onTypeaheadInputChanged should active the filter. it will make a much more reasonable flow.

metalice: image

metalice: and will allow us to sperate between logic and elements in a single function

pcbailey: If I understand correctly, you're saying to update the values passed as the children of the `Select` component by filtering on the input parameter inside `onTypeaheadInputChanged`. 

From my testing, the typeahead functionality doesn't work here. I found that if I removed the `ResourceLink` child in the `SelectOption` components it worked as expected, but I haven't been able to figure out a workaround to make it work while keeping the `ResourceLink` component. If you have any ideas I'd be happy to try them.

metalice: Cool, is make a working version :)

src/utils/components/SSHSecretSection/utils/components/SelectSecret.tsx
```diff
@@ -0,0 +1,94 @@
...
+        const secretName = getName(secret);
+        return (
+          <SelectOption key={secretName} value={secretName}>
+            <ResourceLink
+              groupVersionKind={getGroupVersionKindForModel(SecretModel)}
+              name={secretName}
+            />
+          </SelectOption>
+        );
+      })}
```
avivtur, line 89: this block and lines 42-49 look very similar, does it make sense to create a function for this? or a component?

pcbailey: That's a good suggestion. Done!

PR #1118, opened 2023-03-08
Bug 2175601: Nads loading - non admin user

metalice:
## 📝 Description

When a non-admin user tries to fetch nads, it fails. When failing the loading variable received from hook is false, so added a check that is load error is existing  the loading will be true

### DIFF
--- a/src/utils/components/NetworkInterfaceModal/components/hooks/useNadsData.ts
+++ b/src/utils/components/NetworkInterfaceModal/components/hooks/useNadsData.ts
@@ -1,6 +1,11 @@
 import { useMemo } from 'react';
 
 import { NetworkAttachmentDefinitionModelGroupVersionKind } from '@kubevirt-ui/kubevirt-api/console';
+import {
+  DEFAULT_NAMESPACE,
+  OPENSHIFT_MULTUS_NS,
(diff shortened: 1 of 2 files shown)


avivtur (approved):

src/utils/components/NetworkInterfaceModal/components/hooks/useNadsData.ts
```diff
@@ -18,11 +18,13 @@ const useNadsData: UseNadsData = (namespace) => {
       namespace: namespace,
     },
     //global ns to get usable nads
-    default: {
-      groupVersionKind: NetworkAttachmentDefinitionModelGroupVersionKind,
-      isList: true,
-      namespace: 'default',
-    },
+    ...(namespace !== 'default' && {
```
avivtur, line 21: can you please use DEFAULT_NAMESPACE const from `@kubevirt-utils/constants/constants`?
does it also make sense to check if this namespace is either `openshift-sriov-network-operator` or `openshift-multus`?
and can you also make consts for other ns used here?

metalice: its not 1 of them, they are both global and need to be checked.
made them all consts

PR #1132, opened 2023-03-10
Bug 2158079: Align Storage expandable section with HelpIcon

hstastna:
## 📝 Description

**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2158079

Align _Storage_ expandable section with its `HelpIcon` "?", in _Customize Template parameters_ when creating a VM.

## 🎥 Screenshots
**Before:**
Storage and "?" not aligned nicely, "?" too far from Storage:
align_before

**After:**
Storage and "?" aligned properly:
align_after

### DIFF
--- a/src/views/catalog/customize/components/CustomizeSource/ExpandableCustomizeSourceSection.scss
+++ b/src/views/catalog/customize/components/CustomizeSource/ExpandableCustomizeSourceSection.scss
@@ -1,0 +1,9 @@
+.expandable-customize-source-section__stack-item-storage {
+  .pf-l-flex {
+    --pf-l-flex--AlignItems: center;
+  }
+
+  .pf-c-expandable-section__toggle {
+    padding-right: var(--pf-c-expandable-section__toggle--PaddingTop);
+  }
+}
--- a/src/views/catalog/customize/components/CustomizeSource/ExpandableCustomizeSourceSection.tsx
+++ b/src/views/catalog/customize/components/CustomizeSource/ExpandableCustomizeSourceSection.tsx
@@ -1,21 +1,21 @@
-import * as React from 'react';
+import React, { FC, useState } from 'react';
 
+import HelpTextIcon from '@kubevirt-utils/components/HelpTextIcon/HelpTextIcon';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
 import {
   ExpandableSection,
   ExpandableSectionToggle,
   Flex,
   FlexItem,
-  Popover,
-  PopoverPosition,
   Stack,
   StackItem,
 } from '@patternfly/react-core';
-import { HelpIcon } from '@patternfly/react-icons';
 
 import { CustomizeSource, CustomizeSourceProps } from './CustomizeSource';
 
-export const ExpandableCustomizeSourceSection: React.FC<CustomizeSourceProps> = ({
+import './ExpandableCustomizeSourceSection.scss';
+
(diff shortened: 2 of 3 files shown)


hstastna:

src/views/catalog/customize/components/CustomizeSource/ExpandableCustomizeSourceSection.scss
```diff
@@ -0,0 +1,7 @@
+.pf-l-flex {
```
avivtur, line 1: Hi :) 
This file is a bit problematic when accessing the PF classes and changing them directly, this will affect the entire plugin and could cause unexpected issues.
Can you try and create new classes instead of changing existing PF classes to achieve the same styling?

hstastna: I can wrap this by some unique custom class name, but the rest needs to remain the same, otherwise we won't achieve the expected result 🤷🏼‍♀️ We do need to override the PF class for this component to get the look and the behavior we need. Another option would be not to use the PF component.

avivtur: wrapping these changes inside a custom classname sound like a great solution :) 
this way we still change what we need but don't affect any other places :100: :1st_place_medal:

hstastna: Well, I have an experience that it may not always work like that, but let's try it out.

src/views/catalog/customize/components/CustomizeSource/ExpandableCustomizeSourceSection.scss
```diff
@@ -0,0 +1,11 @@
...
+  .pf-l-flex {
+    --pf-l-flex--AlignItems: center;
+  }
+}
+
+.expandable-customize-source-section__expandable-section-storage {
+  .pf-c-expandable-section__toggle {
+    padding-right: var(--pf-c-expandable-section__toggle--PaddingTop);
+  }
+}
```
avivtur, line 11: this looks good and definitely will work
another suggestion to improve, so we could remove the className from the ExpandableSectionToggle component:
```
.expandable-customize-source-section {
  .pf-l-flex {
    --pf-l-flex--AlignItems: center;
  }

  .pf-c-expandable-section__toggle {
    padding-right: var(--pf-c-expandable-section__toggle--PaddingTop);
  }
}
```

src/views/catalog/customize/components/CustomizeSource/ExpandableCustomizeSourceSection.scss
```diff
@@ -0,0 +1,9 @@
+.expandable-customize-source-section__stack-item-storage {
+  .pf-l-flex {
+    --pf-l-flex--AlignItems: center;
```
metalice, line 3: if you change it to baseline instead of center does it look better?

src/views/catalog/customize/components/CustomizeSource/ExpandableCustomizeSourceSection.tsx
```diff
@@ -1,20 +1,20 @@
 import * as React from 'react';
```
metalice, line 1: please remove *

src/views/catalog/customize/components/CustomizeSource/ExpandableCustomizeSourceSection.scss
hstastna: I think baseline was default..

PR #1133, opened 2023-03-10
Bug 2176725: Carry over "Start this VirtualMachine after creation" value

hstastna:
## 📝 Description

**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2176725

Carry over "Start this VirtualMachine after creation" checkbox value to the last step of creating VM when customizing VM.

Notes:
For achieving the expected behavior `useWizardVMContext` hook was used.
Also note that changes in _src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx_ look a bit wild, but that's just because indentation was updated a bit, due to importing `memo` from `'react'`, and only [this](https://github.com/kubevirt-ui/kubevirt-plugin/compare/main...hstastna:BZ_Start_this_VirtualMachine_after_creation_not_carried_over?expand=1#diff-fcad411937b404f59f942a70de7a8de426f001e371ff8bdaef5151d8409569baR48) and  [this piece of code](https://github.com/kubevirt-ui/kubevirt-plugin/compare/main...hstastna:BZ_Start_this_VirtualMachine_after_creation_not_carried_over?expand=1#diff-fcad411937b404f59f942a70de7a8de426f001e371ff8bdaef5151d8409569baR99-R103) was added.

## 🎥 Screenshots
Unchecking the checkbox, when creating a VM from the chosen template in Catalog:
unchecking

**Before:**
The checkbox is checked no matter we un-checked it in the catalog drawer earlier:
ch_before

**After:**
The checkbox is unchecked as expected :
ch_after

### DIFF
--- a/src/views/catalog/customize/utils.ts
+++ b/src/views/catalog/customize/utils.ts
@@ -40,7 +40,7 @@ export const overrideVirtualMachineDataVolumeSpec = (
 };
 
 export const setTemplateParameters = (template: V1Template, formData: FormData): V1Template => {
-  template.parameters = template.parameters.map((parameter) => {
+  template.parameters = template?.parameters?.map((parameter) => {
     const formParameter = formData.get(parameter.name) as string;
     if (formParameter.length > 0) parameter.value = formParameter;
 
--- a/src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
+++ b/src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { FC, memo, MouseEvent, useState } from 'react';
 import { useHistory } from 'react-router-dom';
 import produce from 'immer';
 
@@ -7,6 +7,7 @@ import {
   replaceTemplateParameterValue,
 } from '@catalog/customize/utils';
 import { quickCreateVM } from '@catalog/utils/quick-create-vm';
+import { useWizardVMContext } from '@catalog/utils/WizardVMContext';
 import { ProcessedTemplatesModel, V1Template } from '@kubevirt-ui/kubevirt-api/console';
 import VirtualMachineModel from '@kubevirt-ui/kubevirt-api/console/models/VirtualMachineModel';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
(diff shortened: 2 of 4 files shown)


hstastna:

src/views/catalog/CreateFromInstanceTypes/components/CreateVMFooter/CreateVMFooter.tsx
```diff
@@ -40,8 +41,10 @@ const CreateVMFooter: FC<CreateVMFooterProps> = ({
   sshSecretCredentials,
 }) => {
   const { t } = useKubevirtTranslation();
+  const { tabsData } = useWizardVMContext();
+
   const history = useHistory();
-  const [startVM, setStartVM] = useState(true);
+  const [startVM, setStartVM] = useState(tabsData?.startVM);
```
avivtur, line 47: the instance types flow and the templates flow is not the same and you should not use the Wizard context here as it's not in the wizard :) 
please revert changes from this file

src/views/catalog/customize/utils.ts
```diff
@@ -40,7 +40,7 @@ export const overrideVirtualMachineDataVolumeSpec = (
 };
 
 export const setTemplateParameters = (template: V1Template, formData: FormData): V1Template => {
-  template.parameters = template.parameters.map((parameter) => {
+  template.parameters = template?.parameters?.map((parameter) => {
```
avivtur, line 43: +1

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
```diff
@@ -40,161 +41,167 @@ type TemplatesCatalogDrawerCreateFormProps = {
...
-                    label={t('Start this VirtualMachine after creation')}
-                  />
-                </StackItem>
-              </>
-            ) : (
+    useEffect(() => {
+      updateTabsData((tabsDataDraft) => {
+        tabsDataDraft.startVM = startVM;
+      });
+    }, [startVM, updateTabsData]);
```
avivtur, line 103: I would take this updateTabsData method and use it in the Checkbox component in the onChange prop something like this:
```
onChange={(checked) => {
  setStartVM(checked);
  updateTabsData((currentTabsData) => {
    return { ...currentTabsData, startVM: checked };
  });
}}
```

src/views/catalog/wizard/components/WizardFooter.tsx
```diff
@@ -59,12 +64,18 @@ export const WizardFooter: React.FC<{ namespace: string }> = ({ namespace }) =>
   const templateNamespace = tabsData?.overview?.templateMetadata?.namespace;
   const loaded = vmContextLoaded && vmCreateLoaded && bootSourceLoaded;
 
-  React.useEffect(() => {
-    if (loaded && startVM !== isBootSourceAvailable) {
+  useEffect(() => {
+    updateTabsData((tabsDataDraft) => {
+      tabsDataDraft.startVM = startVM;
+    });
+  }, [startVM, updateTabsData]);
```
avivtur, line 71: same here for this useEffect, would using the update method in the checkbox onChange prop
```
onChange={(checked) => {
  setStartVM(checked);
  updateTabsData((currentTabsData) => {
    return { ...currentTabsData, startVM: checked };
  });
}}
```

src/views/catalog/wizard/components/WizardFooter.tsx
```diff
@@ -59,12 +64,18 @@ export const WizardFooter: React.FC<{ namespace: string }> = ({ namespace }) =>
...
+
+  useEffect(() => {
+    if (loaded && startVM && !isBootSourceAvailable) {
+      // change startVM to false only if boot source is not available
       setStartVM(isBootSourceAvailable);
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [loaded, isBootSourceAvailable]);
+  }, [loaded, isBootSourceAvailable, startVM]);
```
avivtur, line 78: pretty sure this useEffect is not needed anymore and can be removed

avivtur (commented):
Hi Hilda :) 
Nice work, left some comments

src/views/catalog/customize/utils.ts
hstastna: Yep, this change is here to prevent the UI to do 💣  unnecessary

src/views/catalog/wizard/components/WizardFooter.tsx
hstastna: What is the reason of being so sure about removing `useEffect` in here?

hstastna: I'd leave `useEffect` in there, maybe also would better add `useCallback` to `onChangeStartVM`. WDYT?

src/views/catalog/CreateFromInstanceTypes/components/CreateVMFooter/CreateVMFooter.tsx
```diff
@@ -3,6 +3,7 @@ import { useHistory } from 'react-router-dom';
 import produce from 'immer';
 
 import { SSHSecretCredentials } from '@catalog/CreateFromInstanceTypes/components/VMDetailsSection/components/SSHKeySection/utils/types';
+import { useWizardVMContext } from '@catalog/utils/WizardVMContext';
```
avivtur: please don't forget to remove this import as well :)

src/views/catalog/wizard/components/WizardFooter.tsx
```diff
@@ -21,15 +21,20 @@ import { clearSessionStorageVM, useWizardVMContext } from '../../utils/WizardVMC
...
+    updateTabsData,
+    disableVmCreate,
+    loaded: vmContextLoaded,
+  } = useWizardVMContext();
   const { isBootSourceAvailable, loaded: bootSourceLoaded } = useWizardSourceAvailable();
   const { createVM, error, loaded: vmCreateLoaded } = useWizardVmCreate();
   const { createModal } = useModal();
 
-  const [startVM, setStartVM] = React.useState(true);
+  const [startVM, setStartVM] = useState<boolean>(tabsData?.startVM || true);
```
avivtur: tabsData?.startVM || true will evaluate to true all the time:
tabsData?.startVM === true => true || true => true
tabsData?.startVM === false => false || true => true
you can remove the || true part as it will always should have the value carried from the drawer

src/views/catalog/wizard/components/WizardFooter.tsx
avivtur: let's focus on the isBootSourceAvailable and startVM.
isBootSourceAvailable - is set to true if the boot source is available and we can see both the start vm checkbox and the quick create vm.
startVM - is now getting the carried value from the checkbox at the drawer, this checkbox exists only when we can see the quick create button - when the boot source is available, which means if the boot source is not available, this checkbox has a falsy value which will give you the exact same thing as the effect is doing, which causing to un-needed renders.

src/views/catalog/wizard/components/WizardFooter.tsx
```diff
@@ -21,15 +21,20 @@ import { clearSessionStorageVM, useWizardVMContext } from '../../utils/WizardVMC
...
+    updateTabsData,
+    disableVmCreate,
+    loaded: vmContextLoaded,
+  } = useWizardVMContext();
   const { isBootSourceAvailable, loaded: bootSourceLoaded } = useWizardSourceAvailable();
   const { createVM, error, loaded: vmCreateLoaded } = useWizardVmCreate();
   const { createModal } = useModal();
 
-  const [startVM, setStartVM] = React.useState(true);
+  const [startVM, setStartVM] = useState<boolean>(tabsData?.startVM ?? true);
```
avivtur, line 37: `tabsData?.startVM ?? true` will evaluate to true all the time:
tabsData?.startVM === true => true ?? true => true
tabsData?.startVM === false => false ?? true => true
you can remove the || true part as it will always should have the value carried from the drawer

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
```diff
@@ -40,161 +41,168 @@ type TemplatesCatalogDrawerCreateFormProps = {
...
+        .catch((err) => {
+          setIsQuickCreating(false);
+          setQuickCreateError(err);
+        });
+    };
 
-      const onCustomize = (e) => {
-        e.preventDefault();
-        let catalogUrl = `templatescatalog/customize?name=${template.metadata.name}&namespace=${template.metadata.namespace}&defaultSourceExists=${canQuickCreate}`;
+    const onCustomize = (e: { preventDefault: () => void }) => {
```
metalice, line 88: the right type is event, there is many events, u just need to find the right one, most of the time typescript will let u know when you highlight where you are using it

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
```diff
@@ -40,161 +41,168 @@ type TemplatesCatalogDrawerCreateFormProps = {
...
-                    onClick={onCustomize}
-                    isDisabled={!processedTemplateAccessReview}
+                    data-test-id="quick-create-vm-btn"
+                    type="submit"
+                    form="quick-create-form"
+                    isLoading={isQuickCreating || modelsLoading}
+                    isDisabled={
+                      !isBootSourceAvailable || isQuickCreating || !vmName || isEmpty(models)
+                    }
+                    onClick={(e) => {
```
metalice, line 179: u can add the event type here also if u would like

src/views/catalog/wizard/components/WizardFooter.tsx
hstastna: > tabsData?.startVM === false => false ?? true => true

Nope. Did you test this PR or just assuming? Because it returns `false` as expected. I did test this.

Note that we do need this "or" part (and using ?? is the best, checks for null and undefined, not for boolean 'false' value), because we need some default value, if `tabsData.startVM` will be undefined, as it can be, because you've suggested me to move `updateTabsData` to `onChange` (and so I did it), but this has consequences that if there's no any change made in the 1st step of creating a VM (catalog drawer), `tabsData.startVM` will be `undefined` for sure, and that's we don't want, it would break the UI. And so according to the original code, `true` was the default value, so I've used it in case that  `tabsData.startVM` is `undefined`.

src/views/catalog/wizard/components/WizardFooter.tsx
hstastna: I think I am not getting it. From the code of _src/views/catalog/wizard/components/WizardFooter.tsx_, I can see the checkbox can be displayed no matter `isBootSourceAvailable` in this file. And according to the code above (the original useEffect block or the edited one), `isBootSourceAvailable` can be false. I still think we need to update `startVM` value to the correct one, if that changes to `true` while `isBootSourceAvailable` is false.
The best solution I can see for this is to completely remove this `useEffect` block while setting the initial value of `startVM` according to `isBootSourceAvailable` value, and also updating `isDisabled` param of the `Checkbox` accodfing to `isBootSourceAvailable`, too. WDYT?

hstastna: I wouldn't rely on that condition `canQuickCreate` in [TemplatesCatalogDrawerCreateForm-L.101](https://github.com/kubevirt-ui/kubevirt-plugin/blob/893005c63fd46134da92cb176991b00319673c75/src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx#L101). Because if there will be some changes and the component will be moved, it stops working as expected.

avivtur: we shouldn't rely on things that may happen in the future or may not.
bottom line this useEffect is not needed and costs performance and component complexity

hstastna:

PR #1134, opened 2023-03-10
Add sabrinajess to reviewers

tnisan:


tnisan:
Added sabrinajess to the organization, if this will not do we can add her later as a reviewer as well

hstastna:
@tnisan Seems that it didn't work, as the PR was closed. IMHO the reason it didn't work is here:
https://github.com/kubevirt-ui/kubevirt-plugin/pull/1134#issuecomment-1464430012
```
sabrinajess

    User is not a member of the org. User is not a collaborator. Satisfy at least one of these conditions to make the user trusted.
```
So maybe we need to make her a collaborator or she needs to be added to the org first, before she will be added to reviewers.

hstastna:
Tal, if she was added to the org already, it should be possible to add her to reviewers as well.

PR #1137, opened 2023-03-13
Bug 2174334: VM's disk is not deleted along with the VM if the VM is created from upload image

avivtur:
## 📝 Description

DataVolumes are now garbage collected, so when trying to update ownerReference for the upload's DV, it will not find it.
updating owner reference on correlated PVC, so the PVC will be GC when deleting VM.
Also fixed an issue where the volume name and the PVC resource name didn't match which caused the VM to fail on start

## 🎥 Demo

Before:

https://user-images.githubusercontent.com/67270715/224706010-c431cc4a-5988-464d-b926-110028c347e5.mp4

https://user-images.githubusercontent.com/67270715/224706017-ff8d7f0f-ac84-4402-a473-feee5d709b48.mp4

After:

https://user-images.githubusercontent.com/67270715/224706060-eecad2db-a54b-473c-af4f-471e1ce60b83.mp4

https://user-images.githubusercontent.com/67270715/224706065-a9cb4f2f-ae54-442c-801e-7103d111c87e.mp4

### DIFF
--- a/src/utils/components/DiskModal/utils/helpers.ts
+++ b/src/utils/components/DiskModal/utils/helpers.ts
@@ -131,7 +131,10 @@ export const getDataVolumeFromState = ({
   createOwnerReference?: boolean;
 }): V1beta1DataVolume => {
   const dataVolume = getEmptyVMDataVolumeResource(vm, createOwnerReference);
-  const dvName = resultVolume?.dataVolume?.name || `${vm?.metadata?.name}-${diskState.diskName}`;
+  const dvName =
+    resultVolume?.dataVolume?.name ||
+    resultVolume?.persistentVolumeClaim?.claimName ||
+    `${vm?.metadata?.name}-${diskState.diskName}`;
 
   dataVolume.metadata.name = dvName;
   dataVolume.spec.storage.resources.requests.storage = diskState.diskSize;
@@ -174,7 +177,6 @@ export const getDataVolumeFromState = ({
     dataVolume.spec.source = {
       upload: {},
     };
-    dataVolume.metadata.name = `uploaded-${diskState.diskName}`;
   }
   return dataVolume;
 };
--- a/src/utils/hooks/useCDIUpload/utils.ts
+++ b/src/utils/hooks/useCDIUpload/utils.ts
@@ -4,13 +4,15 @@ import produce from 'immer';
 import DataVolumeModel from '@kubevirt-ui/kubevirt-api/console/models/DataVolumeModel';
 import { V1beta1DataVolume } from '@kubevirt-ui/kubevirt-api/containerized-data-importer/models';
 import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
-import { UploadTokenRequestModel } from '@kubevirt-utils/models';
+import { PersistentVolumeClaimModel, UploadTokenRequestModel } from '@kubevirt-utils/models';
 import { buildOwnerReference, getAPIVersionForModel } from '@kubevirt-utils/resources/shared';
-import { getDataVolume } from '@kubevirt-utils/resources/template/hooks/useVmTemplateSource/utils';
+import {
+  getDataVolume,
+  getPVC,
+} from '@kubevirt-utils/resources/template/hooks/useVmTemplateSource/utils';
 import {
   k8sCreate,
   k8sDelete,
-  k8sGet,
   k8sPatch,
   K8sResourceCommon,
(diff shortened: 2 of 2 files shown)


src/utils/components/DiskModal/utils/helpers.ts
```diff
@@ -174,7 +178,7 @@ export const getDataVolumeFromState = ({
     dataVolume.spec.source = {
       upload: {},
     };
-    dataVolume.metadata.name = `uploaded-${diskState.diskName}`;
+    // dataVolume.metadata.name = `uploaded-${diskState.diskName}`;
```
metalice, line 181: ?

metalice:
please note that DV creating can fail; for instance, register URL isn't correct and the import failed, DV will still be alive and not GC. need to try deleting DV exists, and then PVC.

avivtur:
> please note that DV creating can fail; for instance, register URL isn't correct and the import failed, DV will still be alive and not GC. need to try deleting DV exists, and then PVC.

that's a valid point, when deleting a VM we search for all DV/PVCs/Snapshot linked by ownerReference to the VM,
even if DV wasn't GC, the ownerRef GC should catch and delete the linked resources

src/utils/components/DiskModal/utils/helpers.ts
avivtur: removed, thanks :)

src/utils/components/DiskModal/utils/helpers.ts
```diff
@@ -130,8 +130,12 @@ export const getDataVolumeFromState = ({
   resultVolume?: V1Volume;
   createOwnerReference?: boolean;
 }): V1beta1DataVolume => {
+  debugger;
```
metalice, line 133: nice

metalice: i would love to get stopped every time :) lol

src/utils/hooks/useCDIUpload/utils.ts
```diff
@@ -150,20 +153,23 @@ export const addUploadDataVolumeOwnerReference = (
   dataVolume: V1beta1DataVolume,
 ) => {
   //checking DV exisit before trying to patch it
```
metalice, line 155: not relevant any more

src/utils/components/DiskModal/utils/helpers.ts
avivtur: you would get stopped only if you open the dev tools, so we can say it's a feature lol

avivtur: removed thanks :)

src/utils/hooks/useCDIUpload/utils.ts
avivtur: changed a bit to clean up :)

src/utils/components/DiskModal/utils/helpers.ts
```diff
@@ -131,7 +131,10 @@ export const getDataVolumeFromState = ({
   createOwnerReference?: boolean;
 }): V1beta1DataVolume => {
   const dataVolume = getEmptyVMDataVolumeResource(vm, createOwnerReference);
-  const dvName = resultVolume?.dataVolume?.name || `${vm?.metadata?.name}-${diskState.diskName}`;
+  const dvName =
+    resultVolume?.dataVolume?.name ||
+    resultVolume?.persistentVolumeClaim?.claimName ||
+    `${vm?.metadata?.name}-${diskState.diskName}`;
```
hstastna, line 137: I think we shouldn't need `?` in `vm?.metadata?.name` here. If yes then something is going wrong in here, because I think that..  if we think that it might be `undefined` (because we add "?" in there), then the name wouldn't look nice/correct. Or would it be correct anyway? Ignore this comment if yes. But if not, it's better to let the UI break here so the issue could be seen and fixed asap.

hstastna (commented):
Left one comment with a question about some detail, otherwise 👍🏽 🎇

src/utils/components/DiskModal/utils/helpers.ts
hstastna: I cannot fully agree, only partially. This is what you want most of the time. But from my experience (and from experience of some other skilled devs), this is too defensive development style that can lead to situations when the UI doesn't crash even when it better should - in cases when something is not done well in that place, and then absence of the UI crash leads to not revealing a bug early enough, and causing more problems that can be hard to find later.

PR #1141, opened 2023-03-14
Bug 2167608: highlight parts

upalatucci:
## 📝 Description

Highlight editor parts in the LIVE YAML

## 🎥 Demo

https://user-images.githubusercontent.com/29160323/225008478-a1458235-6d86-4a1c-b70d-17644c59fa3f.mp4

### DIFF
--- a/src/utils/components/SidebarEditor/SidebarEditor.tsx
+++ b/src/utils/components/SidebarEditor/SidebarEditor.tsx
@@ -1,6 +1,7 @@
 import React, { ReactNode, Suspense, useContext, useMemo, useState } from 'react';
 import { dump } from 'js-yaml';
 
+import { PATHS_TO_HIGHLIGHT } from '@kubevirt-utils/resources/vm/utils/constants';
 import { K8sResourceCommon, YAMLEditor } from '@openshift-console/dynamic-plugin-sdk';
 import {
   Alert,
@@ -20,6 +21,7 @@ import {
 import Loading from '../Loading/Loading';
 
 import { SidebarEditorContext } from './SidebarEditorContext';
+import { useEditorHighlighter } from './useEditorHighlighter';
 import { safeLoad } from './utils';
 
 import './sidebar-editor.scss';
@@ -29,6 +31,7 @@ type SidebarEditorProps<Resource> = {
   onResourceUpdate?: (newResource: Resource) => Promise<Resource | void>;
   children: ReactNode | ReactNode[] | ((resource: Resource) => ReactNode);
   onChange?: (resource: Resource) => void;
+  pathsToHighlight?: string[];
 };
 
 const SidebarEditor = <Resource extends K8sResourceCommon>({
@@ -36,6 +39,7 @@ const SidebarEditor = <Resource extends K8sResourceCommon>({
   resource,
   onResourceUpdate,
   onChange,
+  pathsToHighlight = PATHS_TO_HIGHLIGHT.DEFAULT,
 }: SidebarEditorProps<Resource>): JSX.Element => {
   const [editableYAML, setEditableYAML] = useState('');
   const [loading, setLoading] = useState(false);
@@ -50,6 +54,7 @@ const SidebarEditor = <Resource extends K8sResourceCommon>({
 
   const { showEditor, isEditable } = useContext(SidebarEditorContext);
   const editedResource = safeLoad<Resource>(editableYAML);
+  const editorRef = useEditorHighlighter(editableYAML, pathsToHighlight, showEditor);
 
   const changeResource = (newValue: string) => {
     setEditableYAML(newValue);
@@ -95,6 +100,7 @@ const SidebarEditor = <Resource extends K8sResourceCommon>({
                   onChange={changeResource}
                   onSave={() => onUpdate(editedResource)}
                   options={{ readOnly: !isEditable }}
+                  ref={editorRef}
                 />
               </Suspense>
             </StackItem>
--- a/src/utils/components/SidebarEditor/useEditorHighlighter.ts
+++ b/src/utils/components/SidebarEditor/useEditorHighlighter.ts
@@ -1,0 +1,39 @@
+import { useEffect, useRef, useState } from 'react';
+
+import { YAMLEditorRef } from '@openshift-console/dynamic-plugin-sdk';
+
+import { createSelection, getLinesToHighlight } from './utils';
+
+export const useEditorHighlighter = (
+  editableYAML: string,
+  pathsToHighlight: string[],
+  showEditor: boolean,
+) => {
+  const [editor, setEditor] = useState<YAMLEditorRef['editor']>();
+  const isHighlighed = useRef(false);
+
+  useEffect(() => {
+    isHighlighed.current = false;
+  }, [pathsToHighlight, showEditor]);
+
+  useEffect(() => {
+    const highlightPaths = async () => {
+      if (editor && editableYAML && pathsToHighlight && !isHighlighed.current) {
+        isHighlighed.current = true;
+        const ranges = getLinesToHighlight(editableYAML, pathsToHighlight);
+
+        await editor.getAction('editor.foldAll').run();
+
+        const selections = ranges.map((range) => createSelection(range));
+
+        editor.setSelections(selections);
+        await editor.getAction('editor.unfoldRecursively').run();
+        setTimeout(() => editor.revealLineInCenter(ranges.at(-1).start), 500);
+      }
+    };
+
+    highlightPaths();
+  }, [editableYAML, editor, pathsToHighlight]);
+
+  return (ref: YAMLEditorRef) => setEditor(ref?.editor);
+};
--- a/src/utils/components/SidebarEditor/utils.ts
+++ b/src/utils/components/SidebarEditor/utils.ts
@@ -8,3 +8,64 @@ export const safeLoad = <Resource>(value: string): Resource | undefined => {
     return;
   }
 };
+
+export type LineRange = { start: number; end: number };
+
+const getLineFromPath = (resourceYAML: string, path): LineRange => {
+  const yamlLines = resourceYAML.split('\n');
+  const range = { start: 0, end: yamlLines.length - 1 };
+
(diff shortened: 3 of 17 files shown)


src/utils/components/SidebarEditor/useEditorHighlighter.ts
```diff
@@ -0,0 +1,43 @@
...
+      window.orientation = editor;
+    }
+  }, [editableYAML, editor, pathsToHighlight]);
+
+  useEffect(() => {
+    isHighlighed.current = false;
+  }, [pathsToHighlight, showEditor]);
+
+  useEffect(() => {
+    highlightPaths();
```
metalice, line 39: can u just put highlightPaths content here? then u dont need the usecallback

src/utils/components/SidebarEditor/utils.ts
```diff
@@ -8,3 +8,65 @@ export const safeLoad = <Resource>(value: string): Resource | undefined => {
...
+      .slice(startRange + 1, endRange)
+      .map((line) => line.replace(replaceIndentationRegex, ''));
+
+    // find the property
+    const startPropertyRange = rangeLines.findIndex((line) => line.startsWith(`${property}:`));
+
+    // find next property at same depth level
+    let endPropertyRange = rangeLines.findIndex(
+      (line, index) => index > startPropertyRange && line.match(/^[A-z]+:/g),
+    );
```
metalice, line 37:
```suggestion
    let endPropertyRange = rangeLines.slice(startPropertyRange).find(
      (line) =>  line.match(/^[A-z]+:/g),
    );
```

src/utils/components/SidebarEditor/utils.ts
```diff
@@ -8,3 +8,65 @@ export const safeLoad = <Resource>(value: string): Resource | undefined => {
...
+
+    if (endPropertyRange === -1) endPropertyRange = rangeLines.length;
```
metalice, line 39: can be removed?

src/views/virtualmachines/details/tabs/details/VirtualMachineDetailsPage.tsx
```diff
@@ -18,6 +18,14 @@ type VirtualMachineDetailsPageProps = RouteComponentProps<{
   obj?: V1VirtualMachine;
 };
 
+const pathsToHighlight = [
```
metalice, line 21: can ALL pathsToHighlights be saved in a central location? so paths can be reused and shared? i think this approve will be better

src/utils/components/SidebarEditor/utils.ts
upalatucci: Considering the previous suggestion i can do something different but in general i need this.
Here i understand that the range is finished because i find another property using the regex `/^[A-z]+:/g`.

So for example if i want to select the property `spec` i know the start because it starts with `spec:` and i know the end because i encounter another property at the same level `status:` with his conditions.

In that case i know that the spec range is finished. 

But if I want to select the property status, i know that the range is finished because there isn't another property after. So here `endPropertyRange` would be -1. the `findIndex` function does not find anything else after.

So at that point, i know that the end is equivalent to the last line

metalice: Ok cool. But what's wrong in the code I suggested?

metalice: Oh I see you took part of it

src/utils/resources/vm/utils/constants.ts
```diff
@@ -1,3 +1,23 @@
 export const NO_DATA_DASH = '-';
 
 export const MILLISECONDS_TO_SECONDS_MULTIPLIER = 1000;
+
+export const PATHS_TO_HIGHLIGH = {
```
hstastna, line 5: Please, use `PATHS_TO_HIGHLIGHT` with 'T' ;)

src/utils/components/SidebarEditor/SidebarEditor.tsx
```diff
@@ -29,13 +30,15 @@ type SidebarEditorProps<Resource> = {
...
   onChange?: (resource: Resource) => void;
+  pathsToHighlight?: string[];
 };
 
 const SidebarEditor = <Resource extends K8sResourceCommon>({
   children,
   resource,
   onResourceUpdate,
   onChange,
+  pathsToHighlight = ['spec.template.spec.domain.devices.disks', 'spec.template.spec.volumes'],
```
hstastna, line 41: Just curious: what is the reason of setting exactly these paths as the default values for the editor? Thanks, Ugo!

hstastna (commented):
The code looks good, +1 for the comments in `getLineFromPath`.

When I tested the code a while ago, the editor did highlight the relevant lines nicely, however when I changed something in the VM details, e.g. changed _Description_ field (by clicking on that field and adding text to the modal and saving), the yaml updated, but didn't highlight the lines again, got stuck. I had to turn the switcher off and on again to make it displayed highlighting the lines again.
Also the description info didn't occur in the yaml the expected way - at least for me - I mean it didn't occurred there easily visible, but hidden, I had to click on > to expand the annotations in the editor to reveal that part of the yaml - and that was a bit annoying to me, especially when I edited info belonging to that part of the yaml. Not sure if this is the behavior we expect. Additionally, the description was not highlighted at all, not before, nor after editing it - that's not what I'd expect for sure. 

To sum up, IMO we should highlight all the relevant lines (e.g. also Description, when being on VM Details tab), not only most of them. PTAL into that a bit. Also I think we should highlight lines also after some editing (not only before) :) And If we should display some yaml parts expanded or not - that's what I'm not sure about. Depends on the design/what user prefer. Otherwise great work, Ugo! ❇️

src/utils/components/SidebarEditor/SidebarEditor.tsx
metalice: PATHS_TO_HIGHLIGH.DEFAULT

src/utils/components/SidebarEditor/useEditorHighlighter.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import { createSelection, getLinesToHighlight } from './utils';
+
+export const useEditorHighlighter = (
+  editableYAML: string,
+  pathsToHighlight: string[],
+  showEditor: boolean,
+) => {
+  const [editor, setEditor] = useState<YAMLEditorRef['editor']>();
+  const isHighlighed = useRef(false);
```
metalice, line 13: why do we need it? if you will use showEditor as your dependency at line 36 we will get same effect?

src/utils/components/SidebarEditor/utils.ts
```diff
@@ -8,3 +8,63 @@ export const safeLoad = <Resource>(value: string): Resource | undefined => {
...
+    // property not found
+    if (startPropertyRange === -1) return undefined;
+
+    startRange += startPropertyRange + 1;
+
+    endRange = startRange + rangeLength;
+  }
+
+  // editor lines starts from 1, array starts from 0
+  return { start: startRange + 1, end: endRange + 1 };
```
metalice, line 50: can the + 1 be add in line 44, 46? here you can simply return ranges ? as suggested in line 17

src/utils/components/SidebarEditor/utils.ts
```diff
@@ -8,3 +8,63 @@ export const safeLoad = <Resource>(value: string): Resource | undefined => {
     return;
   }
 };
+
+export type LineRange = { start: number; end: number };
+
+const getLineFromPath = (resourceYAML: string, path): LineRange => {
+  const yamlLines = resourceYAML.split('\n');
+  let startRange = 0;
+  let endRange = yamlLines.length - 1;
```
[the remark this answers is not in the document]
upalatucci: nice thanks

src/utils/components/SidebarEditor/utils.ts
upalatucci: No, because all the previous calculations are done considering that the arrays start from 0, and only at the end, we need to add the +1.

If I do +1 on 44, and 46, I should also do some -1 in other places

hstastna:
> @hstastna i've addressed your issue and added `metadata.annotations` to the selection in the details tab. Unfortunately, if you change something, the highlight goes away because its not really an highlight, its more like a selection. But we can find other solutions after this pr if you want

I am not sure if this actual version of the solution wouldn't add just more confusion to the users, if it worked only before they change anything. Maybe we could simulate that my "quick fix" of turning off and on the YAML switcher, so the fields would be selected/highlighted again? Or to reload the editor somehow after the user changes something?

metalice:
ok by me , @hstastna when you are satisfied with this pr give it a go :)

hstastna:
>I am not sure if this actual version of the solution wouldn't add just more confusion to the users, if it worked only before they change anything. Maybe we could simulate that my "quick fix" of turning off and on the YAML switcher, so the fields would be selected/highlighted again? Or to reload/rerender the whole editor somehow after the user changes something?

I'd give this mentioned by me above a try at least, before I give this PR a go. WDYT, @upalatucci ?

upalatucci:
Guys i don't know. At this point, I'll contact Yifat and see what she thinks. I'll share with here what i think we can do with the editor, and let here figure out what we should do with that.

I can't re-highlight the editor when something gets edited because its a live editor, so you can edit things from the editor and also from the form. For example in the Environment tab, you can write down envs values in the yaml and also in the form and you'll see in real time the result in the yaml. So i don't think we want the yaml to highlight at every char pressed.

hstastna:
> Guys i don't know. At this point, I'll contact Yifat and see what she thinks. I'll share with her what i think we can do with the editor, and let her figure out what we should do with that.
> 
> I can't re-highlight the editor when something gets edited because its a live editor, so you can edit things from the editor and also from the form. For example in the Environment tab, you can write down envs values in the yaml and also in the form and you'll see in real time the result on the other side. So i don't think we want the yaml to highlight at every char pressed.

Ok, great, so let's ask Yifat and we'll see. If she's gonna be okay with what we actually have, let's merge this. Just make sure with her. Thanks! :)

PR #1144, opened 2023-03-15
Bug 2175976: 'Select InstanceType' should show the volume's default instanceType

avivtur:
## 📝 Description

on selecting a volume in the IT(=instnace type) VM create flow, we now also see the default IT of the volume.
If a default IT does exists, the name will be shown on 3rd section, VM details. If the default IT also belong to one of the 4 common IT series, it will also reflect on the 2nd section of IT select.

## 🎥 Demo

https://user-images.githubusercontent.com/67270715/225216924-8d7423d5-3e5c-4d90-993e-d915804265cf.mp4


src/views/catalog/CreateFromInstanceTypes/utils/utils.ts
```diff
@@ -8,11 +8,28 @@ import { DEFAULT_NAMESPACE } from '@kubevirt-utils/constants/constants';
...
+  DEFAULT_INSTANCETYPE_LABEL,
+  DEFAULT_PREFERENCE_LABEL,
+  InstanceTypeState,
+} from './constants';
 
 const generateCloudInitPassword = () =>
   `${getRandomChars(4)}-${getRandomChars(4)}-${getRandomChars(4)}`;
 
+export const getInstanceTypeState = (defaultInstanceTypeName: string): InstanceTypeState => {
+  const [prefix, size] = defaultInstanceTypeName?.split('.');
```
metalice, line 24:
```suggestion
  const [prefix, size]: [string, InstanceTypeSize] = defaultInstanceTypeName?.split('.');
```

can work?

metalice: then u can remove the as below

avivtur: will change it in a follow-up PR :) good suggestion

avivtur: @metalice adding your suggestion is not allowed from TS as string[] cannot be converted, giving following error:
```
Type 'string[]' is not assignable to type '[string, InstanceTypeSize]'.
  Target requires 2 element(s) but source may have fewer. 
  ```

PR #1150, opened 2023-03-15
Bug 2175983: Update text on Delete modal for bootable volumes

hstastna:
## 📝 Description

**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2175983

Update the title and the text of the _Delete_ modal for bootable volumes the following way:

Modal title text: 
_"Delete volume parameters?"_

Modal content: 
_"Deleting the parameters will mark this volume as non-bootable and remove it from the bootable volumes list. The volume will still be available in the cluster._

This change of the text is made according to the design changes, as the original text and the title of the modal weren't quite appropriate or clear enough to the user.

_Related design doc:_
https://docs.google.com/document/d/15Cv54978AZKeukkurvkFN1YEx_kBsm_3XcmLey_cgZY/edit?disco=AAAAq58TGPw

_Note:_
"Are you sure you want to delete the parameters from bootable volume <volume_name>?" was not included in the text, because as discussed in the design doc, it wasn't necessary as this is not a destructive action (we're not deleting the whole resource/object, only a few params). Later when this part of the UI will be revisited regarding the question if we could completely delete the volume, then this text will be more relevant. 

## 🎥 Screenshots
**Before:**
del_before

**After:**
del_afterr


hstastna:

hstastna:
/hold

Seems that the discussion about the proper text of the modal is still continuing...

hstastna:
I'm removing hold from this PR, as the discussions regarding the text of the modal seem kinda like never ending story so let's merge what we have, it's already an improvement.

PR #1151, opened 2023-03-15
Bug 2175986: Improve message for different StorageClass

hstastna:
## 📝 Description

**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2175986

Update the message displayed in _Add volume to boot from_ modal when uploading volume, for case that a different StorageClass is selected than the default one, to the following one:

_This StorageClass might cause slower cloning._

So the message is simplified and explains what a "poor performance" (mentioned in the original message) means, when selecting such a StorageClass.

## 🎥 Demo
**Before:**
stor_before

**After:**
stor_after


hstastna:

PR #1158, opened 2023-03-17
Bug 2175977: Disable Create VM button if not everything selected

hstastna:
## 📝 Description

**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2175977

Disable _Create VirtualMachine_ button when creating a VM from InstanceType, if not everything needed is selected, for example when a VM name is missing or InstanceType selection (in case that a chosen volume is missing label with InstanceType name).

## 🎥 Screenshots
**Before**:
Name and InstanceType missing but _Create VirtualMachine_ button still enabled: 
no_before

**After:**
Name and InstanceType missing and _Create VirtualMachine_ button disabled, as expected: 
no_after

### DIFF
--- a/src/views/catalog/CreateFromInstanceTypes/components/CreateVMFooter/CreateVMFooter.tsx
+++ b/src/views/catalog/CreateFromInstanceTypes/components/CreateVMFooter/CreateVMFooter.tsx
@@ -1,8 +1,9 @@
-import React, { FC, useState } from 'react';
+import React, { FC, useMemo, useState } from 'react';
 import { useHistory } from 'react-router-dom';
 import produce from 'immer';
 
 import { SSHSecretCredentials } from '@catalog/CreateFromInstanceTypes/components/VMDetailsSection/components/SSHKeySection/utils/types';
+import { DEFAULT_INSTANCETYPE_LABEL } from '@catalog/CreateFromInstanceTypes/utils/constants';
 import VirtualMachineModel from '@kubevirt-ui/kubevirt-api/console/models/VirtualMachineModel';
 import { V1beta1DataSource } from '@kubevirt-ui/kubevirt-api/containerized-data-importer/models';
 import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
@@ -53,6 +54,14 @@ const CreateVMFooter: FC<CreateVMFooterProps> = ({
     namespace: vm?.metadata?.namespace,
     group: VirtualMachineModel.apiGroup,
   });
+
+  const hasNameAndInstanceType = useMemo(
+    () =>
+      !isEmpty(vm?.metadata?.name) &&
+      (!isEmpty(selectedBootableVolume?.metadata?.labels?.[DEFAULT_INSTANCETYPE_LABEL]) ||
+        !isEmpty(vm?.spec?.instancetype?.name)),
+    [selectedBootableVolume, vm],
+  );
 
   const handleSubmit = async () => {
     setIsSubmitting(true);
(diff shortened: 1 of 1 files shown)


hstastna:

src/views/catalog/CreateFromInstanceTypes/CreateFromInstanceType.tsx
```diff
@@ -56,6 +57,14 @@ const CreateFromInstanceType: FC = () => {
 
   const namespace = ns === ALL_NAMESPACES_SESSION_KEY ? DEFAULT_NAMESPACE : ns;
 
+  const hasNameAndInstanceType = useMemo(
+    () =>
+      !isEmpty(vmName) &&
+      (!isEmpty(selectedBootableVolume?.metadata?.labels[DEFAULT_INSTANCETYPE_LABEL]) ||
```
metalice, line 63: !isEmpty(selectedBootableVolume?.metadata?.labels?.[DEFAULT_INSTANCETYPE_LABEL]

src/views/catalog/CreateFromInstanceTypes/CreateFromInstanceType.tsx
```diff
@@ -56,6 +57,14 @@ const CreateFromInstanceType: FC = () => {
 
   const namespace = ns === ALL_NAMESPACES_SESSION_KEY ? DEFAULT_NAMESPACE : ns;
 
+  const hasNameAndInstanceType = useMemo(
```
metalice, line 60: why did u choose to calc hasNameAndInstanceType here and not in CreateVMFooter ? 
CreateVMFooter already get all those props of vmName, selectedbootablevolume and  , selectedInstanceType ?
why add more props instead of getting the needed value calc in there final location where its needed..

hstastna: The answer is right in front of you, see [this code change](https://github.com/kubevirt-ui/kubevirt-plugin/pull/1158/files/84b9f37427132340b2c28af146f1117cbdb66527#diff-b3e791410ef7aa2ba5068a197aaa56b96553c0c8207d308ed31ce7962cfef9bfR64) :)
To explain more: info about selected InstanceType was missing in `CreateVMFooter`.
So I had 2 options: to make the required calculations in `CreateVMFooter` and add another arg representing selected InstanceType to be able to do those calcs in there, or to make it one component above, where all the info I need is, and just to pass already calculated data to `CreateVMFooter`.
What does it mean? It means we'd have to add another arg to `CreateVMFooter` anyway.

metalice: well im guessing that from the vm u can get the missing info @avivtur which fields those she needs?

avivtur: Hi :) 
as I can see in according to [this](https://github.com/kubevirt-ui/kubevirt-plugin/pull/1158/files#diff-b3e791410ef7aa2ba5068a197aaa56b96553c0c8207d308ed31ce7962cfef9bfR146-R152)
the generated VM should have the required fields to check under the VM in the footer component.
vm.spec.instancetype.name should hold the name for the IT
and as vm name I think is quiet clear how to get this :)

hstastna: I think I need only InstanceType name to get properly from the `vm`, in my case. Thanks so much for your help, Aviv and Matan! ❇️

PR #1164, opened 2023-03-20
Bug 2168111: fix customize default source available

upalatucci:
## 📝 Description

`defaultSourceExists` should depend on just `isBootSourceAvailable` instead of `canQuickCreate` as `canQuickCreate` depends on required parameters with no default values and `isBootSourceAvailable`. 

The template with no `defaultSourceAvailable` does not get default storage configuration in the customization as the template cannot start with the storage resource configured initially.

But the template in the bug has the source available but has also a required parameter.

The user should be able to use default storage configuration when template has source available and required parameters


src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
```diff
@@ -85,9 +85,9 @@ export const TemplatesCatalogDrawerCreateForm: FC<TemplatesCatalogDrawerCreateFo
         });
     };
 
-    const onCustomize = (e: MouseEvent) => {
+    const onCustomize = (e) => {
```
metalice, line 88: why removing the event?

upalatucci: Sorry, bad conflict resolution

upalatucci: resolved

PR #1167, opened 2023-03-20
Bug 2174619: all devices are boot source, just needs to arrange them

upalatucci:
## 📝 Description

Every device is a boot source in theory. So no need to 'add source'. 
The user should just reorder the devices.
Because of that, the `-` symbol will just de-prioritize the device and put it as last device in the order

## 🎥 Demo

**Before**
image

**After**

Screenshot from 2023-03-20 15-41-57

https://user-images.githubusercontent.com/29160323/226893166-5e1c3e81-11a8-4fcd-bb85-5c8c5eea5e7c.mp4


gouyang:
@upalatucci We need to remove the '-' symbol, it's not clear to user if it means de-prioritize the device, most of the time, it means delete.

PR #1168, opened 2023-03-20
Bug 2179626: Removing filtering from diagnostic volume status

metalice:
## 📝 Description

Removing filter and column management for diagnostic volume status

## 🎥 Demo

After:
image

Before:
image

### DIFF
--- a/src/views/virtualmachines/details/tabs/diagnostic/hooks/useDiagnosticVolumeStatusTableColumns.ts
+++ b/src/views/virtualmachines/details/tabs/diagnostic/hooks/useDiagnosticVolumeStatusTableColumns.ts
@@ -12,11 +12,11 @@ type DiagnosticColumn = {
   sort: (columnIndex: any) => ThSortType;
 };
 
-type UseDiagnosticVolumeStatusTableColumns = () => [
-  columns: TableColumn<DiagnosticColumn>[],
-  activeColumns: TableColumn<DiagnosticColumn>[],
-  sort: DiagnosticSort,
-];
+type UseDiagnosticVolumeStatusTableColumns = () => {
+  columns: TableColumn<DiagnosticColumn>[];
+  activeColumns: TableColumn<DiagnosticColumn>[];
+  sorting: DiagnosticSort;
+};
 const useDiagnosticVolumeStatusTableColumns: UseDiagnosticVolumeStatusTableColumns = () => {
   const { t } = useKubevirtTranslation();
   const { getSorting, sort } = useDiagnosticSort();
@@ -45,7 +45,7 @@ const useDiagnosticVolumeStatusTableColumns: UseDiagnosticVolumeStatusTableColum
     columnManagementID: 'diagnostic-tab-volume',
   });
 
-  return [columns, activeColumns, sort];
+  return { columns, activeColumns, sorting: sort };
 };
 
 export default useDiagnosticVolumeStatusTableColumns;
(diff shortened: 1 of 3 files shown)


avivtur (approved):

hstastna (changes requested):
@metalice Are you sure you are fixing the BZ the correct way? I am not sure. In the BZ, the problem is with filters not working correctly, not with their existence. There is:

_Expected results: Filters are cleared_

It's because "Clear all filters" button does not work, and those filters are somehow connected and they shouldn't be, they should work independently, each for its section IMO.
It doesn't mean to remove some filter (as it's done in this PR), it means "to make both filters work as expected, to make 'Clear all filters' button work". 
But if there was some design change that would suggest to remove some filters completely (which IMO doesn't make sense to me, we do want to filter both subsections' data), there should be some info about that in the BZ or in this PR, but isn't. Thanks.

metalice:
> @metalice Are you sure you are fixing the BZ the correct way? I am not sure. In the BZ, the problem is with filters not working correctly, not with their existence. There is:
> 
> _Expected results: Filters are cleared_
> 
> It's because "Clear all filters" button does not work, and those filters are somehow connected and they shouldn't be, they should work independently, each for its section IMO. It doesn't mean to remove some filter (as it's done in this PR), it means "to make both filters work as expected, to make 'Clear all filters' button work". But if there was some design change that would suggest to remove some filters completely (which IMO doesn't make sense to me, we do want to filter both subsections' data), there should be some info about that in the BZ or in this PR, but isn't. Thanks.

Yes I'm sure.

short story: that what me and Gu agreed upon as a bug fix.

long story:
there is a bug in the console component listfilter that on the same page cannot be 2 tables with filters as the filter is using URL query to interpolate the filters needed, and there isn't an id or something to distinguish in the query so this causes the filters to always sync which is the bug.

hstastna:
> short story: that what me and Gu agreed upon as a bug fix.
> 
> long story: there is a bug in the console component listfilter that on the same page cannot be 2 tables with filters as the filter is using URL query to interpolate the filters needed, and there isn't an id or something to distinguish in the query so this causes the filters to always sync which is the bug.

Thank you very much for the explanation, Matan. And does the UX know about that? IMO they should know this to understand why we did this change, and to update the design accordingly. Also IMO it would be good to put this info to bugzilla to prevent any confusions for others that might access the BZ.

metalice:
> > short story: that what me and Gu agreed upon as a bug fix.
> > long story: there is a bug in the console component listfilter that on the same page cannot be 2 tables with filters as the filter is using URL query to interpolate the filters needed, and there isn't an id or something to distinguish in the query so this causes the filters to always sync which is the bug.
> 
> @metalice Thank you very much for the explanation, Matan! Now I understand. And does the UX know about that? IMO they should know this to understand why we did this change, and to update the design accordingly. Also IMO it would be good to put this info to bugzilla to prevent any confusions for others that might access the BZ. Could you, please, put some info there? Thanks!

In the original design, there wasn't 2 tables, that came later after some conversation with Ux and the architects , the filter was a nice addition from my side, that ux liked, removing it , means simply go back to original design with was fine with everyone from the start, so no issue there

hstastna:
> In the original design, there wasn't 2 tables, that came later after some conversation with Ux and the architects , the filter was a nice addition from my side, that ux liked, removing it , means simply go back to original design with was fine with everyone from the start, so no issue there

Thanks for this great explanation of the context around. It's good to know! Maybe next time, it could be great to add a few words like this to the description of the PR to get the PR merged faster ;)

PR #1170, opened 2023-03-20
Bug 2177578: Set columns' width for volumes in Catalog > InstanceTypes

hstastna:
## 📝 Description

**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2177578

Set the minimum width for each of the columns in the volumes list in _Catalog_ > _InstanceTypes_. To be able to do that, update `TableData` component to be able to get `width` as an optional prop, to get styled according to requirements.

## 🎥 Screenshots
**Before:**
Not enough space for the volume name, in case that the description is very long:
row_before1
If more columns displayed, same problem as for the volume name:
row_before2

**After:**
Enough space for the volume name, in case that the description is very long:
a1
If more columns displayed, each column has an appropriate width suitable for its content:
a2

### DIFF
--- a/src/views/catalog/CreateFromInstanceTypes/components/BootableVolumeList/components/BootableVolumeRow/BootableVolumeRow.tsx
+++ b/src/views/catalog/CreateFromInstanceTypes/components/BootableVolumeList/components/BootableVolumeRow/BootableVolumeRow.tsx
@@ -39,6 +39,7 @@ const BootableVolumeRow: FC<BootableVolumeRowProps> = ({
   const bootVolumeName = bootableVolume?.metadata?.name;
   const pvcDiskSize = pvcSource?.spec?.resources?.requests?.storage;
   const sizeData = pvcDiskSize && humanizeBinaryBytes(pvcDiskSize);
+
   return (
     <Tr
       isHoverable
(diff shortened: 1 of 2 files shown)


hstastna:

metalice:
hey
Did u run this design with Yifat? I think that making such big lines is not something very nice, probably better to shorten the description with '...'

hstastna:
> hey Did u run this design with Yifat? I think that making such big lines is not something very nice, probably better to shorten the description with '...'

We are not adding "big lines" in this PR, this PR adds missing default width of each of the columns, so the volume name and any other info is well readable in the table now, no matter the length of description.

We've talked with Yifat, of course, and the design defintely requires some improvements. However, we haven't come up with any solution that wouldn't make other places of the UI inconsistent with this one. So this is still WIP, but probably we will simply go for shortening the description and adding '...', as you've mentioned, but that will be needed to do also in some other places, to keep the UI's consistent look. So I'd go for this improvement in a followup PR. It's a bit far from the original problem we're trying to solve in this PR. WDYT, @metalice ?

metalice:
> m we're trying to solve in this PR. WDY

IMHO best appearance will be 1-2 lines of description that end with '...' and a tooltip on hover with full content.
Where else in the code will we need to adjust? Do we have any other table with long descriptions or strings in cells?

hstastna:
> IMHO best appearance will be 1-2 lines of description that end with '...' and a tooltip on hover with full content.

Thank you very much for your opinion and suggestion to improve this part of the UI. I agree with that. And I am still in contact with Yifat, and we're trying to come up with the best option (details about the mentioned approach, 1 or 2 lines? popover? etc...). There can be a problem with this approach, as it IMO would look bad if there would be a huuge popover with all that hidden text..

Anyway, IMO I'd implement that in a followup PR, not here. WDYT about that?

> Where else in the code will we need to adjust? Do we have any other table with long descriptions or strings in cells?

Bootable volumes list actually for sure, and I am not sure about DataSources list, I remember there were more list views with Description column, but it may have changed as time goes by. Checking it...

PR #1171, opened 2023-03-21
Bug 2168484: Add dates to VM metric charts labels

pcbailey:
## 📝 Description

This PR adds dates to the x-axis labels of the VM metric charts for durations that span multiple days and changes the time format for durations that do not.

### Multiple days format
Selection_229

### Single day format
Selection_230


src/views/virtualmachines/details/tabs/metrics/MigrationCharts/MigrationCharts.tsx
```diff
@@ -3,36 +3,32 @@ import React from 'react';
 import { V1VirtualMachineInstance } from '@kubevirt-ui/kubevirt-api/kubevirt';
 import MigrationThresholdChart from '@kubevirt-utils/components/Charts/MigrationUtil/MigrationThresholdChart';
 import MigrationThresholdChartDiskRate from '@kubevirt-utils/components/Charts/MigrationUtil/MigrationThresholdChartDiskRate';
-import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
+import { t } from '@kubevirt-utils/hooks/useKubevirtTranslation';
```
metalice, line 6: why did u did this change?

pcbailey: When I first started looking at the bug I thought I was going to be changing this file and went ahead and made this change to make it look cleaner. Is it a problem? They should be equivalent, right?

metalice: yes its equal, but until now we preferred using the hook method on tsx files and simply t on ts files, to keep it organized

pcbailey: Hrm. That's the first I've heard of that. How is unnecessary boilerplate more organized?

metalice: Simply as a hook for tsx and t for ts because u cannot use hooks. 
Why do we not simply use t instead of a hook also for tsx? 
t will be static and not rerender like t from a hook; how much its import?
 I don't know... but this is why we kept it like this so if a new reactive version of t will be needed for some reason the hook will make sure of it. 
maybe static t will cause problems in static utils files in the future? I also dont know, but we can always go back to re adding t as arg to function. hope this makes sense to u

pcbailey: Sure. The short answer is "we don't know, but better safe than sorry". lol

I'll change it back in a few.

PR #1173, opened 2023-03-21
fix not migratable label

upalatucci:
## 📝 Description

> Add a brief description

## 🎥 Demo

**Before**

Screenshot from 2023-03-21 12-12-04
image

**After**

Screenshot from 2023-03-21 12-24-56

Screenshot from 2023-03-21 12-25-05


metalice:
is there a BZ to it? if so please update the title.

/lgtm

PR #1175, opened 2023-03-21
Bug 2172375: Error on deleting secret

upalatucci:
## 📝 Description

**Cause**
`produce` inside another `produce` cause some issues with `immer` apparently.

**Solution**
Anyways, the function `removeSecretToVM` was a little bit useless as its just one line so i copied it without the additional `produce`
If the VM does not have other accessCredentials, just don't try to remove it another time.

Why that change in TabModal?
This bug would be more easily solved if the TabModal would throw the error in the submit catch.
Even to just log that something bad happened.

### DIFF
--- a/src/utils/components/TabModal/TabModal.tsx
+++ b/src/utils/components/TabModal/TabModal.tsx
@@ -69,7 +69,10 @@ const TabModal: TabModalFC = React.memo(
 
       onSubmit(obj)
         .then(onClose)
-        .catch(setError)
+        .catch((submitError) => {
+          setError(submitError);
+          console.error(submitError);
+        })
         .finally(() => setIsSubmitting(false));
     };
 
--- a/src/utils/components/VMAuthorizedSSHKeyModal/utils.ts
+++ b/src/utils/components/VMAuthorizedSSHKeyModal/utils.ts
@@ -8,7 +8,6 @@ import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
 import {
   addSecretToVM,
   createVmSSHSecret,
-  removeSecretToVM,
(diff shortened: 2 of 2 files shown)


src/utils/components/TabModal/TabModal.tsx
```diff
@@ -69,7 +69,10 @@ const TabModal: TabModalFC = React.memo(
 
       onSubmit(obj)
         .then(onClose)
-        .catch(setError)
+        .catch((submitError) => {
+          setError(submitError);
+          console.error(submitError);
```
avivtur, line 74: ??

src/utils/components/VMAuthorizedSSHKeyModal/utils.ts
```diff
@@ -55,11 +54,9 @@ export const changeVMSecret = async (
     });
 
     return await createVmSSHSecret(vm, newSSHKey, sshSecretName);
-  } else {
+  } else if (!!vm?.spec?.template?.spec?.accessCredentials) {
     await produceAndUpdate(vm, (vmDraft) => {
-      const produced = removeSecretToVM(vm);
-      vmDraft.spec = produced.spec;
-
+      vmDraft.spec.template.spec.accessCredentials = null;
```
avivtur, line 59: delete vmDraft.spec.template.spec.accessCredentials

upalatucci: okay

src/utils/components/TabModal/TabModal.tsx
avivtur: I meant why use console.error? we shouldn't use it like we shouldn't leave console.log :)

avivtur (approved):

PR #1176, opened 2023-03-21
Bug 2180553: Make removing description from bootable volume work

hstastna:
## 📝 Description

**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2180553

Make removing description from a volume in _Bootable volumes_ list work correctly, in case when the user deletes the description text in _Edit_ modal and saves the change.

## 🎥 Demo
**Before:**
Not possible to remove description if some already set:
https://user-images.githubusercontent.com/13417815/226709640-3078fdab-bd98-4298-92f9-af4636b52809.mp4

**After:**
Editing description works as expected, incl. description deletion:
https://user-images.githubusercontent.com/13417815/226709686-d6f58b3e-58bb-4169-a1f0-4f8f0e3792d2.mp4


hstastna:

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
```diff
@@ -83,7 +83,9 @@ const EditBootableVolumesModal: FC<EditBootableVolumesModalProps> = ({
     const instanceLabel = instanceType && {
       [DEFAULT_INSTANCETYPE_LABEL]: `${instanceType}.${size}`,
     };
-    const descriptionAnnotation = description && { [ANNOTATIONS.description]: description };
+    const descriptionAnnotation = description
```
metalice, line 86: i wonder if simply 
const descriptionAnnotation = { [ANNOTATIONS.description]: description };
will do the trick?

hstastna: No, because, when user deletes the description, it becomes an empty string `''`. And that's what we don't wanna save. We want not to see description annotation in the yaml in this case. And `undefined` does the trick for us in this case perfectly. Having `description: ''` set for the volume does not make sense. It's unnecessary.

metalice: Ok, so, in this case, I recommend using our isEmpty function on the description or trim, as a user enters simply space it will result in a truthy value.

metalice: or we are ok with simply empty space(s) as a description? i think we better off.

hstastna: > or we are ok with simply empty space(s) as a description? i think we better off.

I think we have some function to remove spaces at the beginning and end of the string, but it doesn't work well, especially if the user enters more spaces and nothing else. I've already tested. But I can add it here to `description` anyway, it's some improvement, for sure.

> Ok, so, in this case, I recommend using our isEmpty function on the description or trim, as a user enters simply space it will result in a truthy value.

I don't understand how this could help us. IMO it doesn't hep us because we still do need the condition I've added, for checking the description's falsey value, because we do need to update the description annotation even when the value is falsey - to be able to remove the annotation in case that there's nothing, no description (or empty string). I'd leave it as it is, as it does exactly what we want, and just would add the function to remove unnecessary spaces from the description string I've already mentioned.

hstastna: Updated the PR according to the suggestions, PTAL ;) Thanks!

PR #1181, opened 2023-03-22
Bug 2180279: VM cannot be started while creating from a template which has 2nd disk added

avivtur:
## 📝 Description

Adding the namespace to the DataVolumeTemplate object causes issues and is not really necessary

## 🎥 Demo
After:

https://user-images.githubusercontent.com/67270715/226923973-38367565-8f98-4345-85c4-f7230bf5bb1d.mp4


upalatucci:
Can the dataVolumeTemplate have also a different namespace? I think it can. I would copy also that info if it's in the metadata

avivtur:
> Can the dataVolumeTemplate have also a different namespace? I think it can. I would copy also that info if it's in the metadata

@upalatucci Hi 
the dataVolumeTemplate is an array to create DataVolumes for the VM. you can't create a DV for a VM not in the same namespace, you can target the DV to be cloned from another DV/PVC that exists on another namespace if you have DV clone role binding.
So no it can't have a different namespace, I hope that makes sense to you

PR #1188, opened 2023-03-24
Bug 2174744: Fix bootable volumes filter chips alignment

pcbailey:
## 📝 Description

This PR fixes the misalignment of the filter chips on the InstanceTypes page.

### Before
Selection_241

### After
Selection_240

### DIFF
--- a/src/views/catalog/CreateFromInstanceTypes/CreateFromInstanceType.scss
+++ b/src/views/catalog/CreateFromInstanceTypes/CreateFromInstanceType.scss
@@ -1,5 +1,10 @@
 .create-vm-instance-type-section {
   padding: var(--pf-global--spacer--xl);
+  &__list {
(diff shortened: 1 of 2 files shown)


hstastna (commented):
The solution doesn't look as in other pages, like below, but it's definitely an improvement. The best would be this:
ee

In your solution, there's still a bit more space above "Operating system" and the chips as it could be. IMO margin/padding top and bottom should have the same size (as in the picture above), so all the elements in the grey rectangle would be nicely centered (within the grey rectangle), not like this:
after

WDYT? Could we find some way to improve this even more (to reach the result as in the 1st picture I've added here)? 🙂

pcbailey:
The scope of the bug is to fix the alignment of the filter chips, not the entire component. I'll fix the other issues in a separate PR.

PR #1194, opened 2023-03-27
Bug 2180664: optimize metrics tab with multiple components

upalatucci:
## 📝 Description

Chart metrics optimization


upalatucci:
#1211 solve the issue

PR #1195, opened 2023-03-27
Disable VM actions for view only users

avivtur:
## 📝 Description

Introducing a new component called ActionDropdownItem, which his single purpose is to load the RBAC access for each action, and disable the action in case the user does not hold permissions for that action

## 🎥 Demo

Before:

Enabled for view-only permission on VM list:
disable-actions-by-RBAC-b4

After:

disabled for view-only permission on VM list:
disable-actions-by-RBAC-2

disabled for view-only permission on single VM overview:
disable-actions-by-RBAC-3

Enabled for project-admin:
disable-actions-by-RBAC

### DIFF
--- a/src/utils/components/ActionDropdownItem/ActionDropdownItem.tsx
+++ b/src/utils/components/ActionDropdownItem/ActionDropdownItem.tsx
@@ -1,0 +1,39 @@
+import React, { Dispatch, FC, SetStateAction } from 'react';
+
+import { Action, useAccessReview } from '@openshift-console/dynamic-plugin-sdk';
+import { DropdownItem } from '@patternfly/react-core';
+
+type ActionDropdownItemProps = {
+  action: Action;
+  setIsOpen: Dispatch<SetStateAction<boolean>>;
+};
+
+const ActionDropdownItem: FC<ActionDropdownItemProps> = ({ action, setIsOpen }) => {
+  const [actionAllowed] = useAccessReview(action?.accessReview);
+
+  const handleClick = () => {
+    if (typeof action?.cta === 'function') {
+      action?.cta();
+      setIsOpen(false);
+    }
+  };
+  return (
+    <DropdownItem
+      data-test-id={`${action?.id}`}
+      key={action?.id}
+      onClick={handleClick}
+      isDisabled={action?.disabled || !actionAllowed}
+      description={action?.description}
+    >
+      {action?.label}
+      {action?.icon && (
+        <>
+          {' '}
+          <span className="text-muted">{action.icon}</span>
+        </>
+      )}
+    </DropdownItem>
+  );
+};
+
+export default ActionDropdownItem;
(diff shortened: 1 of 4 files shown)


src/utils/components/ActionDropdownItem/ActionDropdownItem.tsx
```diff
@@ -0,0 +1,39 @@
...
+      key={action?.id}
+      onClick={handleClick}
+      isDisabled={action?.disabled || !actionAllowed}
+      description={action?.description}
+    >
+      {action?.label}
+      {action?.icon && (
+        <>
+          {' '}
+          <span className="text-muted">{action.icon}</span>
```
hstastna, line 32: Shouldn't we move also the css class/to create new .scss file for "text-muted", that was originally used in _src/views/virtualmachines/actions/components/VirtualMachineActions/VirtualMachineActions.tsx_? Aand maybe the class deserves some better name, more aligned with BEM? 🤔 👼🏽  WDYT?

hstastna (commented):
Left one comment, otherwise looks great to me! :)

src/utils/components/ActionDropdownItem/ActionDropdownItem.tsx
avivtur: The `text-muted` class is defined from console and is inherited by the plugin on runtime.
I can copy the css definitions from console to a new file in our project and give it a more accurate name, but it seems a bit redundant IMO, and surely if we do want to do something like that, it should have its own PR as this class been used all over the app.

hstastna: 🤔 interesting. Note that I was able to find that class defined also in _src/views/virtualmachines/details/tabs/overview/components/VirtualMachinesOverviewTabSnapshots/virtual-machines-overview-tab-snapshots.scss_
Not sure about the consequences of that.. @metalice WDYT?

avivtur: What you saw is `text-muted` but only if it's nested under an element with the `VirtualMachinesOverviewTabSnapshots--main` classname, which means to add/update those css definitions to the `text-muted` class only if it's under the mentioned class

hstastna: Yeah, but not sure if that's a correct approach. I'd better see a different name for the class, if wanted to avoid confusions and potential problems, for example if someone moved the component, so it wouldn't be under another classname anymore => it would stop working as expected. Also it may not be clear when looking into tsx file, if that's some our css class or that one from console..

avivtur: > Yeah, but not sure if that's a correct approach. I'd better see a different name for the class, if wanted to avoid confusions and potential problems, for example if someone moved the component, so it wouldn't be under another classname anymore => it would stop working as expected.

if you mean the different name for the `text-muted` class, then I still would have to copy the css from console to here and override/add the needed properties. I expect if someone changes the component structure to look also at the styles too and change it accordingly to not break anything else.

> Also it may not be clear when looking into tsx file, if that's some our css class or that one from console..

I agree that this is not a standart console css name, but the structure of putting a wrapper class over other classes is a pretty good hint for overriding a class we can't directly edit.

hstastna (approved):

PR #1200, opened 2023-03-28
Bug 2145101: Fixing the StorageClass warning

DanaOrr:


src/utils/components/DiskModal/EditDiskModal.tsx
```diff
@@ -72,6 +73,8 @@ const EditDiskModal: React.FC<DiskModalProps> = ({
     diskSourceReducer,
     initialDiskSourceState,
   );
+  const isBootSourceAvailable = getVMBootSourceType(vm)?.type !== 'NONE' ? true : false;
```
avivtur, line 76: please remove the part after the `?` it's not needed
you can also use BOOT_SOURCE.NONE instead of hard coding `'None'`
and are you sure this is the way the check if the boot source is available?

PR #1201, opened 2023-03-28
Bug 2182056: Cloned VM should not use the same PVC of the source VM

avivtur:
## 📝 Description

When cloning a new VM we also need to clone the pvcs, and not use the original ones  

## 🎥 Demo

After:

https://user-images.githubusercontent.com/67270715/228240033-a45dbc2f-f227-4afd-8db4-aac0b63db9b2.mp4

### DIFF
--- a/src/views/virtualmachines/actions/components/CloneVMModal/utils/helpers.tsx
+++ b/src/views/virtualmachines/actions/components/CloneVMModal/utils/helpers.tsx
@@ -3,10 +3,14 @@ import produce from 'immer';
 import { WritableDraft } from 'immer/dist/internal';
 
 import DataVolumeModel from '@kubevirt-ui/kubevirt-api/console/models/DataVolumeModel';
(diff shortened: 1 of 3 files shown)


src/views/virtualmachines/actions/components/CloneVMModal/utils/helpers.tsx
```diff
@@ -68,7 +72,7 @@ export const produceCleanClonedVM = (
     }
 
     delete draftVM?.status;
-    draftVM.spec.dataVolumeTemplates = [];
+    // draftVM.spec.dataVolumeTemplates = [];
```
metalice, line 75: Woops

pcbailey:
This looks good to me, but I'd like another reviewer as well since this is really important functionality.

avivtur:
> I assume the fix applys while cloning VM from catalog and instanceType, right?

Yes both worked as expected :)

PR #1203, opened 2023-03-28
Bug 2182362: no dedicated resources with instancetype

upalatucci:
## 📝 Description

Disabled editing Dedicated resources on vm with instance type. `DedicatedResourcesModal` edit the cpu `dedicatedCpuPlacement` property that goes in conflict with the instance type
Showing tooltip on vm with instancetype 

We cannot 

## 🎥 Demo

image


pcbailey:
The i18n test keeps failing. Could you try running `yarn i18n` again to see if it adds any changes that need to be pushed?

PR #1224, opened 2023-03-30
Bug 2182938: clone revisions alongside vm

upalatucci:
## 📝 Description

Clone `ControllerRevision` for `instancetype` and `preference` in the new namespace as suggested by @lyarwood 

**The flow**
 (Only In case of cloning vm created through instance type) 
 
- if in the destination namespace there is no ControllerRevision with the same name, we'll create it. (for instance type and prefernece)

- clone the vm

- in case the controller revision was not created, the cloned vm will be added as owner. In case the controlelr revision was created, does not have owner and we'll add the cloned vm as only owner

### DIFF
--- a/src/views/virtualmachines/actions/components/CloneVMModal/utils/helpers.tsx
+++ b/src/views/virtualmachines/actions/components/CloneVMModal/utils/helpers.tsx
@@ -2,9 +2,14 @@ import * as React from 'react';
 import produce from 'immer';
 import { WritableDraft } from 'immer/dist/internal';
 
+import ControllerRevisionModel from '@kubevirt-ui/kubevirt-api/console/models/ControllerRevisionModel';
 import DataVolumeModel from '@kubevirt-ui/kubevirt-api/console/models/DataVolumeModel';
-import { IoK8sApiCoreV1PersistentVolumeClaim } from '@kubevirt-ui/kubevirt-api/kubernetes/models';
+import {
+  IoK8sApiAppsV1ControllerRevision,
+  IoK8sApiCoreV1PersistentVolumeClaim,
+} from '@kubevirt-ui/kubevirt-api/kubernetes/models';
 import { V1DataVolumeTemplateSpec, V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import { buildOwnerReference } from '@kubevirt-utils/resources/shared';
 import {
   getDataVolumeTemplates,
   getDisks,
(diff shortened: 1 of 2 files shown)


avivtur (approved):

lyarwood (changes requested):
This will pull down the latest revision of the instance type or preference and stash that in a fresh controller revision. This could easily cause the VirtualMachine to change as a result so I'd highly recommend cloning the referenced controller revisions into the new namespace instead.

lyarwood:
> This will pull down the latest revision of the instance type or preference and stash that in a fresh controller revision. This could easily cause the VirtualMachine to change as a result so I'd highly recommend cloning the referenced controller revisions into the new namespace instead.

I should add that the OwnerReference will need to be updated to point to the new VirtualMachine in the new namespace.

src/views/virtualmachines/actions/components/CloneVMModal/utils/helpers.tsx
```diff
@@ -172,3 +177,70 @@ export const updateClonedDataVolumes = (
...
+) => {
+  try {
+    return await k8sGet<IoK8sApiAppsV1ControllerRevision>({
+      model: ControllerRevisionModel,
+      name: revisionName,
+      ns: destinationNamespace,
+    });
+  } catch (error) {
+    if (error.code !== 404) {
+      throw error;
```
metalice, line 204: You are throwing exceptions but not handling them on the other files when you use it...
I would have handled it here and reduced the possibility someone won't add a try-catch block or concating a .catch

upalatucci: The error would be caught by TabModal and it would display the error with the nice UI.

upalatucci: here I'm trying to fetch a ControllerVersion in another namespace using the referenceName.
I'll create it only if it's not there.

The fact that k8sGet throws a 404 error, its not actually an error. It just means that does not exist in that namespace.

upalatucci: Probably in the next version, the backend will handle cloning in different namespaces and we'll remove all the logic. But for now we need that

metalice: im sorry, but it isn't making sense to me. Please clarify in slack, thank you :)

PR #1231, opened 2023-03-31
Bug 2180666: Display PVC size correctly in Add disk modal

hstastna:
## 📝 Description

**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2180666

Display _Persistent Volume Claim size_ correctly, readable for a user, in _Add disk_ modal accessible from VM _Disks_ page, when choosing _PVC (creates PVC)_ for _Source_ field. Display the value in appropriate size and unit (for example _GiB_ instead of big number + _B_ as bytes unit).

## 🎥 Demo
The actual info about the selected PVC as an example of the expected size when selecting that PVC later in Add disk modal:
pvc_info

**Before:**
Not very user friendly PVC size (in bytes):
pvc_before

**After:**
Nice, readable size - smaller number with _GiB_ units instead of bytes:
pvc_after


hstastna:

PR #1232, opened 2023-03-31
Bug 2175990: Improve Default InstanceType field in Edit modal

hstastna:
## 📝 Description

**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2175990

Improve _Default InstanceType_ field/dropdown in _Edit volume metadata_ modal, accessible in _Bootable volumes_ list, according to the design update: add descriptive text to each one, omit "highperformance" and "server" options from the dropdown.

Improve also related 'Size' dropdown to display items with more info about number of CPUs and Memory size, sorted according to the size and not alpabetically, as expected.

_See more for Default InstanceType and Size drop downs design here:_
https://docs.google.com/document/d/1HTF6_H2WDXwlVy7RnrHH28dzFlhNIa15Hhpeoyg2R0Q/edit

## 🎥 Screenshots
**Before:**
Not much info about options for _Default InstanceType_, also the selected option not marked as selected in the dropdown:
cx1_before
Sizes sorted alphabetically:
cx2_before

**After:**
cx1_after
cx2_after

### DIFF
--- a/src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
+++ b/src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
@@ -1,14 +1,16 @@
-import React, { FC, useCallback, useMemo, useState } from 'react';
+import React, { ChangeEvent, FC, useCallback, useMemo, useState } from 'react';
 
 import FilterSelect from '@catalog/CreateFromInstanceTypes/components/AddBootableVolumeModal/components/FilterSelect/FilterSelect';
+import {
+  CategoryDetails,
+  InstanceTypeCategory,
+} from '@catalog/CreateFromInstanceTypes/components/SelectInstanceTypeSection/utils/types';
+import { categoryDetailsMap } from '@catalog/CreateFromInstanceTypes/components/SelectInstanceTypeSection/utils/utils';
 import {
   DEFAULT_INSTANCETYPE_LABEL,
   DEFAULT_PREFERENCE_LABEL,
 } from '@catalog/CreateFromInstanceTypes/utils/constants';
-import {
-  VirtualMachineClusterInstancetypeModelGroupVersionKind,
-  VirtualMachineClusterPreferenceModelGroupVersionKind,
-} from '@kubevirt-ui/kubevirt-api/console';
+import { VirtualMachineClusterPreferenceModelGroupVersionKind } from '@kubevirt-ui/kubevirt-api/console';
 import { V1beta1DataSource } from '@kubevirt-ui/kubevirt-api/containerized-data-importer/models';
 import { V1alpha2VirtualMachineClusterPreference } from '@kubevirt-ui/kubevirt-api/kubevirt';
 import ExternalLink from '@kubevirt-utils/components/ExternalLink/ExternalLink';
@@ -17,18 +19,28 @@ import TabModal from '@kubevirt-utils/components/TabModal/TabModal';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
 import { convertResourceArrayToMap } from '@kubevirt-utils/resources/shared';
 import { ANNOTATIONS } from '@kubevirt-utils/resources/template';
+import { readableSizeUnit } from '@kubevirt-utils/utils/units';
 import { K8sResourceCommon } from '@openshift-console/dynamic-plugin-sdk';
-import { Form, FormGroup, Grid, GridItem, PopoverPosition, TextArea } from '@patternfly/react-core';
+import {
+  Form,
+  FormGroup,
+  Grid,
+  GridItem,
+  PopoverPosition,
+  Select,
+  SelectOption,
+  SelectVariant,
+  TextArea,
+} from '@patternfly/react-core';
 
 import { BootableVolumeMetadata } from '../../utils/types';
-import { changeBootableVolumeMetadata, getInstanceTypesToSizesMap } from '../../utils/utils';
+import { changeBootableVolumeMetadata } from '../../utils/utils';
 
 type EditBootableVolumesModalProps = {
   dataSource: V1beta1DataSource;
   isOpen: boolean;
   onClose: () => void;
   preferences: V1alpha2VirtualMachineClusterPreference[];
-  instanceTypesNames: string[];
 };
 
 const EditBootableVolumesModal: FC<EditBootableVolumesModalProps> = ({
@@ -36,54 +48,67 @@ const EditBootableVolumesModal: FC<EditBootableVolumesModalProps> = ({
   isOpen,
   onClose,
   preferences,
-  instanceTypesNames,
 }) => {
   const { t } = useKubevirtTranslation();
+
   const preferencesNames = useMemo(
     () => Object.keys(convertResourceArrayToMap(preferences)).sort((a, b) => a.localeCompare(b)),
     [preferences],
   );
 
-  const instanceTypesToSizesMap = useMemo(
-    () => getInstanceTypesToSizesMap(instanceTypesNames),
-    [instanceTypesNames],
-  );
-
   const initialParams = useMemo(() => {
     const instanceTypeLabel =
       dataSource?.metadata?.labels?.[DEFAULT_INSTANCETYPE_LABEL]?.split('.');
+    const initialCategory = Object.entries(categoryDetailsMap).find(
+      (category) => category[1].prefix === instanceTypeLabel?.[0],
+    );
 
     return {
       preference: dataSource?.metadata?.labels?.[DEFAULT_PREFERENCE_LABEL],
-      instanceType: instanceTypeLabel?.[0],
-      size: instanceTypeLabel?.[1],
+      instanceType: initialCategory,
+      size: initialCategory && instanceTypeLabel?.[1],
       description: dataSource?.metadata?.annotations?.[ANNOTATIONS.description],
     };
   }, [dataSource]);
 
   const [preference, setPreference] = useState<string>(initialParams.preference);
-  const [instanceType, setInstanceType] = useState<string>(initialParams.instanceType);
+  const [isInstanceTypeOpen, setIsInstanceTypeOpen] = useState(false);
+  const [instanceType, setInstanceType] = useState<string>(initialParams.instanceType?.[0]);
+  const [isSizeOpen, setIsSizeOpen] = useState(false);
   const [size, setSize] = useState<string>(initialParams.size);
   const [description, setDescription] = useState<string>(initialParams.description);
 
-  const sizeOptions = useMemo(
-    () => instanceTypesToSizesMap[instanceType] || [],
-    [instanceType, instanceTypesToSizesMap],
+  // update options for 'Size' dropdown according to chosen 'instanceType' in 'Default InstanceType' dropdown
+  const { instanceTypes }: CategoryDetails = useMemo(
+    () => (instanceType ? categoryDetailsMap[instanceType] : {}),
+    [instanceType],
   );
 
-  const onInstanceTypeChange = useCallback(
-    (instanceTypeName: string) => {
-      setInstanceType(instanceTypeName);
-      setSize(instanceTypesToSizesMap[instanceTypeName][0]);
-    },
-    [instanceTypesToSizesMap],
-  );
+  const onInstanceTypeSelect = (
+    _event: ChangeEvent<HTMLSelectElement>,
+    newInstanceType: string,
+  ) => {
+    setInstanceType(newInstanceType);
+    setIsInstanceTypeOpen(false);
 
-  const onChangeVolumeParams = useCallback(() => {
+    const newCategoryObject = categoryDetailsMap[newInstanceType];
+    const newCategorySize = newCategoryObject.instanceTypes[0].label;
+    setSize(newCategorySize);
+  };
+
+  const onSizeSelect = (_event: ChangeEvent<HTMLSelectElement>, newSize: string) => {
+    setSize(newSize);
+    setIsSizeOpen(false);
+  };
+
+  const onSubmitVolumeParams = useCallback(() => {
     const preferenceLabel = preference && { [DEFAULT_PREFERENCE_LABEL]: preference };
+
+    const categoryObject = categoryDetailsMap[instanceType];
     const instanceLabel = instanceType && {
-      [DEFAULT_INSTANCETYPE_LABEL]: `${instanceType}.${size}`,
+      [DEFAULT_INSTANCETYPE_LABEL]: `${categoryObject.prefix}.${size}`,
     };
+
     const descriptionAnnotation = description?.trim()
       ? { [ANNOTATIONS.description]: description.trim() }
       : { [ANNOTATIONS.description]: undefined }; // we do want undefined here to get the annotation removed from the resource, if description not provided
@@ -109,7 +134,7 @@ const EditBootableVolumesModal: FC<EditBootableVolumesModalProps> = ({
       isOpen={isOpen}
       onClose={onClose}
       headerText={t('Edit volume metadata')}
-      onSubmit={onChangeVolumeParams()}
+      onSubmit={onSubmitVolumeParams()}
     >
       <Form>
         <FormGroup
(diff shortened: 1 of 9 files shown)


hstastna:

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
```diff
@@ -17,73 +19,99 @@ import TabModal from '@kubevirt-utils/components/TabModal/TabModal';
...
     const instanceTypeLabel =
       dataSource?.metadata?.labels?.[DEFAULT_INSTANCETYPE_LABEL]?.split('.');
 
     return {
       preference: dataSource?.metadata?.labels?.[DEFAULT_PREFERENCE_LABEL],
-      instanceType: instanceTypeLabel?.[0],
+      instanceType: () => {
+        for (const category of Object.entries(categoryDetailsMap)) {
+          if (category[1].prefix === instanceTypeLabel?.[0]) {
+            return category;
```
avivtur, line 68: nit suggestion:
`Object.entries(categoryDetailsMap).map(category => category[1].prefix === instanceTypeLabel?.[0] && category)`

avivtur (commented):
Hi :) 
looks really good, just one comment, `EditBootableVolumesModal.tsx` is getting too big, please try to break it down into smaller components so it would be easier to maintain them.

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
metalice: i think find is more suitable

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
```diff
@@ -157,24 +185,55 @@ const EditBootableVolumesModal: FC<EditBootableVolumesModalProps> = ({
...
             <FormGroup label={t('Size')}>
-              <FilterSelect
-                selected={size}
-                setSelected={setSize}
-                options={sizeOptions}
-                groupVersionKind={VirtualMachineClusterInstancetypeModelGroupVersionKind}
-                optionLabelText={t('size')}
-              />
+              <Select
+                menuAppendTo="parent"
```
metalice, line 216: please check if this doesn't cause any bugs, such as the dropdown sometimes opening up instead of down, anyway, I'm not sure u need this line

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
```diff
@@ -17,73 +19,99 @@ import TabModal from '@kubevirt-utils/components/TabModal/TabModal';
...
+  const { instanceTypes }: CategoryDetails = useMemo(
+    () => (instanceType ? categoryDetailsMap[instanceType] : {}),
+    [instanceType],
   );
 
-  const onInstanceTypeChange = useCallback(
-    (instanceTypeName: string) => {
-      setInstanceType(instanceTypeName);
-      setSize(instanceTypesToSizesMap[instanceTypeName][0]);
+  const onInstanceTypeSelect = useCallback(
```
metalice, line 90: There is no benefit here for useCallback as there are no dependencies nor hard calculations, but useCallback does make a comparison each time which adds complex

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
```diff
@@ -17,73 +19,99 @@ import TabModal from '@kubevirt-utils/components/TabModal/TabModal';
...
+      const newCategoryObject = categoryDetailsMap[newInstanceType];
+      const newCategorySize = newCategoryObject.instanceTypes[0].label;
+      setSize(newCategorySize);
     },
-    [instanceTypesToSizesMap],
+    [],
   );
 
-  const onChangeVolumeParams = useCallback(() => {
+  const onSizeSelect = useCallback((_event: ChangeEvent<HTMLSelectElement>, newSize: string) => {
```
metalice, line 102: There is no benefit here for useCallback as there are no dependencies nor hard calculations, but useCallback does make a comparison each time which adds complex

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
hstastna: Yep, now I can see I could simply use `find`.

> nit suggestion:
Object.entries(categoryDetailsMap).map(category => category[1].prefix === instanceTypeLabel?.[0] && category)

Not sure why but this didn't work for all the cases I've tested. So I better used the less optimized version that works perfectly.

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
hstastna: Originally, there were some dependencies. Once I found I don't need them, maybe I just forgot to remove useCallback after that.

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
metalice: Great just make sure that it ain't open up on some resolution on chrome and firefox. I encounter this bug with menu append

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
metalice: There isn't an issue here of less or more optimized. If u are using find u don't need && category. Find callback just need to meet the condition.

metalice: It will return the first element that meets the condition successfully

hstastna: Yep, right, I've realized that just after I sent the comment :D 
Just a note I am still working on that, want to test my improved code but I am blocked by some other bug - I cannot create a bootable volume (and I do need, with specific labels, and cannot edit labels of other existing ones), that has occurred unexpectedly just from today on our main branch..

hstastna: Tested in various browsers, different resolution and size of the browser window, works as expected.

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
hstastna: Tested in various browsers - Firefox, Chrome, Chromium, different resolutions and size of the browser window, works as expected.

hstastna: Tested in various browsers - Firefox, Chrome, Chromium, different resolutions and size of the browser window, works as expected.

hstastna:

pcbailey:
/hold

I have a few comments, but haven't finished my review.

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
```diff
@@ -17,73 +19,96 @@ import TabModal from '@kubevirt-utils/components/TabModal/TabModal';
...
+
+  const onSizeSelect = (_event: ChangeEvent<HTMLSelectElement>, newSize: string) => {
+    setSize(newSize);
+    setIsSizeOpen(false);
+  };
+
+  const onSubmitVolumeParams = useCallback(() => {
     const preferenceLabel = preference && { [DEFAULT_PREFERENCE_LABEL]: preference };
+
+    const categoryObject = categoryDetailsMap[instanceType];
```
pcbailey, line 107: Using `Object` at the end of the sentence usually doesn't communicate much useful information. It's better to just use a name that helps to identify what type of data that object holds. In this case, this object is an instance of a type, `CategoryDetails`, so it makes since to use the type name as the variable name.

pcbailey (commented):
@hstastna I didn't get to finish my review before it was merged. I will likely open a PR to make the changes suggested here.

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
```diff
@@ -17,73 +19,96 @@ import TabModal from '@kubevirt-utils/components/TabModal/TabModal';
...
       dataSource?.metadata?.labels?.[DEFAULT_INSTANCETYPE_LABEL]?.split('.');
+    const initialCategory = Object.entries(categoryDetailsMap).find(
+      (category) => category[1].prefix === instanceTypeLabel?.[0],
+    );
 
     return {
       preference: dataSource?.metadata?.labels?.[DEFAULT_PREFERENCE_LABEL],
-      instanceType: instanceTypeLabel?.[0],
-      size: instanceTypeLabel?.[1],
+      instanceType: initialCategory,
```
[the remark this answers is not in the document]
hstastna: I was already addressing this problem of naming these kinds of data to Matan. The problem is we are not sure how to call what. So each of us - devs, uses their own namings... And this feature is still evolving... And the UI is misleading with this, too, as "Default InstanceType" field in the Add volume and Edit volume metadata modals refers to only first part of the whole name of the InstanceType resource (before "."). But that was not our decision....
We should come up with some convention for us about how to name specific data related to InstanceTypes in our codebase, so we could understand the code better, at least us. And then I could make an update, of course.

IMO the core of all these problems is the design of the InstanceType concept, where 2 different data is merged in one name for InstanceType resource, while in the UI it is shown separately (which can IMO produce just more misleading feelings for our users, as soon as they realize that those 2 "things" are just one resource). It makes our lives in the UI harder. But I was told it was like that because it is easier for the API like that 🤷🏼‍♀️

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
```diff
@@ -17,73 +19,96 @@ import TabModal from '@kubevirt-utils/components/TabModal/TabModal';
...
+  const [isSizeOpen, setIsSizeOpen] = useState(false);
   const [size, setSize] = useState<string>(initialParams.size);
   const [description, setDescription] = useState<string>(initialParams.description);
 
-  const sizeOptions = useMemo(
-    () => instanceTypesToSizesMap[instanceType] || [],
-    [instanceType, instanceTypesToSizesMap],
+  // update options for 'Size' dropdown according to chosen 'instanceType' in 'Default InstanceType' dropdown
+  const { instanceTypes }: CategoryDetails = useMemo(
+    () => (instanceType ? categoryDetailsMap[instanceType] : {}),
```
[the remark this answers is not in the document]
hstastna: This is the problem of naming the data we are working with, as I've already mentioned in another comment... I'm not gonna change that till we get to some convention for names of the data we are working with, regarding InstanceTypes.

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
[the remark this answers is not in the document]
pcbailey: I definitely think more discussion should have been had around the nomenclature for this feature before it was implemented. I've been going by Fabian's documentation for the feature, here: https://github.com/fabiand/instanceTypes/blob/main/instanceTypes.md. It's the best I've seen so far in terms of explaining the terminology.

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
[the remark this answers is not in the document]
pcbailey: I understand that you don't like the naming as it is right now. That's still no excuse to use comments of self-documenting code. The naming can be updated if/when the nomenclature is clarified.

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
```diff
@@ -17,73 +19,96 @@ import TabModal from '@kubevirt-utils/components/TabModal/TabModal';
...
 
-  const instanceTypesToSizesMap = useMemo(
-    () => getInstanceTypesToSizesMap(instanceTypesNames),
-    [instanceTypesNames],
-  );
-
   const initialParams = useMemo(() => {
     const instanceTypeLabel =
       dataSource?.metadata?.labels?.[DEFAULT_INSTANCETYPE_LABEL]?.split('.');
+    const initialCategory = Object.entries(categoryDetailsMap).find(
```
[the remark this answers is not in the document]
hstastna: Looks perfect. Just a little issue in here: not sure why (I don't remember now already) but this wasn't working as expected for all the cases I was testing. There was something problematic about how the `categoryDetailsMap` is built (and IMHO it is a bit more complex than needed).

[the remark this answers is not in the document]
pcbailey: I pulled your PR and tested it myself, but I don't know what all you did to test it. `categoryDetailsMap` definitely holds a lot of information. In retrospect, I probably should have created one map for the category information and another one for the instanceTypes. I think that would help a bit. I'm open to other suggestions.

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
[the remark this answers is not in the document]
hstastna: Seems like we have different opinion on if my comment in this place is appropriate or not. I was thinking about these names and stuff for a long time, when working on this PR. And I decided to better put the comment in this place. I would go against myself and other good habits of one developer if I simply ignored the fact of our naming issue and not added that comment at all. Or if I used long ugly unreadable name like `DefaultInstanceTypeDropdownOptions`, because that would be the most appropriate to describe what is going on there 😇 With such long names everywhere in our code, we wouldn't help us much. I'll better add one readable comment, in this specific case. Do we have to make it ourselves even harder if we don't have to?

Once we get to some clear naming convention for InstanceTypes related data, the comment will become unnecessary and will be removed, together with updating related namings we use in our code.

We can change everything we want. That means we can add or remove any comment we want, too. That's nothing permanent in our code. So please take it as a temporary improvement of what we have just now. Btw, I'd like to see more comments in our code than we have now. Because there are the same naming issues also in other areas, and absence of comments in some places makes (not just) my life just harder, slows down the development... Using comments isn't bad. It's a tool. We should use the tools we have if they help us, if the actual upside of the used tool is higher than its downside. When using too many unnecessary comments - that's not helpful, of course. Unfortunately, this is obviously subjective when some comment is good to use and when not. I think there is no any clear way how to solve that universally, except hiring a different dev who would be more aligned with opinions and experience of other devs in the team 😄

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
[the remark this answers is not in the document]
hstastna: No worries, we will figure that out. And I'll try to test it again to make sure. And will create a followup PR if needed, for sure.

src/views/bootablevolumes/actions/components/EditBootableVolumesModal.tsx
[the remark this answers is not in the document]
pcbailey: We should avoid long names if we can use a shorter one that communicates the necessary information, but using a longer name that communicates that information is superior to a comment as it will clear anywhere it's used in the code and not just in that one spot. In general, if a comment is just stating what the code is doing and the code isn't something that's unusually hard to parse (like some type of non-idiomatic code or regex), the solution is to refactor, not comment.

I just looked at it again and realized this block of code could be clearer if you renamed the `instanceType` state to `category` as suggested above.

```
  const instanceTypes: InstanceTypeSizeDetails[] = useMemo(
    () => (category ? categoryDetailsMap[category].instanceTypes : {}),
    [category],
  );
```

> I'll better add one readable comment, in this specific case. Do we have to make it ourselves even harder if we don't have to?

I see you use this reasoning a lot. It's better to avoid using a comment, if possible, than use it as a crutch to avoid using practices that aid code readability like good naming, extracting logic into methods that can benefit from good naming, etc.

> We can change everything we want. That means we can add or remove any comment we want, too. That's nothing permanent in our code. So please take it as a temporary improvement of what we have just now.

It's true that we can modify the code as we see fit. We should still always opt to make the best changes possible. In this case, there were better choices than the comment as I pointed out above.

> I'd like to see more comments in our code than we have now. Because there are the same naming issues also in other areas

If there are areas that you think need comments or that suffer from naming issues, then that means the code likely needs to be refactored. It's more likely that the issue is a code smell than a scenario where a comment is the best solution. I suggest you put together a PR with the comments you feel are necessary and let the team review it. It will likely result in lots of opportunity for refactoring.

PR #1233, opened 2023-04-03
Bug 2184098: Fix size in add volume modal

avivtur:
## 📝 Description

when getting a size from the pvc we need to remove the `B` suffix more info on: https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/

## 🎥 Demo

After:

https://user-images.githubusercontent.com/67270715/229574555-2f907a8f-6258-4d92-a182-817d60f306ee.mp4


avivtur:
@metalice I believe we do need to backport this

metalice:
So please open a bug for it

PR #1234, opened 2023-04-03
Bug 2184058: Display 'Add network interface' button only once

hstastna:
## 📝 Description

**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2184058

Remove _Add network interface_ button from VM's _Network interfaces_ tab in case there are no any NICs present, according to the UX suggestions. Display the button only once - in the left up, next to the YAML switcher.

_Notes:_
Note that the solution is not 100% consistent with "empty state" but it’s more important that all the buttons in the VM details page are aligned (discussed that with the UX).

Also note that within the BZ https://bugzilla.redhat.com/show_bug.cgi?id=2176797, the title of the page _Network interfaces_ will be added and then the button will be positioned under that title.

## 🎥 Screenshots
**Before:**
_Add network interface_ button  displayed twice:
nic_before

**After:**
_Add network interface_ button  displayed once - position of the button same as in other VM's tabs:
nic_after

### DIFF
--- a/src/views/virtualmachines/details/tabs/configuration/network/copmonents/list/AutoAttachedNetworkEmptyState.tsx
+++ b/src/views/virtualmachines/details/tabs/configuration/network/copmonents/list/AutoAttachedNetworkEmptyState.tsx
@@ -1,6 +1,5 @@
 import React, { FC } from 'react';
 
-import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
 import {
   Bullseye,
@@ -11,18 +10,15 @@ import {
 } from '@patternfly/react-core';
 import { NetworkIcon } from '@patternfly/react-icons';
 
-import AddNetworkInterfaceButton from '../AddNetworkInterfaceButton';
-
 type AutoAttachedNetworkEmptyStateProps = {
-  vm: V1VirtualMachine;
   isAutoAttached: boolean;
 };
 
 const AutoAttachedNetworkEmptyState: FC<AutoAttachedNetworkEmptyStateProps> = ({
-  vm,
   isAutoAttached,
 }) => {
   const { t } = useKubevirtTranslation();
+
   return (
     <Bullseye>
       <EmptyState variant={EmptyStateVariant.small}>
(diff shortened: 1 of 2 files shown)


hstastna:

avivtur:
/lgtm

> Note that the solution is not 100% consistent with "empty state" but it’s more important that all the buttons in the VM details page are aligned (discussed that with the UX).

just out of curiosity is there any plan to make the `empty state` consistent for also VM list or other place we have duplicated buttons according to UX?

hstastna:
> just out of curiosity is there any plan to make the empty state consistent for also VM list or other place we have duplicated buttons according to UX?

If we hit such duplicated buttons somewhere else, we definitely want to make it consistent. I haven't found such duplications yet, requires more testing. If you are aware of such pages, let me know.

avivtur:
@hstastna I aware of these two
mp-list
vm-list

hstastna:
> > Yes, we need to remove unnecessary buttons from those pages you've pointed out. And according to the discussion with Yifat, we want to remove upper right buttons, not from the middle, because it's consistent with the Patternfly empty state. That's the default convention for empty state. I'm gonna do it in a followup PR. Thanks for pointing that out, Aviv! sparkle
> 
> This may be consistent with PF, but it is not consistent with console as all resources pages have the button at the upper right corner, I'm not sure we want to go in that direction

I am not sure if it is a good approach to look at console project when looking for some "source of truth", applicable to our actual one. We know there were many issues and inconsistencies...

As I've said, this is a bit problematic, as we have some specific screens and cannot be consistent with both approaches I've mentioned, at once, approaches that are different. We can be consistent only with one of them. We have to choose one to improve the screen/screens at least a bit, move towards better. Our project is still fresh and fast evolving. We may not know everything right now, like future plans etc, so we also may not understand all the UX decisions. Because this (a.k.a. the result of this PR and the followup ones that are gonna be created for MigrationPolicies and VMs empty states) was the decision made by the UX, not myself. 

We should follow the PF guidelines for sure, if possible. Of course, for specific situations we have specific UX decision that may not be aligned with PF for 100%.

Anyway, if you do think it was not a good decision, please, feel free to come up with some other suggestions for the mentioned screens, and discuss that with me together with Yifat or anyone from UX. I did this already and this was the decision, I didn't come up with any better ideas to suggest. UX are experts for the UX part of our project anyway. We are aware of the fact that the decision wasn't 100% consistent with everything. We (I and the UX) do realize that. It's probably not the final state of those screens, as I've already said, this project is still evolving and very fast...

PR #1237, opened 2023-04-05
Remove extra buttons to create VM or MigrationPolicy

hstastna:
## 📝 Description

This is a followup PR for the comment:
https://github.com/kubevirt-ui/kubevirt-plugin/pull/1234#issuecomment-1496124398

Don't display the unnecessary buttons to create VirtualMachine or MigrationPolicy in the empty state pages (no resources present), as there is the button to create the appropriate resource present also in the center of the page, which is the default for Patternfly empty state. Make this change according to the UX suggestions.

Create new `MigrationPoliciesCreateButton` component to be used where needed.
Update `MigrationPoliciesEmptyState` component: use preferred `ExternalLink` component instead of `a` in there.

Also remove unnecessary pagination for VM's empty state page, as it doesn't make sense for this case.

## 🎥 Screenshots
**Before:**
MigrationPolicies empty state - 2 buttons to create the resource present in the page (and no options to create a policy from in the middle button):
mm_before
VirtualMachines empty state - 2 buttons to create the resource present in the page (and no options to create a VM from in the middle button):
vm_before

**After**:
MigrationPolicies empty state - only 1 button to create the resource present in the page (together w all the options to create a policy from):
mig_after
VirtualMachines empty state - only 1 button to create the resource present in the page (plus options to create a VM from present in the middle button):
vm-after

---

TODO:

- [x] figure out what to do with the fact, that the button in the center of the page contains only one option to create the resource, comparing to the removed button from upper right of the page, ask the UX about that - done, missing options added to the buttons for empty states
- [x] remove unnecessary pagination for VM's empty state

### DIFF
--- a/src/utils/components/ExternalLink/ExternalLink.tsx
+++ b/src/utils/components/ExternalLink/ExternalLink.tsx
@@ -1,31 +1,32 @@
-import * as React from 'react';
-import classNames from 'classnames';
+import React, { FC } from 'react';
+
+import { ExternalLinkAltIcon } from '@patternfly/react-icons';
 
 type ExternalLinkProps = {
   href: string;
   text?: React.ReactNode;
-  additionalClassName?: string;
+  className?: string;
   dataTestID?: string;
   stopPropagation?: boolean;
 };
 
-const ExternalLink: React.FC<ExternalLinkProps> = ({
+const ExternalLink: FC<ExternalLinkProps> = ({
   children,
   href,
   text,
-  additionalClassName = '',
+  className = '',
   dataTestID,
   stopPropagation,
 }) => (
   <a
-    className={classNames('co-external-link', additionalClassName)}
+    className={className}
     href={href}
     target="_blank"
     rel="noopener noreferrer"
     data-test-id={dataTestID}
     {...(stopPropagation ? { onClick: (e) => e.stopPropagation() } : {})}
   >
-    {children || text}
+    {children || text} <ExternalLinkAltIcon />
   </a>
 );
 
--- a/src/views/migrationpolicies/list/components/MigrationPoliciesCreateButton/MigrationPoliciesCreateButton.tsx
+++ b/src/views/migrationpolicies/list/components/MigrationPoliciesCreateButton/MigrationPoliciesCreateButton.tsx
@@ -1,0 +1,34 @@
+import React, { FC } from 'react';
+import { useHistory } from 'react-router-dom';
+
+import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
+import { ListPageCreateDropdown } from '@openshift-console/dynamic-plugin-sdk';
+
+import { createItems, migrationPoliciesPageBaseURL } from '../../utils/constants';
+
+type MigrationPoliciesCreateButtonProps = {
+  kind: string;
+};
+
+const MigrationPoliciesCreateButton: FC<MigrationPoliciesCreateButtonProps> = ({ kind }) => {
+  const { t } = useKubevirtTranslation();
+  const history = useHistory();
+
+  const onCreate = (type: string) => {
+    return type === 'form'
+      ? history.push(`${migrationPoliciesPageBaseURL}/form`)
+      : history.push(`${migrationPoliciesPageBaseURL}/~new`);
+  };
+
+  return (
+    <ListPageCreateDropdown
+      items={createItems}
+      onClick={onCreate}
+      createAccessReview={{ groupVersionKind: kind }}
+    >
+      {t('Create MigrationPolicy')}
+    </ListPageCreateDropdown>
+  );
+};
+
+export default MigrationPoliciesCreateButton;
--- a/src/views/migrationpolicies/list/utils/constants.ts
+++ b/src/views/migrationpolicies/list/utils/constants.ts
@@ -1,3 +1,9 @@
 import { MigrationPolicyModelRef } from '@kubevirt-ui/kubevirt-api/console';
+import { t } from '@kubevirt-utils/hooks/useKubevirtTranslation';
 
 export const migrationPoliciesPageBaseURL = `/k8s/cluster/${MigrationPolicyModelRef}`;
+
+export const createItems = {
+  form: t('With form'),
+  yaml: t('With YAML'),
+};
--- a/src/views/virtualmachines/list/components/VirtualMachinesCreateButton/VirtualMachinesCreateButton.tsx
+++ b/src/views/virtualmachines/list/components/VirtualMachinesCreateButton/VirtualMachinesCreateButton.tsx
@@ -1,0 +1,64 @@
+import React, { FC, useCallback, useMemo } from 'react';
+import { useHistory } from 'react-router-dom';
+
+import { VirtualMachineModelRef } from '@kubevirt-ui/kubevirt-api/console';
+import DeveloperPreviewLabel from '@kubevirt-utils/components/DeveloperPreviewLabel/DeveloperPreviewLabel';
+import { DEFAULT_NAMESPACE } from '@kubevirt-utils/constants/constants';
+import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
+import { ListPageCreateDropdown } from '@openshift-console/dynamic-plugin-sdk';
+
+type VirtualMachinesCreateButtonProps = {
+  namespace: string;
+  buttonText?: string;
+};
+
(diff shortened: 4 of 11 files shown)


src/views/migrationpolicies/list/components/MigrationPoliciesCreateButton/MigrationPoliciesCreateButton.tsx
```diff
@@ -0,0 +1,39 @@
+import React, { FC } from 'react';
+import { useHistory } from 'react-router-dom';
+
+import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
+import { ListPageCreateDropdown } from '@openshift-console/dynamic-plugin-sdk';
+
+import { migrationPoliciesPageBaseURL } from '../../utils/constants';
+
+type MigrationPoliciesCreateButtonProps = {
+  kind: string;
```
avivtur, line 10: can you remove this prop and import MigrationPolicyModelRef instead? it will make it a bit more clear what is this actually, and it will allow you to use this with no prop which will be easier if needed elsewhere :)

src/utils/components/ExternalLink/ExternalLink.tsx
```diff
@@ -18,14 +19,14 @@ const ExternalLink: React.FC<ExternalLinkProps> = ({
   stopPropagation,
 }) => (
   <a
-    className={classNames('co-external-link', additionalClassName)}
+    className={additionalClassName}
```
avivtur, line 22: +1 to the `co-...` classname removal, can you also rename `additionalClassName` to be `className`?

src/utils/components/ExternalLink/ExternalLink.tsx
```diff
@@ -1,5 +1,6 @@
 import * as React from 'react';
```
avivtur, line 1: can you please get rid of the * here?

avivtur (commented):
This looks nice, added some suggestions :)

src/views/virtualmachines/list/components/VirtualMachinesCreateButton/VirtualMachinesCreateButton.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+import { VirtualMachineModelRef } from '@kubevirt-ui/kubevirt-api/console';
+import DeveloperPreviewLabel from '@kubevirt-utils/components/DeveloperPreviewLabel/DeveloperPreviewLabel';
+import { DEFAULT_NAMESPACE } from '@kubevirt-utils/constants/constants';
+import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
+import { ListPageCreateDropdown } from '@openshift-console/dynamic-plugin-sdk';
+
+type VirtualMachinesCreateButtonProps = {
+  namespace: string;
+  buttonText?: string;
```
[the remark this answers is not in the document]
hstastna: For such decisions, we have the UX folks. I just wanted to keep the original naming of the buttons in specific situations (such as presence of some Vms vs. no VMs at all) in this PR.

src/views/migrationpolicies/list/components/MigrationPoliciesCreateButton/MigrationPoliciesCreateButton.tsx
hstastna: I am not sure if this is a good idea, because if you check the related parent components, they also take `kind` as a prop. If it was so simple and safe to do it, I am sure the parent components wouldn't get this at all, but used what you suggest.

src/utils/components/ExternalLink/ExternalLink.tsx
hstastna: Yep, will do it. Thanks!
Just a note, maybe you've already realized, too, that we have sooo much stuff like this across whole codebase and I am not sure why we haven't imported only what we need from the beginning. Our whole codebase deserves refactoring, and not only regarding this, but we also have so much repeating code everywhere :D

upalatucci: we had an issue with typescript at the beginning

src/views/migrationpolicies/list/components/MigrationPoliciesCreateButton/MigrationPoliciesCreateButton.tsx
metalice: I looked in the code, and the parent is getting it free from the console, as it's a menu item.
Getting the kind its not a network request or IO action that takes time. It's a simple function. From the name of the component MigrationPoliciesCreateButton, I can see it's not a generic component, and it's dedicated to migration policy, u can emit the prop and use it as Aviv suggested or leave it as it is. There isn't a big difference here in this case, it simply makes it a bit cleaner if u getting it as Aviv suggested because the component doesn't depend on anything, but again there isnt a much difference.

src/utils/components/ExternalLink/ExternalLink.tsx
```diff
@@ -1,31 +1,32 @@
...
-    className={classNames('co-external-link', additionalClassName)}
+    className={className}
     href={href}
     target="_blank"
     rel="noopener noreferrer"
     data-test-id={dataTestID}
     {...(stopPropagation ? { onClick: (e) => e.stopPropagation() } : {})}
   >
-    {children || text}
+    {children || text} <ExternalLinkAltIcon />
```
metalice, line 29: This is a generic file we are using across our app does everything looks the same after the change? 
co-external-link === ExternalLinkAltIcon , from an appearance perspective?

src/views/migrationpolicies/list/components/MigrationPoliciesCreateButton/MigrationPoliciesCreateButton.tsx
hstastna: So then if there isn't much difference, shouldn't we omit getting kind also in parent components?

src/utils/components/ExternalLink/ExternalLink.tsx
hstastna: With this change, the icon has finally the correct size, and everywhere. It was too small and everywhere, according to the designs we have.

src/views/migrationpolicies/list/components/MigrationPoliciesCreateButton/MigrationPoliciesCreateButton.tsx
metalice: Actually, I would leave it as it is.. u did it ok; I just wanted to let you both know a bit more information pros and cons of the suggest change.

src/views/migrationpolicies/list/components/MigrationPoliciesCreateButton/MigrationPoliciesCreateButton.tsx
```diff
@@ -0,0 +1,39 @@
...
+};
+
+const MigrationPoliciesCreateButton: FC<MigrationPoliciesCreateButtonProps> = ({ kind }) => {
+  const { t } = useKubevirtTranslation();
+  const history = useHistory();
+
+  const createItems = {
```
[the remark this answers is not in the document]
hstastna: done

src/views/migrationpolicies/list/utils/constants.ts
```diff
@@ -1,3 +1,10 @@
 import { MigrationPolicyModelRef } from '@kubevirt-ui/kubevirt-api/console';
 
 export const migrationPoliciesPageBaseURL = `/k8s/cluster/${MigrationPolicyModelRef}`;
+
+// t('With form')
```
metalice, line 5: you can import t and avoid using comments for translation

hstastna: I am confused. Some people prefer this, some that.. We should decide what is preferred form to use t function in our codebase and we all should follow it.

metalice: It's not preference. At the begining this was the only option using comment. Now for not so short time we are using t. Give us more control on our code. U already used it sever times in utils files I'm guessing this is no exception.

metalice: https://github.com/kubevirt-ui/kubevirt-plugin/pull/901 - nothing new :)

hstastna: Well, I often experience very very different opinions on one topic, depending on who reviews the PR.. Sometimes it is just impossible to make everybody happy.

hstastna:

PR #1238, opened 2023-04-05
Bug 2183491: Truncate PVC name in Add volume modal

hstastna:
## 📝 Description

**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2183491

Truncate the PVC name options for the _PVC name_ drop down in _Add volume_ modal (when selecting _Use existing volume_ radio button) in case when the name is too long according to the width of the modal. Display the whole name in the popover displayed on mouseover.

Note that the decision to make this change to fix the problem of not displaying the modal nicely when PVC name is too long, was made by the UX, of course.

## 🎥 Screenshots
**Before:**
trun_before

**After:**
trun_after


hstastna:

PR #1245, opened 2023-04-12
Bug 2027959: Stop with grace period

upalatucci:
## 📝 Description

Modal for stop with grace period

## 🎥 Demo

Screenshot from 2023-04-12 15-44-57
Screenshot from 2023-04-12 15-44-53


pcbailey:
Why does the modal say "Are you sure you want to **_delete_**..."?

PR #1247, opened 2023-04-13
Fix missing CPU memory info in the instance type VM create details section

avivtur:
## 📝 Description

This bug is happening because we rely on 4 constant serieses instead of reading data from the IT resource.

## 🎥 Demo
Before:
CPU-memory-exists-b4
After:
CPU-memory-exists

### DIFF
--- a/src/utils/hooks/useKubevirtTranslation.ts
+++ b/src/utils/hooks/useKubevirtTranslation.ts
@@ -1,4 +1,5 @@
 import { getI18n, useTranslation } from 'react-i18next';
+import { TOptions } from 'i18next';
 
 /**
  * A Hook for using the i18n translation.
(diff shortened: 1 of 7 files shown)


src/utils/hooks/useKubevirtTranslation.ts
```diff
@@ -10,4 +10,4 @@ export const useKubevirtTranslation = () => useTranslation('plugin__kubevirt-plu
  * @param value string to translate
  */
 // skipcq: JS-C1002
-export const t = (value: string) => getI18n().t(value, { ns: 'plugin__kubevirt-plugin' });
+export const t: TFunction<'plugin__kubevirt-plugin', undefined> = getI18n().t;
```
metalice, line 13: why remove the namespace? what is the default name space?

avivtur: the namespace is declared in the type, I changed it so we can also send translations like these:
`"{{cpu}} CPUS", { cpu: some_value }"`

PR #1249, opened 2023-04-13
Bug 2186592: Display titles in VMI Disks tab in correct size

hstastna:
## 📝 Description

**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2186592

Display VirtualMachineInstance's "Disks" and "File systems" in _Disks_ tab smaller as they were, so they are now more consistent with titles in other tabs.

Additionally, place both "?" icons (Patternfly React `HelpIcon` component) to their correct positions - to the left and next to the titles, and make them displayed enabled, not greyed out and disabled (as they were before this change), so the user can see them easily and click on them.

## 🎥 Screenshots
**Before:**
vmi_before

**After:**

vmi_after


hstastna:

upalatucci (approved):

src/views/virtualmachinesinstance/details/tabs/disks/VirtualMachinesInstancePageDisksTab.scss
```diff
@@ -0,0 +1,5 @@
+.VirtualMachinesInstancePageDisksTab {
+  td#drive {
```
upalatucci, line 2: Why this?

hstastna: This was already there, I just renamed the file and removed unnecessary parts of it. However, this wasn't unnecessary, it does the following: disks -> Disks. Not sure why it was done this way, but I've better left it like that. Or we could ask Matan why...

PR #1257, opened 2023-04-17
Bug 2183915: "No data available" shows on Virtualization overview metrics chart

avivtur:
## 📝 Description

Giving the charts an arbitrary timespan of one week is not getting any data in namespaces we just created new VMs.
useTimespan hook introduced to get all VMs and check if the newest VM is alive less than a day, if so the timespan changes to one day instead of one week, otherwise the timespan will remain one week.

## 🎥 Demo

Before:
overview-graphs-b4

After:
overview-graphs


src/views/clusteroverview/OverviewTab/metric-charts-card/utils/hooks/useTimespan.ts
```diff
@@ -0,0 +1,27 @@
+import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import DurationOption from '@kubevirt-utils/components/DurationOption/DurationOption';
+import { VirtualMachineModelGroupVersionKind } from '@kubevirt-utils/models';
+import { useK8sWatchResource } from '@openshift-console/dynamic-plugin-sdk';
+
+const useTimespan = (namespace?: string): number => {
+  const [vms] = useK8sWatchResource<V1VirtualMachine[]>({
+    groupVersionKind: VirtualMachineModelGroupVersionKind,
+    isList: true,
+    namespace,
```
upalatucci, line 10: if `namespace` is `null`, here we'll fetch all vms in the cluster. 
Can this page be reached by non-priv user?

avivtur: Yes this page can be reached by non-admin, I know this logic isn't optimal, but deciding on 7 days as an arbitrary timespan will get buggy for namespaces that just created vms.
We need to discuss this issue with UX to think about redesigning this part to maybe not render to users without `list` permissions

avivtur: maybe we should add a time range input, and this would be the best solution IMO
@upalatucci @metalice @pcbailey @hstastna 
I would like your opinions as well on this subject.
/wip

metalice: U can simply default it to a day or an hour.. Tell us the issue I'm sure they will agree

avivtur: That's also a valid option, will discuss this with them as well, thanks :) @metalice

avivtur: Discussed with Ronen, we keep the 7 days timespan and changing the messages a bit

PR #1261, opened 2023-04-18
Bug 2176746: Distinguish templates with same display name

upalatucci:
## 📝 Description

Add `template.metadata.name` after the display name annotation for the templates with the same display name.

## 🎥 Demo

**Before**
image

**After**
Screenshot from 2023-04-19 17-21-44


avivtur:
I have a few things that bother me with this fix:
1. Concating the display name with the resource name - it creates a long and not very nice name column, I suggest using only the resource name (as we do on Templates view) OR if the display name is too important to give up on, I'd add a display name column as 2nd column from the right and the name as the 1st column.

2. Using the validation on the display name - IMO this extra logic is not really needed since we can still differ between resources with their names and it can be easily worked around it with CLI (also verifying a label to be unique is not something we should do in general :) )

PR #1263, opened 2023-04-18
Bug 2176797: Move YAML switcher out of VM tabs

hstastna:
## 📝 Description

**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2176797

Related design doc:
https://docs.google.com/document/d/1VOjS0M-8YF_pefXaLWcqFYtoJGWFAQdSOJ3CsxnpY7M/edit#heading=h.ozvp0t9knadz

Move the YAML switcher out of the VM tabs/pages, display it to the left of _Actions_ menu and only for relevant pages, that contained the switcher also before this change. Make this change according to the recent design changes.

Add missing titles to the following VM tabs/pages: _Events_, _Metrics_, _Network interfaces_, _Snapshots_.

Make some css/styling updates for those pages to achieve more consistent look, proper aligning of the components in the page to be displayed as expected in the design doc:
- improve left margin of the _Diagnostics_ page - for sub-titles and filtering
- update margin/padding of the sidebar editor in _Disks_ page
  - same for _NICs_ tab
- for _Metrics_ tab, fix incorrect position of "Time range" text
- change empty state message in _Snapshots_ tab to "No snapshots found", also add Patternfly `Search` icon above the text, center align
- for _Scripts_ tab, make the sub-titles bigger, remove unnecessary "Scripts" title (as also other pages like this one don't have one main title, like in _Disks_ or _Diagnostics_ tab).

Rename "Scheduling and resources requirements" title of _Scheduling_ tab to correct "Scheduling and resource requirements", also rename "VirtualMachine Details" to correct "VirtualMachine details".

---

_Notes:_
I had to make a small change in `EnvironmentFormTitle` component - to add one optional prop, to keep the original design of Environment tab displayed when customizing VM. This change is temporary, to let the component know to display the switcher or not - as it was originally part of this component used in more places. Feel free to add any suggestions to solve this issue in a different way.

In addition, I renamed the "copmonents" dir to correct "components" in _src/views/virtualmachines/details/tabs/configuration/network/_.

I also had to fix failing cypress test, when "VirtualMachine details" breadcrumb was not found in case I manually haven't scrolled up in the page, when running the appropriate test.

---

## 🎥 Screenshots
The following screenshots are displayed as pairs - before and after, for better visual comparison:

VM _Details_ tab:
d_before
d_after

VM _Diagnostics_ tab (_Diagnostic_ was renamed to _Diagnostics_ in some other PR in the meantime):
diag_before
diag_after

VM _Disks_ tab:
disks_before
disks_after

VM _Environment_ tab:
env_before
env_after

VM _Events_ tab:
events_before
events_after

VM _Metrics_ tab:
metrics_before
metrics_after

_Network interfaces_ (a.k.a NICs) tab:
nics_before
nics_after

VM _Scheduling_ tab:
sched_before
sched_after

VM _Scripts_ tab:
cloud_before
scriptss-aafterr

VM _Snapshots_ tab:
snap_before
no_snapshots


hstastna:

src/views/virtualmachines/details/tabs/configuration/scheduling/components/SchedulingSection/SchedulingSection.tsx
```diff
@@ -50,12 +49,7 @@ const SchedulingSection: React.FC<SchedulingSectionProps> = ({ vm, pathname }) =
         <LinkIcon size="sm" />
       </a>
       <Title headingLevel="h2" className="co-section-heading">
-        <Flex>
-          <FlexItem>{t('Scheduling and resources requirements')}</FlexItem>
-          <FlexItem>
-            <SidebarEditorSwitch />
-          </FlexItem>
-        </Flex>
+        {t('Scheduling and resources requirements')}
```
pcbailey, line 52: I believe this should be 'resource' instead of 'resources'.

hstastna:
> Have you considered just disabling the switch instead of removing it in the tabs that do not have the yaml editor?

I think I don't understand this question. If you check the related design doc, you will find that the decision was to move the switch to some other place.

src/views/virtualmachines/details/tabs/configuration/scheduling/components/SchedulingSection/SchedulingSection.tsx
hstastna: It was always "resources". See the original line of the code just above this change. Also see the VM Template's Scheduling tab. Thanks.

avivtur:
Hi @hstastna thank you for this PR looks good :) 
Is the customize VM flow also affected? If so, can you please update the screenshots for there too?

hstastna:
> Hi @hstastna thank you for this PR looks good :) Is the customize VM flow also affected? If so, can you please update the screenshots for there too?

Hi @avivtur. The customize flow is going to be changed, too, but not in this PR. In this PR only VM tabs. This is why I also made a small change in `EnvironmentFormTitle` component, as I've already written in the description of this PR, to keep the customize flow same as it is and not to mix it with VM pages. Otherwise the PR would be even bigger and less reviewable than it is now.

src/views/virtualmachines/details/tabs/snapshots/components/list/SnapshotList.tsx
```diff
@@ -39,6 +44,14 @@ const SnapshotsList: React.FC<SnapshotsListProps> = ({
         columns={columns}
         Row={SnapshotRow}
         rowData={{ restores: restoresMap, isVMRunning }}
+        NoDataEmptyMsg={() => (
+          <>
+            <Bullseye>
+              <SearchIcon />
+            </Bullseye>
+            <Bullseye>{t('No snapshots found')}</Bullseye>
+          </>
```
avivtur, line 53: consider using:
```
<EmptyState variant={EmptyStateVariant.xs}>
  <EmptyStateIcon icon={SearchIcon} />
  <EmptyStateBody>
    {t('No snapshots found')}
  </EmptyStateBody>
</EmptyState>
```

src/views/virtualmachines/details/tabs/configuration/scheduling/components/SchedulingSection/SchedulingSection.tsx
avivtur: I believe he meant that it's not right, even if it used to be like that

src/views/virtualmachines/utils/constants.ts
```diff
@@ -27,3 +27,8 @@ export const booleanTextIds = {
 };
 
 export const NETWORK = 'network';
+
+export enum VM_TABS_WITH_YAML {
+  DETAILS = '/details',
+  CONFIGURATIONS = '/configurations',
+}
```
avivtur, line 34: export const vmTabsWithYAML = [
  VirtualMachineDetailsTab.Configurations,
  VirtualMachineDetailsTab.Details,
];

src/views/virtualmachines/details/VirtualMachineNavPageTitle.tsx
```diff
@@ -34,6 +35,9 @@ const VirtualMachineNavPageTitle: React.FC<VirtualMachineNavPageTitleProps> = ({
 
   const StatusIcon = getVMStatusIcon(vm?.status?.printableStatus);
 
+  const vmTabNameRegExp = new RegExp(`(?<=\/kubevirt.io~v1~VirtualMachine\/${name}).[a-z]*[^\/]`);
+  const vmTabName = window.location.pathname.match(vmTabNameRegExp)?.[0];
```
avivtur, line 39: I'd try to avoid RegExp as they are not clear and hard to maintain, try to see if this works: (please note also change for VM_TABS_WITH_YAML):
 ```
const vmTabName =  vmTabsWithYAML.find((tab) => window.location.pathname.includes(`/${tab}`));
going to line 58
now you only need 
vmTabName && ...
```

File: src/views/virtualmachines/details/tabs/snapshots/components/list/SnapshotList.tsx (hunk 2 above)
Comment (hstastna) [reply src/views/virtualmachines/details/tabs/snapshots/components/list/SnapshotList.tsx:53]: Great! That's what I needed to figure out not to use Bullseye twice.

File: src/views/virtualmachines/details/VirtualMachineNavPageTitle.tsx (hunk 4 above)
Comment (hstastna) [reply src/views/virtualmachines/details/VirtualMachineNavPageTitle.tsx:39]: Using `vmTabsWithYAML.find((tab) => window.location.pathname.includes(`/${tab}`))` is not sufficient. We need to check exact place of the url, to get the proper info. Or try to create some VM called "configuration" and you'll see your suggestion stops working as expected immediately.

Comment (avivtur) [reply src/views/virtualmachines/details/VirtualMachineNavPageTitle.tsx:39]: and if you use name as well won't it work? also for VM called configuration,
vmTabsWithYAML.find((tab) => window.location.pathname.includes(`/${name}/${tab}`));

Comment (hstastna) [reply src/views/virtualmachines/details/VirtualMachineNavPageTitle.tsx:39]: Not sure if that was sufficient enough.

File: src/views/virtualmachines/details/tabs/configuration/scheduling/components/SchedulingSection/SchedulingSection.tsx (hunk 1 above)
Comment (pcbailey) [reply src/views/virtualmachines/details/tabs/configuration/scheduling/components/SchedulingSection/SchedulingSection.tsx:52]: Aviv is correct. In this case, making 'resource' plural isn't proper English.

**Comment by upalatucci** [comment]
> > Have you considered just disabling the switch instead of removing it in the tabs that do not have the yaml editor?
> 
> I think I don't understand this question. If you check the related design doc, you will find that the decision was to move the switch to some other place.

I see that you hide the switch when the user is not in a tab where the editor is supported.
I was just asking if you guys already explored the possibility to disable the switch instead of hiding it in this case.
As this was something on the table when we were discussing the switch position in the past

**Comment by hstastna** [comment]
> I see that you hide the switch when the user is not in a tab where the editor is supported. I was just asking if you guys already explored the possibility to disable the switch instead of hiding it in this case. As this was something on the table when we were discussing the switch position in the past

I don't see any benefit of just disabling the switcher (no any benefit for both the user and us - the devs). We'd still have to carefully check in which tab we actually are.
I'm not aware of all the discussions taken place around this UX decision, but most of the discussions I've been involved regarding the YAML switcher, was about where exactly to move it. And then there was a decision to simply not display it if not relevant, and I think I understand that because it leads to displaying less stuff in the UI, better readability, better UX, IMHO.

File: src/views/virtualmachines/details/VirtualMachineNavPageTitle.tsx (hunk 4 above)
Comment (pcbailey) [reply src/views/virtualmachines/details/VirtualMachineNavPageTitle.tsx:39]: I agree with Aviv. We should try to avoid using regexes where possible. If his suggestion might work, then it's worth trying.

Comment (avivtur) [reply src/views/virtualmachines/details/VirtualMachineNavPageTitle.tsx:39]: If you have a problem and you solve it wite regex, now you have 2 problems, that's my way of thinking, And it's not because I don't know regex, it's because what I said about they are not easy to maintain or understand.

Comment (hstastna) [reply src/views/virtualmachines/details/VirtualMachineNavPageTitle.tsx:39]: Thank you for sharing your thoughts on this. I am taking note of it.
Seems that I should update the PR using solution that doesn't check the url at all, which would be the best anyway.

File: src/views/virtualmachines/details/tabs/configuration/scheduling/components/SchedulingSection/SchedulingSection.tsx (hunk 1 above)
Comment (apinnick) [reply src/views/virtualmachines/details/tabs/configuration/scheduling/components/SchedulingSection/SchedulingSection.tsx:52]: It's "resource requirements" in this context because "resource" is an adjective, modifying the noun "requirements". Adjectives do not have singular or plural forms. 

Examples: blue rooms, wide streets, restaurant receipts, lunch bags, workplace policies, human resources.

**Comment by apinnick** [comment]
This PR will affect docs. I will do the docs update myself.

File: src/views/virtualmachines/details/tabs/configuration/scheduling/components/SchedulingSection/SchedulingSection.tsx (hunk 1 above)
Comment (hstastna) [reply src/views/virtualmachines/details/tabs/configuration/scheduling/components/SchedulingSection/SchedulingSection.tsx:52]: Thank you very much, Avital!

File: src/views/virtualmachines/details/tabs/configuration/scripts/scripts-tab.scss [hunk 5]
```diff
@@ -4,4 +4,17 @@
...
+      font-family: var(--pf-global--FontFamily--heading--sans-serif);
+      font-size: var(--pf-global--FontSize--xl);
+      font-weight: var(--pf-global--FontWeight--normal);
+    }
+  }
+
+  .pf-c-divider {
+    margin-left: -1.5rem;
+    width: 102.5%;
+  }
```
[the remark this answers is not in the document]
hstastna: Yes. All dividers of amount 1 that are children of  components with `vm-scripts-tab` class.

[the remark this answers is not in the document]
upalatucci: I don't think we should have that.
In my opinion the current dividers are great and with this style they seems weird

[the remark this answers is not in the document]
hstastna: We need it to achieve the expected look according to the design doc I've received.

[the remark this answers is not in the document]
upalatucci: I've reviewed the doc: https://docs.google.com/document/d/1VOjS0M-8YF_pefXaLWcqFYtoJGWFAQdSOJ3CsxnpY7M/edit#heading=h.ozvp0t9knadz

I'm okay with the fact that the divider should take all the width as in the environment tab. I've added a max-width to that page just to not have the title and texts shifting between editor on/off.
But what you are trying to do here is to stretch the divider on the left and I don't think it's what Yifat requested.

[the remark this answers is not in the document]
hstastna: Yes, she did.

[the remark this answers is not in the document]
upalatucci: Yes, I've seen that and I'm saying that for `dividers length` I'm pretty sure she means the width on the right. Currently, the content of that page is limited approximately by `700px` and the divider does not take the whole page width for this reason.

As we can see from her design of the script tab, the divider is not shifted to the left with respect to the current implementation and takes all the width until the end without any limitations. 

Disks and Diagnostics tabs have tables and those lines are to limit rows. They are not Dividers.

upalatucci:
> I don't see any benefit of just disabling the switcher (no any benefit for both the user and us - the devs). We'd still have to carefully check in which tab we actually are.

The benefit is that the user would notice the switch even if it's in another tab. So it will know that there is the functionality but it's not enabled.
Yes sure we'd still have to carefully check in which tab we actually are as you do in this pr.

But if you guys do not want that, it's okay. This was the discussion if you are curious.

src/views/virtualmachines/details/tabs/configuration/scripts/scripts-tab.scss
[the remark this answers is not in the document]
hstastna: > Yes, I've seen that and I'm saying that for dividers length I'm pretty sure she means the width on the right. 

The width is usually is related to the content of the page. If that takes just a small part of the whole page (as in our case), the divider simply is shorter. You can see that also in other pages.

Maybe it could have been a bit longer than I've set (just a bit more than 100%, to keep the original length on the right because of moving it to the left, however the exact expected length wasn't explicitly specified), but the main issue there was that the divider wasn't long enough in the left. PTAL at Disks or Diagnostics tabs again to see where the dividers start.

I can eventually ask Yifat about the exact length of the divider and to force it to be even longer despite its default behavior, but we definitely need to start it from the left - same as in other pages with more subsections I've mentioned earlier, for consistency reasons.

> As we can see from her design of the script tab, the divider is not shifted to the left with respect to the current implementation and takes all the width until the end without any limitations.

Thanks for mentioning it. Unfortunately, not everything is always properly updated and displayed in the design docs, as these things are changing often. This is why I talk to Yifat on a daily basis so I know what to do.

> Disks and Diagnostics tabs have tables and those lines are to limit rows. They are not Dividers.

From the visual point of view, it's the same component.

> PS: Diagnostics does not have any padding like what we have on other pages.

Not sure what exactly you mean by this but that page has more issues that should be fixed (followup PR), according to how the components are displayed in that page.

upalatucci:
> > The benefit is that the user would notice the switch even if it's in another tab. So it will know that there is the functionality but it's not enabled. Yes sure we'd still have to carefully check in which tab we actually are as you do in this pr.
> > But if you guys do not want that, it's okay. This was the discussion if you are curious.
> 
> Yes, probably the user would notice the switcher. However, there is an effort to declutter the UI a bit, which means to better not display what isn't needed or super useful for the user. So it was decided not to display the switcher in the pages that are not relevant for it. Please, reach Ronen if you have any concerns, complaints or suggestions about the decision being made. It was not an easy decision about where to place the YAML switcher, and also about not displaying vs. disabling it. Thanks.

okay thanks. My simple question was 'do you discussed about that? It was an option?'. So happy to hear that was on the table. I don't have any complain

hstastna:
> okay thanks. My simple question was 'do you discussed about that? It was an option?'. So happy to hear that was on the table. I don't have any complain

There were more options and changes on the table. Anyway, if you'll try to think a bit more about your suggestion, you will come to the point where you'll realize that it's hard to clearly distinguish between disabled switcher and turned off switcher, as both look grey. How would you do it to prevent any confusion of the user? Wouldn't be better to simply not display it (which, in addition, makes the UI a bit cleaner and simpler)? WDYT?

PR #1270, opened 2023-04-21
Bug 2187971: guard for template parameter on disk moda

upalatucci:
## 📝 Description

No UI diff.

Do not use `vm.metadata.name` if contains a parameter to create the DV NAME. Like ${NAME}
Disk name is provided by the user so no need to guard on that.

### DIFF
--- a/src/utils/components/DiskModal/DiskModal.tsx
+++ b/src/utils/components/DiskModal/DiskModal.tsx
@@ -12,6 +12,7 @@ import {
   getDisks,
   getVolumes,
 } from '@kubevirt-utils/resources/vm';
+import { getRandomChars } from '@kubevirt-utils/utils/utils';
 import { Form } from '@patternfly/react-core';
 
 import AccessMode from './DiskFormFields/AccessMode';
@@ -37,6 +38,7 @@ import {
   getDiskFromState,
   getPersistentVolumeClaimHotplugPromise,
   getVolumeFromState,
+  nameWithoutParameter,
   produceVMDisks,
   requiresDataVolume,
 } from './utils/helpers';
@@ -87,11 +89,11 @@ const DiskModal: React.FC<DiskModalProps> = ({
         );
       }
       if (diskState.diskSource === sourceTypes.UPLOAD) {
-        return getPersistentVolumeClaimHotplugPromise(
-          vmObj,
+        const pvcName = nameWithoutParameter(
           `${vm?.metadata?.name}-${diskState.diskName}`,
-          resultDisk,
+          `${diskState.diskName}-${getRandomChars()}`,
         );
+        return getPersistentVolumeClaimHotplugPromise(vmObj, pvcName, resultDisk);
       }
       const resultDataVolume = getDataVolumeFromState({
         vm: vmObj,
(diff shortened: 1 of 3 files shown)


src/utils/components/DiskModal/DiskModal.tsx
```diff
@@ -87,11 +88,12 @@ const DiskModal: React.FC<DiskModalProps> = ({
         );
       }
       if (diskState.diskSource === sourceTypes.UPLOAD) {
-        return getPersistentVolumeClaimHotplugPromise(
-          vmObj,
-          `${vm?.metadata?.name}-${diskState.diskName}`,
-          resultDisk,
-        );
+        const isNameTemplateParameter = /[${}]/.test(vm?.metadata?.name);
```
hstastna, line 91: Are you sure this kind of check is gonna work well everywhere we use DiskModal? I can see we deal with disks when dealing with VMs and also Templates.. And when you get here data of a template that has been already processed, there's no any chars like `${}` in the name even when the name was generated according to the `NAME` param.

Buut... maybe this is not a problem? I am not sure because this PR lacks more info about the problem in its description. Maybe the problem is just we don't wanna get `${}` chars into the `pvcName`?

src/utils/components/DiskModal/utils/helpers.ts
```diff
@@ -136,7 +136,14 @@ export const getDataVolumeFromState = ({
     resultVolume?.persistentVolumeClaim?.claimName ||
     `${vm?.metadata?.name}-${diskState.diskName}`;
 
-  dataVolume.metadata.name = dvName;
+  const isTemplateParameter = /[${}]/.test(dvName);
+
+  if (isTemplateParameter) {
```
hstastna, line 141: Maybe the following could work, too?
```
dataVolume.metadata.name = isTemplateParameter
  ? `${diskState.diskName}-${getRandomChars()}`
  : dvName;
```

hstastna (commented):
Left some comment, otherwise 👍🏽 :) Thanks!

src/utils/components/DiskModal/DiskModal.tsx
upalatucci: As you can see from the bugzilla ticket, there is also a screenshot of the problem.
The description is explanatory to me. The Diskmodal is trying to create a disk with `${NAME}`. `$`, `{` and `}` cannot be used as names so whatever name it is, if contains those characters, the name contains a parameter. 

The template at this point is not processed.

src/utils/components/DiskModal/utils/helpers.ts
upalatucci: Yes we can do that

upalatucci: fixed

src/utils/components/DiskModal/DiskModal.tsx
hstastna: Thanks for the explanation, Ugo. So you are sure it's gonna work well for both VM and Templates Disks pages?

src/utils/components/DiskModal/utils/helpers.ts
```diff
@@ -136,7 +136,11 @@ export const getDataVolumeFromState = ({
     resultVolume?.persistentVolumeClaim?.claimName ||
     `${vm?.metadata?.name}-${diskState.diskName}`;
 
-  dataVolume.metadata.name = dvName;
+  const isTemplateParameter = /[${}]/.test(dvName);
```
metalice, line 139: Please consider exporting to a function as the same code is repeated three times.
nameWithoutParameter(name, diskStateDiskName, defaultValue?)
abusively any other way of function can work as u choose.

upalatucci: done thanks for the suggestion

PR #1273, opened 2023-04-24
Bug 2188886: Pending loading on detach modal

upalatucci:
## 📝 Description

**Cause**
We transform volume in `persistentVolumeClaim` with `convertDataVolumeToPVC` even when the volume is not a DV.

**Solution** 
Do not transform the volume if it's not a datavolume.

To prevent further issues, let's show the fetch error to the user when something goes wrong.
If we fail to load the resource, we'll not be able to delete it or detach it
I've added a prop to the TabModal as in this component the error alert is already showed for API errors and can be useful for other modals that use tabmodal

### DIFF
--- a/src/utils/components/TabModal/TabModal.tsx
+++ b/src/utils/components/TabModal/TabModal.tsx
@@ -1,4 +1,4 @@
-import * as React from 'react';
+import React, { ComponentType, memo, MouseEventHandler, ReactNode, useState } from 'react';
 
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
 import { K8sResourceCommon } from '@openshift-console/dynamic-plugin-sdk';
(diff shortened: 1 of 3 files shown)


src/utils/components/TabModal/TabModal.tsx
```diff
@@ -56,35 +57,39 @@ const TabModal: TabModalFC = React.memo(
     positionTop = true,
     submitBtnVariant,
     titleIconVariant,
+    modalError,
   }) => {
     const { t } = useKubevirtTranslation();
 
     const [isSubmitting, setIsSubmitting] = React.useState(false);
-    const [error, setError] = React.useState(undefined);
+    const [apiError, setApiError] = React.useState(undefined);
```
metalice, line 65: can u please remove all * as React form this file?
Also can u set a type for this state?

src/utils/components/TabModal/TabModal.tsx
```diff
@@ -56,35 +57,39 @@ const TabModal: TabModalFC = React.memo(
...
       setIsSubmitting(false);
 
       const promise = onClose();
 
-      if (promise) promise?.catch(setError);
+      if (promise) promise?.catch(setApiError);
     };
 
+    const error = apiError || modalError;
+    console.log(error);
```
avivtur, line 91: intentional?

upalatucci: no

src/utils/components/TabModal/TabModal.tsx
upalatucci: unfortunately, there is no type available for the error in `useK8sWatchResource` or any other sdk method.

PR #1274, opened 2023-04-24
Bug 2187664: vm with no namespace means default

upalatucci:
## 📝 Description

The namespace was not propagated to the `SelectSecret`. With no namespace, the select was not fetching any namespace. 
This is to prevent the `useK8sWatchResources` to fetch all secrets from all namespaces


avivtur (commented):
looks really good :) 
can you please also change the `import * as React... ` ?

avivtur (approved):

PR #1275, opened 2023-04-24
Bug 2172390: Enable CPU, Memory editing in the drawer

hstastna:
## 📝 Description

**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2172390

_Related design doc:_
https://docs.google.com/document/d/134raBlCMKZJvpUDV0zb2gBqSoa4aRuB2OZVp-LrG2yU

Enable CPU and Memory editing in the side drawer when creating a VM from selected template.

## 🎥 Screenshots
**Before**:
No ability to edit _CPU | Memory_ field values:
vm_beforee

**After:**
User can clearly see which fields they can edit immediately, and after clicking on blue editable CPU | Memory values, the modal to edit the values is opened, see the quick creation of a VM:

https://user-images.githubusercontent.com/13417815/236003141-e2104708-b0b0-42d0-bd85-322830ed11e9.mp4

If the user decides to customize VM, the values from the initial drawer are kept, of course:

https://user-images.githubusercontent.com/13417815/236003168-b433d138-cb28-404e-af0e-6f36c1137e39.mp4

TODO:
- [x] solve saving the CPU/Memory data from the `CPUMemoryModal` by using context hook `useWizardVMContext` (and not put it to the url, as there's already a lot of data that doesn't have to be there) to be used later out of the drawer, if the user decides to customize VM (and not to quick create the VM), so they can see the same data as set in the drawer, process the template and edit the VM object from the processed template properly to enable using `useWizardVMContext` hook
- [x] set the correct CPU/Memory data when quick creating VM
- [x] display the same CPU/Memory data when customizing VM
- [x] refactor the existing related code a bit, especially code of `TemplatesCatalogDrawerPanel` component, as it's getting too long
- [x] fix bug with undefined vm in case the user won't change "CPU | Memory" field in the drawer at all
- [x] fix failing test


hstastna:

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
```diff
@@ -68,16 +69,27 @@ export const TemplatesCatalogDrawerCreateForm: FC<TemplatesCatalogDrawerCreateFo
...
           replaceTemplateParameterValue(draftTemplate, parameterForName, vmName);
+
+        const vmObject = getTemplateVirtualMachineObject(draftTemplate);
+        ensurePath(vmObject, [
+          'spec.template.spec.domain.resources',
+          'spec.template.spec.domain.cpu',
+        ]);
+        vmObject.spec.template.spec.domain.resources.requests = {
+          ...vmObject?.spec?.template?.spec?.domain?.resources?.requests,
+          memory: `${vm.spec.template.spec.domain.resources.requests['memory']}`,
```
avivtur, line 80: Hi :) 
to update the VM's memory request the update should go to: 
`vmObject.spec.template.spec.domain.memory` 
Changing the `spec.template.spec.domain.resources.requests['memory']` is the virt pod limitations,
We probably do it like this in other places and that's wrong so please fix it as well

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerPanel.tsx
```diff
@@ -75,11 +89,45 @@ export const TemplatesCatalogDrawerPanel: React.FC<TemplatesCatalogDrawerPanelPr
...
+        ensurePath(vmDraft, [
+          'spec.template.spec.domain.resources',
+          'spec.template.spec.domain.cpu',
+        ]);
+
+        vmDraft.metadata.namespace = ns || DEFAULT_NAMESPACE;
+        vmDraft.spec.template.spec.domain.resources.requests = {
+          ...vmDraft?.spec?.template?.spec?.domain?.resources?.requests,
+          memory: `${vmObj.spec.template.spec.domain.resources.requests['memory']}`,
+        };
```
avivtur, line 122: same here :)

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerPanel.tsx
```diff
@@ -75,11 +89,45 @@ export const TemplatesCatalogDrawerPanel: React.FC<TemplatesCatalogDrawerPanelPr
...
+    useEffect(() => {
+      k8sCreate<V1Template>({
+        model: ProcessedTemplatesModel,
+        data: setValueForRequiredParams(template),
+        queryParams: {
+          dryRun: 'All',
+        },
+      }).then((processedTemplate) => {
+        setUpdatedVMObject(getTemplateVirtualMachineObject(processedTemplate));
+      });
```
avivtur, line 107: any specific reason to run this on init render and not on submit function?
you are in a race condition here if the template wasn't loaded on the initial render you will still have undefined in your updatedVMObject

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerPanel.tsx
```diff
@@ -75,11 +89,45 @@ export const TemplatesCatalogDrawerPanel: React.FC<TemplatesCatalogDrawerPanelPr
...
     const hostDevicesCount = getHostDevices(vmObject)?.length || 0;
     const gpusCount = getGPUDevices(vmObject)?.length || 0;
     const hardwareDevicesCount = hostDevicesCount + gpusCount;
 
+    const [updatedVMObject, setUpdatedVMObject] = useState<V1VirtualMachine>(undefined);
+
+    useEffect(() => {
+      k8sCreate<V1Template>({
+        model: ProcessedTemplatesModel,
+        data: setValueForRequiredParams(template),
```
avivtur, line 101: can you please explain to me why is setValueForRequiredParams needed?
we should get the PARAM values when processing the template, is that not working?

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
metalice: hmm... i think u are reffring to a case the VM is already created and not before?

second, BE have a different way to change memory,CPU , in different fields, please refer to - https://bugzilla.redhat.com/show_bug.cgi?id=2174892

look at the comments, we should not change to their suggested yet, so meanwhile I think this is ok.

src/views/catalog/templatescatalog/utils/helpers.ts
```diff
@@ -46,3 +49,19 @@ export const filterTemplates = (templates: V1Template[], filters: TemplateFilter
 
       return aName?.localeCompare(bName);
     });
+
+export const setValueForRequiredParams = (template: V1Template): V1Template => {
+  const requiredParamsIndexes = findAllIndexes(
+    template?.parameters || [],
```
metalice, line 55: i think u are running over the array of template?.parameters twice when u can do it only once? just run over draftTemplate.map why also the findallindexes needed here? just make check for parameter?.required in the map run and set the value if needed and return the item...

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
avivtur: ah yes I thought I saw this somewhere but couldn't find it, thank you :)

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerPanel.tsx
hstastna: I've experienced some scenarios, when for some templates, processing the template failed due to not having the param values already there. Not sure why this happens and only for some templates.

src/views/catalog/templatescatalog/utils/helpers.ts
hstastna: Thanks, will take a look how to optimize this function.

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
hstastna: > We probably do it like this in other places and that's wrong so please fix it as well

Yep, this is why I did it the same way also here. Wasn't aware of the fact this is gonna change. Sooo what is the result? Are we okay with this for now? If yes, then, please, let me know once it will be good to change that, so I could make a followup PR :)

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerPanel.tsx
hstastna: Just a note, I needed to process the template to be able to use the VM object that is suitable for using it with our context hook for saving the VM data and being able to get the data later if user decides to customize the VM. And also to be able to use the CPU/Mem modal we already have implemented. I expect that more stuff is gonna be soon editable in the drawer, so putting everything to url won't be a good idea anymore (and IMHO it never was a good idea but we do it for some data already anyway).

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
avivtur: I think you can leave it as you did for now, @metalice other thoughts?

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerPanel.tsx
avivtur: > What I can see here I am just missing adding the case when the template doesn't load or process successfully on the first render. WDYT?

```
useEffect(() => {
  if(loaded && isEmpty(loadError) && !isEmpty(template)) {
    k8sCreate<V1Template>({
      model: ProcessedTemplatesModel,
      data: setValueForRequiredParams(template),
      queryParams: {
        dryRun: 'All',
       },
      }).then((processedTemplate) => {
        setUpdatedVMObject(getTemplateVirtualMachineObject(processedTemplate));
      });
}, [loaded, loadError, template]);
```
adding a condition should do the trick - the useEffect will run a few times but will process the template only once because of the condition which guarantees that the template is loaded without an error

metalice: Hey,

1. we better not use "// eslint-disable-next-line react-hooks/exhaustive-deps"
as next react versions will be without a dependency array and will auto-add dependency, and it will cause bugs. better just add more protection to the code block and all needed dependencies. 

2. this file is getting very long already, about 250 lines of code. Those changes also add 50+ lines of code just for this file, please export functions and inner components to different utils/components files.

3. u are importing VirtualMachineDescriptionItem which belongs to VM pages into catalog, if you would like to use it please move VirtualMachineDescriptionItem into a more proper location under our shared folder. The same goes to CPUMemory component.

4. why do u need to process the template before any change has done at all?
you are getting the template as a prop which is good, the proccessedtemplate should just check the values that have been changed are ok. why processing without any value change?

hstastna: 1,2,3 yes.

4. because without that we wont be able to use the existing cpu/mem modal and also the context hook. the option to avoid that is to put the data to the url, which I am not going to do.

metalice: The thing with URL is that it refresh resistance, and will keep the user flow with the same data when the page is refreshing. If I'm not mistaken, the context is also saved in localstroage so we have some good behavior that we want, just make sure your changes are refresh resistance. 

processedTemplate- gives u a VM with params set in VM? That's what u are looking for?

hstastna: Seems that processing a template does more than just setting some params or generating a VM name (which is not interesting for us now). It makes the VM object a "proper" VM object, usable for the context hook and for `CPUMemoryModal` component. I simply cannot use `updateVM` with the VM object taken from a template if the template wasn't already processed. It knows that it isn't really a VM.

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
metalice: yes, yes, that what i was saying :)

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerPanel.tsx
```diff
@@ -75,11 +69,51 @@ export const TemplatesCatalogDrawerPanel: React.FC<TemplatesCatalogDrawerPanelPr
...
+
+        vmDraft.metadata.namespace = ns || DEFAULT_NAMESPACE;
+        vmDraft.spec.template.spec.domain.resources.requests = {
+          ...vmDraft?.spec?.template?.spec?.domain?.resources?.requests,
+          memory: `${vmObj.spec.template.spec.domain.resources.requests['memory']}`,
+        };
+        vmDraft.spec.template.spec.domain.cpu.cores = vmObj.spec.template.spec.domain.cpu.cores;
+      });
+
+      setUpdatedVMObject(updatedVM);
```
metalice, line 112: maybe we can extract it to util? just to make the file smaller?
to a func updateVMCPU(VM, callback, ns?) and can be used as  updateVMCPU(vmObj, setUpdatedVMObject, ns)

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerPanel.tsx
metalice: Before this addition, we also used ProcessedTemplates for editing, and updaing the name of VM which is also available in the drawer? should work the same way ?

hstastna: My solution for updating the name in the drawer did require processing the template for sure. However, somebody has changed that to putting the name to url to avoid that (and there's more data in the url, when being in the drawer.. data that could be saved and used later by the same context hook I am using in this PR. We could use the hook for all the data from drawer, not just this. Just the code around is not super short, but we could figure that out.. ).

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerPanel.tsx
hstastna: Maybe.. I was thinking about refactoring this code the way that it could be used also in other places where we update cpu and memory values, but I am not sure if it worth, because we wouldn't save much lines of the code anyway.. But maybe the code would be more readable at least..

avivtur (approved):
Hi @hstastna 
looks good, I have one comment but it's not something important so if you can please try to do that in a follow-up PR that would be great! :smile: :+1:

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
```diff
@@ -68,16 +69,29 @@ export const TemplatesCatalogDrawerCreateForm: FC<TemplatesCatalogDrawerCreateFo
...
+          const vmObject = getTemplateVirtualMachineObject(draftTemplate);
+          ensurePath(vmObject, [
+            'spec.template.spec.domain.resources',
+            'spec.template.spec.domain.cpu',
+          ]);
+          vmObject.spec.template.spec.domain.resources.requests = {
+            ...vmObject?.spec?.template?.spec?.domain?.resources?.requests,
+            memory: `${vm.spec.template.spec.domain.resources.requests['memory']}`,
+          };
+          vmObject.spec.template.spec.domain.cpu.cores = vm.spec.template.spec.domain.cpu.cores;
```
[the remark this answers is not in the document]
hstastna: This is a very good point! I already tried but I've hit some issues, so I better leave it for followup PR for now :)

PR #1276, opened 2023-04-25
CNV-27327: Implement the UI to enable users to create a VM from a PVC

avivtur:
## 📝 Description

Enabling user to select either a DS or a PVC and create a VM from it with IT and prefs.

Will get only PVCs that are not represented by a DataSource and have the default preference label

## 🎥 Demo

https://user-images.githubusercontent.com/67270715/234808328-93106942-8a86-4cf6-a3a2-45fea89efb5d.mp4


hstastna (commented):
Just some nitpicks, otherwise 👍🏽

PR #1277, opened 2023-04-26
Bug 2189744: no docker prefix for container image

upalatucci:
## 📝 Description

remove docker prefix when using containerDisk.

ContainerDisk already expect a docker image URL so using `docker://` prefix is wrong

### DIFF
--- a/src/views/catalog/customize/cd.ts
+++ b/src/views/catalog/customize/cd.ts
@@ -2,6 +2,7 @@ import { produceVMDisks } from '@catalog/utils/WizardVMContext';
 import { V1Template } from '@kubevirt-ui/kubevirt-api/console';
 import {
   V1beta1DataVolumeSpec,
+  V1ContainerDiskSource,
   V1DataVolumeTemplateSpec,
   V1VirtualMachine,
   V1Volume,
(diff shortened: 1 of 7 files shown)


src/views/catalog/customize/cd.ts
```diff
@@ -25,7 +26,7 @@ export const addInstallationCDRom = (
     cdVolume = {
       name: INSTALLATION_CDROM_NAME,
       containerDisk: {
-        image: cdSource.source?.registry.url,
+        image: removeDockerPrefix(cdSource.source?.registry.url),
```
avivtur, line 29: do we add this prefix somewhere? if yes, it's better in my opinion avoid adding something not needed instead of adding than removing. if not please ignore this comment :)

avivtur (commented):
Hi Ugo :) 
left you a conditional comment :grin:

src/views/catalog/customize/cd.ts
metalice: yes, we do add, i think there is a function addDockerPrefix, maybe it can be done before calling this function in case of a container disk

src/views/catalog/customize/cd.ts
```diff
@@ -11,24 +12,36 @@ import {
...
   let cdDataVolumeTemplate: V1DataVolumeTemplateSpec = undefined;
   const dataVolumeName = `${virtualMachine?.metadata?.name}-${INSTALLATION_CDROM_NAME}`;
 
-  if (cdSource?.source?.registry) {
+  if ((cdSource as V1ContainerDiskSource)?.image) {
     cdVolume = {
       name: INSTALLATION_CDROM_NAME,
       containerDisk: {
-        image: cdSource.source?.registry.url,
+        image: removeDockerPrefix((cdSource as V1ContainerDiskSource).image),
```
upalatucci, line 30: let's have this `removeDockerPrefix` in case the user types the prefix anyways.
Here the prefix will cause an error in the VM start.

PR #1279, opened 2023-04-26
Bug 2182172: Add alert for clound init and ssh auth keys

upalatucci:
- Add lines and units to VM metric charts y-axis
- Bug 2182172: Add alert for clound init and ssh auth keys

## 📝 Description

Cloud-init runs only the first time and editing this property alongside auth keys is useless for now.

Unfortunately, following the bug conversation, there is no good way to understand if was already executed and applied.
We could use the status.created of the VM to disable cloud init and ssh key editing but it's not accurate as the cloud-init could not be executed yet.

The good news: the backend is implementing a way to fix that and allow the user to change the configuration after the first tun.

## 🎥 Demo

Screenshot from 2023-04-26 12-09-14

### DIFF
--- a/locales/en/plugin__kubevirt-plugin.json
+++ b/locales/en/plugin__kubevirt-plugin.json
@@ -231,6 +231,7 @@
   "Cloning": "Cloning",
   "Closing it will cause the upload to fail. You may still navigate the console.": "Closing it will cause the upload to fail. You may still navigate the console.",
   "Cloud-init": "Cloud-init",
+  "Cloud-init and SSH key configurations will be applied to the VirtualMachine only at the first boot.": "Cloud-init and SSH key configurations will be applied to the VirtualMachine only at the first boot.",
   "Cluster scope migrations": "Cluster scope migrations",
   "Collapse": "Collapse",
   "Completion timeout": "Completion timeout",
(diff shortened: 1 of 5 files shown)


src/views/catalog/wizard/tabs/scripts/WizardScriptsTab.tsx
```diff
@@ -29,6 +36,20 @@ const WizardScriptsTab: WizardTab = ({ vm, updateVM }) => {
...
+                'Cloud-init and SSH key configurations will be applied to the VirtualMachine only at the first boot.',
+              )}
+              isInline
+              variant={AlertVariant.warning}
+            >
+              {t(
+                'Please, make sure the configuration is correct before starting the VirtualMachine.',
+              )}
+            </Alert>
+          </DescriptionListDescription>
```
avivtur, line 52: I see you use this exact alert in 2 different places, please create as a component and use it in both places

locales/en/plugin__kubevirt-plugin.json
```diff
@@ -754,6 +755,7 @@
   "Please don’t close this browser tab": "Please don’t close this browser tab",
   "Please provide hostname.": "Please provide hostname.",
   "Please wait, once the Data Volume has been created the data will start uploading into this Persistent Volume Claims.": "Please wait, once the Data Volume has been created the data will start uploading into this Persistent Volume Claims.",
+  "Please, make sure the configuration is correct before starting the VirtualMachine.": "Please, make sure the configuration is correct before starting the VirtualMachine.",
```
apinnick, line 758: We do not use "please" in our UIs or documentation. Also, the comma is incorrect. 

Suggested rewrite: "You must ensure that the configuration is correct before starting the VirtualMachine."

PR #1287, opened 2023-05-02
Bug 2181920: Display "UEFI" for Boot mode for "efi: {}" in yaml

hstastna:
## 📝 Description

**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2181920

This is a followup of https://github.com/kubevirt-ui/kubevirt-plugin/pull/1236 introducing the change that `efi: { secureBoot: false }` was written to VM's yaml when changing _Boot mode_ field value to "UEFI", because that was correct representation of "UEFI" boot mode (secure boot disabled).

**In this PR, the following remaining issues are fixed:**
- for existing VM:
  - it shows "UEFI" for _Boot mode_ field in _Details_ for `efi: {}` in the yaml's bootloader section
- in Create VM wizard:
  - it shows "UEFI" for _Boot mode_ in _Review and create VirtualMachine_ screen for `efi: {}` in the yaml's bootloader section
- for existing template:
  - it shows "UEFI" for _Boot mode_ in _Details_ for `efi: {}` in the yaml's bootloader section
  - after changing _Boot mode_ value to "UEFI", `efi: {}` occurs in the yaml (this doesn't happen for existing VMs or when creating a VM)

Part of this PR is also a bit of refactoring - removing duplicated code for updating the VM with selected boot mode data.
The result of refactoring is new `updatedVMBootMode` function that returns the updated VM object.

**Expected results:**
The correct value in the yaml after selecting UEFI should be `efi: { secureBoot: false }`. Also for `efi: {}`  in the YAML (as many default templates come with this), secure boot is enabled and "UEFI (secure)" should be displayed in _Boot mode_ field in the UI.

_Notes about the moot modes:_
Both `efi: {}` and `efi: { secureBoot: true }` in the yaml are equivalent (at least now, maybe this will change) and refer to secure boot enabled displayed as "UEFI (secure)" option in the UI. If no `efi` present, then "BIOS" is displayed as a boot mode, in the Details tab. This was the original implementation at least. I was not dealing with this third option too much, the main problem was the secure boot option, that was enabled while the UI showed just "UEFI".

_How to reproduce the bug:_
https://github.com/kubevirt-ui/kubevirt-plugin/pull/1236#issue-1655828779

_Note:_
It is possible this issue will be fixed by backend eventually, probably the way that `efi: {}` in the yaml won't represent secure boot anymore (but it's possible they will come up with something else). But till then, we need this PR to make this work as expected.

## 🎥 Screenshots
Example of a VM yaml:
vm_yaml

**Before:**
"UEFI" is displayed for _Boot mode_ field in VM _Details_ for `efi: {}` (see the yaml example above):
vm_before
After changing _Boot mode_ value to "UEFI", `efi: {}` occurs in the VM Template's yaml:
template_before

**After:**
"UEFI (secure)" is displayed for _Boot mode_ field in VM _Details_ for `efi: {}` as expected:
vm_after
After changing _Boot mode_ value to "UEFI", `efi: { secureBoot: false }` occurs in the VM Template's yaml:
template_after

### DIFF
--- a/src/utils/components/FirmwareBootloaderModal/utils/utils.ts
+++ b/src/utils/components/FirmwareBootloaderModal/utils/utils.ts
@@ -1,26 +1,58 @@
-import { TFunction } from 'i18next';
+import produce from 'immer';
 
 import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import { ensurePath, isEmpty } from '@kubevirt-utils/utils/utils';
 
-import { BootloaderOptionValue } from './constants';
+import { BootMode, BootModeTitles } from './constants';
+import { BootloaderOptionValue } from './types';
+
+export const isObjectEmpty = (obj: object): boolean => obj && isEmpty(obj);
 
 export const getBootloaderFromVM = (vm: V1VirtualMachine): BootloaderOptionValue => {
-  if (vm?.spec?.template?.spec?.domain?.firmware?.bootloader?.efi?.secureBoot) {
-    return 'uefiSecure';
+  const secureBoot = vm?.spec?.template?.spec?.domain?.firmware?.bootloader?.efi;
+
+  if (secureBoot?.secureBoot === true || isObjectEmpty(secureBoot)) {
+    return BootMode.uefiSecure;
   }
-  if (vm?.spec?.template?.spec?.domain?.firmware?.bootloader?.efi) {
-    return `uefi`;
+  if (secureBoot?.secureBoot === false) {
+    return BootMode.uefi;
   }
-  return 'bios';
+  return BootMode.bios;
 };
 
-export const getBootloaderTitleFromVM = (vm: V1VirtualMachine, t: TFunction): string => {
+export const getBootloaderTitleFromVM = (vm: V1VirtualMachine): string => {
   const bootloader = getBootloaderFromVM(vm);
-  const titles = {
-    uefiSecure: t('UEFI (secure)'),
-    uefi: t('UEFI '),
-    bios: t('BIOS'),
-  };
 
-  return titles[bootloader];
+  return BootModeTitles[bootloader];
 };
+
+/**
+ * A function to return the VirtualMachine object updated with a given boot mode
+ * @param {V1VirtualMachine} vm - VirtualMachine object
+ * @param {BootloaderOptionValue} firmwareBootloader - selected boot mode
+ * @returns {V1VirtualMachine} updated VirtualMachine object
+ */
+export const updatedVMBootMode = (
+  vm: V1VirtualMachine,
+  firmwareBootloader: BootloaderOptionValue,
+) =>
+  produce<V1VirtualMachine>(vm as V1VirtualMachine, (vmDraft: V1VirtualMachine) => {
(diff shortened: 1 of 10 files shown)


hstastna:

src/utils/components/FirmwareBootloaderModal/utils/utils.ts
```diff
@@ -1,14 +1,20 @@
 import { TFunction } from 'i18next';
+import produce from 'immer';
 
 import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import { ensurePath, isEmpty } from '@kubevirt-utils/utils/utils';
 
 import { BootloaderOptionValue } from './constants';
 
+export const isObjectEmpty = (obj: object): boolean => isEmpty(obj) && obj !== undefined;
```
metalice, line 9: What about case null? Is possible? i think isEmpty(obj) && obj can work also?

hstastna: In the place I am using this function, null isn't possible. But if someone decided to use this somewhere else, probably `isEmpty(obj) && obj` is better to use.

avivtur: Can you explain to me why just isEmpty(obj) is not enough? it should return true if it's either one of these:
`null / undefind / [] / {} / ''`

hstastna: Because we don't want to receive true if the object is null or undefined.

avivtur: so `efi: {}` and `efi: { secureBoot: true }` is equivalent right? and if `efi` does not exist we want bios?

hstastna: Yes, they both refer to secure boot enabled displayed as "UEFI (secure)" option in the UI. And yes, if no efi, then show "BIOS". This was the original implementation at least. I was not dealing with this third option too much, the main problem was the secure boot option.

avivtur (commented):
looks good just one nit and I think it's good to go

src/utils/components/FirmwareBootloaderModal/utils/utils.ts
```diff
@@ -24,3 +30,34 @@ export const getBootloaderTitleFromVM = (vm: V1VirtualMachine, t: TFunction): st
...
+  vm: V1VirtualMachine,
+  firmwareBootloader: BootloaderOptionValue,
+) =>
+  produce<V1VirtualMachine>(vm as V1VirtualMachine, (vmDraft: V1VirtualMachine) => {
+    ensurePath(vmDraft, 'spec.template.spec.domain.firmware.bootloader');
+    ensurePath(vmDraft, 'spec.template.spec.domain.features.smm');
+    vmDraft.spec.template.spec.domain.features.smm = { enabled: true };
+
+    switch (firmwareBootloader) {
+      case 'uefi':
```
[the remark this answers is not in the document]
hstastna: Seems that I am doing more refactoring in this PR than bug fixing ...

avivtur (approved):

PR #1291, opened 2023-05-08
[WIP] CNV-27316: Code change to show the cli command in the "Create VM" flow

avivtur:
## 📝 Description

Adding a button to show to YAML / CLI command for creating the VM.
button is disabled if no volume selected or if there's no name for the vm

issue: https://issues.redhat.com/browse/CNV-27316

## 🎥 Demo

https://user-images.githubusercontent.com/67270715/236885882-ff838fa0-238e-4152-8fd1-b49bdc0d591a.mp4


src/views/catalog/CreateFromInstanceTypes/components/CreateVMFooter/components/YamlAndCLIViewerModal/YamlAndCLIViewerModal.tsx
```diff
@@ -0,0 +1,61 @@
+import React, { FC, KeyboardEvent, MouseEvent as ME, Suspense, useState } from 'react';
```
metalice, line 1: MouseEvent as ME ?

avivtur: since we have [MouseEvent](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent) from JS
and MouseEvent from React, I assigned an alias to avoid React.MouseEvent on line 29.

metalice: as we talked u can remove some of them, please change in another PR

PR #1295, opened 2023-05-10
CNV-27319: Add pagination to IT and Pref list views

avivtur:
## 📝 Description

This PR includes:
- Add pagination component to InstanceType and Preferences list views.
- Add Developer preview badge to page's title
- Relocation constants and types related to pagination in utils/hooks/usePagination
- Relocation `columnSorting` function to utils/utils 
- Resetting VM name on create VM with IT flow after VM create/navigating between pages.

## 🎥 Demo

After: 
it-pagi

pref-pagi


src/utils/hooks/usePagination/usePagination.ts
```diff
@@ -1,11 +1,9 @@
...
 
-import { paginationInitialState, PaginationState } from '@virtualmachines/utils';
+import { paginationInitialState } from './utils/constants';
+import { PaginationState, UsePaginationValues } from './utils/types';
 
-type UsePagination = () => [
-  pagination: PaginationState,
-  onPaginationChange: (newPagination: PaginationState) => void,
-];
+type UsePagination = () => UsePaginationValues;
```
metalice, line 6: u can move UsePagination to tils/types instead of importing usepaginationvalues and declaring usepagination here.. just import usepagination

src/views/preferences/list/ClusterPreferenceList.scss
```diff
@@ -0,0 +1,9 @@
+.cluster-preference-list {
```
metalice, line 1: this is identical to cluster-instancetype-list - maybe just use one of them? with a more appropriate name?

src/views/catalog/CreateFromInstanceTypes/state/utils/utils.ts
```diff
@@ -0,0 +1,7 @@
+import { adjectives, animals, uniqueNamesGenerator } from 'unique-names-generator';
+
+export const getRandomVMName = () =>
```
metalice, line 3: i already think we have this function- please check

src/views/preferences/list/ClusterPreferenceList.scss
avivtur: we this code repeating in few more places, any sense in util/scss ?

src/views/catalog/CreateFromInstanceTypes/state/utils/utils.ts
avivtur: we have - https://github.com/kubevirt-ui/kubevirt-plugin/blob/fad61ff751cd239de2b99b8a2cdf7ca84a47ed57/src/utils/resources/template/utils/helpers.ts#L45
which requires a template, I'll create a similar instance for bootable volume, sounds good?

PR #1298, opened 2023-05-11
CNV-23128: Add OpenShift Virtualization Welcome modal

hstastna:
## 📝 Description

**Fixes:**
https://issues.redhat.com/browse/CNV-23128

_Design doc:_
https://docs.google.com/document/d/1-SiRh40laizKsMBDuZgaExODuxJzD0UqwKkMb5yXUnw/edit#

Add this welcome modal to improve the UX for new users in the OpenShift Virtualization space. The modal provides useful information such as the most important/common actions users may take next and acts as shortcut. It also helps users understand they just landed in the right place.

Currently the modal contains "Create VirtualMachine" button that leads to Catalog page.
It also contains the following links:
- "Create a virtual machine from a template (Quick start)" (leads to the appropriate Quick start),
- "Upload a boot source (Quick start)" (also leads to the appropriate Quick start),
- ~~"Migration Toolkit for Virtualization Operator" (leads to the appropriate operator in OperatorHub).~~

## 🎥 Demo
The actual welcome modal:

https://github.com/kubevirt-ui/kubevirt-plugin/assets/13417815/14b009a7-30ff-433f-8490-523ff5f309d3

Clicking on "Create VirtualMachine" button leads to Catalog page, like this:
create-vm

Clicking on "Create a virtual machine from a template (Quick start)" leads to the related Quick start:
template-qs

"Upload a boot source (Quick start)" leads to the related Quick start:
boot-qs

~~"Migration Toolkit for Virtualization Operator" leads to the appropriate operator:~~

---

The remaining TODO:
- [x] make the modal displayed when user accesses Virtualization section in the menu, as expected
- [x] save the information to show/not to show the modal again for a specific user, using `useKubevirtUserSettings` hook


src/views/welcome/WelcomeModal.scss
```diff
@@ -0,0 +1,27 @@
+.WelcomeModal {
+  &__grid {
+    margin-top: var(--pf-global--spacer--2xl);
+    margin-bottom: var(--pf-global--spacer--md);
+    margin-left: var(--pf-global--spacer--xl);
+    margin-right: var(--pf-global--spacer--xl);
```
avivtur, line 6: you can use the shortend of margin like so:
`margin: var(--pf-global--spacer--2xl) var(--pf-global--spacer--xl) var(--pf-global--spacer--md) var(--pf-global--spacer--xl);  /* top, right, bottom, left */`

another little improvement to that, you can also omit the value for left margin and will be as right margin:
`margin: var(--pf-global--spacer--2xl) var(--pf-global--spacer--xl) var(--pf-global--spacer--md); /* top, right+left, bottom`
from MDN docs: https://developer.mozilla.org/en-US/docs/Web/CSS/margin

src/views/welcome/WelcomeModal.tsx
```diff
@@ -0,0 +1,77 @@
...
+        <GridItem span={8}>
+          <Stack>
+            <Trans t={t} ns="plugin__kubevirt-plugin">
+              <Title headingLevel="h2">Welcome to</Title>
+              <Title headingLevel="h1">OpenShift Virtualization</Title>
+
+              <Text className="text-muted WelcomeModal__text" component={TextVariants.p}>
+                OpenShift Virtualization allows you to run and manage virtualized workloads
+                alongside container workloads. You can use it to manage both Linux and Windows
+                virtual machines.
```
avivtur, line 53: shouldn't it be VirtualMachines?

src/views/welcome/WelcomeModal.tsx
```diff
@@ -0,0 +1,77 @@
...
+              </Text>
+
+              <Title headingLevel="h3">What do you want to do next?</Title>
+
+              <WelcomeButtons onClose={onClose} />
+
+              <Checkbox
+                className="WelcomeModal__checkbox"
+                id="welcome-modal-checkbox"
+                label={'Do not show this again'}
```
avivtur, line 63: please change from `{'Do not show this again'}` to `"Do not show this again"` for consistency like `className` and `id` props :)

src/views/welcome/components/WelcomeButtons.tsx
```diff
@@ -0,0 +1,37 @@
...
+};
+
+const WelcomeButtons: FC<WelcomeButtonsProps> = ({ onClose }) => {
+  const history = useHistory();
+  const { ns } = useParams<{ ns: string }>();
+
+  return (
+    <>
+      {welcomeModalButtons(ns).map((modalButton: WelcomeModalButtonsData, index: React.Key) => (
+        <StackItem key={index}>
```
avivtur, line 20: using the index as a key is not recommended and should be avoided, you can use `modalButton.name` as unique key in this case.
If you want to read more about why this should be avoided you can go to react docs here and keep reading until the "Pitfall" yellow part below: https://react.dev/learn/rendering-lists#where-to-get-your-key

src/views/welcome/utils/constants.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const createVMQuickUrl = '/quickstart?quickstart=creating-virtual-machine';
+const uploadBootUrl = '/quickstart?quickstart=upload-boot-source';
+
+export const welcomeModalButtons = (ns: string): WelcomeModalButtonsData[] => {
+  const namespace = ns ? `ns/${ns}` : ALL_NAMESPACES;
+
+  return [
+    {
+      name: 'Create VirtualMachine',
```
avivtur, line 14: should the button names in this file be translated?

src/views/clusteroverview/ClusterOverviewPage.tsx
```diff
@@ -38,6 +44,17 @@ const ClusterOverviewPage: React.FC = () => {
...
+  const showModal = useMemo(
+    () => !quickStarts?.dontShowWelcomeModal,
+    [quickStarts?.dontShowWelcomeModal],
+  );
+
+  useEffect(() => {
+    showModal &&
+      createModal(({ isOpen, onClose }) => <WelcomeModal isOpen={isOpen} onClose={onClose} />);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []); // we do need to render this only once
```
avivtur, line 56: Adding the deps here is causing the modal to spawn more than one?
What happens if you do something like:
```
useEffect(() => {
  !quickStarts?.dontShowWelcomeModal && createModal....
}, [quickStarts?.dontShowWelcomeModal, createModal]);
```
and remove the useMemo above?

src/views/welcome/utils/constants.ts
metalice, line 1: please change the name to utils, probably at the beginning constants was appropriate :)

src/views/clusteroverview/ClusterOverviewPage.tsx
metalice, line 1: I would have considered using the loading that returns from useKubevirtUserSettings , as it's a network call and can take a sec to load so you won't get a blinking modal..

src/views/welcome/WelcomeModal.scss
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  &__image {
+    width: 260;
```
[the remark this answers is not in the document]
hstastna: Yes, it is very specific as the content of the modal is. I haven't found any PF variable good enough for this, so the explicit value has occurred here.

src/views/welcome/WelcomeModal.scss
hstastna: Tried, of course, but margin: `var(--pf-global--spacer--2xl) var(--pf-global--spacer--xl) var(--pf-global--spacer--md) var(--pf-global--spacer--xl);` simply didn't work and I haven't found the root cause of that problem. Spent a lot of time on that. Additionally, I think it is more readable like it is now.
Since the 1st option didn't work, I haven't tried what you suggest as the 2nd option, and directly went for the long version.

src/views/welcome/WelcomeModal.tsx
hstastna: I think I am missing `t` function here. That's a bigger problem, IMO :D

src/views/welcome/components/WelcomeButtons.tsx
hstastna: Yeah, I know. Just sometimes it is hard to see what else I can use instead of index...

src/views/clusteroverview/ClusterOverviewPage.tsx
hstastna: It happens that the modal won't occur, but the blank screen, and your browser gets stuck. I recommend you to try it. This is why I also added the comment there, to prevent someone accidentally changing that and adding deps to the array.

src/views/clusteroverview/ClusterOverviewPage.tsx
hstastna: Yep. I am still figuring out how to use all the benefits of that powerful hook we have - useKubevirtUserSettings ;)

src/views/welcome/WelcomeModal.tsx
avivtur: since it's under the Trans component I assumed it will cover it too, I think the closing Trans needs to be relocated and than you can simply use the `t` here

hstastna: I'm not gonna change these lines of the code, due to position of Trans component. I think it is fine. and `'` vs `"`, well.. some of our rules make it `' `from `"` once I save the code in vs code.

src/views/clusteroverview/ClusterOverviewPage.tsx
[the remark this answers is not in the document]
hstastna: Tested and this patch doesn't work as expected. And is missing "loading that returns from useKubevirtUserSettings".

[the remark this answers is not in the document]
metalice: > Also you cannot mix isOpen or onClose with the information if the modal should be showed again or not. This simply won't work.

what do u mean by that?

> Tested and this patch doesn't work as expected

what isnt working for u?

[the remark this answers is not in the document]
metalice:
```typescript
import React, { FC, useState } from 'react';
import { Helmet } from 'react-helmet';

import { useIsAdmin } from '@kubevirt-utils/hooks/useIsAdmin';
import useKubevirtUserSettings from '@kubevirt-utils/hooks/useKubevirtUserSettings/useKubevirtUserSettings';
import { HorizontalNav, NavPage } from '@openshift-console/dynamic-plugin-sdk';

import WelcomeModal from '../welcome/WelcomeModal';

import ClusterOverviewPageHeader from './Header/ClusterOverviewPageHeader';
import MigrationsTab from './MigrationsTab/MigrationsTab';
import OverviewTab from './OverviewTab/OverviewTab';
import SettingsTab from './SettingsTab/SettingsTab';
import TopConsumersTab from './TopConsumersTab/TopConsumersTab';

const ClusterOverviewPage: FC = () => {
  const isAdmin = useIsAdmin();
  const [quickStarts, , loaded] = useKubevirtUserSettings('quickStart');
  const [isOpen, setIsOpen] = useState<boolean>(true);

  const overviewTabs: NavPage[] = [
    {
      href: '',
      name: t('Overview'),
      component: OverviewTab,
    },
    {
      href: 'top-consumers',
      name: t('Top consumers'),
      component: TopConsumersTab,
    },
    {
      href: 'migrations',
      name: t('Migrations'),
      component: MigrationsTab,
    },
    {
      href: 'settings',
      name: t('Settings'),
      component: SettingsTab,
    },
  ];

  return (
    <>
      <Helmet>
        <title>{t('Virtualization')}</title>
      </Helmet>
      {loaded && !quickStarts?.dontShowWelcomeModal && (
        <WelcomeModal isOpen={isOpen} onClose={() => setIsOpen(false)} />
      )}
      <ClusterOverviewPageHeader />
      {isAdmin ? <HorizontalNav pages={overviewTabs} /> : <OverviewTab />}
    </>
  );
};

export default ClusterOverviewPage;

```

I've added the missing loaded and tested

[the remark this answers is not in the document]
hstastna: Also you cannot mix isOpen or onClose with the information if the modal should be showed again or not. This simply won't work.

> what do u mean by that?

I mean those are different information, not the same.

> what isnt working for u?

Mainly the functionality of the checkbox. 

> I've added the missing loaded and tested

Thanks, I will definitely use "loaded" the way you showed in the patch. Anyway, I still need to solve the checkbox not working as expected, so I am working on this.

src/views/welcome/WelcomeModal.scss
hstastna: Using the 2nd option - providing left+right value for margin as just one value did work. Thanks!

src/views/clusteroverview/ClusterOverviewPage.tsx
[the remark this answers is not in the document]
hstastna: Matan, finally I've decided to add your suggestions to the PR just as you've suggested, as it does make sense to me, just I am the only one for who it does not work well. I think I have some specific issues with my env very probably :D We still can push more commits if improvements will be needed in the future. Removed WIP from this PR. Thanks so much!

hstastna:

PR #1302, opened 2023-05-21
CNV-16349: Adding Force stop action

DanaOrr:
## 📝 Description
Adding force stop action in case vm isn't stopped.
Deleting the StopVMModal and using a grace period - 0 in case of a force stop.

## 🎥 Demo
AddForceStop


src/views/virtualmachines/actions/VirtualMachineActionFactory.tsx
avivtur, line 1: can you please try and remove t as a param from all functions in this file, and simply import `t`?

avivtur (commented):
Hi @DanaOrr this looks good :) 
I added a couple of suggestions to maybe remove some code we might not really need

avivtur (approved):

PR #1311, opened 2023-05-29
Bug 2203187: Make adding disk by cloning PVC work

hstastna:
## 📝 Description

**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2203187

Make adding disk in _Add disk_ modal when choosing _PVC (creates PVC)_ as a source work. Make it work by setting the PVC size unit correctly for this option.

_Details:_
The problem was that when setting disk size for _PVC (creates PVC)_ source (`setDiskSize`), the disk size unit was set to one of the following: MiB, GiB, TiB. The disk was not able to be created when using these units (that we prefer when displaying size in the UI). The problem was presence of "B" at the end of the unit (a.k.a. IEC units) when using `bytesFromQuantity` to get the correct size and unit.
Originally, `bytesFromQuantity` was used to prevent getting the size in bytes in some situations (for example when deleting the size number in the appropriate field in the modal). Now it was replaced by a new function `bytesToDiskSize` that works brilliant, provides units without unnecessary "B" at the end of the unit string, and so it provides the expected result.

## 🎥 Screenshots
**Before:**
Adding disk to an existing VM fails when choosing _PVC (creates PVC)_ for _Source_ field:
disk_before

**After:**
Disk added successfully:
disk_after

### DIFF
--- a/src/utils/components/DiskModal/DiskFormFields/DiskSourceFormSelect/components/DiskSourcePVCSelect.tsx
+++ b/src/utils/components/DiskModal/DiskFormFields/DiskSourceFormSelect/components/DiskSourcePVCSelect.tsx
@@ -1,6 +1,6 @@
 import React, { FC, useCallback, useMemo } from 'react';
 
-import { bytesFromQuantity } from '@catalog/utils/quantity';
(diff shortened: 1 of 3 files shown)


hstastna:

PR #1305, opened 2023-05-25
CNV-27213: Allow create VM with user defined IT

avivtur:
## 📝 Description

This PR includes:

- Refactoring the "Select Instace Type" section to rely on the cluster's existing IT resources instead of hard-coded values. 
  Also Adding Tabs layout for Red Hat ITs as card menus and Users ITs as a table with selectable rows (similar to volume 
  select)
- Exposing and improving useClickOutside to utils
- Changing the IT state to hold only the name of IT as a string since category and size are not relevant for user ITs

NOTE: Red Hat series are not finalized, the dev cluster currently using the upstream series in the `After` screenshot which will change.
## 🎥 Demo
it-section-b4

### After
Users:
it-section-2
RH Provided:
it-section

PR #1316, opened 2023-05-31
Change hard-coded cloud-init username

avivtur:
## 📝 Description

When creating a VM from IT, we want the cloud-init username would be similar to what we had when we create a VM from template.

### Before:

https://github.com/kubevirt-ui/kubevirt-plugin/assets/67270715/711c255e-29c6-42cc-92e5-cc09df7af159

### After:

https://github.com/kubevirt-ui/kubevirt-plugin/assets/67270715/a5d58776-30c2-4020-a03d-85a800981af4


src/views/catalog/CreateFromInstanceTypes/utils/utils.ts
```diff
@@ -103,7 +115,7 @@ export const generateVM = (
       },
       preference: {
         // inferFromVolume: `${virtualmachineName}-disk`,
```
metalice, line 117: ?

avivtur: removed in here https://github.com/kubevirt-ui/kubevirt-plugin/pull/1305

PR #1323, opened 2023-06-05
Bug 2196161: Display Bootable volumes Name column correctly

hstastna:
## 📝 Description

**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2196161

Display the _Name_ column correctly in case when the "Clone in progress" label is present next to the name while the browser window is very small.

_Solution:_
Using `inline` prop with `ResourceLink` instead of `Split` and `SplitItem` components for achieving displaying name with the label at the same line of the row solved the problem.
Additionally, the name with the label are more nicely aligned, when browser window has its normal/common size.

## 🎥 Screenshots
**Before:**
before2
Name with the label not nicely aligned, when browser window has its normal/common size:
before1

**After:**
after2
Name with the label both nicely aligned to the center of the row:
after1

### DIFF
--- a/src/views/bootablevolumes/list/components/BootableVolumesRow.tsx
+++ b/src/views/bootablevolumes/list/components/BootableVolumesRow.tsx
@@ -9,7 +9,7 @@ import { ANNOTATIONS } from '@kubevirt-utils/resources/template';
 import { isDataSourceCloning } from '@kubevirt-utils/resources/template/hooks/useVmTemplateSource/utils';
 import { NO_DATA_DASH } from '@kubevirt-utils/resources/vm/utils/constants';
 import { ResourceLink, RowProps, TableData } from '@openshift-console/dynamic-plugin-sdk';
-import { Label, Split, SplitItem } from '@patternfly/react-core';
+import { Label } from '@patternfly/react-core';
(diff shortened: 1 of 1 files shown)


hstastna:

upalatucci (approved):

src/views/bootablevolumes/list/components/BootableVolumesRow.tsx
```diff
@@ -29,19 +29,14 @@ const BootableVolumesRow: FC<
...
-          {obj.kind === DataSourceModel.kind && isDataSourceCloning(obj as V1beta1DataSource) && (
-            <SplitItem>
-              <Label>{t('Clone in progress')}</Label>
-            </SplitItem>
-          )}
-        </Split>
+        <ResourceLink
+          groupVersionKind={getBootableVolumeGroupVersionKind(obj)}
+          name={obj?.metadata?.name}
+          inline
```
avivtur, line 35: can you also add namespace prop to here? if you click on any pvc it moves to it's page and getting a 404 error for a second

hstastna: Ok. PTAL: https://github.com/kubevirt-ui/kubevirt-plugin/pull/1324

PR #1327, opened 2023-06-07
Bug 2207468: Clone vm without CR

upalatucci:
## 📝 Description

Bug: https://bugzilla.redhat.com/show_bug.cgi?id=2207468
The main issue was that the `k8sGet` method return a CR even if gets called with `undefined` name.
We should not clone the CR if the VM has no instancetype or preference.
So guard on this situation and make the CRs requests in parallel as they are not related to each other.

### DIFF
--- a/src/views/virtualmachines/actions/components/CloneVMModal/CloneVMModal.tsx
+++ b/src/views/virtualmachines/actions/components/CloneVMModal/CloneVMModal.tsx
@@ -87,21 +87,25 @@ const CloneVMModal: React.FC<CloneVMModalProps> = ({ vm, isOpen, onClose }) => {
       await stopVM(vm);
     }
 
-    const cloneRevisionInstanceType = await cloneControllerRevision(
-      updatedVM?.spec?.instancetype?.revisionName,
-      updatedVM.metadata.namespace,
-      vm.metadata.namespace,
-    );
-    const cloneRevisionPreference = await cloneControllerRevision(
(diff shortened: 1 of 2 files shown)


src/views/virtualmachines/actions/components/CloneVMModal/CloneVMModal.tsx
```diff
@@ -87,21 +87,29 @@ const CloneVMModal: React.FC<CloneVMModalProps> = ({ vm, isOpen, onClose }) => {
...
+      ),
+    ]);
 
     const createdVM = await k8sCreate({ model: VirtualMachineModel, data: updatedVM });
 
-    await updateControllerRevisionOwnerReference(cloneRevisionInstanceType, createdVM);
-    await updateControllerRevisionOwnerReference(cloneRevisionPreference, createdVM);
+    await Promise.all([
+      cloneRevisionInstanceType
+        ? updateControllerRevisionOwnerReference(cloneRevisionInstanceType, createdVM)
```
metalice, line 107: Maybe in updateControllerRevisionOwnerReference check the first arg exists if not return resolve?

upalatucci: fixed

metalice:
they want it backported so please do it once its merged https://bugzilla.redhat.com/show_bug.cgi?id=2207468

PR #1330, opened 2023-06-08
CNV-28829: Enable IT feature modal - part 2

avivtur:
## 📝 Description

Introducing a new modal that opens only on Virtualization -> Catalog -> InstanceTypes tab if the IT feature is disabled (feature is disabled by default) - The modal purpose is allowing only cluster-admin users to enable this feature to all users. closing this modal without enabling the feature will redirect the user to templates catalog.

TODO: update Learn more link when 4.14 docs are ready

## 🎥 Demo

non-priviliged:
non-admin-modal

cluster-admin:
admin-modal


src/views/catalog/EnableInstanceTypeTechPreviewModal/EnableInstanceTypeTechPreviewModal.tsx
```diff
@@ -0,0 +1,143 @@
...
+type EnableInstanceTypeTechPreviewModalProps = {
+  navigateToCatalog: () => void;
+};
+
+const EnableInstanceTypeTechPreviewModal: FC<EnableInstanceTypeTechPreviewModalProps> = ({
+  navigateToCatalog,
+}) => {
+  const { t } = useKubevirtTranslation();
+
+  const { instanceTypesEnabled, loading, toggleInstanceTypesFeature, error } = usePreviewFeatures();
```
metalice, line 33: on each modal open it will make the 3 network calls?

avivtur: no, only if CM is not found on the first-time admin visits this tab it will run 4 network calls, 1 for watch CM and 3 for create, and if exists only 1 call for watching the CM

src/views/catalog/EnableInstanceTypeTechPreviewModal/EnableInstanceTypeTechPreviewModal.tsx
```diff
@@ -0,0 +1,149 @@
...
+              </Split>
+            </StackItem>
+          </Stack>
+        </SplitItem>
+        <SplitItem isFilled />
+        <SplitItem>
+          <img
+            className="pf-u-pr-lg pf-u-pb-lg"
+            src={instanceTypeTechPreviewModalImage}
+            width={320}
```
hstastna, line 141: I'd suggest to explicitly provide also height of the image. It helps browsers to load the images faster. You could add a css class where you specify it, no need to add width as `img` element prop. But it's up to you.

Also I am not sure what exactly those patternfly classes mean, but those are not saying info about the dimensions of the image for sure. Maybe padding :) Maybe it could be all in some css file?

src/views/catalog/EnableInstanceTypeTechPreviewModal/EnableInstanceTypeTechPreviewModal.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+  const onClose = useCallback(() => {
+    setIsOpen(false);
+    navigateToCatalog();
+  }, [navigateToCatalog]);
+
+  return (
+    <Modal
+      header={
+        <Title headingLevel="h1">{t('Creating a VirtualMachine from an InstanceType')}</Title>
```
hstastna, line 57: I think that "Creating **a** VirtualMachine" or "an InstanceType" is not very consistent with the rest of the text of our UI. But I've better commented the design doc, as this is more for a discussion with the UX/doc team. Maybe this is gonna change in the future 😇 😅

src/views/catalog/CreateFromInstanceTypes/components/SelectInstanceTypeSection/components/RedHatSeriesMenuCard/RedHatSeriesMenuCard.tsx
```diff
@@ -69,31 +69,32 @@ const RedHatSeriesMenuCard: FC<RedHatSeriesMenuCardProps> = ({
...
-        >
-          <Card className="instance-type-series-menu-card__toggle-card">
-            <div className="instance-type-series-menu-card__card-icon">
-              <Icon />
-            </div>
-            <CardBody>
-              <div className="instance-type-series-menu-card__card-title">{seriesTitle}</div>
-              <div className="instance-type-series-menu-card__card-toggle-text">
-                {seriesLabel || classAnnotation} <AngleDownIcon />
+        <div className="instance-type-series-menu-card">
```
hstastna, line 72: Hmm, what is the reason of adding this extra div with this class here? I don't see any reason for it... 
The change you've added `color: var(--pf-global--Color--100);` is applicable only for "instance-type-series-menu-card__toggle-card" class.

src/views/catalog/CreateVMHorizontalNav/CreateVMHorizontalNav.tsx
```diff
@@ -19,22 +20,24 @@ const CreateVMHorizontalNav: FC<RouteComponentProps<{ ns: string }>> = ({
   match,
 }) => {
   const { t } = useKubevirtTranslation();
+
   const [currentTab, setCurrentTab] = useState<CREATE_VM_TAB>(
-    location.pathname.includes(CREATE_VM_TAB.INSTANCE_TYPES)
+    location.pathname.endsWith(CREATE_VM_TAB.INSTANCE_TYPES)
```
hstastna, line 25: 👍🏽

src/views/catalog/EnableInstanceTypeTechPreviewModal/EnableInstanceTypeTechPreviewModal.tsx
```diff
@@ -0,0 +1,149 @@
...
+              <Text component={TextVariants.p}>to the settings tab in the overview page.</Text>
+            </Text>
+          </StackItem>
+        </Stack>
+      </Trans>
+      <br />
+      <Split>
+        <SplitItem>
+          <Stack hasGutter>
+            <StackItem>
```
hstastna, line 94: I'd better avoid using `Split` and `Stack` components like this. I think it is unnecessary to achieve the expected look.

When I was implementing Welcome modal, I used `Grid` component to divide the space into 2 columns. In your case, the 2 links could go to the 1st column, and the image to the second one. Simple as that. Maybe I also needed to use `Stack` component to get better look, but only one time. The result was much simple, more readable code.

With this approach you could also avoid using `<br />` a few lines above. Using `Grid` + eventually some css if needed would do the thing. You don't have to do it, take it just as my suggestion ;) Thanks.

hstastna (commented):
Just some nitpicks, otherwise great work, Aviv! 👍🏽

src/views/catalog/EnableInstanceTypeTechPreviewModal/EnableInstanceTypeTechPreviewModal.tsx
avivtur: > I'd suggest to explicitly provide also height of the image. It helps browsers to load the images faster. 
> You could add a css class where you specify it, no need to add width as `img` element prop. But it's up to you.

will add the height, just a quick note - you're right that it might help to load faster, but there's some some difference between adding the dimensions as props or as css attr - as props we get slightly better performence than in css file, but css file allows to handle dynamic sizes , I will add as prop for now.

> Also I am not sure what exactly those patternfly classes mean, but those are not saying info about the dimensions of the image for sure. Maybe padding :) Maybe it could be all in some css file?

both classes is for padding-right and padding bottom I know that these css classes [from PF](https://www.patternfly.org/v4/utilities/spacing/#usage) are not easy to read, but this is planned to be backported to 4.13.1 and I wanted to simplify the backport process by having less files to backport, hope that makes sense to you

src/views/catalog/EnableInstanceTypeTechPreviewModal/EnableInstanceTypeTechPreviewModal.tsx
avivtur: responded in the doc :)

src/views/catalog/CreateFromInstanceTypes/components/SelectInstanceTypeSection/components/RedHatSeriesMenuCard/RedHatSeriesMenuCard.tsx
avivtur: I made a lot of changes with css in this PR, probably just forgot to remove this, thanks :)

src/views/catalog/EnableInstanceTypeTechPreviewModal/EnableInstanceTypeTechPreviewModal.tsx
avivtur: the welcome modal is looking a bit different than this one by design, I tried to use grid but it wasn't fitting as I wanted and needed to make more css modifications that it's not worth it. about readability, I can make this component a bit easier on the eyes with extracting components to separated files

PR #1331, opened 2023-06-08
CNV-28953: Virtualization Settings section enhancements - part 1

avivtur:
## 📝 Description

Changing Overview -> Settings tab layout and adding a new section to opt-in and out of the IT feature.

[related UX doc](https://docs.google.com/document/d/1bsz4QjvDMumv3AUaigLXNZDpNBa49U9ccRls_9rcWng/edit#)

[Jira issue](https://issues.redhat.com/browse/CNV-28953)

## 🎥 Demo

enhanced-settings-tab3-include-feature-checkbox
enhanced-settings-tab2
enhanced-settings-tab
enhanced-settings-tab4


src/views/clusteroverview/SettingsTab/ClusterTab/ClusterTab.tsx
```diff
@@ -0,0 +1,63 @@
...
+import TemplatesProjectTab from '../TemplatesProjectTab/TemplatesProjectTab';
+
+import ExpandSection from './ExpandSection';
+
+import './ClusterTab.scss';
+
+const ClusterTab: FC = () => {
+  const { t } = useKubevirtTranslation();
+
+  const { instanceTypesEnabled, loading, toggleInstanceTypesFeature } = usePreviewFeatures();
```
metalice, line 20: Hmm it don't feel comfortable to me that this hook need to be combined with admin hook all the time. Maybe add useadmin to this hook and return canEdit or something like that?
Another thing is that this hook is very specific to instance type which in my opinion is a miss. As tomorrow can be another feature with preview tech that we may want to enable or disable. Think about making this hook more generic by getting a key and returning based on that key. So in this configmap not only one preview tech can live.

metalice: This can be done in a follow up or for sure

src/views/clusteroverview/SettingsTab/ClusterTab/ClusterTab.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  const [isChecked, setIsChecked] = useState(false);
+
+  useEffect(() => {
+    if (!loading) {
+      setIsChecked(instanceTypesEnabled);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
avivtur: removed :)

src/views/clusteroverview/SettingsTab/ClusterTab/ClusterTab.tsx
avivtur: you're right, modifed to accept a featureName as key to have a more generic approach, also added isAdmin hook inside, thank you very much! :)

PR #1333, opened 2023-06-08
CNV-29237: Enable Restore template settings" in Catalog CPU modal

hstastna:
## 📝 Description

**Fixes:**
https://issues.redhat.com/browse/CNV-29237

Enable _Restore template settings_ button displayed in _Edit CPU | Modal_ accessible in _Catalog_ drawer and make it work correctly, by providing a correct template namespace to get the original CPU/Memory data, needed for restoring template CPU/Memory values.

_Notes:_
_src/utils/components/CPUMemoryModal/constants.ts_ file was deleted as it has no use anymore. It contained only one constant used in only [one place](https://github.com/kubevirt-ui/kubevirt-plugin/compare/main...hstastna:kubevirt-plugin:BZ_Restore_template_settings_is_disabled_catalog?expand=1#diff-604e7268a1183adcb2fe6c4ae607baf9b22253a3a0856c27675fd94317aa803fL50). We have some other constant with the same string if needed in future. No need to define more constants containing `'openshift'`.

Also note that `COMMON_TEMPLATE_DEFAULT_NAMESPACE` was replaced by `DEFAULT_NAMESPACE` because if the namespace is not provided at all, usually `'default'` one is the correct one, not `'openshift'`. However this approach was not possible to use in _Catalog_ drawer and the template namespace had to be provided explicitly. We could avoid providing the template namespace explicitly but we would have to do some other (more expensive) operations in Catalog drawer instead, like processing template again or already creating VM, which I consider unnecessary.

## 🎥 Screenshots
**Before:**
The button is disabled, not possible to restore the values to the original template's ones:
edit_before

**After:**
The button is enabled and clicking on it changes the values to the original ones (displayed in the drawer - orange arrow):
edit_after

### DIFF
--- a/src/utils/components/CPUMemoryModal/CpuMemoryModal.tsx
+++ b/src/utils/components/CPUMemoryModal/CpuMemoryModal.tsx
@@ -2,6 +2,7 @@ import React, { ChangeEvent, FC, useEffect, useMemo, useState } from 'react';
 import produce from 'immer';
 
 import { V1VirtualMachine, V1VirtualMachineInstance } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import { DEFAULT_NAMESPACE } from '@kubevirt-utils/constants/constants';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
 import { getLabel } from '@kubevirt-utils/resources/shared';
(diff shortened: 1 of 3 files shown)


hstastna:

upalatucci (approved):

PR #1337, opened 2023-06-09
CNV-29304: Make more space for VirtualMachine details in step 3

hstastna:
## 📝 Description

**Fixes:**
https://issues.redhat.com/browse/CNV-29304

Make more space for "VirtualMachine details" - step 3 of VM creation from InstanceTypes (_Catalog > InstanceTypes_ tab), especially for a long VM name and small screens.

## 🎥 Screenshots
**Before:**
vm_before

**After:**
vm_after


hstastna:

upalatucci (approved):

pcbailey:
It's hard to tell since the text in the field is different, but it looks like the implementation is wider than the mockup provided by Yifat. Could you please take another screenshot with the text from her example 'my-awesome-virtual-machine' so we can compare?

hstastna:
> It's hard to tell since the text in the field is different, but it looks like the implementation is wider than the mockup provided by Yifat. Could you please take another screenshot with the text from her example 'my-awesome-virtual-machine' so we can compare?

**After:**
my-awesome

PR #1344, opened 2023-06-12
CNV-28377: Add help text for volume creation and empty state

hstastna:
## 📝 Description

**Fixes:**
https://issues.redhat.com/browse/CNV-28377

_Design doc:_
https://docs.google.com/document/d/1R2TY2fbXRxxWIikK-ssava27kI6s6tvEE0kJpSLsxM4/edit#

Add **help icon "?" displaying popover** with the following help text to help user to add volumes to boot from when creating VM from InstanceTypes:
_"Select a volume (image) to start your VirtualMachine from. You can click the “Add volume” button to add another OS image that doesn’t appear on the list."_

Add **missing empty state** to explain to the user how to add a volume to boot from:
_"No volumes found. Click the “Add volume” button to add a volume to boot from."_

Hide the unnecessary list components when displaying the empty state: filtering, pagination, "Show all" button and list column names ("Volume project" and "Add volume" button remain displayed).

## 🎥 Screenshots
**Before:**
popo_before
If no volumes available:
before_emptystate

**After:**
Help icon "?" displaying popover with the appropriate help text:
popo_after

Empty state (unnecessary filtering, pagination, "Show all" and names of the columns are hidden):
emptystate

---

TODO:
- [x] wait for Yifat to agree on the proposed design (by me) for the empty state, as the original one in the design doc does not look consistent with other empty states we display in our UI
- [x] discuss w Yifat the need of displaying pagination and names of the columns for empty state
- [x] remove empty state for volumes
- [x] add empty state for volumes again, because of the design changes


src/views/catalog/CreateFromInstanceTypes/components/BootableVolumeList/BootableVolumeList.tsx
```diff
@@ -151,20 +160,29 @@ const BootableVolumeList: FC<BootableVolumeListProps> = ({
...
+                key={getName(bs)}
+              />
+            ))
+          ) : (
+            <EmptyState>
+              <Trans ns="plugin__kubevirt-plugin" t={t}>
+                <Title headingLevel="h3">No volumes found</Title>
+                Click the <b> Add volume</b> button to add a volume to boot from
+              </Trans>
+            </EmptyState>
```
avivtur, line 184: instead of putting this empty state here I would make before the return statment the following:
if(isEmpty(bootableVolumes) && loaded) return <EmptyState>....;
This way when there are no bootable volumes we hide the pagination and filters as well, which is better as they are not needed if there are not volumes

src/views/catalog/CreateFromInstanceTypes/CreateFromInstanceType.scss
```diff
@@ -24,6 +24,16 @@
   }
   &__header {
     font-size: var(--pf-global--FontSize--xl);
+
+    .help-icon__cursor {
+      color: var(--pf-global--Color--dark-100);
+      font-size: var(--pf-global--FontSize--sm);
+      vertical-align: 0rem !important;
+    }
```
avivtur, line 32: why do we want to change the cursor?

src/views/catalog/CreateFromInstanceTypes/CreateFromInstanceType.tsx
```diff
@@ -29,8 +30,19 @@ const CreateFromInstanceType: FC = () => {
...
             <List className="create-vm-instance-type-section__list" isPlain>
               <SectionListItem
+                headerText={
+                  <>
+                    {t('Select volume to boot from')}{' '}
+                    <HelpTextIcon
+                      bodyContent={t(
+                        'Select a volume (image) to start your VirtualMachine from. You can click the "Add volume" button to add another OS image that doesn\'t appear on the list.',
+                      )}
+                      maxWidth="425px"
```
avivtur, line 40: can you please check if it's possible to use the helpIconClassName prop instead of adding this maxWidth prop?

src/views/catalog/CreateFromInstanceTypes/components/BootableVolumeList/BootableVolumeList.tsx
hstastna: Actually, I've realized that the design doc really doesn't show filtering, but it does show pagination and project name, plus it shows names of the columns. I have no idea why and also how to achieve this expected look :/

src/views/catalog/CreateFromInstanceTypes/CreateFromInstanceType.scss
hstastna: We don't change the cursor even when it looks like that. We change size and color of "?" icon with this change, to achieve the expected look as in the design doc.

src/views/catalog/CreateFromInstanceTypes/CreateFromInstanceType.tsx
hstastna: Yes, I did and it didn't work as we think it could.

src/views/catalog/CreateFromInstanceTypes/components/BootableVolumeList/BootableVolumeList.tsx
hstastna: @avivtur and WDYT about project name? should we display it for empty state or not? And what about names of the columns?

avivtur: if there's no table I think all of those are not needed, including pagination and project name, I think you tell that to Yifat she would agree with you :)

hstastna: So it was decided to hide everything but the project name. It was discussed with Yifat and also with Ronen. The update is in the design doc already. I'm gonna try to implement this decision asap.

hstastna:

src/views/catalog/CreateFromInstanceTypes/CreateFromInstanceType.tsx
hstastna: So finally I was able to achieve this by using `!important` in the related css file and adding appropriate class to `HelpTextIcon` component. 
Note that suggested using `helpIconClassName` is not appropriate as it is related to the icon itself, not the text that is showed after the icon is clicked.

src/views/catalog/CreateFromInstanceTypes/CreateFromInstanceType.scss
```diff
@@ -36,4 +42,7 @@
     --pf-c-divider--BorderWidth--base: var(--pf-global--spacer--form-element);
     --pf-c-divider--after--BackgroundColor: var(--pf-global--palette--black-200);
   }
+  &__HelpTextIcon {
+    max-width: 425px !important;
```
metalice, line 46: same as above

src/views/catalog/CreateFromInstanceTypes/components/BootableVolumeList/BootableVolumeList.tsx
```diff
@@ -99,74 +108,89 @@ const BootableVolumeList: FC<BootableVolumeListProps> = ({
...
+        <EmptyState className="bootable-volume-list-empty-state">
+          <Trans ns="plugin__kubevirt-plugin" t={t}>
+            <Title headingLevel="h3">No volumes found</Title>
+            Click the <b> Add volume</b> button to add a volume to boot from
+          </Trans>
+        </EmptyState>
+      )}
     </>
   );
 };
```
metalice, line 196: this file is getting 200 lines of code almost, please consider splitting it up

src/views/catalog/CreateFromInstanceTypes/CreateFromInstanceType.scss
```diff
@@ -24,6 +24,12 @@
   }
   &__header {
     font-size: var(--pf-global--FontSize--xl);
+
+    .help-icon__cursor {
+      color: var(--pf-global--Color--dark-100);
+      font-size: var(--pf-global--FontSize--sm);
+      vertical-align: 0 !important;
```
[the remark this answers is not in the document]
hstastna: Without it, we are unable to achieve the expected look, unfortunately.

src/views/catalog/CreateFromInstanceTypes/CreateFromInstanceType.scss
hstastna: Without it, we are unable to achieve the expected look, unfortunately.

src/views/catalog/CreateFromInstanceTypes/CreateFromInstanceType.scss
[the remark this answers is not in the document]
metalice: ok, in looked in this is inline styling which we can override with CSS unless we are using important but we can change the inline styling, but just leave it like this as the CSS is scoped. btw when seeing the difference with and without vertical-aling: 0 i think its better from an appearance perspective to not add vertical-align:0 , but this is up to u

src/views/catalog/CreateFromInstanceTypes/components/BootableVolumeList/BootableVolumeList.tsx
```diff
@@ -99,74 +108,89 @@ const BootableVolumeList: FC<BootableVolumeListProps> = ({
...
-      </TableComposable>
+          </Tbody>
+        </TableComposable>
+      ) : (
+        <EmptyState className="bootable-volume-list-empty-state">
+          <Trans ns="plugin__kubevirt-plugin" t={t}>
+            <Title headingLevel="h3">No volumes found</Title>
+            Click the <b> Add volume</b> button to add a volume to boot from
+          </Trans>
+        </EmptyState>
```
avivtur, line 192: @hstastna This empty state is no longer required for this version, you can remove it.
Please read last comment by UX in here: https://docs.google.com/document/d/1R2TY2fbXRxxWIikK-ssava27kI6s6tvEE0kJpSLsxM4/edit?disco=AAAAzAgx0g8

src/views/catalog/CreateFromInstanceTypes/CreateFromInstanceType.scss
hstastna: I checked that for Firefox browser, it doesn't work without `!important`, and I haven't found any other way to achieve the expected look, except using a component prop `maxWidth` instead (but I was told not to do so, and better to style it in css file). However, in Chrome, it works without `!important`. 

Another way is to omit this styling totally, but the popover text container will be shorter and looking less nice, not as recommended via design doc.

src/views/catalog/CreateFromInstanceTypes/components/BootableVolumeList/BootableVolumeList.tsx
hstastna: Hmm, changes again? 🤦🏽

src/views/catalog/CreateFromInstanceTypes/components/BootableVolumeList/BootableVolumeList.tsx
```diff
@@ -77,6 +78,8 @@ const BootableVolumeList: FC<BootableVolumeListProps> = ({
     }));
   };
 
+  const displayVolumes = !isEmpty(bootableVolumes) && loaded && false;
```
avivtur, line 81: I think you missed removing the `&& false` at the end here

PR #1345, opened 2023-06-13
Adding support for kubevirt proxy pod

metalice:
## 📝 Description

Adding support for kubevirt proxy pod including controlling URL for dev or prod and websockets

## 🎥 Demo


src/utils/hooks/useKubevirtDataPod/useKubevirtDataPod.ts
```diff
@@ -0,0 +1,105 @@
...
+  const socket = useWebSocket<{ object: K8sResourceCommon; type: string }>(
+    `ws://${window.location.host}${url}`,
+    {
+      onClose: () => console.log('websocket closed kubevirt: ', url),
+      onOpen: () => console.log('websocket open kubevirt: ', url),
+      queryParams: {
+        cluster: 'local-cluster',
+        resourceVersion,
+        watch: 'true',
+        // fieldSelector: 'metadata.name=?',
```
metalice, line 45: its ok

src/utils/hooks/useKubevirtDataPod/hooks/useKubevirtDataPodHealth.ts
```diff
@@ -0,0 +1,25 @@
+import { useEffect, useRef } from 'react';
+
+import { consoleFetch } from '@openshift-console/dynamic-plugin-sdk';
+
+import { PROXY_KUBEVIRT_URL, PROXY_KUBEVIRT_URL_HEALTH_PATH } from '../utils/constants';
+
+const useKubevirtDataPodHealth = () => {
```
avivtur, line 7: possible to add return type here?

src/utils/hooks/useKubevirtDataPod/hooks/useKubevirtDataPodHealth.ts
```diff
@@ -0,0 +1,25 @@
...
+  const alive = useRef(null);
```
upalatucci, line 8: Don't we need a `useState` here?
This still works because the  `useKubevirtWatchResource` trigger the re-render several times and in the meantime, this `consoleFetch` resolve before the `useWatchResource` in `useKubevirtWatchResource` could end fetching data. Is there a case where we can have problems?

src/utils/hooks/useKubevirtDataPod/useKubevirtDataPod.ts
```diff
@@ -0,0 +1,105 @@
...
+  K8sResourceCommon,
+  WatchK8sResource,
+} from '@openshift-console/dynamic-plugin-sdk';
+
+import useDeepCompareMemoize from '../useDeepCompareMemoize/useDeepCompareMemoize';
+
+import { constructURL, getResourceVersion, registerResourceVersion } from './utils/utils';
+import useKubevirtDataPodFilters from './useKubevirtDataPodFilters';
+
+type UseKubevirtDataPod = <T extends K8sResourceCommon>(
```
avivtur, line 16: can you use `IoK8sApiCoreV1Pod` instead of `K8sResourceCommon`?

src/utils/hooks/useKubevirtDataPod/useKubevirtDataPod.ts
```diff
@@ -0,0 +1,105 @@
...
+  watchOptions: WatchK8sResource,
+  filterOptions?: { [key: string]: string },
+) => [T, boolean, Error];
+
+const useKubevirtDataPod: UseKubevirtDataPod = <T extends K8sResourceCommon>(
+  watchOptions: WatchK8sResource,
+  filterOptions?: { [key: string]: string },
+) => {
+  const [data, setData] = useState<T>((<unknown>[]) as T);
```
avivtur, line 25: Can you please explain to me why you used `(<unknown>[]) as T`?

src/utils/hooks/useKubevirtDataPod/useKubevirtDataPod.ts
avivtur, line 45: intentional?

src/utils/hooks/useKubevirtDataPod/useKubevirtDataPod.ts
```diff
@@ -0,0 +1,105 @@
...
+  const url = useMemo(
+    () => constructURL(watchOptionsMemoized, query),
+    [query, watchOptionsMemoized],
+  );
+  const [shouldConnect, setShouldConnect] = useState<boolean>(false);
+  const socket = useWebSocket<{ object: K8sResourceCommon; type: string }>(
+    `ws://${window.location.host}${url}`,
+    {
+      onClose: () => console.log('websocket closed kubevirt: ', url),
+      onOpen: () => console.log('websocket open kubevirt: ', url),
```
avivtur, line 40: just making sure, this looks intentional, right?

src/utils/hooks/useKubevirtDataPod/utils/utils.ts
```diff
@@ -0,0 +1,23 @@
...
+  rsMapper[key] = rs > rsMapper[key] ? rsMapper[key] : rs;
+  return;
+};
+export const getResourceVersion = (key: string) => rsMapper[key];
+
+export const constructURL = (watchOptions: WatchK8sResource, query?: string) => {
+  const { groupVersionKind, name, namespace } = watchOptions || {};
+  const baseUrl = `${PROXY_KUBEVIRT_URL}apis/${groupVersionKind?.group}/${groupVersionKind?.version}/`;
+  const namespaceUrl = `${namespace ? `namespaces/${namespace}/` : ''}`;
+  const nameUrl = `${!name ? 's' : `/${name}`}`;
```
avivtur, line 18: why `s`?

src/utils/hooks/useKubevirtDataPod/hooks/useKubevirtDataPodHealth.ts
metalice: yes will add return bool

src/utils/hooks/useKubevirtDataPod/hooks/useKubevirtDataPodHealth.ts
metalice: u mean ref ? 
Because of the ref, it will trigger /health call only once for each load of hook and won't be affected by rerenders. on line 19 I'm checking before calling that ref has no results, ref will be declared and used on all lifecycles of calling component with the same result and isn't affected by rerenders

src/utils/hooks/useKubevirtDataPod/useKubevirtDataPod.ts
metalice: well i looked on IoK8sApiCoreV1Pod , its a completely different animal. why do u think it can be replaced?

src/utils/hooks/useKubevirtDataPod/useKubevirtDataPod.ts
metalice: TS is quite annoying regarding default values which are empty, so example if I set a default value of [object kind of k8scommon] it will pass without need of unknown, but I dont want to give false object, so TS i forcing me to do double casting here

src/utils/hooks/useKubevirtDataPod/useKubevirtDataPod.ts
metalice: yes, same as console, they are printing open and close minus the kubevirt word :)

src/utils/hooks/useKubevirtDataPod/utils/utils.ts
metalice: url of /virtualmachines will give all
but if you want a specific machine, meaning have a name, the URL is changed to /virtualmachine/name

so "s" is for "get me all" I dont use names

src/utils/hooks/useKubevirtDataPod/hooks/useKubevirtDataPodHealth.ts
upalatucci: Cool!!

src/utils/hooks/useKubevirtDataPod/useKubevirtDataPod.ts
upalatucci: `IoK8sApiCoreV1Pod` is the type for the Pod resource. But this hook does not return the Pod information. It uses the dedicated Pods to have filtered, paginated, and sorted resources like VMs.

avivtur: I thought we are creating a proxy pod so you can you Kubernetes API type, but I might be wrong :)

PR #1346, opened 2023-06-13
Fix user setting for non admin users

avivtur:
## 📝 Description

This PR includes these changes to user settings mechanism:
- We create one CM for all users instead of one per user, named `kubevirt-user-settings` on `default` namespace, and the CMs data is looking like: `username: userSettings`
- We create a Role and RoleBinding CRs to allow all users `get` / `list` / `update` / `patch` actions only on the `kubevirt-user-settings` CM.

Current issues: 
- In some edge cases a user is logging into a cluster before any cluster admin is, this  feature won't work, an initial cluster-admin login is required. (We should push to get this CM on Kubevirt deployment from BE on the next version to avoid this issue).
- It is possible for any user to edit this CM manually (not directly through the UI) which can impact other users settings. 

## 🎥 Demo

After:
user-settings-CM

### DIFF
--- a/src/utils/hooks/useKubevirtUserSettings/useKubevirtUserSettings.ts
+++ b/src/utils/hooks/useKubevirtUserSettings/useKubevirtUserSettings.ts
@@ -1,11 +1,17 @@
-import { useEffect, useMemo, useState } from 'react';
+import { useEffect, useState } from 'react';
 
 import {
   ConfigMapModel,
   modelToGroupVersionKind,
+  RoleBindingModel,
+  RoleModel,
   UserModel,
 } from '@kubevirt-ui/kubevirt-api/console';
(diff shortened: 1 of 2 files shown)


src/utils/hooks/useKubevirtUserSettings/useKubevirtUserSettings.ts
```diff
@@ -44,12 +52,26 @@ const useKubevirtUserSettings: UseKubevirtUserSettings = (key) => {
     fetchUserName();
   }, []);
 
+  useEffect(() => {
+    if (userName && !isAdmin) {
```
metalice, line 56: u always try to create to a non admin user before knowing is the other user has default access. i think it will be better to first check if the user has default access , if not trying to create on is own space

avivtur: The solution to creating a project for each user is not relevant anymore, can you please review it again? :)

PR #1347, opened 2023-06-13
CNV-29788: Remove titles from Environment, Disks tab for customizing VM

hstastna:
## 📝 Description

**Fixes:**
https://issues.redhat.com/browse/CNV-29788

Remove titles from _Environment_ and _Disks_ tabs in _Review and create VirtualMachine_ page to make the tabs more consistent with each other. 

Add _The following information is provided by the OpenShift Virtualization operator._ text to _Disks_ tab instead of help text icon displaying the popover containing this text, similarly as it is already done in the _Environment_ tab, to keep the original information available to the user.

## 🎥 Screenshots
**Before**:
disk_before
en_before

**After:**
_Disks_ title removed, text of the original popover added:
disk_after
_Environment_ title removed:
en_after

### DIFF
--- a/src/views/catalog/wizard/tabs/disks/WizardDisksTab.tsx
+++ b/src/views/catalog/wizard/tabs/disks/WizardDisksTab.tsx
@@ -1,7 +1,6 @@
-import * as React from 'react';
+import React from 'react';
 
 import { WizardTab } from '@catalog/wizard/tabs';
-import DiskListTitle from '@kubevirt-utils/components/DiskListTitle/DiskListTitle';
 import DiskModal from '@kubevirt-utils/components/DiskModal/DiskModal';
 import { useModal } from '@kubevirt-utils/components/ModalProvider/ModalProvider';
 import SidebarEditor from '@kubevirt-utils/components/SidebarEditor/SidebarEditor';
@@ -41,8 +40,9 @@ const WizardDisksTab: WizardTab = ({ loaded, tabsData, updateTabsData, updateVM,
           pathsToHighlight={PATHS_TO_HIGHLIGHT.DISKS_TAB}
           resource={vm}
         >
-          <DiskListTitle />
-
+          <Flex className="wizard-disk-tab__flex">
+            {t('The following information is provided by the OpenShift Virtualization operator.')}
+          </Flex>
           <ListPageCreateButton
             onClick={() =>
               createModal(({ isOpen, onClose }) => (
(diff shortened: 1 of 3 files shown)


hstastna:

src/views/catalog/wizard/tabs/disks/WizardDisksTab.tsx
```diff
@@ -41,8 +40,9 @@ const WizardDisksTab: WizardTab = ({ loaded, tabsData, updateTabsData, updateVM,
           pathsToHighlight={PATHS_TO_HIGHLIGHT.DISKS_TAB}
           resource={vm}
         >
-          <DiskListTitle />
-
+          <Flex className="wizard-disk-tab__flex">
```
metalice, line 43: u removed DiskListTitle component, can this file be removed completely? any other component is using it?

hstastna: > u removed DiskListTitle component, can this file be removed completely? 

No.

> any other component is using it?

Yes. VMs and Templates Disks tabs.

metalice: shouldn't we be consistent in those 2 places also? why is only wizard a different?

hstastna: Check the jira ticket, please. Thanks.

metalice: yes i did check this, it's a jira ticket you opened so you should know best. 
Because of this PR we have 2 separate looks one for wizard and second for templates and vms. why is that? In the jura it only referred to a clean wizard. Why no info about maybe fixing templates and vms to be consistent? you had 2 solutions there - make as vms and template or change to a clean look. They go for a clean look, but no referring if it should be applied also for vms and templates - this is exactly why I'm asking here - if this look should be applied to others as well? (vms and templates)

avivtur: @metalice Hi :) 
Yifat and I were also discussing this topic.
the reason for different looks is because under the customize wizard the title is not really needed as each page has a single context which is the tab's text.
However some template and vm tabs have more than one thing in there page (disks tab has disks table and filesystem table for example) and Yifat felt the titles were needed.
I hope that makes sense to you :)

PR #1348, opened 2023-06-14
CNV-26881: Handling bootable volume access

upalatucci:
## 📝 Description

Handle create permissions for user.

If has access to at least DS and PVC, show the create button
Disable the select type based on the permissions

## 🎥 Demo

Screenshot from 2023-06-14 10-52-58
Screenshot from 2023-06-14 10-52-32
Screenshot from 2023-06-14 10-52-13

### DIFF
--- a/src/utils/components/AddBootableVolumeModal/components/SourceTypeSelection/SourceTypeSelection.tsx
+++ b/src/utils/components/AddBootableVolumeModal/components/SourceTypeSelection/SourceTypeSelection.tsx
@@ -1,17 +1,24 @@
-import React, { FC, useCallback, useState } from 'react';
+import React, { FC, useCallback, useEffect, useState } from 'react';
 
 import { t } from '@kubevirt-utils/hooks/useKubevirtTranslation';
+import useCanCreateBootableVolume from '@kubevirt-utils/resources/bootableresources/hooks/useCanCreateBootableVolume';
 import { FormGroup, Select, SelectOption, SelectVariant } from '@patternfly/react-core';
 
 import { DROPDOWN_FORM_SELECTION } from '../../utils/constants';
 
 type SourceTypeSelectionProps = {
   formSelection: DROPDOWN_FORM_SELECTION;
+  namespace: string;
   setFormSelection: (value: DROPDOWN_FORM_SELECTION) => void;
 };
 
-const SourceTypeSelection: FC<SourceTypeSelectionProps> = ({ formSelection, setFormSelection }) => {
+const SourceTypeSelection: FC<SourceTypeSelectionProps> = ({
+  formSelection,
+  namespace,
+  setFormSelection,
+}) => {
   const [isOpen, setIsOpen] = useState(false);
+  const { canCreateDS, canCreatePVC, loading } = useCanCreateBootableVolume(namespace);
 
   const onSelect = useCallback(
     (event, value) => {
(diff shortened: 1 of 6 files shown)


src/utils/components/AddBootableVolumeModal/components/SourceTypeSelection/SourceTypeSelection.tsx
```diff
@@ -22,6 +29,12 @@ const SourceTypeSelection: FC<SourceTypeSelectionProps> = ({ formSelection, setF
     [setFormSelection],
   );
 
+  useEffect(() => {
+    if (!loading && !canCreatePVC) {
+      setFormSelection(DROPDOWN_FORM_SELECTION.USE_REGISTRY);
+    }
+  }, [canCreatePVC, loading, setFormSelection]);
+
```
avivtur, line 37: should we also make same check for canCreateDS?
I think maybe it's best the get the useCanCreateBootableVolume values as props and this way on initial render you would have a correct display of what is disabled what is not

avivtur (commented):
Hi Ugo :) 
Thank you for this PR it looks good.
I didn't see any code changes for this in here, do you intend to disable the `Add volume` button on Catalog > InstanceType page? as you are checking in bootable resource view list

src/utils/resources/bootableresources/hooks/useAccessCreateBootableVolume.ts
```diff
@@ -0,0 +1,35 @@
...
+  });
+
+  const [canCreateDIC, loadingDIC] = useAccessReview({
+    group: DataImportCronModel.apiGroup,
+    namespace: namespace,
+    resource: DataImportCronModel.plural,
+    verb: 'create' as K8sVerb,
+  });
+
+  return [canCreateDS && canCreateDIC, canCreatePVC, loadingPVC || loadingDS || loadingDIC];
```
metalice, line 32: loading shoul be &&
true | true | false will give true and u want to give false..
true & true & false will give right false

upalatucci: With `useAccessReview` you don't get the `loaded` boolean as in `useK8sWatchResource`.
With this hook, you get `loading` as a boolean. So yes, it's an OR operator.

I want that the result is false when all the loading is false. It means that all the access has been fetched successfully

src/utils/components/AddBootableVolumeModal/components/SourceTypeSelection/SourceTypeSelection.tsx
upalatucci: I make this check just because the select has UPLOAD as default value. 
So only in that case we need to change it.

src/views/bootablevolumes/list/BootableVolumesList.tsx
```diff
@@ -43,6 +44,9 @@ const BootableVolumesList: FC<BootableVolumesListProps> = ({ namespace }) => {
   const { createModal } = useModal();
   const { bootableVolumes, error, loaded } = useBootableVolumes(namespace);
 
+  const [canCreateDS, canCreatePVC] = useCanCreateBootableVolume(namespace);
+  const canCreate = canCreateDS || canCreatePVC;
```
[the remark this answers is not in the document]
upalatucci: I'll create the object because i think it's more descriptive. Good idea. But this OR operation is something peculiar about this page and don't think we need that in other places.

upalatucci:
@pcbailey @metalice @avivtur fixed

metalice:
lgtm by me. @pcbailey @avivtur

pcbailey:
lgtm from me. @avivtur

avivtur (approved):
Great work!

PR #1350, opened 2023-06-14
Adding eslint plugin to notify on export but unused objects

metalice:
## 📝 Description
Adding eslint rule to notify on dead code of exported objects but that arent consumed by anything 
## 🎥 Demo


src/utils/utils/humanize.js
avivtur, line 1: this entire file is not used?

metalice: Yes

PR #1354, opened 2023-06-19
CNV-29224: Load balancer config

upalatucci:
## 📝 Description

Add a settings flag to enable Loadbalancer even if MetalLB is not installed

We want the user to be able to create a LoadBalancer service for the vms even when there si a different service backend operator. 
 
 
 As discussed with Aviv, `usePreviewFeatures`  is a generic hook that can handle different feature flags.
 To use it here we renamed it with `useFeatures` and also the config-map used to handle those flags.

The component `EnableFeatureCheckbox` would be useful in the future to add other features checkbox easily

## 🎥 Demo

image


metalice:
There is no check to see if load balancer is available at cluster? user can check/uncheck this even if load balancer isn't existing?

upalatucci:
Yes. I also asked other people and does not seem to be a way to know if the cluster has some backend service for that. We have a specific code to check if MetalLB is installed, but for other backend there is no way.

PR #1361, opened 2023-06-21
Bug 2190448: Disable Resume, Pause buttons on stopped VM in Overview

hstastna:
## 📝 Description

**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2190448
https://issues.redhat.com/browse/CNV-28439

Disable unnecessary _Pause_ button on _Details_ card present in a VM's _Overview_ page for a stopped VM. 
Rename _Resume_ button to _Start_, to be more aligned with VM Actions dropdown.

How the buttons in the kebab menu should look like:
- if VM is stopped -> show Start VM (rename the resume to start), and Pause VM (disabled)
- if VM is running -> show Stop VM and Pause VM both active
- if VM is paused -> you can either Unpause or Stop the VM

## 🎥 Screenshots
**Before:**
Stopped VM:
stop_before

**After:**
Stopped VM:
stop_afterr

_Additional screenshots (there wasn't any visual change using code changes in this PR):_
Running VM:
running_vm

Paused VM:
paused_vm


src/views/virtualmachines/details/tabs/overview/components/VirtualMachinesOverviewTabDetails/components/VirtualMachinesOverviewTabDetailsTitle.tsx
```diff
@@ -46,11 +46,18 @@ const VirtualMachinesOverviewTabDetailsTitle: React.FC<
...
+          >
+            {t('Stop VirtualMachine')}
+          </DropdownItem>,
+          <DropdownItem isDisabled={!isMachinePaused} key="resume" onClick={() => unpauseVM(vm)}>
+            {t('Resume VirtualMachine')}
           </DropdownItem>,
-          <DropdownItem key="pause" onClick={() => (isMachinePaused ? unpauseVM(vm) : pauseVM(vm))}>
-            {isMachinePaused ? t('Unpause VirtualMachine') : t('Pause VirtualMachine')}
+          <DropdownItem isDisabled={!isMachineRunning} key="pause" onClick={() => pauseVM(vm)}>
+            {t('Pause VirtualMachine')}
```
avivtur, line 60: the start VM is removed? why?
That's one of the most common actions (if not the most common of them all) that can be done in a VM

hstastna: If you look closely, you will see that the "Start VirtualMachine" button was not there at all even before this code change. You think start VM was removed because I deleted `startVM` call, which was used for resuming VM - and that is what I consider not appropriate, because there is also `unpauseVM` - I consider that one more accurate.

avivtur: I'm not sure why it was named `Resume` and I honestly don't think that's important.
I think there should be 3 options as there were before. (the SSH copy command should stay always)
If VM is stopped - you should show Start VM (rename the resume to start), and Pause VM (disabled).
If VM is running - you should show Stop VM and Pause VM both active
if VM is paused - you can either Unpause or Stop the VM.
This will also be consistent with current VM actions dropdown

hstastna: Maybe there was some reason to use Resume, originally? I think we'd need some design and discussion with involved people, to make sure we are fixing this the correct way...

metalice: i think i initially implemented this feature, you can always look on git blame :) if I remember correctly I thought it is strange it was a resume and when asked about it they told me that what they currently want. i dont think there is really a big issue to change it to something more appropriate

avivtur (approved):

PR #1370, opened 2023-06-25
Fix InstanceType tab loading state

avivtur:
## 📝 Description

This Pr fixes:
- loading state of InstanceType tab under the catalog - instead of sections load one by one and modal to pop, I add a loading spinner to make the page loading look better
- By getting inside dev tools it was possible to remove the enable IT modal and create a VM, so adding another defense to disable the create button if the feature is disabled.

## 🎥 Demo

Before:

https://github.com/kubevirt-ui/kubevirt-plugin/assets/67270715/397e684a-bc24-422c-aba5-5c5fea09d07d

After:

https://github.com/kubevirt-ui/kubevirt-plugin/assets/67270715/6df48245-d00e-4163-96be-276f1d5d738c

### DIFF
--- a/src/views/catalog/CreateFromInstanceTypes/CreateFromInstanceType.tsx
+++ b/src/views/catalog/CreateFromInstanceTypes/CreateFromInstanceType.tsx
@@ -3,26 +3,59 @@ import { Trans } from 'react-i18next';
 
 import SelectInstanceTypeSection from '@catalog/CreateFromInstanceTypes/components/SelectInstanceTypeSection/SelectInstanceTypeSection';
 import VMDetailsSection from '@catalog/CreateFromInstanceTypes/components/VMDetailsSection/VMDetailsSection';
+import EnableInstanceTypeTechPreviewModal from '@catalog/EnableInstanceTypeTechPreviewModal/EnableInstanceTypeTechPreviewModal';
 import HelpTextIcon from '@kubevirt-utils/components/HelpTextIcon/HelpTextIcon';
+import Loading from '@kubevirt-utils/components/Loading/Loading';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
-import { Card, Divider, Grid, GridItem, List, PopoverPosition } from '@patternfly/react-core';
+import {
+  Bullseye,
+  Card,
+  Divider,
+  Grid,
+  GridItem,
+  List,
+  PopoverPosition,
+} from '@patternfly/react-core';
 
 import AddBootableVolumeButton from './components/AddBootableVolumeButton/AddBootableVolumeButton';
 import BootableVolumeList from './components/BootableVolumeList/BootableVolumeList';
 import CreateVMFooter from './components/CreateVMFooter/CreateVMFooter';
 import SectionListItem from './components/SectionListItem/SectionListItem';
+import useEnableInstanceTypeModal from './hooks/useEnableInstanceTypeModal';
 import { useInstanceTypeVMStore } from './state/useInstanceTypeVMStore';
 import { INSTANCE_TYPES_SECTIONS } from './utils/constants';
 
 import './CreateFromInstanceType.scss';
 
-const CreateFromInstanceType: FC = () => {
+type CreateFromInstanceTypeProps = {
+  isInstanceTypeTab: boolean;
+  navigateToCatalog: () => void;
+};
+
+const CreateFromInstanceType: FC<CreateFromInstanceTypeProps> = ({
+  isInstanceTypeTab,
+  navigateToCatalog,
+}) => {
   const { t } = useKubevirtTranslation();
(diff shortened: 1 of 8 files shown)


src/views/catalog/CreateFromInstanceTypes/CreateFromInstanceType.tsx
```diff
@@ -3,27 +3,59 @@ import { Trans } from 'react-i18next';
...
+  const { loading, ...enableITModalProps } = useEnableInstanceTypeModal(
+    isInstanceTypeTab,
+    navigateToCatalog,
+  );
 
   useEffect(() => resetInstanceTypeVMState(), [resetInstanceTypeVMState]);
 
+  if (loading || !bootableVolumesData?.loaded || !instanceTypesAndPreferencesData?.loaded)
+    return (
+      <Bullseye>
```
metalice, line 54: maybe center this also in the height? like height: 100% or something?

src/views/catalog/CreateFromInstanceTypes/hooks/useEnableInstanceTypeModal.ts
```diff
@@ -0,0 +1,41 @@
+import { useCallback } from 'react';
+
+import { INSTANCE_TYPE_ENABLED } from '@kubevirt-utils/hooks/useFeatures/constants';
+import { useFeatures } from '@kubevirt-utils/hooks/useFeatures/useFeatures';
+
+export type EnableInstanceTypeTechPreviewModalProps = {
```
metalice, line 6: this props or return value?

src/views/catalog/CreateFromInstanceTypes/hooks/useEnableInstanceTypeModal.ts
```diff
@@ -0,0 +1,41 @@
...
+  canEdit: boolean;
+  isOpen: boolean;
+  onClose: () => void;
+  onEnableInstanceTypeFeature: () => void;
+};
+
+type UseEnableInstanceTypeModal = (
+  isInstanceTypeTab: boolean,
+  navigateToCatalog: () => void,
+) => EnableInstanceTypeTechPreviewModalProps & { loading: boolean };
```
metalice, line 16: loading here? why not line 6?

src/views/catalog/CreateFromInstanceTypes/hooks/useEnableInstanceTypeModal.ts
```diff
@@ -0,0 +1,41 @@
...
+) => {
+  const { canEdit, featureEnabled, loading, toggleFeature } = useFeatures(INSTANCE_TYPE_ENABLED);
+
+  const onClose = useCallback(() => {
+    navigateToCatalog();
+  }, [navigateToCatalog]);
+
+  const onEnableInstanceTypeFeature = useCallback(() => {
+    toggleFeature(true);
+  }, [toggleFeature]);
```
metalice, line 30: i dont think usecallbak is adding any values here.. no calculation at all... and especially when line 34 isn't :) can be removed

src/views/catalog/CreateFromInstanceTypes/hooks/useEnableInstanceTypeModal.ts
avivtur: since I wanted to re-use the Props from above, but since I'm changing, I'll change that too :)

src/views/catalog/CreateFromInstanceTypes/CreateFromInstanceType.tsx
avivtur: height 100% is not good as the PF Tab component has a class that I need to override which is causing the pages to have bigger height than wanted, I went for 50vh to get a result that looks fine at different screen heights

src/views/catalog/CreateFromInstanceTypes/hooks/useEnableInstanceTypeModal.ts
avivtur: Actually this saves me a few not needed re-renders, since I pass these as props, the child component will re-render these callback on every child re-render. so it's true it's not crucial here, but I preffered to add this.

avivtur: >and especially when line 34 isn't :) can be removed

Do you suggest to no return these methods at all if the modal is not shown?

metalice: No, I just think, and please test, that it will recalc when props or hook will send updates, so callbacks here don't save anything. Callbacks can save rerenders of useState changes, which affect the whole component rerender, but here u don't have any state, just passing props and hook return, so rerender will occur anyway, usecallback won't save anything but more checks :)

avivtur: It is saving a few initial renders, but it still changes on every state change of the child component so removed

PR #1375, opened 2023-06-28
Bug 2211057: Display only Default templates in Catalog

hstastna:
## 📝 Description

**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2211057

_Jira:_
https://issues.redhat.com/browse/CNV-29281

In Templates catalog, some more than just default templates were shown when clicking on "Default templates". Now only default templates will be displayed with this change, consistently with the templates list that is filtered by "Default templates".

In addition, change "Default Templates" to "Default templates" - sentence case, for consistency reasons. Same for "All Items".

Note that there was a lot of discussion about the expected result of this bug I opened some time ago, various approaches, opinions, but it was decided to simply fix the filter at the end.

## 🎥 Screenshots
Default templates in the Templates list (before; correct number of default templates):
default_templates

**Before:**
Incorrect number and cards displayed for "Default templates":
cat_beforee

**After:**
def_after


hstastna:

avivtur (approved):

PR #1377, opened 2023-06-28
CNV-30327: Add identity flag to ssh command

avivtur:
## 📝 Description

This PR fixes:
- Allow editing the SSH command in the details tab to insert the identity file (ssh private key) path.
- Adding an example of how an SSH command should look like with virtctl.
- Disable the option to copy the SSH command if the VM is missing an SSH secret attached to it.

## 🎥 Demo

Before: 
actions enabled and VM does not have secret with ssh public key:
copy-ssh-disabled-b4
copy-ssh-disabled-b4-2

showing ssh command:
ssh-command-available-b4

After:
show ssh command only if ssh secret is available
Screenshot - 2023-06-29T145932 871

ssh secret not configured:
ssh-command-not-available

action disabled only if vm missing ssh secret configured:
copy-ssh-disabled
copy-ssh-disabled-2

### DIFF
--- a/src/utils/components/SSHAccess/constants.ts
+++ b/src/utils/components/SSHAccess/constants.ts
@@ -13,3 +13,5 @@ export enum SERVICE_TYPES {
 }
 
 export const METALLB_GROUP = 'metallb.io';
+
+export const exampleIdentityFilePath = '--identity-file=/home/jdoe/.ssh/id_rsa';
--- a/src/utils/components/SSHAccess/utils.ts
+++ b/src/utils/components/SSHAccess/utils.ts
@@ -5,7 +5,8 @@ import VirtualMachineInstanceModel from '@kubevirt-ui/kubevirt-api/console/model
 import VirtualMachineModel from '@kubevirt-ui/kubevirt-api/console/models/VirtualMachineModel';
 import { IoK8sApiCoreV1Service } from '@kubevirt-ui/kubevirt-api/kubernetes';
 import { V1VirtualMachine, V1VirtualMachineInstance } from '@kubevirt-ui/kubevirt-api/kubevirt';
-import { ensurePath, getRandomChars } from '@kubevirt-utils/utils/utils';
+import { getCloudInitCredentials } from '@kubevirt-utils/resources/vmi';
+import { ensurePath, getRandomChars, isEmpty } from '@kubevirt-utils/utils/utils';
 import {
   k8sCreate,
   k8sDelete,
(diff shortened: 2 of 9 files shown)


src/utils/components/SSHAccess/utils.ts
```diff
@@ -106,5 +107,14 @@ export const createSSHService = async (
...
-  `virtctl ${vmNamespace ? `-n ${vmNamespace}` : ''} ssh ${userName}@${vmName}`;
+export const getConsoleVirtctlCommand = (vm: V1VirtualMachine, identityFlag?: string) => {
+  const [vmName, vmNamespace, userName] = [
+    getName(vm),
+    getNamespace(vm),
+    getCloudInitCredentials(vm)?.users?.[0]?.name,
+  ];
+
+  return `virtctl -n ${vmNamespace} ssh ${userName}@${vmName} ${
+    !isEmpty(identityFlag) ? identityFlag : '[--identity-file]'
```
apinnick, line 118: I do not think you should put the flag in square brackets because that implies that the flag is optional. When I tested the 'virtctl ssh' commands, I was unable to connect without specifying the '--identity-file' flag. The VM simply could not find the right key without that flag, even if the key was in the default location (/home/user/.ssh). I recommend you change it to something like '--identity-file=<ssh_key>'.

src/utils/components/SSHAccess/constants.ts
```diff
@@ -13,3 +13,5 @@ export enum SERVICE_TYPES {
 }
 
 export const METALLB_GROUP = 'metallb.io';
+
+export const exampleIdentityFilePath = '--identity-file=/home/jdoe/.ssh/id_rsa';
```
upalatucci, line 17: What do you think of:  '--identity-file=~/.ssh/id_rsa'?

upalatucci (commented):
Great Aviv!

src/utils/components/SSHAccess/constants.ts
avivtur: this example was take from the cli help text, so I think it's better to stay synced

src/utils/components/SSHAccess/utils.ts
```diff
@@ -106,5 +107,14 @@ export const createSSHService = async (
...
 };
 
-export const getConsoleVirtctlCommand = (userName: string, vmName: string, vmNamespace?: string) =>
-  `virtctl ${vmNamespace ? `-n ${vmNamespace}` : ''} ssh ${userName}@${vmName}`;
+export const getConsoleVirtctlCommand = (vm: V1VirtualMachine, identityFlag?: string) => {
+  const [vmName, vmNamespace, userName] = [
+    getName(vm),
+    getNamespace(vm),
+    getCloudInitCredentials(vm)?.users?.[0]?.name,
+  ];
```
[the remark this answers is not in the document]
avivtur: it's just a way to be lazy and write const only once :)

metalice:
ok by me, @upalatucci is up to u

upalatucci (approved):

PR #1379, opened 2023-06-29
CNV-29392: VM actions menu jumps when vms on list change state

avivtur:
## 📝 Description

When appending the dropdown to the outer page, every re-render on that page will cause the dropdown to re-render and change location.
When using the Dropdown component from PF we can use the menuAppentTo prop to append the dropdown to the parent, however this presents a scrolling issue on the firefox browser, as the dropdown keeps getting focus and the scorlling would seem stuck. this is a PF issue that needs to be resolved as the focus is coming from the dropdown component.

## 🎥 Demo

Before:

https://github.com/kubevirt-ui/kubevirt-plugin/assets/67270715/483f3437-e26c-4b95-ba1b-5ad3e59fe30c

After:

https://github.com/kubevirt-ui/kubevirt-plugin/assets/67270715/25afa42a-7c70-4758-841c-191d53a770cc

### DIFF
--- a/src/views/virtualmachines/actions/components/VirtualMachineActions/VirtualMachineActions.scss
+++ b/src/views/virtualmachines/actions/components/VirtualMachineActions/VirtualMachineActions.scss
@@ -1,0 +1,5 @@
+.VirtualMachineActions.pf-c-dropdown.pf-m-expanded {
+  ul {
+    transform: translate(-150px, 0);
+  }
+}
--- a/src/views/virtualmachines/actions/components/VirtualMachineActions/VirtualMachineActions.tsx
+++ b/src/views/virtualmachines/actions/components/VirtualMachineActions/VirtualMachineActions.tsx
@@ -1,40 +1,23 @@
-import React, { useState } from 'react';
+import React, { FC, memo, useState } from 'react';
 
-import {
-  V1VirtualMachine,
-  V1VirtualMachineInstanceMigration,
-} from '@kubevirt-ui/kubevirt-api/kubevirt';
 import ActionDropdownItem from '@kubevirt-utils/components/ActionDropdownItem/ActionDropdownItem';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
-import { getContentScrollableElement } from '@kubevirt-utils/utils/utils';
-// import { LazyActionMenu } from '@openshift-console/dynamic-plugin-sdk-internal';
-import { Dropdown, DropdownPosition, DropdownToggle, KebabToggle } from '@patternfly/react-core';
-import useVirtualMachineActionsProvider from '@virtualmachines/actions/hooks/useVirtualMachineActionsProvider';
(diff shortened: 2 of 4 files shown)


hstastna (commented):
The code changes look good to me, make sense. I've tested them both in Chromium and Firefox browsers (up to date). In Chromium it works fine, however in Firefox I wasn't able to reach some actions in the drop down at all, for smaller browser window:

https://github.com/kubevirt-ui/kubevirt-plugin/assets/13417815/be33e74b-7579-44fc-a448-c3af40a02516

IMHO we need to find a way to prevent scrolling back up automatically when trying to reach some actions at the end of the menu.

upalatucci:
`getContentScrollableElement` was there to prevent the issue that Hilda is pointing out.

avivtur:
@upalatucci @hstastna thanks for reminding me that issue :) 
changed the PR and description accordingly please review again

avivtur:
@upalatucci @hstastna @metalice 
Since the behavior of the dropdown is happening only in firefox, and is caused by the dropdown that keeps focusing on the menu only in firefox, I've opened an issue for that in PF https://github.com/patternfly/patternfly-react/issues/9326

hstastna:
> @upalatucci @hstastna @metalice Since the behavior of the dropdown is happening only in firefox, and is caused by the dropdown that keeps focusing on the menu only in firefox, I've opened an issue for that in PF [patternfly/patternfly-react#9326](https://github.com/patternfly/patternfly-react/issues/9326)

Sounds reasonable to me. But as fixing bug in PF usually takes very long time, what would be cost of simply not using PF component (but some other one, maybe implemented by us) in this specific scenario? WDYT, Aviv?

avivtur:
> Sounds reasonable to me. But as fixing bug in PF usually takes very long time, what would be cost of simply not using PF component (but some other one, maybe implemented by us) in this specific scenario? WDYT, Aviv?

@hstastna I'd say it can take at least a day to make a new dropdown component (assuming I can make one based on menu component from PF, if not that can take more to copy all the css)
I also noticed this issue is depending on the window's zoom level (90% and below this bug is not happening for me) so maybe it could be just a known issue that is written in the release notes

upalatucci:
Okay for me @hstastna

hstastna:
> @hstastna I'd say it can take at least a day to make a new dropdown component (assuming I can make one based on menu component from PF, if not that can take more to copy all the css) I also noticed this issue is depending on the window's zoom level (90% and below this bug is not happening for me)...

Of course it depends on the zoom level. And unfortunately, on the size of the window, too. This PR improves the jumping issue a little bit, but still not fixed completely. I'd suggest to merge this PR as is, to contact PF folks to fix this issue in the component asap. 

>  ...so maybe it could be just a known issue that is written in the release notes

Yes, this one, too, please. However I am not sure how to achieve this. Hopefully you know :D

avivtur:
PF issue link: https://github.com/patternfly/patternfly-react/issues/9326

PR #1380, opened 2023-06-29
Fix VM details action crash when vm not loaded yet

avivtur:
## 📝 Description

When getting inside any VM and clicking on actions will crash as the VM is not loaded yet

## 🎥 Demo

Before:

https://github.com/kubevirt-ui/kubevirt-plugin/assets/67270715/9586fd45-b9ac-41ee-8981-8ec3d2123e0a

After:

https://github.com/kubevirt-ui/kubevirt-plugin/assets/67270715/9040270a-ebb4-4b0b-bb26-e331827922db


hstastna (approved):
Nice catch, Aviv! 👍🏽

PR #1392, opened 2023-07-11
Adding ability to filter api server results

metalice:
## 📝 Description

This PR will add the options to filter using our proxy, means filtering happen on BE side,
this is a temp solution to vm list, as this component should be rewritten as the component form console need to have a lot of adjustment in order to filter in BE side.

## 🎥 Demo

filtering-be


orenc1:
updated the CI cluster to ocp 4.13 in https://github.com/openshift/release/pull/41304
/retest

PR #1395, opened 2023-07-11
Bug 2221939: Add help icon with text to SSH key field in  M Sctips tab

hstastna:
## 📝 Description

**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=2221939

_Jira:_
https://issues.redhat.com/browse/CNV-30864

Add "?" help icon with the popover containing the following text to help the user find where to add static public SSH key:
_"Static SSH keys can be added on Overview > Settings > User > Manage SSH keys."_

Add it to "Authorized SSH key" field/section in VM _Configuration > Scripts_ tab, according to the suggestions from the doc team and UX.

## 🎥 Screenshots
**Before:**
ss_before

**After:**
ss_after

### DIFF
--- a/src/utils/components/DescriptionItem/DescriptionItemHeader.tsx
+++ b/src/utils/components/DescriptionItem/DescriptionItemHeader.tsx
@@ -15,6 +15,7 @@ type DescriptionItemHeaderProps = {
   bodyContent: ReactNode;
   breadcrumb?: string;
   descriptionHeader: string;
+  iconContent?: ReactNode;
   isPopover: boolean;
   label?: ReactNode;
   maxWidth?: string;
@@ -25,6 +26,7 @@ export const DescriptionItemHeader: FC<DescriptionItemHeaderProps> = ({
   bodyContent,
   breadcrumb,
   descriptionHeader,
+  iconContent,
   isPopover,
   label,
   maxWidth,
@@ -66,7 +68,7 @@ export const DescriptionItemHeader: FC<DescriptionItemHeaderProps> = ({
 
   return (
     <DescriptionListTerm>
-      {descriptionHeader} {label}
+      {descriptionHeader} {iconContent} {label}
     </DescriptionListTerm>
   );
 };
--- a/locales/en/plugin__kubevirt-plugin.json
+++ b/locales/en/plugin__kubevirt-plugin.json
@@ -985,6 +985,7 @@
   "Start this VirtualMachine in pause mode": "Start this VirtualMachine in pause mode",
   "Start VirtualMachine": "Start VirtualMachine",
   "Start VirtualMachine on clone": "Start VirtualMachine on clone",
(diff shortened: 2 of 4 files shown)


hstastna:

src/utils/components/DescriptionItem/DescriptionItemHeader.tsx
```diff
@@ -15,6 +15,7 @@ type DescriptionItemHeaderProps = {
   bodyContent: ReactNode;
   breadcrumb?: string;
   descriptionHeader: string;
+  hasIcon?: boolean;
```
avivtur, line 18: Hi :) 
nit suggestion: instead of passing hasIcon as boolean, add it as a ReactNode with null default value.
It's hard to guess that `bodyContent` is also meant to be an Icon.

avivtur: The word `content` refers to text, I'd go with just `icon`, that should be clear enough :)

apinnick:
I can see a couple problems, not related to the help text. 

1. The banner text "SSH keys ... applied at first boot" only refers to static keys. Dynamic keys can be applied after first boot. 
2. The user will have no idea what a static key is or how it is different from a dynamic key.
3. The user will have no idea that dynamic keys are only supported by RHEL9.2+. All other OSes have to have static keys applied before first boot.

hstastna:
> I can see a couple problems, not related to the help text.
> 
>     1. The banner text "SSH keys ... applied at first boot" only refers to static keys. Dynamic keys can be applied after first boot.
> 
>     2. The user will have no idea what a static key is or how it is different from a dynamic key.
> 
>     3. The user will have no idea that dynamic keys are only supported by RHEL9.2+. All other OSes have to have static keys applied before first boot.

Hi @apinnick, all these problems do seem to be related to the text displayed in appropriate pages as they all refer to one common thing: not informing the user well. Please, suggest better text for the "banner text" or the popover added to SSH key section, so the problems can be solved. Thanks in advance.

src/utils/components/DescriptionItem/DescriptionItemHeader.tsx
hstastna: Well, I am not sure. It provides icon together with popover text, so using just `icon` doesn't feel to me as good enough. And iconWithPopover feels to me unnecessary long :D

avivtur:
> I can see a couple problems, not related to the help text.
> 
> 1. The banner text "SSH keys ... applied at first boot" only refers to static keys. Dynamic keys can be applied after first boot.
> 2. The user will have no idea what a static key is or how it is different from a dynamic key.
> 3. The user will have no idea that dynamic keys are only supported by RHEL9.2+. All other OSes have to have static keys applied before first boot.

@apinnick Hi Avital, you're right about the banner text is not accurate thanks to the new feature of dynamic SSH key injection. You're also right the it is not related to this PR. Please open a Jira ticket to report the issue about the text in the banner and it will be fixed under that scope, Thanks!

/lgtm

PR #1396, opened 2023-07-12
CNV-20469: Apply activation key and organiztion ID to QuickCreate VM - part2

avivtur:
## 📝 Description

Apply activation key and organization ID to cloud-init config when creating RHEL VM with Quick Create VM flow.
Also Changing the cloud-init volume to Cloud Drive

## 🎥 Demo

rhel-subscribed-vm

upgrade successful:

https://github.com/kubevirt-ui/kubevirt-plugin/assets/67270715/37c29be1-1017-4337-9712-0c5e55f5e47c

### DIFF
--- a/src/utils/components/CloudinitModal/utils/cloudinit-utils.ts
+++ b/src/utils/components/CloudinitModal/utils/cloudinit-utils.ts
@@ -1,13 +1,16 @@
-import { dump, load } from 'js-yaml';
+import produce from 'immer';
+import { dump, load, LoadOptions } from 'js-yaml';
 
 import {
   V1CloudInitConfigDriveSource,
   V1VirtualMachine,
   V1Volume,
 } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import { RHELAutomaticSubscriptionData } from '@kubevirt-utils/hooks/useRHELAutomaticSubscription/utils/types';
 import { getVolumes } from '@kubevirt-utils/resources/vm';
+import { isEmpty } from '@kubevirt-utils/utils/utils';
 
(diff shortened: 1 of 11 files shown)


metalice:
Hey does it supporr config drive and also no cloud ?

avivtur:
> Hey does it supporr config drive and also no cloud ?

Not sure, and is this relevant for this PR? if we will need to switch to a cloud drive or to the opposite, we will probably do it in a separate PR

metalice:
> > Hey does it supporr config drive and also no cloud ?
> 
> Not sure, and is this relevant for this PR? if we will need to switch to a cloud drive or to the opposite, we will probably do it in a separate PR

Yes it does, we currently using both ways and its wrong, so we need to know which type it support

src/utils/components/CloudinitModal/utils/cloudinit-utils.ts
```diff
@@ -100,10 +106,54 @@ export const createDefaultCloudInitYAML = () =>
...
+  const cloudInitVol = getCloudInitVolume(vm);
+  const restVolumes = getVolumes(vm).filter((vol) => vol?.name !== cloudInitVol?.name);
+  const cloudInitVolData = getCloudInitData(cloudInitVol);
+
+  const userDataObject = convertYAMLUserDataObject(cloudInitVolData?.userData);
+
+  const { activationKey, organizationID } = subscriptionData || {};
+
+  const updatedUserDataObject = produce(userDataObject, (draftUserDataObject) => {
+    if (!isEmpty(organizationID) && !isEmpty(activationKey)) {
```
metalice, line 122: i think this check should be much higher-- if its false it can avoid all this convert and functions below

PR #1406, opened 2023-07-17
CNV-29219: Hide namespace related info in Manage columns modal

hstastna:
## 📝 Description

**Fixes:**
https://issues.redhat.com/browse/CNV-29219

Hide unnecessary namespace related info text displayed usually in _Manage columns_ modal, part of `ListPageFilter` component, for list of bootable resources when creating VM from InstanceTypes, as namespace is not relevant for this page. Specifically, hide the following in the modal:

_The namespace column is only shown when in "All projects"_

## 🎥 Screenshots
**Before:**
col_before

**After:**
col_after


hstastna:

metalice:
so no change using css for the columns number?

/lgtm

hstastna:
> so no change using css for the columns number?

@metalice I am not sure. probably in followup PR. Asked the UX in the meantime if we really need to change that text in that modal, waiting for the response. If they say yes, then followup PR. If no, I'd be happy we won't have to do that hacky stuff with our css.

PR #1413, opened 2023-07-20
CNV-31184: Add an option to filter DS and PVC in bootable volumes page

avivtur:
## 📝 Description

Add resource kind filter to bootable volumes page

## 🎥 Demo

Before:
filter_ds_pvc

After:

https://github.com/kubevirt-ui/kubevirt-plugin/assets/67270715/326c5470-ba22-48f8-ad90-dda7ec38adff


metalice:
i think its better changing form DS to DataSource ? make any sense?

avivtur:
> i think its better changing form DS to DataSource ? make any sense?

Since we only have the abbreviation icon at the right UX figured it would be more user-friendly to be DS rather that DataSource, I hope that makes sense to you

PR #1415, opened 2023-07-23
CNV-30889: Displaying disabled dynamic injection and changing the text in the popover

DanaOrr:
## 📝 Description
https://issues.redhat.com/browse/CNV-30889

Before this change we couldn't see the dynamic ssh key injection section in vm creation if the OS doesn't support dynamic ssh key injection.
Now we can see this section disabled for non-supporting OS as well.
Also, if the vm doesn't support dynamic injection, the popover will contain different text.

## 🎥 Demo
Before: 
image

After: 
Popover for an os that supports dynamic ssh key injection:
image

Popover for an os that doesn't support dynamic ssh key injection:
image


src/views/catalog/wizard/tabs/scripts/components/DynamicSSHKeyInjectionWizard.tsx
```diff
@@ -41,22 +44,22 @@ const DynamicSSHKeyInjectionWizard = () => {
...
+              ? t(
+                  'Only RHEL 9 supports dynamic SSH key injection. Instead, you must add an authorized SSH key before the VirtualMachine boots for the first time.',
+                )
+              : t(
+                  'Requires RHEL9 and QEMU guest agent. If enabled, SSH keys can be changed while the VirtualMachine is running. Otherwise, the VirtualMachine inherits the key injection setting of its image.',
+                )
+          }
           className="dynamic-ssh-key"
           position={PopoverPosition.right}
         />
```
metalice, line 60: this is used twice, can be extracted to a component and consumed twice?

locales/en/plugin__kubevirt-plugin.json
```diff
@@ -864,6 +864,7 @@
   "Requested of ": "Requested of ",
   "Required": "Required",
   "Required when automatic update is enabled": "Required when automatic update is enabled",
+  "Requires RHEL9 and QEMU guest agent. If enabled, SSH keys can be changed while the VirtualMachine is running. Otherwise, the VirtualMachine inherits the key injection setting of its image.": "Requires RHEL9 and QEMU guest agent. If enabled, SSH keys can be changed while the VirtualMachine is running. Otherwise, the VirtualMachine inherits the key injection setting of its image.",
```
hstastna, line 867: I can see "RHEL9" is used here as a part of the popover text, but another popover text uses "RHEL 9" (you can see that text a few lines above, line 758) which looks inconsistent. Could you please, find out what is the correct form (maybe asking Avital from doc team or Yifat someone from UX) and update the code? Thanks! :)

src/views/catalog/CreateFromInstanceTypes/components/VMDetailsSection/components/DynamicSSHKeyInjectionIntanceType.tsx
```diff
@@ -11,6 +12,9 @@ import { PopoverPosition } from '@patternfly/react-core';
 const DynamicSSHKeyInjectionIntanceType = () => {
   const { instanceTypeVMState, setInstanceTypeVMState } = useInstanceTypeVMStore();
   const { sshSecretCredentials } = instanceTypeVMState;
+  const isDynamicSSHInjectionLabelExist =
```
hstastna, line 15: `isDynamicSSHInjectionLabelExist` feels to me kinda long and kinda weird. In English, we never ask the way "is something exist?", but "does something exist?" So having verbs "is" and "exist" in one name of the variable can be confusing. We could also ask if something exists like "is something present?" or ".. available"?

So I'd better suggest to name the variable for example `isDynamicSSHInjectionLabelPresent` (even if it is even a bit longer, but makes more sense in English and makes the code clearly understandable), or maybe something like `dynamicSSHKeyLabelExist` or `hasDynamicSSHLabel` (which would say the information that some related resource _has_ the appropriate related label or not). Up to you :) 

Note that this is not super important however I'd recommend to improve the name of this variable a little bit :)

hstastna (commented):
Nice PR, Dana! I like how you've provided a nice and explaining description of the PR + you've also commented the screenshots, so you make it much easier for the reviewer. Thank you! 👍🏽

locales/en/plugin__kubevirt-plugin.json
apinnick: "RHEL 9", not "RHEL9"

src/views/catalog/CreateFromInstanceTypes/components/VMDetailsSection/components/DynamicSSHKeyInjectionIntanceType.tsx
DanaOrr: Ok Hilda 👍

PR #1418, opened 2023-07-23
CNV-31249: Bootable volume list does not listen to namespace bar

avivtur:
## 📝 Description

Fix bootable volumes list to listen to the namespace bar

## 🎥 Demo

Before:
volumes-list-b4

After:
volumes-list

### DIFF
--- a/src/views/bootablevolumes/list/BootableVolumesList.tsx
+++ b/src/views/bootablevolumes/list/BootableVolumesList.tsx
@@ -1,5 +1,5 @@
 import React, { FC, useState } from 'react';
-import { useHistory } from 'react-router-dom';
+import { RouteComponentProps, useHistory } from 'react-router-dom';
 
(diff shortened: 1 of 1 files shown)


src/utils/resources/bootableresources/hooks/useBootableVolumes.ts
```diff
@@ -27,6 +27,7 @@ type UseBootableVolumes = (namespace?: string) => {
 };
 
 const useBootableVolumes: UseBootableVolumes = (namespace) => {
+  console.log('🚀 ~ file: useBootableVolumes.ts:30 ~ namespace:', namespace);
```
upalatucci, line 30: Do you want to leave these `console.log` ?

avivtur: lol removed

PR #1428, opened 2023-07-26
CNV-31216: Back/Next workflow does not persist storage setting

avivtur:
## 📝 Description

on the customize VM flow, we initially see the customize parameters page, if clicking next and than back, the storage changes from taking the template's default to PVC source, and it makes the user to fill the data (and no option to re-select the default option).
In this PR I change the back button to the reset button and fixed the button's behavior to get the initial state.

## 🎥 Demo

After:

https://github.com/kubevirt-ui/kubevirt-plugin/assets/67270715/3a81b9c6-9301-4d25-af04-a2f7eaf3c5ce


src/views/catalog/wizard/tabs/overview/components/VMNameModal/VMNameModal.tsx
```diff
@@ -0,0 +1,52 @@
...
+      onSubmit={onSubmit}
+    >
+      <Form>
+        <FormGroup
+          fieldId="vm-name"
+          helperText={t('Please provide name to VirtualMachine.')}
+          helperTextInvalid={isEmpty(vmName) && t('VirtualMachine name can not be empty.')}
+          isRequired
+          label={t('VirtualMachine name')}
+          validated={isEmpty(vmName) ? ValidatedOptions.default : ValidatedOptions.error}
```
metalice, line 43: this is right?

avivtur: it should be the other way around, nice catch will edit this in a follow-up PR

PR #1434, opened 2023-07-28
CNV-31327: Add option to filter by user created templates

hstastna:
## 📝 Description

**Fixes:**
https://issues.redhat.com/browse/CNV-31327

Add new filter _User templates_ to templates catalog page, along with _Default templates_, so the filters will be the following:
- _All items_
- _Default templates_
- _User templates_

Move `TemplateFilters` type to a separate file _types.ts_.

Additionally, fix _Clear All Filters_ button text to sentence case _Clear all filters_.

TODO:
- [x] fix failing unit test

## 🎥 Demo
**Before:**
Not possible to filter by user templates:
default_before

**After:**

https://github.com/kubevirt-ui/kubevirt-plugin/assets/13417815/3cfaed58-0a96-411a-8251-a05264fdf80d


hstastna:

src/views/catalog/templatescatalog/utils/helpers.ts
```diff
@@ -28,12 +28,14 @@ export const filterTemplates = (templates: V1Template[], filters: TemplateFilter
 
       const defaultVariantFilter = filters?.onlyDefault ? isDefaultVariantTemplate(tmp) : true;
 
+      const userFilter = filters?.onlyUser ? !isDefaultVariantTemplate(tmp) : true;
```
metalice, line 31: Could this work also?
!filters.onlyUser || !isDefaultVariantTemplate

hstastna: Yes, this would work, too. However, I used ternary operator to make the statement more readable, understandable and consistent with other ones, for example look at `defaultVariantFilter` above. I believe, it was implemented and approved in this form for good reasons. Or not?

metalice: well for me its not clearer , as it doing : true which can be simply avoided as it not really needed, also for the lines above and below

hstastna: Well, interesting that somebody approved such code in the past here.

metalice: sometimes we approve and then improve later

hstastna:

PR #1437, opened 2023-08-01
CNV-29095: Do not allow editor without anything

upalatucci:
## 📝 Description

Do not allow cloud-init with no data.

Having a cloud-init with no data should result in removing the cloud-init volume and disk. But this can led to some issues with ssh credentials. 

## 🎥 Demo

Screenshot from 2023-08-01 10-57-02

Screenshot from 2023-08-01 10-57-08

### DIFF
--- a/locales/en/plugin__kubevirt-plugin.json
+++ b/locales/en/plugin__kubevirt-plugin.json
@@ -539,6 +539,7 @@
   "Invalid certificate, please visit the following URL and approve it": "Invalid certificate, please visit the following URL and approve it",
   "Invalid MAC address format": "Invalid MAC address format",
   "Invalid URL": "Invalid URL",
+  "Invalid YAML": "Invalid YAML",
   "Inventory": "Inventory",
   "IOPS total: {{input}}": "IOPS total: {{input}}",
   "IP address": "IP address",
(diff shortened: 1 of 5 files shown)


locales/en/plugin__kubevirt-plugin.json
```diff
@@ -539,6 +539,7 @@
   "Invalid certificate, please visit the following URL and approve it": "Invalid certificate, please visit the following URL and approve it",
   "Invalid MAC address format": "Invalid MAC address format",
   "Invalid URL": "Invalid URL",
+  "Invalid YAML": "Invalid YAML",
```
hstastna, line 542: @apinnick PTAL about this. Do you agree on adding this error message for invalid YAML? Thanks.

src/utils/components/CloudinitModal/CloudInitEditor/CloudInitEditor.tsx
```diff
@@ -1,10 +1,10 @@
-import * as React from 'react';
+import React, { FC, useLayoutEffect, useRef, useState } from 'react';
```
hstastna, line 1: 👍🏽

hstastna (commented):
LGTM, just.. I think Avital should know about texts we add to the UI. So I've pinged her to let her know.

PR #1444, opened 2023-08-02
CNV-28248: Display CPU and memory from VMI on VM Details page

pcbailey:
## 📝 Description

For VMs created from InstanceTypes the CPU and memory on the VM Details page is currently being read from the InstanceType, which leads to a mismatch on the VM Details page if the Instancetype is updated. The value circled on the graph shows the actual value.

## 🎥 Screenshots

The InstanceType used to create the VM starts off having 1 CPU and 4GiB of RAM. Between the "Before InstanceType update" and "After InstanceType update" screenshots, that InstanceType is edited to have 2 CPU and 2GiB of RAM.

#### Before InstanceType update
Selection_251

#### After InstanceType update
Selection_250

#### Before InstanceType update
Selection_252

#### After InstanceType update
Selection_253

### DIFF
--- a/src/utils/resources/vm/hooks/useVMI.ts
+++ b/src/utils/resources/vm/hooks/useVMI.ts
@@ -1,0 +1,29 @@
+import { VirtualMachineInstanceModelGroupVersionKind } from '@kubevirt-ui/kubevirt-api/console';
+import { V1VirtualMachineInstance } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import { useK8sWatchResource } from '@openshift-console/dynamic-plugin-sdk';
+
+type UseVMI = (
+  vmName: string,
+  vmNamespace: string,
+) => {
+  vmi: V1VirtualMachineInstance;
+  vmiLoaded: boolean;
+  vmiLoadError: Error;
+};
+
+const useVMI: UseVMI = (vmName, vmNamespace) => {
+  const [vmi, vmiLoaded, vmiLoadError] = useK8sWatchResource<V1VirtualMachineInstance>({
+    groupVersionKind: VirtualMachineInstanceModelGroupVersionKind,
+    isList: false,
+    name: vmName,
+    namespace: vmNamespace,
+  });
+
+  return {
+    vmi,
+    vmiLoaded,
+    vmiLoadError,
+  };
+};
+
+export default useVMI;
--- a/src/utils/components/CPUMemory/CPUMemory.tsx
+++ b/src/utils/components/CPUMemory/CPUMemory.tsx
@@ -1,20 +1,14 @@
 import React, { FC } from 'react';
 
-import {
-  VirtualMachineClusterInstancetypeModelGroupVersionKind,
-  VirtualMachineInstancetypeModelGroupVersionKind,
-} from '@kubevirt-ui/kubevirt-api/console';
-import {
-  V1beta1VirtualMachineInstancetype,
-  V1InstancetypeMatcher,
-  V1VirtualMachine,
-} from '@kubevirt-ui/kubevirt-api/kubevirt';
+import { V1InstancetypeMatcher, V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
 import MutedTextSpan from '@kubevirt-utils/components/MutedTextSpan/MutedTextSpan';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
+import { getName, getNamespace } from '@kubevirt-utils/resources/shared';
 import { vCPUCount } from '@kubevirt-utils/resources/template/utils';
+import useInstanceType from '@kubevirt-utils/resources/vm/hooks/useInstanceType';
+import useVMI from '@kubevirt-utils/resources/vm/hooks/useVMI';
 import { readableSizeUnit } from '@kubevirt-utils/utils/units';
 import { isEmpty } from '@kubevirt-utils/utils/utils';
-import { useK8sWatchResource } from '@openshift-console/dynamic-plugin-sdk';
 import { Skeleton } from '@patternfly/react-core';
 
 type CPUMemoryProps = {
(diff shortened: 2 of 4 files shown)


src/utils/resources/vm/hooks/useVMIForVM.ts
```diff
@@ -0,0 +1,29 @@
...
+const useVMIForVM: UseVMIForVM = (vmName, vmNamespace) => {
+  const [vmi, vmiLoaded, vmiLoadError] = useK8sWatchResource<V1VirtualMachineInstance>({
+    groupVersionKind: VirtualMachineInstanceModelGroupVersionKind,
+    isList: false,
+    name: vmName,
+    namespace: vmNamespace,
+  });
+
+  return {
+    vmi: vmName === vmi?.metadata?.name && vmNamespace === vmi?.metadata?.namespace && vmi,
```
metalice, line 23: i dont think its necessary as u already sending the name and namespace in the request.. 
btw i recommend the name of the hook to be something like useGetVMI (as this is what it does)
it can also get the VM instead of 2 args and just extract here from metadata

src/utils/resources/vm/hooks/useVMIForVM.ts
```diff
@@ -0,0 +1,29 @@
...
+    vmiLoaded: vmiLoaded,
```
metalice, line 24: can be simply

{vmi, vmiLoaded, vmiLoadError} no need for {[name]: [name]}

src/utils/resources/vm/hooks/useVMIForVM.ts
```diff
@@ -0,0 +1,29 @@
...
+import { V1VirtualMachineInstance } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import { useK8sWatchResource } from '@openshift-console/dynamic-plugin-sdk';
+
+type UseVMIForVM = (
+  vmName: string,
+  vmNamespace: string,
+) => {
+  vmi: V1VirtualMachineInstance;
+  vmiLoaded: boolean;
+  vmiLoadError: any;
```
metalice, line 11: instead of any u can use Error

src/utils/resources/vm/hooks/useInstanceTypeForVM.ts
```diff
@@ -0,0 +1,36 @@
...
+  V1beta1VirtualMachineInstancetype,
+  V1InstancetypeMatcher,
+} from '@kubevirt-ui/kubevirt-api/kubevirt';
+import { isEmpty } from '@kubevirt-utils/utils/utils';
+import { useK8sWatchResource } from '@openshift-console/dynamic-plugin-sdk';
+
+type UseInstanceTypeForVM = (instanceTypeMatcher: V1InstancetypeMatcher) => {
+  instanceType: V1beta1VirtualMachineInstancetype;
+  instanceTypeLoaded: boolean;
+  instanceTypeLoadError: any;
```
metalice, line 15: can use Error type

src/utils/resources/vm/hooks/useInstanceTypeForVM.ts
```diff
@@ -0,0 +1,36 @@
...
+    );
+
+  return {
+    instanceType,
+    instanceTypeLoaded,
+    instanceTypeLoadError,
+  };
+};
+
+export default useInstanceTypeForVM;
```
metalice, line 36: i can recommend changing the name to useGetInstanceType

src/utils/resources/vm/hooks/useVMIForVM.ts
pcbailey: Why not just `useVMI`? The `Get` makes no sense to me with the prefix `use`. I was only using for the `forVM` convention because it's what was being used already.

This hook is use by `useVMIAndPodsForVM` which doesn't have access to the VM. I check its usages and there are components that use it that only have the VM name and namespace that they get from URL parameters.

src/utils/resources/vm/hooks/useVMIForVM.ts
pcbailey: Fixed!

src/utils/resources/vm/hooks/useVMIForVM.ts
pcbailey: Fixed!

src/utils/resources/vm/hooks/useInstanceTypeForVM.ts
pcbailey: Fixed!

src/utils/resources/vm/hooks/useInstanceTypeForVM.ts
pcbailey: I changed it to `useInstanceType`.

src/utils/resources/vm/hooks/useVMIForVM.ts
metalice: the use is mandatory by react to be declared as a hook. other wise we will remove it also :) 

the get simple indicates that its only fetches the vmi, but useVMI can work also.


> This hook is use by useVMIAndPodsForVM which doesn't have access to the VM. I check its usages and there are components that use it that only have the VM name and namespace that they get from URL parameters.

ok

src/utils/resources/vm/hooks/useVMI.ts
```diff
@@ -0,0 +1,29 @@
...
+type UseVMIForVM = (
+  vmName: string,
+  vmNamespace: string,
+) => {
+  vmi: V1VirtualMachineInstance;
+  vmiLoaded: boolean;
+  vmiLoadError: Error;
+};
+
+const useVMIForVM: UseVMIForVM = (vmName, vmNamespace) => {
```
metalice, line 14: forgot the name

src/utils/resources/vm/hooks/useVMI.ts
```diff
@@ -0,0 +1,29 @@
...
+  const [vmi, vmiLoaded, vmiLoadError] = useK8sWatchResource<V1VirtualMachineInstance>({
+    groupVersionKind: VirtualMachineInstanceModelGroupVersionKind,
+    isList: false,
+    name: vmName,
+    namespace: vmNamespace,
+  });
+
+  return {
+    vmi: vmName === vmi?.metadata?.name && vmNamespace === vmi?.metadata?.namespace && vmi,
```
metalice, line 23: only VMI i think is enough , as the filtering of vmname and vmnamespace is done in the request

src/utils/resources/vm/hooks/useVMI.ts
pcbailey: Fixed!

src/utils/resources/vm/hooks/useVMI.ts
pcbailey: Fixed!

PR #1449, opened 2023-08-07
CNV-31619: Adding a popover to Descheduler in wizard

DanaOrr:
## 🎥 Demo
After:
image


src/views/catalog/wizard/tabs/scheduling/components/WizardSchedulingGrid.tsx
```diff
@@ -1,4 +1,5 @@
 import * as React from 'react';
```
metalice, line 1: if u already touching this file please fix all the imports of react

src/views/catalog/wizard/tabs/scheduling/components/WizardSchedulingGrid.tsx
```diff
@@ -99,6 +100,21 @@ const WizardSchedulingGrid: React.FC<WizardSchedulingGridProps> = ({ updateVM, v
...
+                <Trans ns="plugin__kubevirt-plugin" t={t}>
+                  <p>
+                    The descheduler can be used to evict a running pod to allow the pod to be
+                    rescheduled onto a more suitable node.
+                  </p>
+                  <br />
+                  <p>
+                    Note: if VirtualMachine has LiveMigratable=False condition, edit is disabled.
+                  </p>
+                </Trans>
```
metalice, line 114: this is already used in another place, consider making it its own component next to Descheduler component and consume it here and in VirtualMachineSchedulingLeftGrid

DanaOrr: Thank you :)

src/views/virtualmachines/details/tabs/configuration/scheduling/components/VirtualMachineSchedulingLeftGrid.tsx
```diff
@@ -1,7 +1,7 @@
 import * as React from 'react';
```
metalice, line 1: same as above

src/views/virtualmachines/details/tabs/configuration/scheduling/components/VirtualMachineSchedulingLeftGrid.tsx
```diff
@@ -26,7 +26,7 @@ type VirtualMachineSchedulingLeftGridProps = {
   vmi?: V1VirtualMachineInstance;
 };
 
-const VirtualMachineSchedulingLeftGrid: React.FC<VirtualMachineSchedulingLeftGridProps> = ({
+const VirtualMachineSchedulingLeftGrid: FC<VirtualMachineSchedulingLeftGridProps> = ({
```
metalice, line 29: also line 39

hstastna (commented):
Looks good to me! I have just one design concern: IMO it would be good if _LiveMigrable=false_ in the popover had kinda different style than rest of the text, because it does not look good and clear just how it is now. it is not a regular name or so, it is a specific variable with its value, so I think it should be displayed a bit differently. WDYT @yfrimanm?

yfrimanm:
+1 to what @hstastna said. I recommend having the `LiveMigrable=false` in a code block style, or if that's not PF supported,  in italics, to make it stand out from the rest of the text.

hstastna:
> +1 to what @hstastna said. I recommend having the `LiveMigrable=false` in a code block style, or if that's not PF supported, in italics, to make it stand out from the rest of the text.

PTAL @DanaOrr :)

PR #1451, opened 2023-08-07
CNV-31551: Edit add bootable volume

upalatucci:
## 📝 Description

Add bootable volume have a graphical structure different from Upload/PVC to registry.

With this pr I align everything.

- Under *Source details* have only the registry URL and move the Volume name under Destination details (re-suing the same `Volume name` field that Upload/PVC use.
- Remove the Size field and use the storageClassname/size fields.
- Create the DataImportCron with the selected storageClassname

## 🎥 Demo

**Upload/Registry**
Screenshot from 2023-08-07 16-33-55
 

**Before**

Screenshot from 2023-08-07 16-33-51

**After**
Screenshot from 2023-08-07 16-30-34


apinnick:
Just a question about the new screenshot. Is "PVC image" really the correct term?

upalatucci:
@apinnick in other pages we have 'Upload a new file to a PVC' 

image

PR #1452, opened 2023-08-08
Update fetch query filters

metalice:
## 📝 Description

Fix query so request will be consist 

## 🎥 Demo

### DIFF
--- a/src/utils/hooks/useKubevirtDataPod/useKubevirtDataPodFilters.ts
+++ b/src/utils/hooks/useKubevirtDataPod/useKubevirtDataPodFilters.ts
@@ -1,35 +1,31 @@
-import { useEffect, useMemo, useState } from 'react';
+import { useEffect, useState } from 'react';
 
 import { debounce } from '@kubevirt-utils/utils/debounce';
+import { isEmpty } from '@kubevirt-utils/utils/utils';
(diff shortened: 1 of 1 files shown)


src/utils/hooks/useKubevirtDataPod/useKubevirtDataPodFilters.ts
```diff
@@ -1,35 +1,31 @@
...
 import { debounce } from '@kubevirt-utils/utils/debounce';
+import { isEmpty } from '@kubevirt-utils/utils/utils';
 import useQuery from '@virtualmachines/details/tabs/metrics/NetworkCharts/hook/useQuery';
 
 import useDeepCompareMemoize from '../useDeepCompareMemoize/useDeepCompareMemoize';
 
 type UseKubevirtDataPodFilters = (filters: { [key: string]: string }) => any;
 
+const updateQueryString = (setQueryString) =>
+  debounce((url) => setQueryString((prevQuery) => (prevQuery !== url ? url : prevQuery)), 700);
```
avivtur, line 12: can you assign a local const for 700 with a name to it?

metalice: i don't think its necessary here as debounce function has a doc with params explaining, and this value isn't used in somewhere else, it's hard coupled.

avivtur (approved):

PR #1463, opened 2023-08-10
CNV-31906: Deleted secret is showing in overview settings

avivtur:
## 📝 Description

When adding a secret to the user settings and then the secret is deleted there's no indication of that in the UI. 
When creating a VM with the template's catalog, it will still try to apply the key from the settings even if the secret is missing.
Now when trying to apply a non existing secret from the settings, we will check if the secret exist and if not it will be removed from the settings when trying to apply it or in the manage SSH keys section

## 🎥 Demo

Before:
dont-apply-ssh-if-secret-missing-b4
missing-secret-alert-b4

After:

https://github.com/kubevirt-ui/kubevirt-plugin/assets/67270715/defeffa3-e296-4d33-8b49-8de31df69967

https://github.com/kubevirt-ui/kubevirt-plugin/assets/67270715/433c12a9-6bc4-437b-a351-d70bcb355502

### DIFF
--- a/src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerFooter.tsx
+++ b/src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerFooter.tsx
@@ -1,6 +1,7 @@
-import * as React from 'react';
+import React, { FC, useEffect } from 'react';
 
-import { V1Template } from '@kubevirt-ui/kubevirt-api/console';
+import { SecretModel, V1Template } from '@kubevirt-ui/kubevirt-api/console';
+import { IoK8sApiCoreV1Secret } from '@kubevirt-ui/kubevirt-api/kubernetes';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
 import useKubevirtUserSettings from '@kubevirt-utils/hooks/useKubevirtUserSettings/useKubevirtUserSettings';
 import useRHELAutomaticSubscription from '@kubevirt-utils/hooks/useRHELAutomaticSubscription/useRHELAutomaticSubscription';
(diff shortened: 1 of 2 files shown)


hstastna (commented):
@avivtur thank you very much for this PR! 
Just.. I don't understand the fix. IMHO this UX behavior is not the best. If something is deleted/removed, it is gone, and that's it. I don't see any reason to inform the user in the settings page, that something was removed some time in the past.  Hence, I don't see any reason to show the warning in the settings page, that the secret does not exist. IMO better is simply not to show it at all in the page. I don't understand why it ever is there, if it does not exist, how it is possible that we show it still there, if it is already gone, if it was removed from the specific project/namespace. IMO this might be very confusing for our users.
Regarding catalog drawer page, I can see ssh key field is not there anymore, if the key is not available. This part of the fix makes sense to me.

avivtur:
@hstastna Hi
The settings are defined in a ConfigMap CR, hence it's not possible to bind a Secret deletion to edit data on that CM.
Please keep in mind that the CM can also be modified outside the intended flow of the `Manage SSH keys` with the YAML editor or the CLI, hence we should present the warning if the secret is missing for the settings.

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerFooter.tsx
```diff
@@ -32,16 +37,31 @@ type TemplateCatalogDrawerFooterProps = {
...
+    authorizedSSHKeys?.[namespace] && {
+      groupVersionKind: getGroupVersionKindForModel(SecretModel),
+      isList: false,
+      name: authorizedSSHKeys?.[namespace],
+      namespace,
+    },
+  );
+
+  useEffect(() => {
+    if (loadError && authorizedSSHKeys?.[namespace]) {
```
metalice, line 61: can u please add a comment if the secret isn't exisit removing the secret from namespace ?

metalice:
Hey @avivtur added a small comment - please explain @hstastna u have taken a different approach now

PR #1478, opened 2023-08-16
CNV-32114: Remove hard-coded namespace from Overview > Settings paths

pcbailey:
## 📝 Description

The namespace in the Overview > Settings link path for the "Go to the Virtualization Settings tab to enable this feature" link in the Add network interface modal was hardcoded to "rhodes". This PR sets the namespace dynamically.


src/utils/components/PreviewFeatureAlert/PreviewFeatureAlert.tsx
```diff
@@ -28,7 +31,7 @@ const PreviewFeatureAlert: FC<PreviewFeatureAlertProps> = ({ onClose }) => {
         </StackItem>
         <StackItem>
           <span onClick={onClose}>
-            <Link to={OVERVIEW_SETTINGS_TAB_PATH}>
+            <Link to={getOverviewSettingsPath(activeNamespace)}>
```
metalice, line 34: working also with all projects?

pcbailey: This link is only accessible via the Network interface list page in VM details. Is it possible to have a VM in the All Projects namespace? I updated it to work with All projects just in case, but it shouldn't matter.

metalice: well if its vm details page so it must have a ns different from all.
But this improvements I no harm

PR #1482, opened 2023-08-18
CNV-30896: Manage SSH settings keeps loading

avivtur:
## 🎥 Demo

Before:
manage-ssh-loading-nonstop

After:
manage-ssh-loaded

### DIFF
--- a/src/views/clusteroverview/SettingsTab/UserTab/components/ManageSSHKeySection/SSHAuthKeysList/hooks/useSSHAuthKeys.ts
+++ b/src/views/clusteroverview/SettingsTab/UserTab/components/ManageSSHKeySection/SSHAuthKeysList/hooks/useSSHAuthKeys.ts
@@ -56,15 +56,12 @@ const useSSHAuthKeys: UseSSHAuthKeys = () => {
           }),
         );
 
-        filterAuthRows(authRows)
-          .then((filteredRows) => {
-            setAuthKeyRows(filteredRows);
-          })
-          .finally(() => {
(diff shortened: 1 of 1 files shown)


src/views/clusteroverview/SettingsTab/UserTab/components/ManageSSHKeySection/SSHAuthKeysList/hooks/useSSHAuthKeys.ts
```diff
@@ -56,15 +56,12 @@ const useSSHAuthKeys: UseSSHAuthKeys = () => {
...
-          })
-          .finally(() => {
-            setLoading(false);
-          });
+        filterAuthRows(authRows).then((filteredRows) => {
+          setAuthKeyRows(filteredRows);
+        });
       }
 
+      setLoading(false);
```
upalatucci, line 64: this way the loading will be called before the `filterAuthRows` promises get fulfilled or rejected

avivtur: you're right I still need the finally as well as I need it here in case of no keys existing

avivtur: But since it's merged already I'll add it in a follow-up PR

PR #1483, opened 2023-08-21
CNV-32102: Changing text in several different places

DanaOrr:
## 🎥 Demo
Template:
image

InstanceTypes:
image

Bootable volumes:
image
image

Preferences: 
image


locales/en/plugin__kubevirt-plugin.json
```diff
@@ -699,6 +699,7 @@
   "No alerts found": "No alerts found",
   "No available boot source": "No available boot source",
   "No boot source": "No boot source",
+  "No Bootable volumes found": "No Bootable volumes found",
```
apinnick, line 702: If you are using generic nouns and not k8s object names, they should follow sentence case: "No bootable volumes found"

locales/en/plugin__kubevirt-plugin.json
```diff
@@ -724,6 +725,7 @@
   "No optimized StorageProfile settings for this StorageClass.": "No optimized StorageProfile settings for this StorageClass.",
   "No owner": "No owner",
   "No parameters found in this template.": "No parameters found in this template.",
+  "No Preferences found": "No Preferences found",
```
apinnick, line 728: => "No preferences found"

locales/en/plugin__kubevirt-plugin.json
```diff
@@ -732,6 +734,7 @@
   "No Template found": "No Template found",
   "No Template items are being shown due to the filters being applied.": "No Template items are being shown due to the filters being applied.",
   "No Template was selected for review, please go to the catalog and select one.": "No Template was selected for review, please go to the catalog and select one.",
+  "No Templates found": "No Templates found",
```
apinnick, line 737: => "No templates found"

locales/en/plugin__kubevirt-plugin.json
gouyang: @DanaOrr @apinnick If we look at OCP pages, they're all using the uppercase, such as "No Pods found", "No StatefulSets found", "No DeploymentConfigs found", should we just align with OCP?

apinnick: "No StatefulSets" and "No DeploymentConfigs" are understandable, but not correct, because StatefulSet and DeploymentConfig are k8s objects and therefore are capitalized. However, the writing guidelines say that we are not to use plural forms of k8s objects -- "StatefulSet" is OK but "StatefulSets" is wrong. 

In the case of "preferences" or "instance types", those terms should not be capitalized because they are [general references](https://github.com/openshift/openshift-docs/blob/main/contributing_to_docs/doc_guidelines.adoc#general-references), not k8s object names.

PR #1503, opened 2023-08-30
CNV-32392: Fix error in catalog drawer for All Projects namespace

pcbailey:
## 📝 Description

This PR fixes a bug where an error alert was displayed in the catalog drawer.

Cause: Using the All Projects namespace session key in a `k8sCreate` call.

Issue: https://issues.redhat.com/browse/CNV-32392

### DIFF
--- a/src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerPanel.tsx
+++ b/src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerPanel.tsx
@@ -16,6 +16,7 @@ import CPUMemoryModal from '@kubevirt-utils/components/CPUMemoryModal/CpuMemoryM
 import HardwareDevices from '@kubevirt-utils/components/HardwareDevices/HardwareDevices';
 import { useModal } from '@kubevirt-utils/components/ModalProvider/ModalProvider';
 import VirtualMachineDescriptionItem from '@kubevirt-utils/components/VirtualMachineDescriptionItem/VirtualMachineDescriptionItem';
+import { DEFAULT_NAMESPACE } from '@kubevirt-utils/constants/constants';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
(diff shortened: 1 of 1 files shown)


hstastna (commented):
IMHO this is unnecessary complicated. You really need only to use `ns || DEFAULT_NAMESPACE` in two places in this file, nothing else. You don't need to check if it is in "All Projects". Unnecessary checks. Really. Because if it is there, `ns` becomes `undefined` for sure.

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerPanel.tsx
```diff
@@ -174,7 +177,7 @@ export const TemplatesCatalogDrawerPanel: FC<TemplatesCatalogDrawerPanelProps> =
                             <CPUMemoryModal
                               isOpen={isOpen}
                               onClose={onClose}
-                              onSubmit={updateVMCPUMemory(ns, updateVM, setUpdatedVM)}
+                              onSubmit={updateVMCPUMemory(namespace, updateVM, setUpdatedVM)}
```
hstastna, line 180: This change is unnecessary.

pcbailey:
@hstastna Thank you for pointing that out. I was thinking of `useActiveNamespace` not `useParams`.

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerPanel.tsx
```diff
@@ -103,8 +105,7 @@ export const TemplatesCatalogDrawerPanel: FC<TemplatesCatalogDrawerPanelProps> =
         .catch((err) => {
           setError(err);
         });
-      // eslint-disable-next-line react-hooks/exhaustive-deps
-    }, [ns, template]);
+    }, [vmNamespace, template]);
```
hstastna, line 108: This is also not good, you still need `// eslint-disable-next-line react-hooks/exhaustive-deps` here. No, we don't want to add `updateVM` to dependency array. IMO it was better to merge https://github.com/kubevirt-ui/kubevirt-plugin/pull/1505 instead.

PR #1512, opened 2023-09-04
CNV-32447: update useActiveNamespace import

metalice:
## 📝 Description

fix import of useActiveNaemsapce
## 🎥 Demo


hstastna (commented):
Looks good. And what about `useLastNamespace`  from '@openshift-console/dynamic-plugin-sdk-internal'? Why do we use also this? Feels to me the same as `useActiveNamespace`. Because the active one is also the last visited one. Isn't it?

metalice:
> Looks good. And what about `useLastNamespace` from '@openshift-console/dynamic-plugin-sdk-internal'? Why do we use also this? Feels to me the same as `useActiveNamespace`. Because the active one is also the last visited one. Isn't it?

useLast.. don't need to change imports as they didn't move it.
About the use case, I didn't have the chance to use it, but by the name, it seems like if u moving between projects and you want the projects before to give a more accurate URL instead of history.goBack()

hstastna (approved):

PR #1513, opened 2023-09-04
CNV-32597: Handling yarn lint warnings

DanaOrr:
Description 
Added React Hooks the missing dependency.

### DIFF
--- a/src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerPanel.tsx
+++ b/src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerPanel.tsx
@@ -105,6 +105,7 @@ export const TemplatesCatalogDrawerPanel: FC<TemplatesCatalogDrawerPanelProps> =
         .catch((err) => {
           setError(err);
         });
+      // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [vmNamespace, template]);
 
     return (
--- a/src/utils/hooks/useAlerts/useAlerts.ts
+++ b/src/utils/hooks/useAlerts/useAlerts.ts
@@ -26,7 +26,7 @@ const useAlerts: UseAlerts = () => {
     const { alerts } = getAlertsAndRules((response as PrometheusRulesResponse)?.data);
 
     return silenceFiringAlerts(alerts, silences);
-  }, [response]);
+  }, [response, silences]);
 
   return { alerts: allAlerts || [], loaded: pollingStatus === SUCCESS };
 };
--- a/src/views/dashboard-extensions/KubevirtHealthPopup/components/HealthPopupChart.tsx
+++ b/src/views/dashboard-extensions/KubevirtHealthPopup/components/HealthPopupChart.tsx
@@ -28,7 +28,7 @@ const HealthPopupChart: React.FC<HealthPopupChartProps> = ({ alerts }) => {
         });
         return acc;
       }, []),
-    [alerts],
+    [alerts, totalNumberAlerts],
   );
 
   return (
--- a/src/views/virtualmachines/details/tabs/configuration/network/components/modal/VirtualMachinesNetworkInterfaceModal.tsx
+++ b/src/views/virtualmachines/details/tabs/configuration/network/components/modal/VirtualMachinesNetworkInterfaceModal.tsx
@@ -65,7 +65,7 @@ const VirtualMachinesNetworkInterfaceModal: FC<VirtualMachinesNetworkInterfaceMo
           });
         }
       },
-    [vm],
+    [nicHotPlugEnabled, vm],
   );
 
   return (


src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerPanel.tsx
```diff
@@ -105,7 +105,7 @@ export const TemplatesCatalogDrawerPanel: FC<TemplatesCatalogDrawerPanelProps> =
         .catch((err) => {
           setError(err);
         });
-    }, [vmNamespace, template]);
+    }, [vmNamespace, template, updateVM]);
```
hstastna, line 108: This is the exact place I was talking about, this is the exception, you don't want to do this here. I am pretty sure :) It would cause bad things in the Catalog drawer. Did you test it? I'd recommend to test catalog drawer in your env locally with and without this change. I am sure you would find unexpected problems.. 

To fix this place I'd recommend to add the comment for this place to prevent warnings, the comment that is missing here. The comment was here, just Phillip removed it recently (not sure why, probably by accident).

DanaOrr: Thank you Hilda :)

hstastna (approved):

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerPanel.tsx
hstastna: Unfortunately, this is a temporary fix, yes. But to avoid this completely, we'd need to completely rebuild the logic around Catalog drawer and customizing VM. Maybe we would have to give up editing any data in the drawer, to prevent using context and processing template etc for the data in the drawer. And that's what we won't fix within this PR, that's for a longer run, unfortunately. 
Also I heard creating VMs from Catalog will soon be deprecated so we maybe don't need to undergo what I've mentioned in this comment earlier.
So I'd suggest to unhold this PR and to leave it as it is. Adding `updateVM` to dependency array will cause not nice behavior in the drawer for sure. We really don't want to check updateVM function for its changes.

metalice: Usually, a function shouldn't cause over-rendering and shouldn't cause any issues while adding it as a dependency.

I did a simple test using console.log in the hook. Without this added dependency, there is only one print; with that added dependency, there is still only one print.

Have you checked if this added dependency breaks something? From what I can see from my soft check, the functionality, performance, and visual  - there aren't any issues.

hstastna: > Have you checked if this added dependency breaks something? From what I can see from my soft check, the functionality, performance, and visual - there aren't any issues.

🤦🏽 I've checked this like millions of times and it does cause visual and performance issues. The related code is executed on and on and CPU | Memory field is blinking in the drawer, with that unnecessary dependency added in the array + problems with editing CPU | Memory field in the drawer. This is true since we added ability to edit CPU | Memory field. However, we haven't found any better way how to achieve the expected without using context in the drawer (while avoiding putting all the data to the url). If you don't experience this problem, I can say it is probably a problem of slow cluster or connectivity to it, because when I simulate slow connectivity in my env, I also cannot see any issue with this in the UI.

metalice: Well, I don't experience it at all, and it isn't related to my network or cluster speed, which is fine, regular, and without any tunning. So please recheck your side, and we should avoid those comments.

hstastna: I rechecked my side, tried on all the available clusters we have and I insist on what I wrote earlier.

metalice: When trying to reproduce the over-rendering when there is the template issue , I couldn't see any over rendering, but when I fixed the template issue I could get over rendering because of updateVM from context. the problem is updateVM isn't wrapped with useCallback as this is a function and need to be wrapped. @avivtur found that wrapping updateVM in useValidateVM with useCallback solve the issue of over-rendering, please check so we can add this fix also useValidateVM and remove the comment.

metalice: Dana please restore the comment for this deps array, lets merge this pr. we should open another bug and fix the usecallback there.

metalice: @hstastna ptal

PR #1514, opened 2023-09-04
CNV-32369: Display proper CPU | Memory values in Catalog drawer

hstastna:
## 📝 Description

**Fixes:**
https://issues.redhat.com/browse/CNV-32369

Display the CPU and Memory values for "CPU | Memory" field in the Catalog drawer, instead of showing "Not available". Make them editable again. 

Also display the values correctly, reflect the latest actual values after editing (this was also not correct) in the drawer.

The root cause of the problem was that to the related CPU/Memory code - to `CPUMemory` component was VMI related logic added without any considering that this code is common for existing VMs/VMIs and also for Catalog drawer and VMs that don't exist yet (so those cannot run yet, obviously). This also caused incorrect CPU value in the drawer (once "Not available" issue was solved by adding necessary conditions to `vmi` checks in the same component).

_Note:_
There is another bug in _Catalog_ drawer, but was hidden by the bug this PR is fixing: once editing _CPU | Memory_ field will work, the expected CPU/Mem values will appear for the newly created VM when quick creating VM, but when customizing VMs, will not. This is fixed separately, within [this](https://github.com/kubevirt-ui/kubevirt-plugin/pull/1515) PR, as this issue is related to the different code.

## 🎥 Screenshots
**Before:**
cpu_before

**After:**
cpu_after

### DIFF
--- a/src/utils/components/CPUMemory/CPUMemory.tsx
+++ b/src/utils/components/CPUMemory/CPUMemory.tsx
@@ -10,6 +10,7 @@ import useVMI from '@kubevirt-utils/resources/vm/hooks/useVMI';
 import { readableSizeUnit } from '@kubevirt-utils/utils/units';
 import { isEmpty } from '@kubevirt-utils/utils/utils';
 import { Skeleton } from '@patternfly/react-core';
+import { isVMRunning } from '@virtualmachines/utils';
 
 type CPUMemoryProps = {
   vm: V1VirtualMachine;
@@ -19,17 +20,17 @@ const CPUMemory: FC<CPUMemoryProps> = ({ vm }) => {
   const { t } = useKubevirtTranslation();
   const itMatcher: V1InstancetypeMatcher = vm?.spec?.instancetype;
   const { instanceType, instanceTypeLoaded, instanceTypeLoadError } = useInstanceType(itMatcher);
+  const isMachineRunning = isVMRunning(vm);
   const { vmi, vmiLoadError } = useVMI(getName(vm), getNamespace(vm));
 
-  if (vmiLoadError || (instanceTypeLoadError && !isEmpty(itMatcher)))
+  if ((isMachineRunning && vmiLoadError) || (!isEmpty(itMatcher) && instanceTypeLoadError))
     return <MutedTextSpan text={t('Not available')} />;
 
-  if (!vmi || !vm || !instanceTypeLoaded) return <Skeleton />;
+  if ((isMachineRunning && !vmi) || !vm || !instanceTypeLoaded) return <Skeleton />;
 
   const cpu =
-    vCPUCount(vmi?.spec?.domain?.cpu) ||
-    instanceType?.spec?.cpu?.guest ||
-    vCPUCount(vm?.spec?.template?.spec?.domain?.cpu);
+    vCPUCount(vmi?.spec?.domain?.cpu || vm?.spec?.template?.spec?.domain?.cpu) ||
+    instanceType?.spec?.cpu?.guest;
 
   const memory = readableSizeUnit(
     vmi?.spec?.domain?.memory?.guest ||
--- a/src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerPanel.tsx
+++ b/src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerPanel.tsx
@@ -166,7 +166,7 @@ export const TemplatesCatalogDrawerPanel: FC<TemplatesCatalogDrawerPanelProps> =
                       <VirtualMachineDescriptionItem
                         bodyContent={
                           <CPUDescription
-                            cpu={vmObject?.spec?.template?.spec?.domain?.cpu}
+                            cpu={updatedVM?.spec?.template?.spec?.domain?.cpu}
                             helperTextResource={CpuMemHelperTextResources.FutureVM}
                           />
                         }
--- a/src/views/catalog/wizard/components/WizardHeader.tsx
+++ b/src/views/catalog/wizard/components/WizardHeader.tsx
@@ -1,4 +1,4 @@
-import * as React from 'react';
+import React, { FC, memo } from 'react';
 import { useHistory } from 'react-router-dom';
 
 import { CUSTOMIZE_TEMPLATE_TITLE } from '@catalog/customize/constants';
@@ -17,7 +17,7 @@ import {
   Title,
 } from '@patternfly/react-core';
 
-export const WizardHeader: React.FC<{ namespace: string }> = React.memo(({ namespace }) => {
+export const WizardHeader: FC<{ namespace: string }> = memo(({ namespace }) => {
   const { t } = useKubevirtTranslation();
   const { tabsData } = useWizardVMContext();
   const history = useHistory();


hstastna:

src/utils/components/CPUMemory/CPUMemory.tsx
```diff
@@ -19,17 +22,15 @@ const CPUMemory: FC<CPUMemoryProps> = ({ vm }) => {
   const { t } = useKubevirtTranslation();
   const itMatcher: V1InstancetypeMatcher = vm?.spec?.instancetype;
   const { instanceType, instanceTypeLoaded, instanceTypeLoadError } = useInstanceType(itMatcher);
+  const isVMRunning = vm?.status?.printableStatus === printableVMStatus.Running;
```
metalice, line 25: there is a function for it in a general util

src/utils/components/CPUMemory/CPUMemory.tsx
```diff
@@ -19,17 +22,15 @@ const CPUMemory: FC<CPUMemoryProps> = ({ vm }) => {
...
   const { vmi, vmiLoadError } = useVMI(getName(vm), getNamespace(vm));
 
-  if (vmiLoadError || (instanceTypeLoadError && !isEmpty(itMatcher)))
-    return <MutedTextSpan text={t('Not available')} />;
+  const cpu = vCPUCount(useCPUData(vm)) || instanceType?.spec?.cpu?.guest;
```
metalice, line 28: u already made a custom hook useCpu.. why not getting instanceType there also and simply return the CPU? u dont need this ||

metalice: also the vcpucount can be used there

src/utils/components/CPUMemory/CPUMemory.tsx
```diff
@@ -19,17 +22,15 @@ const CPUMemory: FC<CPUMemoryProps> = ({ vm }) => {
   const { t } = useKubevirtTranslation();
   const itMatcher: V1InstancetypeMatcher = vm?.spec?.instancetype;
   const { instanceType, instanceTypeLoaded, instanceTypeLoadError } = useInstanceType(itMatcher);
+  const isVMRunning = vm?.status?.printableStatus === printableVMStatus.Running;
   const { vmi, vmiLoadError } = useVMI(getName(vm), getNamespace(vm));
```
metalice, line 26: u are doing 2 network calls for vmi with your ned hook, we dont want this. u can return vmi from your hook.

src/utils/components/CPUMemory/CPUMemory.tsx
```diff
@@ -19,17 +22,15 @@ const CPUMemory: FC<CPUMemoryProps> = ({ vm }) => {
...
+  if ((isVMRunning && vmiLoadError) || (!isEmpty(itMatcher) && instanceTypeLoadError))
+    return <MutedTextSpan text={t('Not available')} />;
 
-  const cpu =
-    vCPUCount(vmi?.spec?.domain?.cpu) ||
-    instanceType?.spec?.cpu?.guest ||
-    vCPUCount(vm?.spec?.template?.spec?.domain?.cpu);
+  if (!vm || !instanceTypeLoaded || (isVMRunning && !vmi)) return <Skeleton />;
 
   const memory = readableSizeUnit(
```
metalice, line 35: u can add this to ur hook and also the instanceType call , just return the data

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerPanel.tsx
```diff
@@ -105,6 +108,7 @@ export const TemplatesCatalogDrawerPanel: FC<TemplatesCatalogDrawerPanelProps> =
         .catch((err) => {
           setError(err);
         });
+      // eslint-disable-next-line react-hooks/exhaustive-deps
```
metalice, line 111: no need , recheck please

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerPanel.tsx
```diff
@@ -166,7 +170,7 @@ export const TemplatesCatalogDrawerPanel: FC<TemplatesCatalogDrawerPanelProps> =
                       <VirtualMachineDescriptionItem
                         bodyContent={
                           <CPUDescription
-                            cpu={vmObject?.spec?.template?.spec?.domain?.cpu}
+                            cpu={useCPUData(updatedVM)}
```
metalice, line 173: please save as a const and use it, doing the hook directly is not something I see in our codebase

src/views/catalog/wizard/tabs/overview/components/WizardOverviewGrid.tsx
```diff
@@ -146,7 +149,7 @@ const WizardOverviewGrid: FC<WizardOverviewGridProps> = ({ tabsData, updateVM, v
             helperPopover={{
               content: (
                 <CPUDescription
-                  cpu={vm?.spec?.template?.spec?.domain?.cpu}
+                  cpu={useCPUData(vm)}
```
metalice, line 152: same as above

src/utils/components/CPUDescription/utils/utils.ts
```diff
@@ -1,8 +1,17 @@
...
+import useVMI from '@kubevirt-utils/resources/vm/hooks/useVMI';
 
 export const CpuMemHelperTextResources = {
   ActualVM: t('This VirtualMachine has'),
   FutureVM: t('The VirtualMachine will have'),
   Template: t('The template has'),
   VMI: t('The VirtualMachineInstance has'),
 };
+
+export const useCPUData = (vm: V1VirtualMachine): V1CPU => {
```
metalice, line 13: please expand this hook and move it to its own file

src/utils/components/CPUMemory/CPUMemory.tsx
hstastna: why not? because I think we don't need to put instanceType check everywhere. Really unnecessary. This is just a small refactoring, I am not changing anything in the logic, except that I am adding extra check to vmi check, in this PR.

src/utils/components/CPUMemory/CPUMemory.tsx
hstastna: Where ate those network calls? I cannot see it. Could you be more specific? Thanks.

src/utils/components/CPUMemory/CPUMemory.tsx
hstastna: We don't want this in "my" hook. Why? PTAl where it is used and why

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerPanel.tsx
hstastna: We DO NEED this.

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerPanel.tsx
hstastna: What exactly is problematic about it? We should have better reasons of doing something than just "not seeing something in our codebase".

src/utils/components/CPUDescription/utils/utils.ts
hstastna: Yes, thanks, makes sense.

src/utils/components/CPUMemory/CPUMemory.tsx
hstastna: I will take a look, thanks. But it is interesting that this line of the code is in so many places of our codebase, and everybody is fine with that, just here it is not fine.

metalice: I added this funciton last week, because I see too many repeat code such as this, so this is the first time I we can use it other the pr created this funciton

src/utils/components/CPUMemory/CPUMemory.tsx
metalice: useVMI, useCPUData both call vmi - its extra call which is uneeded

src/utils/components/CPUMemory/CPUMemory.tsx
metalice: ur hook names useCPUData but it not retrieve, instanceType cpu data, if you already have a chance to improve this file and ur hook to retrieve also the instanceType cpu i think you should get from this hook the cpu and memory data instead of getting half and then in the component to check the other half.. just make more sense that you have 1 place to get the data

src/utils/components/CPUMemory/CPUMemory.tsx
metalice: ur hook can be something to improve this whole modal as it will sperate between logic and display, if u move memory to there this component will only be in charge of rendering.

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerPanel.tsx
metalice: please see solution in Dana pr

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerPanel.tsx
metalice: i understand what you are saying, but this is much less readable than save the return data as const and placing it as cpu={cpu}

src/utils/components/CPUMemory/CPUMemory.tsx
hstastna: We need a lot of refactoring in our codebase :)

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerPanel.tsx
hstastna: I think Dana's PR is not that solution, it's what you've mentioned regarding Aviv's finding and adding missing useCallback to appropriate places. I haven't seen the PR yet.

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerPanel.tsx
hstastna: This is kinda tricky because under `vmObject?.spec?.template?.spec?.domain?.cpu` we have cpu data, which is an object containing number of sockets, cores, threads, then we have vcpu count, which is just a number, calculated one. Sometimes we use this, sometimes we need that number. We should distinguish between those. Using a variable called just `cpu` does not say enough, nothing about if it represents a number - vcpu count, or an object containing cpu values (sockets, cores, threads).

PR #1515, opened 2023-09-04
CNV-32609: Keep the CPU value from Catalog drawer for customizing page

hstastna:
## 📝 Description

**Fixes:**
https://issues.redhat.com/browse/CNV-32609

Save correctly the CPU value from _Catalog_ drawer, especially if the user has edited this value in the drawer, so it can be displayed in _Customize and create VirtualMachine_ page correctly, same as it was set when the user has left the drawer.

_Note:_
This PR can be merged now, but it is kinda problematic to reproduce the bug without https://github.com/kubevirt-ui/kubevirt-plugin/pull/1514 being merged. 

## 🎥 Demo
**Before:**

https://github.com/kubevirt-ui/kubevirt-plugin/assets/13417815/310234ff-65d8-4c75-9b71-aabc870826c0

**After:**

https://github.com/kubevirt-ui/kubevirt-plugin/assets/13417815/d72aaff2-e7e7-4696-8e77-76aa6ae94dfb

### DIFF
--- a/src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
+++ b/src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
@@ -135,6 +135,7 @@ export const TemplatesCatalogDrawerCreateForm: FC<TemplatesCatalogDrawerCreateFo
 
     const onCustomize = (e: MouseEvent) => {
       e.preventDefault();
+
       if (isEmpty(template?.parameters)) {
         return k8sCreate<V1Template>({
           data: { ...template, metadata: { ...template?.metadata, namespace } },
@@ -159,8 +160,7 @@ export const TemplatesCatalogDrawerCreateForm: FC<TemplatesCatalogDrawerCreateFo
               ...vmDraft?.spec?.template?.spec?.domain?.resources?.requests,
               memory: `${vm.spec.template.spec.domain.resources.requests['memory']}`,
             };
-            vmDraft.spec.template.spec.domain.cpu.cores =
-              vmObject.spec.template.spec.domain.cpu.cores;
+            vmDraft.spec.template.spec.domain.cpu.cores = vm.spec.template.spec.domain.cpu.cores;
 
             const updatedVolumes = applyCloudDriveCloudInitVolume(vmObject);
             vmDraft.spec.template.spec.volumes = isRHELTemplate(processedTemplate)
--- a/src/views/catalog/customize/components/CustomizeForms/useCustomizeFormSubmit.ts
+++ b/src/views/catalog/customize/components/CustomizeForms/useCustomizeFormSubmit.ts
@@ -100,7 +100,7 @@ export const useCustomizeFormSubmit = ({
           ...vmDraft?.spec?.template?.spec?.domain?.resources?.requests,
           memory: `${vm.spec.template.spec.domain.resources.requests['memory']}`,
         };
-        vmDraft.spec.template.spec.domain.cpu.cores = vmObj.spec.template.spec.domain.cpu.cores;
+        vmDraft.spec.template.spec.domain.cpu.cores = vm.spec.template.spec.domain.cpu.cores;
 
         const updatedVolumes = applyCloudDriveCloudInitVolume(vmObj);
         vmDraft.spec.template.spec.volumes = isRHELTemplate(processedTemplate)


hstastna:

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
```diff
@@ -95,14 +97,16 @@ export const TemplatesCatalogDrawerCreateForm: FC<TemplatesCatalogDrawerCreateFo
 
       const nameParameterExists = isNameParameterExists(template);
 
-      const templateToProcess = produce(template, (draftTemplate) => {
+      const templateToProcess = produce(template, (draftTemplate: V1Template) => {
```
metalice, line 100: If you are adding types, what about template and function?

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
```diff
@@ -95,14 +97,16 @@ export const TemplatesCatalogDrawerCreateForm: FC<TemplatesCatalogDrawerCreateFo
...
         if (nameParameterExists)
           replaceTemplateParameterValue(draftTemplate, NAME_INPUT_FIELD, vmName);
 
         const vmObject = getTemplateVirtualMachineObject(draftTemplate);
 
         if (!isEmpty(authorizedSSHKey)) {
-          draftTemplate.objects = template.objects.filter((obj) => obj?.kind !== SecretModel.kind);
+          draftTemplate.objects = template.objects.filter(
+            (obj: { kind: any }) => obj?.kind !== SecretModel.kind,
```
metalice, line 108: Why use any? obj probably can be ResourceCommon

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
```diff
@@ -143,10 +148,10 @@ export const TemplatesCatalogDrawerCreateForm: FC<TemplatesCatalogDrawerCreateFo
           queryParams: {
             dryRun: 'All',
           },
-        }).then(async (processedTemplate) => {
+        }).then(async (processedTemplate: V1Template) => {
           const vmObject = getTemplateVirtualMachineObject(processedTemplate);
 
-          const updatedVM = produce(vmObject, (vmDraft) => {
+          const updatedVM = produce(vmObject, (vmDraft: V1VirtualMachine) => {
```
metalice, line 154: same as above

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
```diff
@@ -159,21 +164,21 @@ export const TemplatesCatalogDrawerCreateForm: FC<TemplatesCatalogDrawerCreateFo
...
 
-          updateTabsData((tabsDataDraft) => {
+          updateTabsData((tabsDataDraft: TabsData) => {
             // additional objects
-            tabsDataDraft.additionalObjects = processedTemplate.objects.filter((obj) =>
-              !isEmpty(authorizedSSHKey)
-                ? obj.kind !== VirtualMachineModel.kind || obj.kind !== SecretModel
-                : obj.kind !== VirtualMachineModel.kind,
+            tabsDataDraft.additionalObjects = processedTemplate.objects.filter(
+              (obj: { kind: any }) =>
```
metalice, line 178: same as above

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
hstastna: Because I cannot know which type could be even better there. I am not familiar with this codebase. I am just improving it. Better than nothing.

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
hstastna: ?

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
metalice: you can simply try, if it wont be the right type ts will complain, ts is scanning all code base with every save

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
hstastna: It is not recommended to add types that TS can easily figure out. Type annotation vs. type inference. Type inference is over annotation, in TS.

metalice: it is true about local vars, function signatures should be explicit. 
basically, it makes the code more readable for functions and return values. instead of the developer trying to figure out the type using IDE.

https://www.fun4code.com/typescript-explicit-vs-implicit/
https://medium.com/@ThiliniWathsala/explicit-vs-implicit-type-in-typescript-5727b5cbde

hstastna: Whole course on Udemy about TS I underwent recently was telling me how and why we should prioritize type inference over type annotation. So it is opinion against opinion. Everything will always have some pros and cons. We can argue about what is better forever while our customers are experiencing serious bug I am trying to fix in this PR. I can see this is getting too far and I regret I did more in this PR than it was necessary to fix the bug.

metalice: you're right that arguing over it is taking too long.

Using "any" as an explicit type or not giving functions signatures the right type if not right. I can recommend watching this part of the course you took a second time, as I'm pretty sure the speaker was referring to local variables or very simple use of functions. Another thing is that 99% of cases there is a difference between theory in course and practice in real-life coding, maybe this is one of those cases referring to the course you watch. 

Any way please remove your WIP so we can get it merged. Thanks

hstastna: > Using "any" as an explicit type or not giving functions signatures the right type is not right. I can recommend watching this part of the course you took a second time, as I'm pretty sure the speaker was referring to local variables or very simple use of functions. Another thing is that 99% of cases there is a difference between theory in course and practice in real-life coding, maybe this is one of those cases referring to the course you watch.

Well, did, and no, I still insist on what I've written earlier, at least for now. And that course was recommended and paid by my employer, they should know what they are doing.

> Any way please remove your WIP so we can get it merged. Thanks

Yes, I'll do, also will remove all the types improvements I've included in this PR. I'll better implement them separately, via followup PR, once we make sure about what and where. It's not good to mix bug fixes with other improvements or refactorings anyway. Thank you for your feedback, opinions and also links to the articles. Of course nothing is black or white, it depends. And sometimes it happens that we deal with something too much, with such details that at the end of the day really don't matter so much.

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
hstastna: "More" doesn't always mean "better".

PR #1521, opened 2023-09-06
CNV-32266: The example of bootable volumes does not show in the list

avivtur:
## 📝 Description

Create a DV example with proper IT and pref labels so it would show up on the Bootable volumes list

Notes:
* The CR created is a DataVolume so we see the breadcrumb on top of created DV to lead back to DV list instead of Bootable volumes list - (The way to go around it is implementing a new DataVolume details page)
* To make this work a cluster should have a SC defined to the cluster as default OR the manually select a storage class for the example DV.
## 🎥 Demo

After:

https://github.com/kubevirt-ui/kubevirt-plugin/assets/67270715/ed274f3e-afa1-4ae3-b02d-1333e20d6eb8


hstastna (commented):
Looks good! Just.. shouldn't we also rename the file _src/templates/datasource-yaml.ts_ appropriately? :)

PR #1522, opened 2023-09-06
Use isRunning VM util

hstastna:
## 📝 Description

Use recently implemented `isVMRunning` util instead of repeatedly defining the same variable across the codebase, rename it to `isRunning`, which is shorter and still self explaining enough, and it also prevents potential bugs that could be created while unnecessary renaming `isVMRunning` variable already used in many components.

## 🎥 Demo

No any visual changes


hstastna:

PR #1525, opened 2023-09-07
CNV-26584: The PVC is not deleted along with the bootable volumes or dataSources

avivtur:
## 📝 Description

When creating a bootable volume with the `Add volume` dialog, we don't have a proper way to delete the volume's CR if wanted, renaming 'Delete' to 'Remove from list' as the operation will not delete the CR but only the preference label
Also adding a 'Delete' action to delete the CR (PVC/DV/DS) 

## 🎥 Demo

Before:
PVC delete:
DeletePVC2-b4
DeletePVC-b4

DS delete: 
DeleteDS-b4

After:
PVC delete:
DeletePVC2
DeletePVC

DS delete: (row actions dropdown looks the same)
DeleteDS

https://github.com/kubevirt-ui/kubevirt-plugin/assets/67270715/ff5a4ede-9990-496f-9aae-04623ffcbacc


src/views/bootablevolumes/actions/hooks/useBootableVolumesActions.tsx
```diff
@@ -36,17 +42,40 @@ const useBootableVolumesActions: BootableVolumesActionsProps = (source, preferen
...
       id: 'edit-bootablevolume',
       label: t('Edit'),
     },
     {
       cta: () =>
         createModal(({ isOpen, onClose }) => (
-          <DeleteBootableVolumesModal isOpen={isOpen} onClose={onClose} source={source} />
+          <RemoveBootableVolumesModal isOpen={isOpen} onClose={onClose} source={source} />
+        )),
+      // description: t('Only the bootable metadata will be deleted'),
```
metalice, line 54: ?

avivtur: Removed, Thanks :smile:

src/views/datasources/actions/DeleteDataSourceModal/DeleteDataSourceModal.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+    if (deletePVC) {
+      // We try to delete the created DV, if already GC, we want to fallback to delete the PVC
+      try {
+        await k8sDelete({ model: DataVolumeModel, resource: dv });
+      } catch {
+        await k8sDelete({ model: PersistentVolumeClaimModel, resource: pvc });
+      }
+    }
+  };
```
metalice, line 63: this block repeat it self, can be extracted?

src/views/datasources/actions/DeleteDataSourceModal/DeleteDataSourceModal.tsx
```diff
@@ -0,0 +1,103 @@
...
+        model: DataImportCronModel,
+        resource: dataImportCron,
+      });
+    } catch (e) {
+      kubevirtConsole.log(e?.message);
+    } finally {
+      await k8sDelete({
+        model: DataSourceModel,
+        resource: dataSource,
+      });
```
metalice, line 52: this wont throw?

src/views/datasources/actions/DeleteDataSourceModal/DeleteDataSourceModal.tsx
```diff
@@ -0,0 +1,103 @@
...
+          />
+        </StackItem>
+      </Stack>
+    </TabModal>
+  );
+};
+
+export default DeleteDataSourceModal;
+{
+  /* <>
```
metalice, line 99: ?

src/views/datasources/actions/DeleteDataSourceModal/hooks/useUnderlyingPVC.ts
```diff
@@ -0,0 +1,36 @@
...
+import { useK8sWatchResource } from '@openshift-console/dynamic-plugin-sdk';
+
+type UseUnderlyingPVC = (dataSource: V1beta1DataSource) => {
+  dv: V1beta1DataVolume;
+  pvc: IoK8sApiCoreV1PersistentVolumeClaim;
+};
+
+const useUnderlyingPVC: UseUnderlyingPVC = (dataSource) => {
+  const name = dataSource?.spec?.source?.pvc?.name;
+  const namespace = dataSource?.spec?.source?.pvc?.namespace;
```
metalice, line 20: u can add selectors for those, might be useful

src/views/datasources/actions/DeleteDataSourceModal/hooks/useUnderlyingPVC.ts
```diff
@@ -0,0 +1,36 @@
...
+    namespace,
+  });
+
+  const [dv, dvLoaded] = useK8sWatchResource<V1beta1DataVolume>({
+    groupVersionKind: modelToGroupVersionKind(DataVolumeModel),
+    name,
+    namespace,
+  });
+
+  return { dv: dvLoaded && dv, pvc: pvcLoaded && pvc };
```
metalice, line 33: what about default data? if someone expect dv and pvc to get an object it will return false

src/views/datasources/actions/DeleteDataSourceModal/DeleteDataSourceModal.tsx
avivtur: I can wrap this in try catch and handle the error, wdyt?

src/views/datasources/actions/DeleteDataSourceModal/hooks/useUnderlyingPVC.ts
avivtur: added a check to make sure the objects aren't empty

src/views/datasources/actions/DeleteDataSourceModal/DeleteDataSourceModal.tsx
metalice: i see from the code that you always want to delete dataSource , if u will change the order?
first, delete dataSource then dataimportcron ? is possible? then u will need only one try catch

src/views/datasources/actions/DeleteDataSourceModal/hooks/useUnderlyingPVC.ts
metalice: better do this check in the hook and not in the consuming component, instead of returning false, better to return empty object and loaded boolean (for loaders)

src/views/datasources/actions/DeleteDataSourceModal/DeleteDataSourceModal.tsx
avivtur: Actually, this code block was taken from the DS actions hook (not copied) and added the part where I delete the PVC and DV, Since we're a bit lacking time, how about I check that as a follow-up task?

metalice: ok but please open a ticket on u so it wont be forgotten :)

avivtur: @metalice I've checked this subject, We are deleting the DIC first because the DIC CR is auto-populating the DS if not exist. Adding this image of my terminal (hopefully it will help clarify this topic):
Screenshot from 2023-09-12 11-16-07

So I guess that's the we should add them together to the same try block, instead of finally that will run even if the DIC deletion fails (which will auto-heal the DS deletion)

PR #1526, opened 2023-09-07
CNV-32596: Display Memory value correctly in the UI

hstastna:
## 📝 Description

**Fixes:**
https://issues.redhat.com/browse/CNV-32596

The recent backend change in https://github.com/kubevirt/common-templates/pull/543 broke displaying memory value in the whole UI. This PR updates our UI codebase to reflect this new change. From now the memory value is read from:

`vm.spec.template.spec.domain.memory.guest`

## 🎥 Screenshots
**Before:**
Catalog drawer, error right after opening the drawer, no values for CPU and Memory:
drawer_before
Template details page - no value for memory:
templates_before

**After:**
drawer_after
templates_after

### DIFF
--- a/src/utils/components/PendingChanges/utils/helpers.ts
+++ b/src/utils/components/PendingChanges/utils/helpers.ts
@@ -43,18 +43,14 @@ export const checkCPUMemoryChanged = (
   if (isEmpty(vm) || isEmpty(vmi)) {
     return false;
   }
-  const vmRequests = vm?.spec?.template?.spec?.domain?.resources?.requests;
+  const vmMemory = vm?.spec?.template?.spec?.domain?.memory?.guest;
   const vmCPU = vm?.spec?.template?.spec?.domain?.cpu?.cores || 0;
 
-  const vmiRequests: { [key in string]?: string } = vmi?.spec?.domain?.resources?.requests || {};
+  const vmiMemory = vmi?.spec?.domain?.memory?.guest || '';
 
   const vmiCPU = vmi?.spec?.domain?.cpu?.cores || 0;
 
-  const memoryChanged = vmRequests
-    ? !isEqualObject(vmRequests, vmiRequests)
-    : vmiRequests?.memory !== vm?.spec?.template?.spec?.domain?.memory?.guest;
-
-  return memoryChanged || vmCPU !== vmiCPU;
+  return vmMemory !== vmiMemory || vmCPU !== vmiCPU;
 };
 
 export const checkBootOrderChanged = (
--- a/src/utils/components/CPUMemory/CPUMemory.tsx
+++ b/src/utils/components/CPUMemory/CPUMemory.tsx
@@ -35,7 +35,7 @@ const CPUMemory: FC<CPUMemoryProps> = ({ vm }) => {
   const memory = readableSizeUnit(
     vmi?.spec?.domain?.memory?.guest ||
       instanceType?.spec?.memory?.guest ||
-      (vm?.spec?.template?.spec?.domain?.resources?.requests as { [key: string]: string })?.memory,
+      vm?.spec?.template?.spec?.domain?.memory?.guest,
   );
 
   return (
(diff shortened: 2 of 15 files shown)


hstastna:

src/utils/components/PendingChanges/utils/helpers.ts
```diff
@@ -43,18 +43,14 @@ export const checkCPUMemoryChanged = (
   if (isEmpty(vm) || isEmpty(vmi)) {
     return false;
   }
-  const vmRequests = vm?.spec?.template?.spec?.domain?.resources?.requests;
+  const vmMemory = vm?.spec?.template?.spec?.domain?.memory?.guest;
```
pcbailey, line 46: I think we should add a selector for VMs and VMIs in `src/utils/resources/vm/utils/selectors.ts` and `src/utils/resources/vmi/utils/selectors.ts` respectively instead of accessing it this way throughout the code. It won't work everywhere, but it should cover most use cases and would make any future changes easier.

I'm sure there are other attributes that we're accessing this way as well, such as CPU, that would benefit from the same treatment. We can address them at appropriate points in the future.

pcbailey (commented):
One suggestion that will help maintainability in the future. Other than that it looks good!

src/utils/components/PendingChanges/utils/helpers.ts
hstastna: Yes and it is not time for cool refactorings, there is much more we could do here. But this is urgent bug we need to fix asap. Cool improvements later.

pcbailey: Following DRY and other best practices aren't "cool refactorings". They are basic expectations for professional software developers and expected by most members of this team. We already have a pattern of using selectors for properties throughout the code base.

We shouldn't have allowed this to happen in the first place, but it did and now we need to fix it when and where it's appropriate. This is a good opportunity to fix this particular issue.

The change shouldn't take more than 10-15 minutes to make and I'm around all day to approve it.

hstastna: This PR should have already been merged!! it is blocking lots of stuff! One should think about DRY way much earlier than now, when copy pasting that code in our codebase.

pcbailey: The bug has been present for some time now. When fixing it it's worth fixing it correctly. I said that we should have done better in the past, but there's nothing we can do about that now. The fact that we didn't do it perfectly in the past is not an excuse to do it poorly now. The goal is incremental improvement. It's a quick change and there's no good reason not to do it.

hstastna: > The bug has been present for some time now..

Yes, hidden behind 2 more bugs and was not urgent as it is now...

>  It's a quick change and there's no good reason not to do it.

So if it is such a quick change, why you simply didn't approved this PR and created another one with improvements you can see in the code? Obviously you are much better dev who can do things much faster...

I personally wouldn't underestimate refactoring, it always produces unnecessary bugs and that's what we really don't need now. Also it is not good to mix bug fixes with refactoring. I have very bad experience doing this.

hstastna:
@avivtur @metalice @upalatucci @vojtechszocs please review

src/utils/components/PendingChanges/utils/helpers.ts
pcbailey: > Maybe I already could have an open followup refactoring PR if this was already merged. Because we already could be happy and finally calm that the urgent bug is gone and so more space for improvements would be here...

Create a second commit making the change. It doesn't need to be a separate PR. An individual commit can be reverted if it causes problems.

We're the only ones here today so it makes more sense to knock this out in a single PR and be done with it.

pcbailey:
/lgtm

I'm merging due to branching on Sunday and will open a PR to address the quality issues in 4.15.

PR #1528, opened 2023-09-10
CNV-32568: Fixing Hardware Devices table in Compute -> Hardware devices

DanaOrr:
## 📝 Description

The Hardware Devices table in Compute -> Hardware Devices shows nothing.
Before the current change, the table was the same as the table that was in the vm overview. 
I changed the table and added the correct values ​​that should be there.

## 🎥 Demo
Before:
gpuBefore

After:
gpuAfter

### DIFF
--- a/src/utils/components/HardwareDevices/HardwareDevicesPageTable.tsx
+++ b/src/utils/components/HardwareDevices/HardwareDevicesPageTable.tsx
@@ -1,0 +1,39 @@
+import React, { FC } from 'react';
+
+import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
+import { VirtualizedTable } from '@openshift-console/dynamic-plugin-sdk';
+
+import MutedTextSpan from '../MutedTextSpan/MutedTextSpan';
+
+import useHardwareDevicesPageColumns from './list/hooks/useHardwareDevicesPageColumns ';
+import { HardwareDevicePageRow } from './utils/constants';
+import HardwareDevicesPageRow from './HardwareDevicesPageRow';
+
+type HardwareDevicesPageTableProps = {
+  devices: HardwareDevicePageRow[];
+  error: Error;
+  loaded: boolean;
+};
+
+const HardwareDevicesPageTable: FC<HardwareDevicesPageTableProps> = ({
+  devices,
+  error,
+  loaded,
+}) => {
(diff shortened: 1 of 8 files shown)


src/utils/components/HardwareDevices/HardwareDevicesPageTable.tsx
```diff
@@ -0,0 +1,37 @@
...
+  loaded,
+}) => {
+  const { t } = useKubevirtTranslation();
+  const columns = useHardwareDevicesPageColumns();
+
+  return (
+    <VirtualizedTable
+      columns={columns}
+      data={devices}
+      EmptyMsg={() => <div>{t('Not available')} </div>}
```
avivtur, line 28: div element is unnecessary, change it to Fragment (<></>) and is the space between text and closing tag intentional?

avivtur (commented):
Looks good, I left one little comment

avivtur:
> Before: gpuBefore

about the column name in the VM details page, Resource name should be just Name (resourceName is an actual property represented by "Hardware device name" column and using it as the column name for the name, is confusing to me) please change that as well

metalice:
> > Before: gpuBefore
> >       
> >         [
> >           
> >         
> >             gpuBefore
> >           ](https://user-images.githubusercontent.com/61961469/266828720-64e87899-cd0e-43d8-8252-c4f5508a242f.gif)
> >         
> >         
> >           
> >             
> >               
> >             
> >             
> >               
> >               
> >             
> >           
> >           [
> >             
> >               
> >             
> >           ](https://user-images.githubusercontent.com/61961469/266828720-64e87899-cd0e-43d8-8252-c4f5508a242f.gif)
> 
> about the column name in the VM details page, "Resource name" should be just "Name" (`resourceName` is an actual property represented by "Hardware device name" column, and using it as the column label for the device's name, is confusing to me) please change that as well

hmm im not sure.. in the device modal we are adding a name of our own, so calling it only a name here might be confusing :)  WDYT?

src/utils/components/HardwareDevices/HardwareDevicesPageTable.tsx
DanaOrr: The space at the end isn't intentional.
But maybe it's better if I use MutedTextSpan?

avivtur (approved):

PR #1530, opened 2023-09-10
CNV-32815: Fix template reggression, cpu and memory changes from catalog and added selectors

metalice:
…log drawer and unify selectors

## 📝 Description

1. Adding selector for CPU and memory so code will be unified
2. Fix regression for templates with requests.memory to be useful.
3.  Fix catalog drawer CPU and memory so it will work and presented values will be passed

## 🎥 Demo

Before:
non-cpu
non-example

After:
working-example

working-cpu


src/utils/components/CPUMemoryModal/hooks/useTemplateDefaultCpuMemory.ts
```diff
@@ -38,9 +39,7 @@ const useTemplateDefaultCpuMemory: UseTemplateDefaultCpuMemory = (
       });
   }, [templateName, templateNamespace]);
 
-  const defaultMemory = getMemorySize(
-    template?.objects?.[0]?.spec?.template?.spec?.domain?.memory?.guest,
-  );
+  const defaultMemory = getMemorySize(getMemory(template?.objects?.[0]));
```
avivtur, line 42: Please use getTemplateVirtualMachineObject method instead of `template?.objects?.[0]` here and below

src/utils/components/Charts/MemoryUtil/MemoryThresholdChart.tsx
```diff
@@ -32,7 +33,7 @@ const MemoryThresholdChart: FC<MemoryThresholdChartProps> = ({ vmi }) => {
   const queries = useMemo(() => getUtilizationQueries({ duration, obj: vmi }), [vmi, duration]);
   const { height, ref, width } = useResponsiveCharts();
 
-  const memory = getMemorySize(vmi?.spec?.domain?.memory?.guest);
+  const memory = getMemorySize(getMemory(vmi?.spec?.domain));
```
avivtur, line 36: shouldn't this be `getMemoty(vmi)` ?

src/utils/components/DedicatedResourcesModal/DedicatedResourcesModal.tsx
```diff
@@ -43,9 +44,7 @@ const DedicatedResourcesModal: React.FC<DedicatedResourcesModalProps> = ({
   vmi,
 }) => {
   const { t } = useKubevirtTranslation();
-  const [checked, setChecked] = React.useState<boolean>(
-    !!vm?.spec?.template?.spec?.domain?.cpu?.dedicatedCpuPlacement,
-  );
+  const [checked, setChecked] = React.useState<boolean>(!!getCPU(vm)?.dedicatedCpuPlacement);
```
avivtur, line 47: Please Change the `import * as React from 'react';` :)

src/utils/components/CPUMemoryModal/hooks/useTemplateDefaultCpuMemory.ts
avivtur: now saw that you also have getTemplateVirtualMachineCPU

src/views/templates/details/tabs/details/components/CPUMemory.tsx
```diff
@@ -59,7 +60,7 @@ const CPUMemory: FC<CPUMemoryProps> = ({ template }) => {
         <Popover
           bodyContent={
             <CPUDescription
-              cpu={getTemplateVirtualMachineObject(template)?.spec?.template?.spec?.domain?.cpu}
+              cpu={getCPU(getTemplateVirtualMachineObject(template))}
```
avivtur, line 63: here too you can use getTemplateVirtualMachineCPU

avivtur (commented):
Looking good, left a few minor comments
I like the thinking of using the same selectors for VM/VMI just need to make sure no unexpected behavior comes from that

avivtur (approved):

src/views/catalog/customize/components/CustomizeForms/useCustomizeFormSubmit.ts
```diff
@@ -96,9 +96,9 @@ export const useCustomizeFormSubmit = ({
         vmDraft.metadata.namespace = ns || DEFAULT_NAMESPACE;
         vmDraft.metadata.labels[LABEL_USED_TEMPLATE_NAME] = processedTemplate.metadata.name;
         vmDraft.metadata.labels[LABEL_USED_TEMPLATE_NAMESPACE] = template.metadata.namespace;
-
-        vmDraft.spec.template.spec.domain.cpu.cores = vm.spec.template.spec.domain.cpu.cores;
-        vmDraft.spec.template.spec.domain.memory.guest = vm.spec.template.spec.domain.memory.guest;
+        const { cpu, memory } = getMemoryCPU(vm);
+        vmDraft.spec.template.spec.domain.cpu.cores = cpu?.cores;
```
hstastna, line 100: in some places in the codebase like exactly here, we don't wanna "?" to add. We want to let the UI get broken if the resource is not there. Because it should ALWAYS be there, in this place of the code, and if it is not there, something in the code around is wrong. We don't wanna not to see the the error we should see Due to "?" here, the UI may act like everything is working in case if not, and that is bad. Worse than showing an error.

PR #1532, opened 2023-09-12
[WIP] CNV-23832: Apply StorageClass from Template's boot source

avivtur:
## 📝 Description

When there's no StorageClass defined in the template or default StorageClass in the cluster, the CDI clone process can't determine which StorageClass to use and the DataVolume does not create its PersistentVolumeClaim.
If a template has an available boot source, we should apply the boot source's StorageClass if exist to allow VM to be created.

Note: This applies to the quick create flow for now, as we some issue with customize flow: 

## 🎥 Demo

Before:

https://github.com/kubevirt-ui/kubevirt-plugin/assets/67270715/3b0a0828-8e1e-4349-9178-cb532fd047d6

After:

https://github.com/kubevirt-ui/kubevirt-plugin/assets/67270715/f25b3270-165f-4a73-9ed2-27885ef72d06

### DIFF
--- a/src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
+++ b/src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
@@ -13,7 +13,9 @@ import {
   V1Template,
 } from '@kubevirt-ui/kubevirt-api/console';
 import VirtualMachineModel from '@kubevirt-ui/kubevirt-api/console/models/VirtualMachineModel';
+import { V1DataVolumeTemplateSpec } from '@kubevirt-ui/kubevirt-api/kubevirt';
 import { updateCloudInitRHELSubscription } from '@kubevirt-utils/components/CloudinitModal/utils/cloudinit-utils';
+import { isEqualObject } from '@kubevirt-utils/components/NodeSelectorModal/utils/helpers';
 import {
   addSecretToVM,
   applyCloudDriveCloudInitVolume,
@@ -25,12 +27,10 @@ import {
   ANNOTATIONS,
   getTemplateOS,
   getTemplateVirtualMachineObject,
-  LABEL_USED_TEMPLATE_NAME,
-  LABEL_USED_TEMPLATE_NAMESPACE,
 } from '@kubevirt-utils/resources/template';
-import { getMemoryCPU } from '@kubevirt-utils/resources/vm';
+import { TemplateBootSource } from '@kubevirt-utils/resources/template/hooks/useVmTemplateSource/utils';
+import { getDataVolumeTemplates } from '@kubevirt-utils/resources/vm';
 import { ensurePath, isEmpty } from '@kubevirt-utils/utils/utils';
-import { k8sCreate } from '@openshift-console/dynamic-plugin-sdk';
 import { useAccessReview, useK8sModels } from '@openshift-console/dynamic-plugin-sdk';
 import {
   Alert,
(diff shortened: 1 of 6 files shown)


metalice:
this is a partial fix? not fixing customize flow ?

src/views/catalog/utils/quick-create-vm.ts
```diff
@@ -27,13 +31,21 @@ type QuickCreateVMType = (inputs: {
...
     subscriptionData: RHELAutomaticSubscriptionData;
+    templateBootSource: TemplateBootSource;
   };
   template: V1Template;
 }) => Promise<V1VirtualMachine>;
 
 export const quickCreateVM: QuickCreateVMType = async ({
   models,
-  overrides: { authorizedSSHKey, name, namespace = DEFAULT_NAMESPACE, startVM, subscriptionData },
+  overrides: {
```
metalice, line 41: the override getting bloated, is there another way to resolve it?

src/views/catalog/utils/quick-create-vm.ts
```diff
@@ -47,6 +59,8 @@ export const quickCreateVM: QuickCreateVMType = async ({
 
   const vm = getTemplateVirtualMachineObject(processedTemplate);
 
+  const hasSC = await hasStorageClass(vm);
```
metalice, line 62: please change var and function name to something more clear, if feels to me like hasSC is very generic and this is not the case here

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/utils/helpers.ts
```diff
@@ -0,0 +1,22 @@
+import { StorageClassModel } from '@kubevirt-ui/kubevirt-api/console';
+import { IoK8sApiStorageV1StorageClassList } from '@kubevirt-ui/kubevirt-api/kubernetes';
+import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import { isDefaultStorageClass } from '@kubevirt-utils/components/DiskModal/DiskFormFields/utils/helpers';
+import { getDataVolumeTemplates } from '@kubevirt-utils/resources/vm';
+import { isEmpty } from '@kubevirt-utils/utils/utils';
+import { k8sGet } from '@openshift-console/dynamic-plugin-sdk';
+
+export const hasStorageClass = async (vm: V1VirtualMachine) => {
```
metalice, line 9: please change the name of the function, From the code I can see that if not all dvt has sc it will fetch and return true if default sc exists. so maybe name it shouldReplaceToTemplateStorageClass (not one of my best) or something better?

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
```diff
@@ -90,7 +93,7 @@ export const TemplatesCatalogDrawerCreateForm: FC<TemplatesCatalogDrawerCreateFo
       verb: 'create',
     });
 
-    const onQuickCreate = () => {
+    const onQuickCreate = async () => {
```
metalice, line 96: u add async but not await? so no benefit here, there still .then() :)

src/views/catalog/utils/quick-create-vm.ts
avivtur: refactored quickCreate to apply all overrides in the onQuickCreate method, for some reason we were modifying the VM before and after processing the template which looked weird to me, tested multiple use cases should work the same as before.

avivtur:
@metalice Currently this PR is using the first SC it finds from the list (as the 4.15 cluster has volume snapshots and not PVC) we should figure out a way to get the SC from the boot source or simply use the first one as I do currently.
BTW this functionality will exist only in UI

PR #1535, opened 2023-09-13
CNV-32665: Fix hot plug NIC not added to VMI after live migration

pcbailey:
## 📝 Description

This fixes a bug wherein a secondary hot-plugged NIC is added to a VM, but it is not added to the VMI after live migration.

jira issue: https://issues.redhat.com/browse/CNV-32665


src/utils/resources/vm/utils/network/constants.ts
```diff
@@ -12,10 +12,10 @@ const typeHandler = {
   },
 };
 
-const types = {
+export const NetworkTypes = {
```
metalice, line 15: no need to export it. u can simply use import interfaceTypes, it is a proxy, you will get a value from type if exisit if not u will get Bridge as the default value.

PR #1537, opened 2023-09-13
CNV-32824: Paste long string into vnc console

metalice:
## 📝 Description

VNC console can't paste long strings in the current implementation.
VNC console is 3rd party component we are consuming and apparently, it can handle fast key sending which causes the paste to break for long strings(probably a buffer issue). I've added a sleep function before each key is sent in order to slow down the rate and make it work.

## 🎥 Demo

Before:
ssh-key-console-before

After:
ssh-key-console-after

### DIFF
--- a/src/utils/components/Consoles/components/vnc-console/VncConsole.tsx
+++ b/src/utils/components/Consoles/components/vnc-console/VncConsole.tsx
@@ -17,6 +17,7 @@ import {
 import { css } from '@patternfly/react-styles';
 import styles from '@patternfly/react-styles/css/components/Consoles/VncConsole';
 
+import { sleep } from '../../utils/utils';
 import { ConsoleState, WS, WSS } from '../utils/ConsoleConsts';
 import useCopyPasteConsole from '../utils/hooks/useCopyPasteConsole';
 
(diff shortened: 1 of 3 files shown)


src/utils/components/Consoles/components/vnc-console/VncConsole.tsx
```diff
@@ -132,12 +133,17 @@ export const VncConsole: FC<VncConsoleProps> = ({
             return;
           }
           const clipboardText = await navigator?.clipboard?.readText?.();
-          [...(clipboardText || pasteText.current)].forEach((char) => {
+          const text = clipboardText || pasteText.current;
+          const lastItem = text.length - 1;
+          [...text].forEach(async (char, i) => {
             const shiftRequired = isShiftKeyRequired(char);
-
+            await sleep(i * 50);
```
pcbailey, line 140: Why is the sleep duration based on the index?

metalice: For each, map, I anything with a callback don't support async await without it.

This is a bit more detailed
https://zellwk.com/blog/async-await-in-loops/

metalice: WDYT?

avivtur: I'm confused, according to the article you suggested it say clearly that it can not do await inside of forEach.
Why not do a simple for-loop?

metalice: Don't be confused, lol. it says await/async doesn't work on callback the regular way you expect, but with adding the index it does, think about it for 2 minutes I'm sure you will understand why.

the forEach was there before so I simply didn't change the loop, of course, I can change the loop to any other kind.

 If I'm changing the loop I won't need to send the index, but leaving the loop I do need to send the index.
seems to like tomato, tomâto from a code perspective as change loop vs. send index, 
but if you guys prefer something else I would be happy to change it.

pcbailey: My only concern is with performance. The wait increases with each iteration. Do you know how long the string would have to get before it becomes noticeable?

metalice: hmm, it's a fair point, although I don't think there is an issue, let's make it easy, ill change the loop and let the event loop handle it as normal. sound ok?

pcbailey: I'm good with whatever solution you think will get it done without bogging it down. If you think this solution will do it then we can push forward with it. We can always change it if we get complaints.

PR #1540, opened 2023-09-14
CNV-32927: Fix instancetypes and preferences for non admin user

metalice:
## 📝 Description

Virtualization -> {instance type, perferences} was selected to show by CAN_LIST_NS flag, which doesn't deliver the right answer when the non-admin user has privileges to view those objects.
This fix is creating a new 2 flags that check the view of a user to those objects and present the menu links accordingly.


src/utils/flags/useEnableKubevirtMenuFlags.ts
```diff
@@ -0,0 +1,24 @@
+import VirtualMachineClusterInstancetypeModel from '@kubevirt-ui/kubevirt-api/console/models/VirtualMachineClusterInstancetypeModel';
+import VirtualMachineClusterPreferenceModel from '@kubevirt-ui/kubevirt-api/console/models/VirtualMachineClusterPreferenceModel';
+import { K8sVerb, SetFeatureFlag, useAccessReview } from '@openshift-console/dynamic-plugin-sdk';
+
+import { FLAG_KUBEVIRT_INSTANCETYPES, FLAG_KUBEVIRT_PREFERENCES } from './consts';
+
+const useKubevirtMenuFlags = (setFeatureFlag: SetFeatureFlag) => {
```
pcbailey, line 7: nitpick: Why are the file name and hook name different?

metalice: U are very right!

metalice: done

avivtur (commented):
I think we should show the pages and if a user can't list them, the page should show an error like other pages we have

PR #1542, opened 2023-09-14
CNV-32920: Disable Stop button for VM being provisioned

hstastna:
## 📝 Description

**Fixes:**
https://issues.redhat.com/browse/CNV-32920

Disable _Stop_ button in _Actions_ drop down for a VM that is being provisioned, as clicking on that button at this stage does not have any meaning, it should not be active.

## 🎥 Screenshots
**Before:**
Possible to click on _Stop_ button, which is unnecessary for a VM with _Provisioning_ status:
prov_before

**After:**
_Stop_ button disabled:
prov_after


hstastna:

PR #1543, opened 2023-09-14
CNV-32958: Rename "From volume" to "From InstanceType"

hstastna:
## 📝 Description

**Fixes:**
https://issues.redhat.com/browse/CNV-32958

When creating a VM in VirtualMachines list page, change creating VM drop down item from "From volume" to "From InstanceType" to align more with "InstanceTypes" tab, to which clicking on that menu item leads.

## 🎥 Screenshots
**Before:**
is_before

**After:**
is_after


hstastna:

PR #1546, opened 2023-09-15
Use getCPUcores selector in checkCPUMemoryChanged

hstastna:
## 📝 Description

Use the existing `getCPUcores` selector in `checkCPUMemoryChanged` where possible, also add other minor improvements to the code like add missing "instance" to the description of some VMI related selectors, also move `getCPUcores` to VM selectors related file.

## 🎥 Demo

No any visual changes

### DIFF
--- a/src/utils/resources/vm/utils/selectors.ts
+++ b/src/utils/resources/vm/utils/selectors.ts
@@ -185,3 +185,10 @@ export const getMemoryCPU = <T>(obj: T): { cpu: V1CPU; memory: string } => ({
   cpu: getCPU(obj),
   memory: getMemory(obj),
 });
+
+/**
+ * A selector that returns number of CPU cores of the resource
+ * @param {T} obj resource such as vm or vmi
+ * @returns {number} the number of CPU cores
+ */
+export const getCPUcores = <T>(obj: T): number => getCPU(obj)?.cores || 1;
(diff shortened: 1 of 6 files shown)


hstastna:

src/utils/resources/vm/utils/selectors.ts
```diff
@@ -185,3 +185,10 @@ export const getMemoryCPU = <T>(obj: T): { cpu: V1CPU; memory: string } => ({
   cpu: getCPU(obj),
   memory: getMemory(obj),
 });
+
+/**
+ * A selector that returns number of CPU cores of the resource
+ * @param {T} obj resource such as vm or vmi
+ * @returns {number} the number of CPU cores
+ */
+export const getCPUcores = <T>(obj: T): number => getCPU(obj)?.cores || 1;
```
pcbailey, line 194: Previously it was `|| 0`. Why change it to `|| 1`?

hstastna: I think it was a bug. It does not make sense to set it to 0. In other places in the code we used to use "1" - it was refactored by Matan recently already. Correct me if I am wrong.

pcbailey: I'm not sure. We should ask him to make sure we're not introducing a regression. Other than that this PR looks good. Thanks for the nice refactor! 😄

metalice: Hey, i think the default value should be 1 , as this is cpu a VM cannot exist without at least 1...
/lgtm

PR #1547, opened 2023-09-15
CNV-33008: Display CPU, Memory VM Template's values first

hstastna:
## 📝 Description

**Fixes:**
https://issues.redhat.com/browse/CNV-33008

Display CPU and Memory values in VM Template _Details_ page before "CPU" and "Memory" words in _CPU | Memory_ field, to align with the rest of the UI.

Additionally, add little bit of refactoring, use `getMemoryCPU` and `getCPUcores` selectors in the related files.

## 🎥 Screenshots
**Before:**
template_cpu_before

**After:**
template_cpu_after


hstastna:

src/views/templates/list/hooks/useVirtualMachineTemplatesCPUMemory.tsx
```diff
@@ -1,23 +1,17 @@
...
+import { getMemoryCPU } from '@kubevirt-utils/resources/vm';
 import { readableSizeUnit } from '@kubevirt-utils/utils/units';
 import { isTemplateParameter } from '@kubevirt-utils/utils/utils';
 import { Popover, PopoverPosition } from '@patternfly/react-core';
 import { HelpIcon } from '@patternfly/react-icons';
 
-export const useVirtualMachineTemplatesCPUMemory = (
-  template: V1Template,
-): React.ReactNode | string => {
+export const useVirtualMachineTemplatesCPUMemory = (template: V1Template): ReactNode | string => {
```
avivtur, line 12: Hi :) 
A little suggestion: Making this a component instead of a hook would be better suited here it returns a component, Please consider to change it.

avivtur (commented):
Hi :) 
just a small comment, other than that it looks great :+1:

src/views/templates/list/hooks/useVirtualMachineTemplatesCPUMemory.tsx
hstastna: I'd better delete that popover related code as I think it is deprecated. I was never able to find any scenario to see that code in the UI, I also have no idea why it is there. But I've found this PR: https://github.com/kubevirt-ui/kubevirt-plugin/pull/331
It looks like that additional code with adding popover to this hook was used to display in the VM Templates list. Now when I look at templates list, it looks like this:
vmt
=> no any popover in the list, regarding CPU, Memory, no matter the Red Hat template or user created one, it looks the same. IMO someone just forgot to delete that code when the UI was changed. WDYT about it?

avivtur: If you're sure that this code is not needed than +1 on removing it

hstastna: Checked with Yifat also and it seems we really can remove that old code :)

PR #1552, opened 2023-09-19
CNV-32957: Add cloud-init basic info to instancetype cli command

metalice:
## 📝 Description

The current cli command in the catalog instancetype was with sshkey but without any other user data that belongs to cloud int.
## 🎥 Demo


avivtur:
@metalice Looks good, Can you please add screenshots of the cloud-init YAML that is modified?

metalice:
> @metalice Looks good, Can you please add screenshots of the cloud-init YAML that is modified?

I guess it will look similar as there is already base64 string there :) but sure why not

PR #1553, opened 2023-09-19
CNV-32923: Fix clicking enter after copy console

metalice:
## 📝 Description

When clicking paste in the console screen the button stays in focus and when clicking enter it pastes again.
the fix is those 2 lines of code:
e.currentTarget.blur();
e.preventDefault();

the rest is to remove React.* 
## 🎥 Demo

### DIFF
--- a/src/utils/components/Consoles/components/vnc-console/VncConsole.tsx
+++ b/src/utils/components/Consoles/components/vnc-console/VncConsole.tsx
@@ -1,4 +1,13 @@
-import React, { FC, ReactNode, useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import React, {
+  FC,
+  MouseEvent,
+  ReactNode,
+  useCallback,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from 'react';
 import cn from 'classnames';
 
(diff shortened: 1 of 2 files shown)


src/utils/components/Consoles/components/vnc-console/VncConsole.tsx
```diff
@@ -201,9 +210,13 @@ export const VncConsole: FC<VncConsoleProps> = ({
               text: 'Ctrl + Alt + 2',
             },
           ]}
+          onInjectTextFromClipboard={(e: MouseEvent<HTMLButtonElement>) => {
+            e.currentTarget.blur();
+            e.preventDefault();
+            rfb?.sendPasteCMD();
```
avivtur, line 216: Is there any way to focus on the console section too?

avivtur (commented):
Looks good, just one question :)

src/utils/components/Consoles/components/vnc-console/VncConsole.tsx
metalice: yes. rfb.focus()
but I didn't like it as it makes the screen jump down

PR #1558, opened 2023-09-20
CNV-33013: Sort the instanceType list by the size of CPU and memory

avivtur:
## 📝 Description

Sorting IT according to their CPUs and Memory

## 🎥 Demo

Before:
it-sorted-b4-2
it-sorted-b4

After:
it-sorted-2
it-sorted


src/utils/components/AddBootableVolumeModal/components/VolumeMetadata/components/InstanceTypeDrilldownSelect/utils/utils.ts
```diff
@@ -85,4 +87,12 @@ export const getInstanceTypeMenuItems = (
 
     return acc;
   }, initialMenuItems);
+
+  return produce(itemsData, (draftItemsData) => {
+    draftItemsData.redHatProvided.items = itemsData.redHatProvided.items.map((series) => {
+      return produce(series, (seriesDraft) => {
```
metalice, line 93: why those produce? what benefit do we get here from using produce inside a simple function? immer , etc , is usually good for shared objects that you don't want to mutate, such as the VM object. here you create itemsData... as reduce return new object.. i don't see any benefit for it

PR #1559, opened 2023-09-20
CNV-33063: Add KSM to Cluster settings

hstastna:
## 📝 Description

**Feature:**
https://issues.redhat.com/browse/CNV-33063

_Design:_
https://docs.google.com/document/d/1f1InBA6gNQafYpvy1xTotDjDVmBwtbsPB8tlSSEQjYo/

_Depends on (still WIP):_
https://issues.redhat.com/browse/CNV-32410

Add KSM (_Kernel Samepage Merging_) configuration to cluster settings, under _Resource management_ expandable section.

Actually, it was decided that this functionality will just enable or disable KSM on all nodes, no any special configuration of specific nodes (yet). If `nodeLabelSelector` is nil, KSM will not be enabled on any nodes. Empty nodeLabelSelector (`{}`) will enable KSM on every node.

_More about KSM:_
https://docs.kernel.org/admin-guide/mm/ksm.html
https://kubevirt.io/user-guide/operations/ksm/

## 🎥 Screenshots
**Before:**
No any _Resource management_ section, no any KSM feature:
ksm_before

**After:**
_Resource management_ expandable section with KSM feature available:
ksm_new
If some error occurs:
ksm_err


src/views/clusteroverview/SettingsTab/ClusterTab/components/ResourceManagementSection/ResourceManagementSection.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+import ExpandSection from '../../../ExpandSection/ExpandSection';
+
+import KernelSamepageMerging from './components/KernelSamepageMerging/KernelSamepageMerging';
+
+const ResourceManagementSection: FC = () => {
+  const { t } = useKubevirtTranslation();
+
+  return (
+    <ExpandSection toggleText={t('Resource management')}>
```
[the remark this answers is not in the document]
hstastna: I don't think so. I think using Stack or other components here is unnecessary. We are adding 2 sub-expandable-sections under this "Resource management" section. It should be fine, we just need to adjust left padding a bit (for example as I did [here](https://github.com/kubevirt-ui/kubevirt-plugin/pull/1559/files#diff-cbf3928574b48e909288f532600beac247b7e1d5db10388a87ed1600ca2f546fR3)), to align with the design.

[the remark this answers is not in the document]
pcbailey: Ok. I added the left padding as well, but did it on the `Stack` so the whole section is indented. We just need to make sure we have the correct vertical spacing when there's more than one item in the section.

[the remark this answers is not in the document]
hstastna: `Stack` component is unnecessary once we update left padding. `ExpandSection` component has already some indentation. I'd avoid adding unnecessary components, it makes some things worse, for example some PF components tend not to working properly once those are too deep in the tree.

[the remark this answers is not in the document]
pcbailey: I understand we need to ensure the left spacing is correct. I meant we need to ensure the vertical spacing between the 'Auto-compute CPU limits' and the 'Kernel Samepage Merging (KSM)' items that will be in the 'Resource management section. The `Stack` component formats items stacked vertically.

If we need to add vertical spacing and the options are to use a layout component or adding additional CSS that does the same thing as the layout component, I think we should use the layout component as that's what it was designed for.

I haven't encountered an issue with PF components not working if they're too deeply nested, so I'm not sure what you mean there. Was a bug opened for the problem?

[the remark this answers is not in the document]
hstastna: > If we need to add vertical spacing and the options are to use a layout component or adding additional CSS that does the same thing as the layout component, I think we should use the layout component as that's what it was designed for.

I think it is too early to know for sure what the best option is for this section. 

> I haven't encountered an issue with PF components not working if they're too deeply nested, so I'm not sure what you mean there. Was a bug opened for the problem?

Simply various behavior not working properly or not at all. Various components or their props, yes, bugs opened but it takes forever for fixes, usually. But I think for now for these features we are okay, don't worry too much. If something occurs, we're gonna solve it eventually.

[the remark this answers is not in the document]
pcbailey: > I think it is too early to know for sure what the best option is for this section.

Agreed. I plan on opening my WIP soon, but I still have work to do. We can talk about it once one of us gets close to finishing their PR.

src/utils/hooks/useHyperConvergeConfiguration.ts
```diff
@@ -7,6 +7,7 @@ import { K8sResourceCommon, useK8sWatchResource } from '@openshift-console/dynam
 
 export type HyperConverged = K8sResourceCommon & {
   spec: {
+    configuration?: { ksmConfiguration: { nodeLabelSelector?: object } };
```
metalice, line 10: instead of object (as if i remember correctly can be a lot of things) and according the pr description this can be simply a {} ?

src/views/clusteroverview/SettingsTab/ClusterTab/components/ResourceManagementSection/components/KernelSamepageMerging/KernelSamepageMerging.tsx
```diff
@@ -0,0 +1,89 @@
...
+  useEffect(() => {
+    if (!hyperLoaded || hyperLoadingError) {
+      setError(hyperLoadingError);
+    } else {
+      const ksmConfiguration = hyperConverge?.spec?.configuration?.ksmConfiguration;
+
+      setError(null);
+      setIsEnabled(!!(ksmConfiguration && isEmpty(ksmConfiguration?.nodeLabelSelector))); // Empty nodeLabelSelector will enable KSM on every node.
+    }
+  }, [hyperConverge, hyperLoaded, hyperLoadingError]);
```
metalice, line 37: I don't think this useeffect is needed. you can wait for hyperLoaded and for error in line 75 u can use error || hyperLoadingError

src/views/clusteroverview/SettingsTab/ClusterTab/components/ResourceManagementSection/components/KernelSamepageMerging/KernelSamepageMerging.tsx
```diff
@@ -0,0 +1,89 @@
...
+        {
+          op: 'replace',
+          path: `/spec/configuration/ksmConfiguration/nodeLabelSelector`,
+          value: value ? {} : null,
+        },
+      ],
+      model: HyperConvergedModel,
+      resource: hyperConverge,
+    })
+      .then(() => setIsEnabled(value))
```
metalice, line 53: u already doing it on line 40 - why u need here also?

src/views/clusteroverview/SettingsTab/ClusterTab/components/ResourceManagementSection/components/KernelSamepageMerging/KernelSamepageMerging.tsx
```diff
@@ -0,0 +1,89 @@
...
+      .catch((err: Error) => setError(err));
```
metalice, line 54: the function this is running is async , u can use await and try and catch

hstastna: I can but I am not sure. I've found 51 results in 44 files of our repo where we deal with async functions like this. And with this it is IMHO more obvious what and when happens.

src/views/clusteroverview/SettingsTab/ClusterTab/components/ResourceManagementSection/components/KernelSamepageMerging/KernelSamepageMerging.tsx
hstastna: Thanks, I'll take care about that!

src/views/clusteroverview/SettingsTab/ClusterTab/components/ResourceManagementSection/components/KernelSamepageMerging/KernelSamepageMerging.tsx
metalice: but u already doing it on line 40 - u just double doing it without a reason it don't make any sense.
in line 40 u set the enable already no matter if it will pass or failed, so if u call will fail u will have false enabled/disabled. u need to make the logic correct here.

src/views/clusteroverview/SettingsTab/ClusterTab/components/ResourceManagementSection/components/KernelSamepageMerging/KernelSamepageMerging.tsx
metalice: i dont know about other places in code, but this is wrong, this is anti pattern, u should remove the async or add await, no benefit at all in this way. don't just look on other places on the code and copy, or use as conformations.

metalice: i suggest looking a bit about promises and seeing the difference before u decide this is a good way to handle this function. as there is no sense in this combination that you claim that is more clear.

src/views/clusteroverview/SettingsTab/ClusterTab/components/ResourceManagementSection/components/KernelSamepageMerging/KernelSamepageMerging.tsx
hstastna: I see the point. There is something I forgot to delete, haha. Thanks!

src/views/clusteroverview/SettingsTab/ClusterTab/components/ResourceManagementSection/components/KernelSamepageMerging/KernelSamepageMerging.tsx
hstastna: Ah, I see. I think we misunderstood each other again about what is the problem, but let me take a closer look at this again, and I'll solve it for sure..

src/utils/hooks/useHyperConvergeConfiguration.ts
hstastna: So as I've expected, this was not good idea. Did you ever try to add {} there? If yes, then you must have seen the following:
```
Don't use `{}` as a type. `{}` actually means "any non-nullish value".
- If you want a type meaning "any object", you probably want `Record<string, unknown>` instead.
- If you want a type meaning "any value", you probably want `unknown` instead.
- If you want a type meaning "empty object", you probably want `Record<string, never>` instead.
```
:)

hstastna:

PR #1565, opened 2023-09-26
[release-4.12] CNV-33504: Clone VM should clone from pvc and not from pvc source

avivtur:
## 📝 Description

Manual backport for #1509


avivtur:
/retitle [release-4.12] CNV-33504: Clone VM should clone from pvc and not from pvc source

PR #1567, opened 2023-09-27
CNV-33388: Developer Preview notation on Instancetype tab hard to read in dark mode

avivtur:
## 📝 Description

Removing our CSS rules allows PF to handle the dark mode for us for the "Developer Preview" label

## 🎥 Demo

Before:
dev-prev-label-dmode

After:
dev-prev-label-dmode-after


metalice:
Please attache a screenshot of the white theme, Is it working as before?

src/utils/components/DeveloperPreviewLabel/DeveloperPreviewLabel.tsx
```diff
@@ -15,7 +13,7 @@ const DeveloperPreviewLabel: React.FC = () => {
         'Developer preview features are not intended to use in production environments. The clusters deployed with the developer preview features are developmental clusters and are not currently supported by Red Hat.',
       )}
     >
-      <Label isCompact icon={<InfoCircleIcon />} className="dev-preview-label">
+      <Label isCompact icon={<InfoCircleIcon />} color="orange">
```
hstastna, line 16: I remember someone was bad to me some time ago exactly because of this change and wanted me to put it all to a separate css file, haha 🥲

hstastna (approved):

src/utils/components/DeveloperPreviewLabel/DeveloperPreviewLabel.tsx
metalice: The preferred way is external css, as we don't do inline styling as it is tough to override when needed, so the color prop is the same. It is more like a pattern fly issue that, on theme change, they don't change the var color. Another solution for this bug can be adding a selector for a dark theme and changing to a different color aviv

metalice: @hstastna please don't approve PR's while I requested more info just 10 minutes before you approved. it's not nice.

hstastna:
> Please attache a screenshot of the white theme, Is it working as before?

After:
dev_after

PR #1575, opened 2023-10-04
CNV-29457: Move cloud-init network data to v2

avivtur:
## 📝 Description

Restructuring cloud-init network data according to v2, for more info: https://cloudinit.readthedocs.io/en/latest/reference/network-config-format-v2.html#addresses-sequence-of-scalars

## 🎥 Demo

Before:
v
v1-cloud-init

After:
v2-cloud-init-form
v2-cloud-init
ip-addr


src/utils/components/CloudinitModal/utils/cloudinit-utils.ts
```diff
@@ -75,27 +75,17 @@ export const convertYAMLToNetworkDataObject = (networkData: string): CloudInitNe
...
-                    gateway: networkData?.gateway,
-                    type: 'static',
-                  },
-                ],
-                type: 'physical',
-              },
-            ],
-            version: '1',
+          ethernets: {
+            [networkData?.name || '']: {
```
metalice, line 83: i dont htink u need || "" as u are checking in hasValue that it isn't empty

metalice:
u sure about removing gateway? there is gateway4 instead
I saw something about deprecated but here https://issues.redhat.com/browse/CNV-28700 , I can see sample its still in use

src/utils/components/CloudinitModal/utils/cloudinit-utils.ts
```diff
@@ -75,27 +75,17 @@ export const convertYAMLToNetworkDataObject = (networkData: string): CloudInitNe
...
+              addresses: (networkData?.addresses || '')?.replace(/\s/g, '').split(','),
+            },
           },
-        })
+          version: 2,
+        } as CloudInitNetwork)
```
metalice, line 88: no v2clooudinitnetwork?

src/utils/components/CloudinitModal/utils/cloudinit-utils.ts
avivtur: I'm checking if either name isn't empty or the addresses which means that the other can be empty, in that case I do need the `|| ''`

avivtur:
> u sure about removing gateway? there is gateway4 instead I saw something about deprecated but here https://issues.redhat.com/browse/CNV-28700 , I can see sample its still in use

No I'm not sure about it, I just saw in the cloud-init docs that it's depracated, and it suggest using routes instead of gateway4 (or 6): https://netplan.readthedocs.io/en/latest/netplan-yaml/#default-routes

For this PR I will expose the gateway4 as you suggested, and I guess in the future we will move to routes

metalice (approved):

PR #1581, opened 2023-10-09
CNV-33095: add volumesnapshot option in the add bootable volume modal

upalatucci:
## 📝 Description

Add the VoumeSnapshot option in the bootable volume.

With this option, we'll create a `DataSource` with `spec.source.snapshot`
UI is the same as the PVC option.

## 🎥 Demo

Screenshot from 2023-10-09 10-58-23
Screenshot from 2023-10-09 10-58-04
Screenshot from 2023-10-09 10-58-01


src/utils/components/SelectSnapshot/SelectSnapshot.tsx
```diff
@@ -0,0 +1,57 @@
...
+  );
+
+  const snapshotNames = useMemo(
+    () =>
+      snapshots?.map((snapshot) => snapshot?.metadata?.name)?.sort((a, b) => a?.localeCompare(b)),
+    [snapshots],
+  );
+
+  return (
+    <div>
```
metalice, line 38: div or fragment?

src/utils/components/SelectSnapshot/SelectSnapshotNamespace.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+type SelectSnapshotNamespaceProps = {
+  isDisabled?: boolean;
+  onChange: Dispatch<SetStateAction<string>>;
+  projectsLoaded: boolean;
+  projectsName: string[];
+  selectedProject: string;
+};
+
+export const FilterSnapshotSelect = (options: string[]) => {
```
metalice, line 17: seems to me like its repeating with the file above

upalatucci: Yes, I didn't notice. I'll do a common utils.

src/utils/components/SelectSnapshot/utils.tsx
```diff
@@ -0,0 +1,18 @@
...
+    let newOptions = options;
+
+    if (value) {
+      const regex = new RegExp(value, 'i');
+      newOptions = options.filter((namespace) => regex.test(namespace));
+    }
+
+    return newOptions.map((namespace) => (
+      <SelectOption key={namespace} value={namespace} />
+    )) as ReactElement[];
```
metalice, line 16: hmm returning elements from util function isnt recommended, what about return the array and in the consuming component just map over it? like we are doing with dropdown items or other select items?

src/utils/components/SelectSnapshot/SelectSnapshotNamespace.tsx
```diff
@@ -0,0 +1,78 @@
+import React, { Dispatch, FC, SetStateAction, useCallback, useState } from 'react';
```
metalice, line 1: cant use a generic component for name and namespace? they are very similar

src/utils/components/SelectSnapshot/useSnapshots.ts
```diff
@@ -0,0 +1,70 @@
...
+
+type UseSnapshotsReturnType = {
+  error: Error;
+  projectsLoaded: boolean;
+  projectsNames: string[];
+  snapshots: VolumeSnapshotKind[];
+  snapshotsLoaded: boolean;
+};
+
+export type VolumeSnapshotKind = K8sResourceCommon & {
```
metalice, line 19: if u are exporting a type better to put in a new tyhpes.ts near util, other components can consume it not from a hook file, just make it more organize

PR #1583, opened 2023-10-12
Add ExpandSectionWithSwitch component

pcbailey:
## 📝 Description

This PR:

- Adds the ExpandSectionWithSwitch component, which will be used by the kernel same-page merging and auto-compute CPU limits features on the Settings page
- Adds the `isIndented` and `isDisabled` options to the `ExpandSection` component
- Adds the `isIndented` prop to all expandable sections on the Settings page

### ExpandSectionWithSwitch example (enabled and switch on)

expand-section-with-switch-example2023-10-12_09-09

### ExpandSectionWithSwitch example (disabled)

expand-section-with-switch-disabled-example-2023-10-12_09-09

### Settings page before (without indents)

settings-page-expanded-no-indent-2023-10-12_09-08

### Settings page after (with indents)

settings-page-expanded-2023-10-12_09-04

### DIFF
--- a/src/utils/components/ExpandSectionWithSwitch/ExpandSectionWithSwitch.tsx
+++ b/src/utils/components/ExpandSectionWithSwitch/ExpandSectionWithSwitch.tsx
@@ -1,0 +1,51 @@
+import React, { FC, FormEvent, ReactNode } from 'react';
+
+import HelpTextIcon from '@kubevirt-utils/components/HelpTextIcon/HelpTextIcon';
+import { PopoverPosition, Split, SplitItem, Switch } from '@patternfly/react-core';
+
+import ExpandSection from '../../../views/clusteroverview/SettingsTab/ExpandSection/ExpandSection';
+
+import './ExpandSectionWithSwitch.scss';
+
+type ExpandSectionWithSwitchProps = {
+  children: ReactNode;
+  helpTextIconContent?: ReactNode;
+  isDisabled?: boolean;
+  switchIsOn: boolean;
+  toggleContent?: ReactNode;
+  toggleText?: string;
+  turnOnSwitch: (checked: boolean, event: FormEvent<HTMLInputElement>) => void;
+};
+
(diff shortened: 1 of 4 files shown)


upalatucci:
for me is lgtm.  @hstastna

src/utils/components/ExpandSectionWithSwitch/ExpandSectionWithSwitch.tsx
```diff
@@ -0,0 +1,58 @@
...
+const ExpandSectionWithSwitch: FC<ExpandSectionWithSwitchProps> = ({
+  children,
+  helpTextIconContent,
+  isDisabled,
+  switchIsOn,
+  toggleContent,
+  toggleText,
+  turnOnSwitch,
+}) => {
+  return (
```
hstastna, line 29: I think you don't need `return` here, you can omit it.

hstastna (commented):
I think adding `isIndented` prop is unnecessary, as I think all that content should have already been intended nicely and by default, without any extra prop. I mean indentation is great, I'd leave it there by default, just adding a prop to do so is unnecessary. 

Just a note that according to the design changes, KSM feature won't need this component anymore.

src/utils/components/ExpandSectionWithSwitch/ExpandSectionWithSwitch.tsx
pcbailey: Done!

pcbailey:
>  I think adding `isIndented` prop is unnecessary, as I think all that content should have already been intended nicely and by default, without any extra prop.

@hstastna `isIndented` defaults to false in the `ExpandableSection` component, so the prop must be added to indent the contents. Check the "Settings page before (without indents)" screenshot in the description. Currently, the contents are left-aligned with the arrow. Adding the `isIndented` prop fixed that and aligned the contents with the label as shown in the "Settings page after (with indents)" screenshot.

hstastna (approved):

PR #1587, opened 2023-10-16
CNV-29738: Adding netwrok ping tool -checkups

metalice:
## 📝 Description

Adding network checkups tool
https://docs.google.com/document/d/15xX-JPNt90bBGEVWhqKe6o-AyeluIVc6519hzXUF0_E/edit
https://issues.redhat.com/browse/CNV-29738

## 🎥 Demo

checkups


src/views/checkups/network/components/CheckupsNetworkStatusIcon.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const statusJob =
+    (job?.status?.succeeded === 1 && 'done') ||
+    (job?.status?.active === 1 && 'running') ||
+    (job?.status?.succeeded === 0 && 'failed');
+
+  const statusConfigMap =
+    (configMap?.data?.[STATUS_SUCCEEDED] === 'true' && 'done') ||
+    ((configMap?.data?.[STATUS_SUCCEEDED] === 'false' || statusJob === 'failed') && 'failed') ||
+    (configMap?.data?.[STATUS_SUCCEEDED] === undefined && statusJob === 'done' && 'failed') ||
+    (configMap?.data?.[STATUS_SUCCEEDED] === undefined && job?.status?.active === 1 && 'running');
```
upalatucci, line 32: Here we can have two functions  maybe?

src/views/checkups/network/components/form/CheckupsNetworkFormNads.tsx
```diff
@@ -0,0 +1,53 @@
...
+          >
+            {selectedNad || t('Select NetworkAttachmentDefinition')}
+          </DropdownToggle>
+        }
+        className="CheckupsNetworkForm--main__nads"
+        disabled={isEmpty(nads)}
+        dropdownItems={nadsDropdownItems}
+        id="nad"
+        isOpen={isOpenDropdown}
+      />
```
upalatucci, line 48: I would use a Select with Typeahead, so that the user can search the right NAD in situations where there are too much

src/views/checkups/network/components/form/CheckupsNetworkFormNodes.tsx
```diff
@@ -0,0 +1,134 @@
...
+                    id="toggle-source-node"
+                  >
+                    {nodeSource || t('Select node')}
+                  </DropdownToggle>
+                }
+                disabled={isEmpty(nodesDropdownItems)}
+                dropdownItems={nodesDropdownItems}
+                id="source-node"
+                isOpen={isNodeSourceOpen}
+              />
```
upalatucci, line 101: Same here. I would use Select with typeahead. We don't know the number of nodes. There can be 200 in the cluster and searching in a dropdown could be awful

src/views/checkups/network/hooks/useCheckupsNetworkPermissions.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  const isConfigMapRoleBinding = useMemo(
+    () => findObjectByName(roleBinding, KIAGNOSE_CONFIGMAP_ACCESS),
+    [roleBinding],
+  );
+
+  const isServiceAccount = useMemo(
+    () => findObjectByName(serviceAccounts, VM_LATENCY_CHECKUP_SA),
+    [serviceAccounts],
+  );
```
upalatucci, line 68: Do we know the namespace of those resources? `default` or maybe the current `namespace`?
In that case, we could watch directly the resource and see if the error code is 404

src/views/checkups/network/components/CheckupsNetworkStatusIcon.tsx
metalice: hmm i will think about something because its ugly lol

src/views/checkups/network/components/form/CheckupsNetworkFormNodes.tsx
metalice: same as above :)

src/views/checkups/network/components/form/CheckupsNetworkFormNads.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
+import { isEmpty } from '@kubevirt-utils/utils/utils';
+import { useActiveNamespace } from '@openshift-console/dynamic-plugin-sdk';
+import { Dropdown, DropdownItem, DropdownToggle, FormGroup } from '@patternfly/react-core';
+
+type CheckupsNetworkFormNadsProps = {
+  selectedNad: string;
+  setSelectedNad: Dispatch<SetStateAction<string>>;
+};
+const CheckupsNetworkFormNads: FC<CheckupsNetworkFormNadsProps> = ({
```
pcbailey, line 13: NAD is an abbreviation. I thought our convention was to use all capitals for abbreviations?

src/views/checkups/network/components/form/CheckupsNetworkFormNodes.tsx
```diff
@@ -0,0 +1,134 @@
...
+  isNodes: boolean;
+  nodeSource: string;
+  nodeTarget: string;
+  setIsNodes: SetStateAction<boolean>;
+  setNodeSource: Dispatch<SetStateAction<string>>;
+  setNodeTarget: Dispatch<SetStateAction<string>>;
+};
+
+const CheckupsNetworkFormNodes = ({
+  isNodes,
```
pcbailey, line 35: This name doesn't make sense to me. Could we rename it to something else?

src/views/checkups/network/details/CheckupsNetworkDetailsPage.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const { t } = useKubevirtTranslation();
+  const { configMaps, error, jobs, loading } = useCheckupsNetworkData();
+  const [activeTabKey, setActiveTabKey] = useState<number>(0);
+
+  const configMap = configMaps.find((cm) => cm.metadata.name === vmName);
+  const jobMatches = getJobByName(jobs, configMap?.metadata?.name);
+
+  if (!configMap)
+    return (
+      <Bullseye>
```
pcbailey, line 29: This is used a lot in the codebase. I suggest we make it a dedicated component.

src/views/checkups/network/list/CheckupsNetworkListEmptyState.tsx
```diff
@@ -0,0 +1,85 @@
...
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+
+  return (
+    <EmptyState variant={EmptyStateVariant.large}>
+      <EmptyStateIcon icon={NetworkIcon} />
+      <Title headingLevel="h4" size="lg">
+        {t('No network latency checkups yet')}
+      </Title>
+      <EmptyStateBody>
+        {t('To get started, install permission and then run a checkup')}
```
pcbailey, line 39: You're missing the 's' on 'permissions'.

src/views/checkups/network/list/CheckupsNetworkListEmptyState.tsx
```diff
@@ -0,0 +1,85 @@
...
+          isDisabled={!nadsInNamespace || isLoading}
+          isLoading={isLoading}
+          variant={ButtonVariant.secondary}
+        >
+          {!isLoading && t(isPermitted ? 'Remove permissions' : 'Install permissions')}
+        </Button>
+      </EmptyStateSecondaryActions>
+      {!nadsInNamespace && (
+        <Title className="CheckupsNetworkListEmptyState--title__namespace" headingLevel="h5">
+          {t('Please add NetworkAttachmentDefinition to this namespace in order to use checkups')}
```
pcbailey, line 70: Add an 'a' before 'NetworkAttachmentDefinition'.

pcbailey (commented):
This looks great! I just have a few comments/suggestions.

src/views/checkups/network/components/CheckupsNetworkStatusIcon.tsx
metalice: yes, i will think about something , a switch , if else (which I don't like - it not pretty and confusion IMHO), function or something else.. :)

src/views/checkups/network/hooks/useCheckupsNetworkPermissions.ts
metalice: its the active/current namespace, for serviceAccounts , the first call, as I used there, the other resources I did tried using the namespace but no luck :( 
I can try again to be sure

src/views/checkups/Storage/CheckupsStorageForm.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+import { Form, FormSection } from '@patternfly/react-core';
+
+const CheckupsStorageForm = () => {
+  return (
+    <Form>
+      <FormSection>Place holder</FormSection>
```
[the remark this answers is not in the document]
metalice: if im already making new commit so why not? :)

src/views/checkups/network/components/CheckupsNetworkActions.tsx
```diff
@@ -0,0 +1,63 @@
...
+  Dropdown,
+  DropdownItem,
+  DropdownPosition,
+  DropdownToggle,
+  KebabToggle,
+} from '@patternfly/react-core';
+
+import { deleteNetworkCheckup, rerunNetworkCheckup, STATUS_SUCCEEDED } from '../utils/utils';
+
+const CheckupsNetworkActions = ({
```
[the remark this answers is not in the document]
metalice: im sure you're right about the English :)
I choose this way because of routing, as checkups are first, then network, then form, etc...
if there is something that is backwards I think its better change it like u mention some are netwrokcheckups 

wdyt?

src/views/checkups/network/components/form/CheckupsNetworkFormNads.tsx
metalice: i saw in a few places in code where we used Nads, but sure, lets cut this mistake

src/views/checkups/network/components/form/CheckupsNetworkFormNodes.tsx
metalice: isNodeChecked?
isNodesSelected?
sure way not.. pick something :)

src/views/checkups/network/details/CheckupsNetworkDetailsPage.tsx
metalice: sure

src/views/checkups/network/list/CheckupsNetworkListEmptyState.tsx
metalice: will add

src/views/checkups/network/list/CheckupsNetworkListEmptyState.tsx
metalice: will add

src/views/checkups/network/details/CheckupsNetworkDetailsPage.tsx
metalice: On second thought, I've checked the appearances of this bullseye loading, and I see it has about 20 appearances in different files, you're right we should change it, but not in this pr which is big enough and I don't want to add 20 more files to it, should be in separate pr

metalice:
@pcbailey @upalatucci ptal

upalatucci:
lgtm for me. @pcbailey it's on you

src/views/checkups/network/components/CheckupsNetworkActions.tsx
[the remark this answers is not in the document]
pcbailey: I don't think it's necessary to constrain component/file names to match the routing path since the way we approach each of those is different. These file names just read wrong to my brain.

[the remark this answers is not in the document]
metalice: I understand what you are saying, and u have a point, but currently, I can see this is how we are using it in many places. It's in our guidelines (need to find the link) that is written it should be like that, VM list, vmi list, and templates are using this convention, so I'm guessing it is okay here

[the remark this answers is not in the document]
pcbailey: I believe you, but don't remember that discussion. Let me know if you find the link.

PR #1589, opened 2023-10-26
Adding platform arg to docker and pod man

metalice:
## 📝 Description

Recent changes in ocp image create in quay, create only linux image, adding this platform arg will make those scripts workable also on macos
## 🎥 Demo

### DIFF
--- a/start-console-auth-mode.sh
+++ b/start-console-auth-mode.sh
@@ -4,9 +4,9 @@ set -euo pipefail
 PATH+=:/usr/bin
 
 oc process -f scripts/oauth-client.yaml | oc apply -f -
-oc get oauthclient console-oauth-client -o jsonpath='{.secret}' > scripts/console-client-secret
-oc get secrets -n default --field-selector type=kubernetes.io/service-account-token -o json | \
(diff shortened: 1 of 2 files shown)


start-console-auth-mode.sh
```diff
@@ -4,9 +4,9 @@ set -euo pipefail
 PATH+=:/usr/bin
 
 oc process -f scripts/oauth-client.yaml | oc apply -f -
-oc get oauthclient console-oauth-client -o jsonpath='{.secret}' > scripts/console-client-secret
-oc get secrets -n default --field-selector type=kubernetes.io/service-account-token -o json | \
-    jq '.items[0].data."ca.crt"' -r | python3 -m base64 -d > scripts/ca.crt
+oc get oauthclient console-oauth-client -o jsonpath='{.secret}' >scripts/console-client-secret
```
pcbailey, line 7: Is there a reason why you're removing the spaces after the angle brackets?

metalice: its the auto save.. didn't touch it just left prettier fixes

PR #1599, opened 2023-11-06
CNV-34765: Add data-test-ids to VM action icons

pcbailey:
## 📝 Description

This PR adds the `data-test-id` attribute to the VM action icons for testing purposes.

Jira issue: https://issues.redhat.com/browse/CNV-34765

### DIFF
--- a/src/views/virtualmachines/actions/components/VMActionsIconBar/components/ActionIconButton.tsx
+++ b/src/views/virtualmachines/actions/components/VMActionsIconBar/components/ActionIconButton.tsx
@@ -22,6 +22,7 @@ const ActionIconButton: FC<VMActionIconDetails> = ({ action, Icon, iconClassname
         <Tooltip content={action?.label}>
           <Button
             className="vm-actions-icon-bar__button"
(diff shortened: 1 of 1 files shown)


src/views/virtualmachines/actions/components/VMActionsIconBar/components/ActionIconButton.tsx
```diff
@@ -7,7 +7,13 @@ import { VMActionIconDetails } from '@virtualmachines/actions/components/VMActio
```
metalice, line 7: data test id if very internal, what do u think about passing id? also you can pass id to the button and if needed create a data test id also

pcbailey: Guohua asked for data-test-ids, so I think I'll pass id and use that for the data-test-id attribute.

pcbailey: I updated it to use the VM action ids with '-button' as a suffix so I don't have to pass anything at all.

PR #1602, opened 2023-11-09
Make running yarn dev with no errors

hstastna:
## 📝 Description

Fix multiple errors occurred after running `yarn dev` for multiple PF components (however this is not reproducible always):
`<ComponentName> cannot be used as a JSX component`

Fix the errors by adding missing `"@types/react-dom"` package, also adding `"@patternfly/react-icons"` and `"@types/react-dom"` to the `resolutions` in the _package.json_ file, as some of our packages use those packages as their dependencies, and that lead to the mess as yarn does not do de-duplication, and so those packages were installed more times, which lead to typescript being confused. And so also `"react": ["./node_modules/react", "./node_modules/@types/react"]` had to be added to _tsconfig.json_ file, too, to tell it explicitly, which package and types to use.

Also remove the duplication of `CheckupsStorageForm.tsx`.

TODO:
- [x] fix broken unit tests

## 🎥 Screenshots
**Before:** 
Terminal after executing `yarn dev`:
errrrrrr
Listing the dependencies for a specific package:
```
yarn list --pattern @types/react
yarn list -- pattern @patternfly/react-icons
```
yarn list before

**After:**
after
yarn list after


tsconfig.json
```diff
@@ -14,7 +14,8 @@
       "@catalog/*": ["src/views/catalog/*"],
       "@kubevirt-utils": ["src/utils"],
       "@kubevirt-utils/*": ["src/utils/*"],
-      "@virtualmachines/*": ["src/views/virtualmachines/*"]
+      "@virtualmachines/*": ["src/views/virtualmachines/*"],
+      "react": ["./node_modules/@types/react"]
```
upalatucci, line 18: this should be imported automatically

hstastna: Unfortunately, it doesn't work well without this change.

upalatucci:
No errors with this code on my PC @hstastna

src/views/checkups/Storage/CheckupsStorageForm.tsx
upalatucci, line 1: Why do you remove this file ?

hstastna: It is extra, probably someone made a copy of the file unnecessary or something forgotten when renaming files. There is the same file, same component, just in the `storage` dir (small "s") - which is the correct one (PTAL at _package.json_).

hstastna:

PR #1616, opened 2023-11-22
CNV-33044: New VM Cloning API

upalatucci:
## 📝 Description

Using the new cloning API

Removing namespace and description fields as they are not supported by the new API

I've simulated an error to show what would look like in that case.
The alert message will show the actual reason instead of `unknown` which is a backup in case there is no reason from the backend. 

Make sure the user does not use an already created VM name. Using an already created VM name will result in a restoring of that VM instead of a cloning. 

## 🎥 Demo

Screenshot from 2023-11-24 10-52-54

Screenshot from 2023-11-24 10-52-57

Screenshot from 2023-11-24 10-53-07

Screenshot from 2023-11-24 11-23-55

**Name already used** 
image


src/views/virtualmachines/actions/VirtualMachineActionFactory.tsx
```diff
@@ -1,5 +1,6 @@
 import * as React from 'react';
```
metalice, line 1: please remove *

src/views/virtualmachines/actions/components/CloneVMModal/components/ConfigurationSummary.tsx
```diff
@@ -1,13 +1,13 @@
 import * as React from 'react';
```
metalice, line 1: please remove *

src/views/catalog/wizard/tabs/scheduling/components/Descheduler.tsx
```diff
@@ -1,7 +1,7 @@
 import * as React from 'react';
```
metalice, line 1: please remove

src/views/virtualmachines/actions/components/CloneVMModal/CloneVMModal.tsx
```diff
@@ -41,90 +26,67 @@ type CloneVMModalProps = {
...
-  const isVMRunning = isRunning(vm);
+  const [vmsInSameNamespace] = useK8sWatchResource<V1VirtualMachine[]>(
+    vm && {
+      groupVersionKind: modelToGroupVersionKind(VirtualMachineModel),
+      isList: true,
+      namespace: vm.metadata.namespace,
+    },
+  );
+
+  const nameAlreadyUsed = Boolean(vmsInSameNamespace?.find((v) => v.metadata.name === cloneName));
```
metalice, line 43: 2 things.
1. why bring them all then filter? try to fetch with the name and namespace if result is 200 its existing. 
2. u really think u need it? what are the chances that machine will be existing? and if existing do u display the error from BE? u will get 409 conflict that you can present.

src/views/virtualmachines/actions/components/CloneVMModal/components/CloningStatus.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+  const [freshVMCloneRequest] = useK8sWatchResource<V1alpha1VirtualMachineClone>(
+    vmCloneRequest && {
+      groupVersionKind: modelToGroupVersionKind(VirtualMachineCloneModel),
+      name: vmCloneRequest.metadata.name,
+      namespace: vmCloneRequest.metadata.namespace,
+    },
+  );
+
+  const cloneRequest = freshVMCloneRequest || vmCloneRequest;
```
metalice, line 35: why? why fresh required if u are getting the promise of vmclone?

src/views/virtualmachines/actions/components/CloneVMModal/components/CloningStatus.tsx
```diff
@@ -0,0 +1,100 @@
...
+} from '@patternfly/react-core';
+
+import { CLONING_STATUSES, STATUS_TO_PROGRESS_VARIANT } from '../utils/constants';
+
+type CloningStatusProps = {
+  onComplete: (status: CLONING_STATUSES) => void;
+  vmCloneRequest: V1alpha1VirtualMachineClone;
+};
+
+const CloningStatus: FC<CloningStatusProps> = ({ onComplete, vmCloneRequest }) => {
```
metalice, line 24: why onclomple is as prop if no other component is using it and can be declared here?

src/views/virtualmachines/actions/components/CloneVMModal/CloneVMModal.tsx
upalatucci: 1. Nice suggestion we can do that even if this means that there will be a loading after typing.
2 There is no error from BE if you use the name of an existing VM. If it exists, the BE will restore the existing VM using the selected VM state and this is something we should absolutely avoid

src/views/virtualmachines/actions/components/CloneVMModal/components/CloningStatus.tsx
upalatucci: `vmCloneRequest` is the clone request that we get from the k8sCreate. The request changes its status over time. So with `freshVMCloneRequest`  we take track of this state. If the request fails, the user will see the failure. If the request is completed, the user will see complete at some point. 
At the beginning `freshVMCloneRequest` is undefined so for this reason I made an `|| ` with the initial clone request.

src/views/virtualmachines/actions/components/CloneVMModal/components/CloningStatus.tsx
upalatucci: `CloneVMModal` use this prop to redirect the user to the right page on succeeded and change the modal state.

PR #1629, opened 2023-12-05
CNV-31107: Enable using SSH-key secret from another Project

hstastna:
## 📝 Description

**Fixes:**
https://issues.redhat.com/browse/CNV-31107

_Design doc:_
https://docs.google.com/document/d/1OVNEuq_ULO-87hhqtyvl6m2aZURXRsgj6MlmaFGi1hs/edit?pli=1

Allow users that are managers of multiple projects to copy their SSH-key secret from one project to another so they can use it when creating new VMs instead of re-entering the SSH key. The original SSH key management mechanism allowed the user to use an SSH key in the current project or to create a new one.

## 🎥 Demo
**Before:**

https://github.com/kubevirt-ui/kubevirt-plugin/assets/13417815/4147abee-445d-4410-b6f8-7e7ee0c325db

**After:**

https://github.com/kubevirt-ui/kubevirt-plugin/assets/13417815/9aa45a16-2715-4747-8950-0d56c78ec83e


hstastna:

This is a work of me and Matan. Could someone, please, take control over this PR and to finish it? I was told I should not work anymore (should have been on PTO). 

Please, make sure it is working as expected also in the user settings. It is more complicated part. Check how it was working there before this PR. When editing some key in the user settings, make sure the correct project and the secret name are already selected in the modal (after clicking on the pencil icon to edit it). The correct project is the one you can already see in the user settings, not any other one, nor the actual selected namespace (but sometimes it can be the same, if you edit the the key with the same namespace as the actual selected one).

However in other places it should work a bit different: the actual namespace should be selected in the project drop down, no matter the key was already set for the specific VM or not.

upalatucci (approved):

PR #1630, opened 2023-12-05
CNV-36065: Disk modal with persistent reservation

upalatucci:
## 📝 Description

Add the 'Share disk between VMs' and 'Persistent Reservation' flags under 'Advanced settings' in the disk modal 

design: https://docs.google.com/document/d/1_fbw2BUZFRmArjhdKc32lTEqU9MRHsj2KYBxbR6ZL_0

If the disk is LUN, enable the persistent reservation flag but disable the 'Share disk' flag .

Share disk for LUN does not make sense as LUN already is a shared disk

Make sure the  `FEATURE_HCO_PERSISTENT_RESERVATION` cluster flag is enabled, otherwise is not possible to create a disk like that.  

## 🎥 Demo

Screenshot from 2023-12-05 14-45-18
Screenshot from 2023-12-05 14-45-04


src/utils/components/DiskModal/DiskFormFields/AdvancedSettings.tsx
```diff
@@ -0,0 +1,67 @@
...
+};
+
+const AdvancedSettings: FC<AdvancedSettingsProps> = ({ diskState, dispatchDiskState }) => {
+  const { t } = useKubevirtTranslation();
+
+  const { featureEnabled } = useFeatures(FEATURE_HCO_PERSISTENT_RESERVATION);
+
+  const isLunType = diskState.diskType === diskTypes.lun;
+  return (
+    <ExpandableSection toggleText={t('Advanced Settings')}>
```
hstastna, line 25: I think we prefer sentence case, so maybe 'Advanced settings' would be more aligned with the rest of the UI.

src/views/clusteroverview/SettingsTab/ClusterTab/components/PersistentReservationSection/PersistentReservationSection.tsx
```diff
@@ -0,0 +1,97 @@
...
+type PersistentReservationSectionProps = {
+  hyperConvergeConfiguration: [hyperConvergeConfig: HyperConverged, loaded: boolean, error: Error];
+};
+
+const PersistentReservationSection: FC<PersistentReservationSectionProps> = ({
+  hyperConvergeConfiguration,
+}) => {
+  const { t } = useKubevirtTranslation();
+  const [hyperConverge, hyperLoaded] = hyperConvergeConfiguration;
+  const persistentReservation = hyperConverge?.spec?.featureGates?.persistentReservation ?? false;
```
hstastna, line 30: Hmm, do we really need the nullish coalescing operator here? Maybe just 
`!!hyperConverge?.spec?.featureGates?.persistentReservation` would do the same?  Or `||` one instead `??` ?

src/views/clusteroverview/SettingsTab/ClusterTab/components/PersistentReservationSection/PersistentReservationSection.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+      await toggleFeature(checked);
+      setIsLoading(false);
+    } catch (err) {
+      setError(err.message);
+    }
+  };
+
+  return (
+    <ExpandSection toggleText={t('SCSI Persistent Reservation')}>
```
hstastna, line 59: Again, it seems to me using the "sentence case" to be more appropriate. Maybe the design doc was not quite precise? I'd better make sure about the texts. Also below, I can see "persistent reservation" to be used, which seems to be in the correct form.

src/views/clusteroverview/SettingsTab/ClusterTab/components/PersistentReservationSection/PersistentReservationSection.tsx
```diff
@@ -0,0 +1,97 @@
...
+      <Split>
+        <SplitItem isFilled>
+          {t('Enable persistent reservation')}{' '}
+          <HelpTextIcon
+            bodyContent={t(
+              'The SCSI reservation for disk makes the disk attached to the VirtualMachine as a SCSI LUN. This option should only be used for cluster-aware applications',
+            )}
+            className="KernelSamepageMerging__HelpTextIcon"
```
hstastna, line 67: I can see this `className` name from kinda a different part of the UI used here. Maybe it should be renamed to something a bit more general and then to be used in both places? Same for `helpIconClassName` below.

src/views/clusteroverview/SettingsTab/ClusterTab/components/PersistentReservationSection/PersistentReservationSection.tsx
upalatucci: Okay i used `Boolean()`

metalice:
the pr lgtm but , i see u are adding the new init values to state and adding actions, but maybe I'm missing something, but I don't see any addition to create disk function to use those new fields? again, maybe logic of current create disk function is ok and can handle it, but I just want to make sure :)

upalatucci:
Now i pushed the logic to actually add the options to the disk

src/utils/components/DiskModal/utils/helpers.ts
```diff
@@ -92,12 +92,24 @@ export const requiresDataVolume = (diskSource: string): boolean => {
...
+    },
+    name: diskState.diskName,
+  };
+
+  if (diskState.lunReservation && disk.lun) {
+    disk.lun.reservation = true;
+  }
+
+  if (diskState.sharable) {
+    disk.shareable = true;
```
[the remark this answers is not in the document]
upalatucci: okay fixed

PR #1642, opened 2023-12-13
CNV-31107: Enable using SSH-key secret from another Project

metalice:
Allow users that are managers of multiple projects to copy their SSH-key secret from one project to another so they can use it when creating new VMs instead of re-entering the SSH key. The original SSH key management mechanism allowed the user to use an SSH key in the current project or to create a new one.

Fixes https://issues.redhat.com/browse/CNV-31107

## 📝 Description

Adding the option to add ssh key form another project.
@hstastna 
## 🎥 Demo

ssh-another-project


src/utils/components/SSHSecretSection/components/SSHOptionUseExisting/SSHOptionUseExisting.tsx
```diff
@@ -0,0 +1,168 @@
...
+              hasInlineFilter
+              inlineFilterPlaceholderText={t('Search project')}
+              isOpen={isOpen}
+              maxHeight={400}
+              onFilter={onFilterProject}
+              onSelect={onSelectProject}
+              onToggle={setIsOpen}
+              selections={selectedProject}
+              variant={SelectVariant.single}
+            >
```
upalatucci, line 123: use 
```
menuAppendTo="parent"
````
to fix selects in modals that get hided

metalice: cool, just checked it does make a different, will push

PR #1656, opened 2023-12-18
CNV-36090: fix permissions

metalice:
## 📝 Description

Fix what are we checking for no prev user permissions
## 🎥 Demo


src/utils/hooks/usePermissions/usePermissions.ts
```diff
@@ -19,57 +14,43 @@ export type UsePermissions = () => {
 
 const usePermissions: UsePermissions = () => {
   const { t } = useKubevirtTranslation();
-  const cdiInstalled = useFlag(FLAG_KUBEVIRT_CDI);
+  const [activeNamespace] = useActiveNamespace();
```
upalatucci, line 17: `activeNamespace` can be `ALL_NAMESPACES_SESSION_KEY` in case of `all-namespaces`. In that case, what should we do?

PR #1658, opened 2023-12-18
Refactor pending changes with live mgiration

upalatucci:
## 📝 Description

Fixing 2 bugs with that:

1. vms created through IT do not show pending changes
2. adding hotplug network triggers also the pending changes on the 'default' network. If vms do not have networks, vmi will have one default network. But no pending change should be triggered in this case.  

## 🎥 Demo

**Before**

No pending changes

**After**

Screenshot from 2023-12-18 18-54-32
Screenshot from 2023-12-18 18-36-14

### DIFF
--- a/src/views/virtualmachines/details/VirtualMachinePendingChangesAlert/VirtualMachinePendingChangesAlert.tsx
+++ b/src/views/virtualmachines/details/VirtualMachinePendingChangesAlert/VirtualMachinePendingChangesAlert.tsx
@@ -1,17 +1,15 @@
 import React, { FC } from 'react';
-import { useHistory } from 'react-router-dom';
 
 import { V1VirtualMachine, V1VirtualMachineInstance } from '@kubevirt-ui/kubevirt-api/kubevirt';
 import { usePendingChanges } from '@kubevirt-utils/components/PendingChanges/hooks/usePendingChanges';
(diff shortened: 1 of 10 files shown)


src/views/virtualmachines/details/VirtualMachinePendingChangesAlert/VirtualMachinePendingChangesAlert.tsx
```diff
@@ -22,40 +18,32 @@ const VirtualMachinePendingChangesAlert: FC<VirtualMachinePendingChangesAlertPro
...
 
-  const sortedNICHotPlugPendingChanges = getSortedNICPendingChanges(vm, vmi, history);
-  const { showLiveMigrateSection, showRestartSection } = showPendingChangesSections(
-    pendingChanges,
-    sortedNICHotPlugPendingChanges,
+  const liveMigrationChanges = pendingChanges.filter(
+    (pendingChange) => pendingChange.appliedOnLiveMigration && pendingChange.hasPendingChange,
+  );
+  const restartChanges = pendingChanges.filter(
+    (pendingChange) => !pendingChange.appliedOnLiveMigration && pendingChange.hasPendingChange,
```
metalice, line 33: u can use 1 reduce for those both

PR #1661, opened 2023-12-20
CNV-36460: Fix console router upgrade causing broken pages

metalice:
## 📝 Description

Console upgraded the router pack and changed some components as HorizontalNav, which acts differently now. 
Some broken pages I managed to fix only by getting history or location from the router package hook, and HorizontalNav were replaced in a newly created version and not consuming from SDK anymore.

## 🎥 Demo

Please look at https://issues.redhat.com/browse/CNV-36460?filter=12417679 for the broken images, the after is just as it was before.


src/utils/components/HorizontalNavbar/HorizontalNavbar.tsx
```diff
@@ -0,0 +1,76 @@
...
+                id={`horizontal-pageHeader-${item.name}`}
+                isActive={activeItem === item.name.toLowerCase()}
+                itemId={item.name.toLowerCase()}
+                key={item.name}
+                preventDefault
+                to={item.href}
+              >
+                {item.name}
+              </NavItem>
+            ) : null;
```
upalatucci, line 55:
```suggestion
            if (item?.isHidden) return null
            
            return (
              <NavItem
                className="horizontal-nav-bar__menu-item"
                id={`horizontal-pageHeader-${item.name}`}
                isActive={activeItem === item.name.toLowerCase()}
                itemId={item.name.toLowerCase()}
                key={item.name}
                preventDefault
                to={item.href}
              >
                {item.name}
              </NavItem>
            );
```

upalatucci (approved):

PR #1663, opened 2023-12-20
CNV-32746: make the pending changes alert expandable in the vm detail…

upalatucci:
## 📝 Description

Make it expandable to occupy less vertical space

## 🎥 Demo

Screenshot from 2023-12-20 16-13-37
Screenshot from 2023-12-20 16-13-34


metalice:
the default will be close? i think default should be expanded and just adding the open to collapse

PR #1664, opened 2023-12-21
CNV-32943: show instancetype in the details page

upalatucci:
## 📝 Description

remove Template info in the overview page
add preference and instancetype links in the overview page

Remove Workload in details page configuration tab

We have two files called headlessMode identical. 
It was raising  errors in the details page. Remove that also 

## 🎥 Demo
Screenshot from 2023-12-21 10-32-10
Screenshot from 2023-12-21 10-27-31

In the search box in case of no workload profile (because the VM is created through instance type)

Screenshot from 2023-12-21 11-12-14


src/views/virtualmachines/details/tabs/configuration/search/VirtualMachineConfigurationTabSearch.tsx
```diff
@@ -33,12 +42,13 @@ const VirtualMachineConfigurationTabSearch = () => {
...
           const match = newValue.toLowerCase();
-          if (title.startsWith(match)) {
+          const isEnabled = !item.element.isDisabled;
+          if (title.startsWith(match) && isEnabled) {
             acc.startWith.push(item);
             return acc;
           }
 
-          if (title.includes(match)) acc.includes.push(item);
+          if (title.includes(match) && isEnabled) acc.includes.push(item);
```
metalice, line 51: u can do if (enabled)  both both if inside instead of twice &&

PR #1676, opened 2023-12-25
CNV-35730: Fix windows sysprep on vm init run tab

metalice:
## 📝 Description

It seems like an old bug before the day that you could create a VM that won't start at the beginning.
The sysprep on the init run tab was missing the modal when clicking edit, and edit was hard coded disabled.

## 🎥 Demo

### DIFF
--- a/src/views/templates/details/tabs/scripts/components/SysPrepItem/sysprep-utils.ts
+++ b/src/views/templates/details/tabs/scripts/components/SysPrepItem/sysprep-utils.ts
@@ -109,4 +109,4 @@ export const updateSysprepObject = (
   }
 };
 
-export const getSyspredConfigMapName = (volume: V1Volume) => volume?.sysprep?.configMap?.name;
+export const getSysprepConfigMapName = (volume: V1Volume) => volume?.sysprep?.configMap?.name;
--- a/src/views/virtualmachines/details/tabs/configuration/initialrun/components/InitialRunTabSysprep.tsx
+++ b/src/views/virtualmachines/details/tabs/configuration/initialrun/components/InitialRunTabSysprep.tsx
@@ -1,27 +1,33 @@
 import React, { FC } from 'react';
-import { getSyspredConfigMapName } from 'src/views/templates/details/tabs/scripts/components/SysPrepItem/sysprep-utils';
+import { getSysprepConfigMapName } from 'src/views/templates/details/tabs/scripts/components/SysPrepItem/sysprep-utils';
 
 import { ConfigMapModel, modelToGroupVersionKind } from '@kubevirt-ui/kubevirt-api/console';
 import { IoK8sApiCoreV1ConfigMap } from '@kubevirt-ui/kubevirt-api/kubernetes';
 import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
 import WindowsLabel from '@kubevirt-utils/components/Labels/WindowsLabel';
+import { useModal } from '@kubevirt-utils/components/ModalProvider/ModalProvider';
 import SearchItem from '@kubevirt-utils/components/SearchItem/SearchItem';
 import { AUTOUNATTEND, UNATTEND } from '@kubevirt-utils/components/SysprepModal/sysprep-utils';
 import { SysprepDescription } from '@kubevirt-utils/components/SysprepModal/SysprepDescription';
+import { SysprepModal } from '@kubevirt-utils/components/SysprepModal/SysprepModal';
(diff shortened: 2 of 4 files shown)


src/views/templates/details/tabs/scripts/components/SysPrepItem/sysprep-utils.ts
```diff
@@ -109,4 +109,4 @@ export const updateSysprepObject = (
   }
 };
 
-export const getSyspredConfigMapName = (volume: V1Volume) => volume?.sysprep?.configMap?.name;
+export const getSysprepConfigMapName = (volume: V1Volume) => volume?.sysprep?.configMap?.name;
```
hstastna, line 112: 👍🏽 awesome! nice catch!

src/views/virtualmachines/details/tabs/configuration/initialrun/components/InitialRunTabSysprep.tsx
```diff
@@ -46,9 +57,22 @@ const InitialRunTabSysprep: FC<InitialRunTabSysprepProps> = ({ canUpdateVM, vm }
...
+            onSysprepSelected={onSysprepSelected}
+            sysprepSelected={sysprepSelected}
+            unattend={unattend}
+          />
+        ))
+      }
       data-test-id="sysprep-button"
       descriptionHeader={<SearchItem id="sysprep">{t('Sysprep')}</SearchItem>}
-      isDisabled
+      isDisabled={!canUpdateVM}
```
hstastna, line 75: 👍🏽

src/views/virtualmachines/details/tabs/configuration/initialrun/utils/utils.ts
```diff
@@ -0,0 +1,97 @@
...
+  const vmDisks = getDisks(vm);
+
+  await k8sPatch<V1VirtualMachine>({
+    data: [
+      {
+        op: 'replace',
+        path: `/spec/template/spec/domain/devices/disks`,
+        value: [
+          ...vmDisks.filter((disk) => disk?.name !== SYSPREP),
+          ...(!isEmpty(name) ? [sysprepDisk()] : []),
```
hstastna, line 31: Hmm, wouldn't this line 31 work simply without spread operator and not putting the result of the condition into an array? (same for the line 39)

hstastna (commented):
Looks 👍🏽, just left one little question :)

src/views/virtualmachines/details/tabs/configuration/initialrun/utils/utils.ts
hstastna: I mean why to put `sysprepDisk()` into an array and then to use spread operator to "unpack" that array.. seems to me unnecessary. Simply using `!isEmpty(name) ? sysprepDisk() : null,` in that line didn't work as expected? For me it seems to work. Am I missing something? :D

hstastna: `undefined` neither?

metalice: Same. It's a value. Its a primitive type.

hstastna: You are right. But.. does it ever matter? Would presence of null or undefined in that array in that place mean anything from the backend perspective? From my experience, sometimes such values are simply ignored or thrown away when sent to backend, and that feels so nice and makes our life easier (but depending if that value has some specific meaning in that specific place or not, depending on the backend, of course, we should not always rely on it, we should know what we do).

metalice: i dont know didnt test it, but I don't see a reason sending a null or undefined value if I have other options.

hstastna: 🆗

src/views/templates/details/tabs/scripts/components/SysPrepItem/sysprep-utils.ts
lkladnit: @metalice can you tell me what was the change here?

hstastna: Leon, it is just one character change in the naming: "d" to "p",  _Syspred_ to _Sysprep_ substring in the name of this util.

PR #1685, opened 2023-12-28
CNV-36794: Update vm example template

metalice:
## 📝 Description

Updating fedora version from 36 to latest

### DIFF
--- a/src/templates/vm-yaml.ts
+++ b/src/templates/vm-yaml.ts
@@ -9,14 +9,14 @@ metadata:
     description: VM example
   labels:
     app: example
-    os.template.kubevirt.io/fedora36: 'true'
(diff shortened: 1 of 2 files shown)


src/templates/vm-yaml.ts
```diff
@@ -60,7 +60,7 @@ spec:
       volumes:
         - name: rootdisk
           containerDisk:
-            image: 'quay.io/containerdisks/fedora:36'
+            image: 'quay.io/containerdisks/fedora'
```
pcbailey, line 63: Did you mean to add the latest tag?

metalice: yes, changing to latest

PR #1686, opened 2023-12-28
CNV-34072: Removing ipv6 form ip's list

metalice:
## 📝 Description

Removing ipv6 form ips list

## 🎥 Demo

After:
image

Before:
image


gouyang:
@metalice Will this fix impact the dual stack network which will has IPv6 address for the NIC?

PR #1695, opened 2024-01-03
CNV-35271: Prevent changing disk size unit to B after PVC selection

hstastna:
## 📝 Description

**Fixes:**
https://issues.redhat.com/browse/CNV-35271

Prevent changing disk size unit to "B" after selecting a PVC when using an existing volume in _Add volume_ modal (_InstanceTypes_ tab in _Catalog_).

## 🎥 Screenshots
**Before:**
Disk size unit suddenly changed to "B" when selecting PVC from the _PVC name_ dropdown:
volume_before

**After:**
Correct disk size:
volume_after

### DIFF
--- a/src/views/catalog/utils/quantity.ts
+++ b/src/views/catalog/utils/quantity.ts
@@ -1,6 +1,6 @@
 import byteSize, { ByteSizeResult } from 'byte-size';
 
-import { binaryUnits, customUnits, multipliers, toIECUnit } from '@kubevirt-utils/utils/units';
+import { customUnits, multipliers, toIECUnit } from '@kubevirt-utils/utils/units';
 
 export const bytesToIECBytes = (
   bytes: number,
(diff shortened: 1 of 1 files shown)


hstastna:

src/views/catalog/utils/quantity.ts
```diff
@@ -15,8 +15,8 @@ export const bytesToIECBytes = (
 };
 
 export const bytesToDiskSize = (size: string) => {
-  const bytesizeresult = bytesToIECBytes(parseFloat(size), 0, binaryUnits);
-  return [bytesizeresult.value, bytesizeresult.unit].join('');
+  const bytesizeresult = bytesFromQuantity(size, 0);
+  return [bytesizeresult[0], bytesizeresult[1]].join('');
```
metalice, line 19: 1. is an array in size of 2 guaranteed? if not please add ?.[x]

2. can u simply return bytessizeresult? or does it contain more items, and u want only the first 2?

hstastna: 1. it is guaranteed (and we always do want and need both items, not only one)
2. no, we need to join the items, we need to return a string. If I edited `bytesFromQuantity` itself for our needs here, it would break some other places in the UI.

metalice: or i see, so you are using join instead of toString here, ok, it not really matter.

hstastna: `bytesizeresult.toString()` returns `30,GiB` (for example) which is not what we want. We want `30GiB` instead. And from what I know, `toString` does not take any params. So we really don't have any other option than to use `join` here.

PR #1697, opened 2024-01-04
CNV-34692: Prevent VM stuck in starting status after detaching disk

hstastna:
## 📝 Description

**Fixes:**
https://issues.redhat.com/browse/CNV-34692

Prevent VM getting stuck in "Starting" status in case of detaching a disk that is in error (if PVC is not in "Bound" state, for example if you added wrong url when creating that disk) by disabling _Detach_ button for that specific disk in the VM's _Disks_ list.

## 🎥 Screenshots
**Before:**
Possible to detach the problematic disk leading to VM stuck in starting status:
detach_before

**After:**
_Detach_ button disabled if the disk is problematic/if PVC is not in Bound state:
detach_after

### DIFF
--- a/src/views/virtualmachines/details/tabs/configuration/storage/components/tables/disk/DiskRow.tsx
+++ b/src/views/virtualmachines/details/tabs/configuration/storage/components/tables/disk/DiskRow.tsx
@@ -1,10 +1,6 @@
-import * as React from 'react';
+import React, { FC } from 'react';
 
 import { V1VirtualMachine, V1VirtualMachineInstance } from '@kubevirt-ui/kubevirt-api/kubevirt';
-import {
-  CONTAINER_EPHERMAL,
-  OTHER,
-} from '@kubevirt-utils/components/DiskModal/DiskFormFields/utils/constants';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
 import { modelToGroupVersionKind, PersistentVolumeClaimModel } from '@kubevirt-utils/models';
 import { DiskRowDataLayout } from '@kubevirt-utils/resources/vm/utils/disk/constants';
@@ -12,14 +8,14 @@ import { readableSizeUnit } from '@kubevirt-utils/utils/units';
 import { ResourceLink, RowProps, TableData } from '@openshift-console/dynamic-plugin-sdk';
 import { Label, Stack, StackItem } from '@patternfly/react-core';
 
+import { isPVCSource } from './utils/helpers';
 import DiskRowActions from './DiskRowActions';
 import { HotplugLabel } from './HotplugLabel';
 
-const DiskRow: React.FC<
+const DiskRow: FC<
   RowProps<DiskRowDataLayout, { vm: V1VirtualMachine; vmi?: V1VirtualMachineInstance }>
 > = ({ activeColumnIDs, obj, rowData: { vm, vmi } }) => {
   const { t } = useKubevirtTranslation();
-  const isPVCSource = ![CONTAINER_EPHERMAL, OTHER].includes(obj?.source);
 
   return (
     <>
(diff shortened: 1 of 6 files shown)


hstastna:

src/views/virtualmachines/details/tabs/configuration/storage/components/hooks/usePVCBound.ts
```diff
@@ -0,0 +1,18 @@
+import {
```
pcbailey, line 1: I think this is a good candidate for `/src/utils/hooks`.

pcbailey (commented):
One small suggestion. Other than that it looks good!

src/views/virtualmachines/details/tabs/configuration/storage/components/hooks/usePVCBound.ts
hstastna: I am not sure because we don't need this in other places than VM's disks list.

src/views/virtualmachines/details/tabs/configuration/storage/components/hooks/usePVCBound.ts
```diff
@@ -0,0 +1,18 @@
...
+  modelToGroupVersionKind,
+  PersistentVolumeClaimModel,
+} from '@kubevirt-ui/kubevirt-api/console';
+import { IoK8sApiCoreV1PersistentVolumeClaim } from '@kubevirt-ui/kubevirt-api/kubernetes';
+import { useK8sWatchResource } from '@openshift-console/dynamic-plugin-sdk';
+
+const usePVCBound = (pvcName: string, pvcNamespace: string): boolean => {
```
metalice, line 8: there is already a call in const useDisksTableData for all pvcs, just expand it to return also the pvcs and in DiskList add to rowData a function to get the PVC by name and namespace , no need for this hook, its just add a call for each row

hstastna: So you are suggesting to use that call for getting all PVCs and just filter them by name, to get the specific PVC we need. Right? Then I'll need some helper function to check the PVC's status anyway.

metalice: Yes u can create a function(kind of selector, no need for hook) or if u prefer, when u return the pvc's you can modify it to an object , in the matter of {[vmname: string]: pvcs[]} , then just pass this value of mapper to rowData as pvcProp={pvcObject[vmName]} , whatever u prefer, there are several example in our code for it, you can look at virtualmachinelist if you like

src/views/virtualmachines/details/tabs/configuration/storage/components/tables/disk/DiskRow.tsx
```diff
@@ -73,7 +74,13 @@ const DiskRow: React.FC<
         className="dropdown-kebab-pf pf-v5-c-table__action"
         id=""
       >
-        <DiskRowActions diskName={obj?.name} pvcResourceExists={isPVCSource} vm={vm} vmi={vmi} />
+        <DiskRowActions
+          diskName={obj?.name}
+          isPVCBound={isPVCStatusBound(obj?.sourceStatus)}
+          pvcResourceExists={isPVCSource}
```
metalice, line 80: I think moving obj as a single prop and extracting diskName, isPVCBound and pvcResourceExists inside DiskRowActions is better then sending 3 props. what do u think?

hstastna: That's a good improvement I think. Thanks!

src/views/virtualmachines/details/tabs/configuration/storage/components/hooks/usePVCBound.ts
hstastna: Yes, thanks very much for the suggestion!

PR #1700, opened 2024-01-08
CNV-35635: Fix instance type different bios

metalice:
## 📝 Description

Fix instance type bios, presenting the correct value.
Instance type can have some fields with values that affect bios type when VM is created.
We cannot check all the fields and make assumptions. We need to use the expand-spec of instance type VM in order to compare to vmi and get right values of bios or anything else. This fix will send instanceTypeExpandedSpec to all tabs and can be used instead of VM when needed.

Also, I changed a bit of the CPU memory component while exploring the code; this component made a lot of network calls to instance type to get CPU and memory; now, it's not needed as instanceTypeExpandedType can deliver the same data.
## 🎥 Demo

After:
image

Before:
image

### DIFF
--- a/src/views/virtualmachines/details/tabs/configuration/details/components/DetailsSectionBoot.tsx
+++ b/src/views/virtualmachines/details/tabs/configuration/details/components/DetailsSectionBoot.tsx
@@ -20,11 +20,17 @@ import { updateBootLoader, updatedBootOrder, updateStartStrategy } from '../util
 
 type DetailsSectionBootProps = {
   canUpdateVM: boolean;
+  instanceTypeVM: V1VirtualMachine;
   vm: V1VirtualMachine;
   vmi: V1VirtualMachineInstance;
 };
 
-const DetailsSectionBoot: FC<DetailsSectionBootProps> = ({ canUpdateVM, vm, vmi }) => {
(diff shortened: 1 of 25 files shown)


src/views/virtualmachines/details/tabs/configuration/details/components/DetailsSectionBoot.tsx
```diff
@@ -46,7 +52,7 @@ const DetailsSectionBoot: FC<DetailsSectionBootProps> = ({ canUpdateVM, vm, vmi
       <VirtualMachineDescriptionItem
         descriptionData={
           <div className={classNames({ 'text-muted': !canUpdateVM })}>
-            {getBootloaderTitleFromVM(vm)}
+            {getBootloaderTitleFromVM(instanceTypeVM)}
```
upalatucci, line 55: Shouldn't we have something like `instanceTypeVM || vm` ? Is instanceTypeVM populated even when the VM do not have instanceType?

metalice: u right i forgot

upalatucci (approved):

PR #1715, opened 2024-01-11
CNV-37075: Fix "Quick create VirtualMachine" is disabled 

upalatucci:
## 📝 Description

`hasVMValidDVSources` returns `undefined` when the VM does not have `dataVolumeTemplates` array. Instead, should return true if there should be no dataVolumeTemplates


metalice:
u added irrelevant commit

PR #1730, opened 2024-01-18
CNV-36780: Validate SSH secret name in Instancetypes SSH modal

pcbailey:
## 📝 Description

This PR fixes a bug wherein no validation error is displayed in the Instancetypes SSH modal when the Use existing option is selected and the name is edited to match an existing secret name in the targeted namespace.

It also:

- Changes the styling of the "New secret name" field to match the one on the "Add new" screen
- Disables the save button when modal is loading
- Corrects the capitalization of 'SSH' in a user-facing message

Jira: https://issues.redhat.com/browse/CNV-36780

#### Before
ssh-modal-loading-BEFORE-2024-01-18_10-40

use-existing-secret-BEFORE-2024-01-18_10-35

#### After
ssh-modal-loading-AFTER-2024-01-18_10-39

use-existing-secret-AFTER-2024-01-18_10-37


src/utils/components/SSHSecretSection/components/SSHKeyUpload/SSHKeyUpload.tsx
```diff
@@ -26,6 +27,7 @@ type SSHKeyUploadProps = {
 
 const SSHKeyUpload: FC<SSHKeyUploadProps> = ({ secrets, setSSHDetails, sshDetails }) => {
   const { t } = useKubevirtTranslation();
+  const [activeNamespace] = useActiveNamespace();
```
upalatucci, line 30: is there a chance that this component is used in a `all-namespaces` page?
In that case, `activeNamespace` will have the value of `ALL_NAMESPACES_SESSION_KEY` and not `undefined`

hstastna (commented):
The code looks good to me. Also tested, works good, just I am not sure what the expected behavior is when being in "All Projects". We should make sure about this.
 
Another thing: when choosing some from the _Public SSH key_ drop down, it may happen that the final _New secret name_ becomes too long (sometimes, depending on what is generated), especially if the original name was kinda long. I think that is kinda annoying. From the UX point of view, I think that if we generate a new secret name, we should automatically take care of this. We should not generate anything which is not valid, so the user would not have to edit the name unnecessary. I'd suggest to check automatically if the generated name is not too long. And if it is, then to try to generate another name or to cut what is extra.

Also when it happens, the input becomes a bit shorter, which could be nicer. But that is a minor issue. The more important is how the name becomes so long as already mentioned.

secret_ long

src/utils/components/SSHSecretSection/components/SSHKeyUpload/SSHKeyUpload.tsx
pcbailey: Yes, there is. I fixed it to account for that. Thanks!

pcbailey:
@hstastna 
> The code looks good to me. Also tested, works good, just I am not sure what the expected behavior is when being in "All Projects". We should make sure about this.

If "All projects" is selected the VM will be created in the default namespace.

> Another thing: when choosing some from the _Public SSH key_ drop down, it may happen that the final _New secret name_ becomes too long (sometimes, depending on what is generated), especially if the original name was kinda long. I think that is kinda annoying. From the UX point of view, I think that if we generate a new secret name, we should automatically take care of this. We should not generate anything which is not valid, so the user would not have to edit the name unnecessary. I'd suggest to check automatically if the generated name is not too long. And if it is, then to try to generate another name or to cut what is extra.

I addressed this. Take a look and tell me what you think.

> Also when it happens, the input becomes a bit shorter on the right side, which could be nicer. But that is a minor issue. The more important is how the name becomes so long as already mentioned.

Strangely, PF created it like that. You're right that it doesn't look nice. I overrode it to make it stay flush with the right side.

src/utils/components/SSHSecretSection/utils/utils.ts
```diff
@@ -169,3 +193,49 @@ export const getPropagationMethod = (
...
+
+const generateShortPrettyName = (prefix?: string): string => {
+  const numberDictionary = NumberDictionary.generate({ length: 2 });
+  const prefixValue = prefix ? `${prefix}-` : '';
+
+  return `${prefixValue}${uniqueNamesGenerator({
+    dictionaries: [animals, numberDictionary],
+    separator: '-',
+  })}`;
+};
```
metalice, line 205: why duplicate it with a minor change?

src/utils/components/SSHSecretSection/utils/utils.ts
```diff
@@ -169,3 +193,49 @@ export const getPropagationMethod = (
...
+      generatedName = generatePrettyName(name);
+    else if (nameLength <= SAFE_NAME_LENGTH_FOR_SHORT_GENERATED_SUFFIX)
+      generatedName = generateShortPrettyName(name);
+    else generatedName = generateMinimalPrettyName(name);
+
+    isValid = validateSecretName(generatedName, vmNamespaceTarget, secrets);
+  } while (!isValid);
+
+  return generatedName;
+};
```
metalice, line 241: Why do all of this, instead of checking if it's above 70 chars, if yes, generate a new name?

metalice:
@pcbailey please be aware this modal isnt only for instance type.
it used everywhere, wizards, VM details, cluster setting. please check all angels, for new and editing in the all appearance across our app, we don't want 1 bug fix will cause others.

src/utils/components/SSHSecretSection/utils/utils.ts
hstastna: I think the max is 51 chars. Isn't it?

metalice: 51 or 70 or 100 isnt really matter, should check the limit :)

hstastna (commented):
The look of the SSH name field looks nice now! Thanks, Phillip! 😊 

I've found kinda weird scenario in the Catalog InstanceTypes page/tab: after I choose some existing SSH from a different namespace and save it, and then I try to change/edit it, the modal reopens with the "Add new" option as the default chosen option, which feels confusing, while for example in the Overview settings page, the modal opens in "Use existing" when trying to edit the key.

src/utils/components/SSHSecretSection/utils/utils.ts
pcbailey: @metalice @hstastna I was trying to keep as much of the original secret name as possible while still adding a suffix to making it unique, but I'm good with simplifying it. 

The name limit is 51 characters, which includes any suffix we add. I generated 5,000 suffixes multiple times. A few times I got a single 9 character suffix, but I always got a few 10 character suffixes, so I think we should use 10 characters as the minimum suffix length.

With that we know that every existing secret name greater than 41 characters will be too long to get a suffix and remain under the character limit. So, are both of you good with the follow algorithm?

if (secretName.length > 41) generate a totally random name
else generate and attach a suffix

pcbailey:
@hstastna That's the behavior we have currently. I agree it seems weird.
@metalice Do you want me to change this behavior in this PR?

pcbailey:
@metalice I checked the behavior everywhere it's used and found no issues.

src/utils/components/SSHSecretSection/utils/utils.ts
pcbailey: This was part of trying to keep as much of the original name as possible.

metalice:
> @hstastna That's the behavior we have currently. I agree it seems weird. @metalice Do you want me to change this behavior in this PR?

better in a new pr thanks

src/utils/components/SSHSecretSection/utils/utils.ts
metalice: yes i think its much better to generate a new name if > 41

hstastna: I am not sure, for me this all seems to be too complicated unnecessary. I'd just generate a suffix as originally and if the final name longer than expected, I'd just cut from the suffix what is extra to achieve being the final name max 51 chars. Or.. or I'd just simply generate a totally random name always, no matter the original secret name. WDYT @pcbailey @metalice ?

pcbailey: I just pushed the updated logic, but I'm fine with either of those if that's what you guys want. 

FYI, the logic I just pushed will ensure that names that can get a suffix will get one that fits and creates a valid name while those too long for a suffix get a newly generated name.

src/utils/components/SSHSecretSection/utils/utils.ts
```diff
@@ -169,3 +188,27 @@ export const getPropagationMethod = (
...
+  if (secretName.length >= MIN_NAME_LENGTH_FOR_GENERATED_SUFFIX) return generatePrettyName();
+
+  const maxSuffixLength = MAX_NAME_LENGTH - secretName.length;
+  let generatedName = [secretName, generateValidLengthSuffix(maxSuffixLength)].join('-');
+
+  while (!validateSecretName(generatedName, vmNamespaceTarget, secrets))
+    generatedName = [secretName, generateValidLengthSuffix(maxSuffixLength)].join('-');
+
+  return generatedName;
+};
```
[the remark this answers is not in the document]
pcbailey: From the tests I ran suffixes have a minimum length of 9 (rare) and a maximum length of 26 (also rare). The most common lengths are 16 and 17. If we only checked if it's greater than 41 we would absolutely have instances where the created name + suffix combination would be too long.

This logic says:
1. If the name is too long for a suffix generate and return a generic name
2. Generate a suffix that will create a valid length name
3. Check the validity of the name in length and uniqueness
4. Is the name valid ? Continue : Run steps 2 and 3 again until a valid name is generated
5. Return the valid name

Chances of having to repeat steps 2 and 3 are very unlikely, but that check makes sure that we always get a valid length name that's unique within the namespace, which removes the possibility of the user seeing our generated name getting flagged as invalid.

I think this logic provides the cleanest user experience as it maintains the original secret name with a suffix format that matches the others in the UI if possible. It just requires a little more work behind the scenes. That being said, if you want something different I'm open to it.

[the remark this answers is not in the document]
metalice: u sure secret name is under 41 ? i think its 70 

so when checking 41 in worst case is 41 + 26 and safe.

this code is under to understand and will be a problem continue forward to maintain it. I prefer a simple solution. if we can get any simple just generate new name each time, its ok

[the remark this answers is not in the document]
pcbailey: @metalice 51 characters is used in the length validation function, so that's what I've been using; however, I just tested it and the actual maximum length is 253 characters. I'm not sure where the 51 character limit came from. I seriously doubt anyone will hit that, but I'll switch it to check for length greater than 227 (253 - 26).

So I know where to improve, what makes the existing code hard to understand? I used constants instead of magic numbers, used descriptive variable names, and extracted logic into a function with a descriptive name where it was reasonable. What else could I have done to make it easier to understand in your opinion?

pcbailey:
@hstastna @metalice PTAL

src/utils/components/SSHSecretSection/hooks/useSecretsData.ts
```diff
@@ -4,45 +4,72 @@ import { modelToGroupVersionKind, SecretModel } from '@kubevirt-ui/kubevirt-api/
...
-const useSecretsData: UseSecretsData = (sshDetails, namespace) => {
-  const [secrets, ...loadedAndErrorData] = useK8sWatchResource<IoK8sApiCoreV1Secret[]>({
-    groupVersionKind: modelToGroupVersionKind(SecretModel),
-    isList: true,
-    ...(namespace !== ALL_NAMESPACES_SESSION_KEY && {
-      namespace,
-    }),
-    limit: 10000,
-  });
+const useSecretsData: UseSecretsData = (sshDetails, secretsNamespace, currentNamespace) => {
```
metalice, line 21: please check if secretsNamespace and currentNamespace arent the same

currentNamespace isn't activeNamespace? 

what happen if they are the same? u just fetching same thing twice

pcbailey: `secretsNamespace` is the namespace selected in the "Use existing" section of the SSH Secret modal. `currentNamesace` is the namespace the VM is being created in or the namespace selected in the "Manage SSH keys" section on the settings page. 

They start off as the same and remain the same unless a different namespace is selected in the "Use existing" section.

I updated it to only pull the `secretsNamespace` secrets if `secretsNamespace` and `currentNamespace` are different.

PR #1731, opened 2024-01-19
CNV-35287: Prevent VM Migration charts to show the sum of values

hstastna:
## 📝 Description

**Fixes:**
https://issues.redhat.com/browse/CNV-35287

Use `rate` instead of `sum_over_time` for the queries:
- `MIGRATION_DATA_PROCESSED`
- `MIGRATION_DATA_REMAINING`
- `MIGRATION_MEMORY_DIRTY_RATE` for the migration graphs, so the curves won't only "go up", which wasn't correct.

_Note:_
This is just a part of the whole jira bug fix, as dealing with VM migrations that takes just a very short time is WIP. It can possibly be a matter of how the metric was implemented.

## 🎥 Demo
coming very soon


hstastna:

PR #1738, opened 2024-01-23
Fix search filters appearing in the dropdown

upalatucci:
## 📝 Description

When `filter.items` is empty for some filters like 'Node' as maybe all vms are stopped, the filters appears in the dropdown.

## 🎥 Demo

**Before**
Screenshot from 2024-01-23 09-09-08

**After**
Screenshot from 2024-01-23 09-14-11


avivtur:
@upalatucci Hi Ugo :) 
Quick question, by the after screenshot it also seems there's no ip address but it still shows in the dropdown, why are we still showing the IP address option?
I also thought in general we don't make option disappear but we disable it, is that possible to disable instead of removing?

upalatucci:
hei @avivtur here in the dropdown we should have only searchable filters. We added recently IP Address so that the user can type the IP and search vms by subnetwork and singular ips. Node is a filter that should be in the 'Filters' section and not in the dropdown

avivtur (approved):

PR #1748, opened 2024-01-25
CNV-37457: Add user provided tab to instance type

metalice:
## 📝 Description

> Add a brief description
Adding user user-provided tab to the instance type 

## 🎥 Demo

After:
image

Before:
image


pcbailey:
It's unfortunate that the acronym for user-provided ITs is VMI since is may cause some confusion for users.

src/views/instancetypes/list/UserInstancetypeList.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+import UserInstancetypeRow from './components/UserInstancetypeRow';
+import useUserInstancetypeListColumns from './hooks/useUserInstancetypeListColumns';
+
+import '@kubevirt-utils/styles/list-managment-group.scss';
+
+const UserInstancetypeList: FC = () => {
+  const { t } = useKubevirtTranslation();
+  const [activeNamespace] = useActiveNamespace();
+  const [instanceTypes, loaded, loadError] = useK8sWatchResource<
```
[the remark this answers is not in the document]
metalice: yes i know, i used it already, but this hook will fetch all except all projects, now I can edit it and add a boolean arg to it but seems redundant and less redable, because It will make branching in the hook, as there isn't any logic in this hook except the ns checking, so when using here directly usek8swatch , is same as using a custom hook, because there is no logic except the fetching.. so i didn't used the hook u recommended and directly fetched.

hstastna (commented):
Your PR looks good! Great work! I like it. I'd just renamed _list-managment-group.scss_ file to _list-management-group.scss_ while editing this file (see the missing 'e' there). But that's a minor thing, up to you, Matan :)

src/views/instancetypes/list/InstanceTypePage.tsx
```diff
@@ -0,0 +1,93 @@
...
+    );
+  }, [activeTabKey, history, urlUserPreference]);
+
+  return (
+    <>
+      <div className={classNames({ 'list-header-spacer': activeTabKey === 0 })}>
+        <ListPageHeader
+          title={
+            activeTabKey === 0
+              ? t('VirtualMachineClusterInstancetype')
```
hstastna, line 61: Why singular form used if it is a list of items we display?

metalice:
> Your PR looks good! Great work! I like it. I'd just renamed _list-managment-group.scss_ file to _list-management-group.scss_ while editing this file (see the missing 'e' there). But that's a minor thing, up to you, Matan :)

u right, but should be in separate pr as this will add 9 more files to this PR

src/views/instancetypes/list/InstanceTypePage.tsx
metalice: sure changed

avivtur (approved):

PR #1750, opened 2024-01-26
CNV-23927: Do not count persistentVolumeClaim volumes in DeleteVMModal

upalatucci:
## 📝 Description

We should not count volumes that use PVC attached with `persistentVolumeClaim`  in the DeleteVM modal.

Those volumes are attached to the VM as `existing PVCs` so they do not have the same VM lifecycle. 

Maybe in the next version we can consider a checkbox list to let the user decide which disks wants to delete. 

## Demo

VM with one volume that use `persistVolumeClaim`

**Before**

image

**After**

     
image

### DIFF
--- a/src/views/virtualmachines/actions/components/DeleteVMModal/hooks/useDataVolumeConvertedVolumeNames.ts
+++ b/src/views/virtualmachines/actions/components/DeleteVMModal/hooks/useDataVolumeConvertedVolumeNames.ts
@@ -5,7 +5,7 @@ import { useK8sWatchResource } from '@openshift-console/dynamic-plugin-sdk';
 
 type UseDataVolumeConvertedVolumeNames = (vmVolumes: V1Volume[]) => {
   dvVolumesNames: string[];
-  pvcVolumesNames: string[];
+  isDataVolumeGarbageCollector: boolean;
 };
 
 const useDataVolumeConvertedVolumeNames: UseDataVolumeConvertedVolumeNames = (vmVolumes) => {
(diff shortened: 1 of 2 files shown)


src/views/virtualmachines/actions/components/DeleteVMModal/hooks/useDataVolumeConvertedVolumeNames.ts
```diff
@@ -21,14 +21,9 @@ const useDataVolumeConvertedVolumeNames: UseDataVolumeConvertedVolumeNames = (vm
...
 
-  const pvcVolumesNames = (vmVolumes || [])
-    .filter((volume) => volume?.persistentVolumeClaim)
-    ?.map((volume) => volume?.persistentVolumeClaim?.claimName);
   return {
     dvVolumesNames,
-    pvcVolumesNames: isDataVolumeGarbageCollector
-      ? pvcVolumesNames.concat(dvVolumesNames)
-      : pvcVolumesNames,
+    pvcVolumesNames: isDataVolumeGarbageCollector ? dvVolumesNames : [],
```
hstastna, line 26: Hmm, does it still make sense to call this "pvcVolumesNames"?

upalatucci: I'll propose a new thing in order to fix that.

avivtur (approved):

PR #1753, opened 2024-01-28
CNV-37373: VM IP address disappears from UI when migrating VM

avivtur:
## 📝 Description

The VM row would show IP and node only if the VM is in "Running" status, we should also show ip and node name even if VM is migrating (still running) or if the VM is paused.

## 🎥 Demo

After:

https://github.com/kubevirt-ui/kubevirt-plugin/assets/67270715/3c0ab398-2f5b-4970-bec3-75a5929b0b06

Before:

https://github.com/kubevirt-ui/kubevirt-plugin/assets/67270715/b76eab52-31d6-41ca-97f0-f76ad2d8d9bd

### DIFF
--- a/src/views/virtualmachines/list/components/VirtualMachineRow/VirtualMachineRow.tsx
+++ b/src/views/virtualmachines/list/components/VirtualMachineRow/VirtualMachineRow.tsx
@@ -1,19 +1,19 @@
-import * as React from 'react';
+import React, { FC } from 'react';
 
 import {
   V1VirtualMachine,
   V1VirtualMachineInstance,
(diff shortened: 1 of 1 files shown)


hstastna (commented):
LGTM! You can also update the 1st line of the edited file, you know...
`import * as React from 'react';` to 
`import React, {FC, ...} from 'react';`
😏

avivtur:
> LGTM! You can also update the 1st line of the edited file, you know... `import * as React from 'react';` to `import React, {FC, ...} from 'react';` 😏

done, thanks!

hstastna (approved):

PR #1779, opened 2024-02-05
CNV-37111: Make Bootable volumes list page column management work

hstastna:
## 📝 Description

**Fixes:**
https://issues.redhat.com/browse/CNV-37111
https://issues.redhat.com/browse/CNV-37821

Make _Bootable volumes_ list page column management work so the settings are saved and applied when user enters the page repeatedly.

_Notes:_
- this change fixes "not saveable" column management also in some other pages, except _Preferences_ list page, where it didn't work at all and still doesn't with this change, probably needs some different fix (followup commit/PR)
- when being in "All namespaces" in _Bootable volumes_ list page, only in this case the _Namespace_ column is displayed (in case it is selected when managing columns, of course), otherwise not

## 🎥 Demo
**Before:**

https://github.com/kubevirt-ui/kubevirt-plugin/assets/13417815/0a60f8cc-15d8-4068-b0d8-63af0b330363

**After:**

https://github.com/kubevirt-ui/kubevirt-plugin/assets/13417815/d5f94938-b359-4a55-a777-4ef10a7fee87

### DIFF
--- a/src/utils/hooks/useKubevirtUserSettings/useKubevirtUserSettingsTableColumns.ts
+++ b/src/utils/hooks/useKubevirtUserSettings/useKubevirtUserSettingsTableColumns.ts
@@ -1,6 +1,7 @@
 import { useCallback, useEffect } from 'react';
 
 import { isEqualObject } from '@kubevirt-utils/components/NodeSelectorModal/utils/helpers';
+import { KUBEVIRT_V1_VIRTUALMACHINE } from '@kubevirt-utils/constants/constants';
 import { TableColumn, useActiveColumns } from '@openshift-console/dynamic-plugin-sdk';
 import { useUserSettings } from '@openshift-console/dynamic-plugin-sdk-internal';
 
@@ -35,7 +36,11 @@ const useKubevirtUserSettingsTableColumns: UseKubevirtUserSettingsTableColumnsTy
     if (!loaded || error) return;
 
     if (
-      !isEqualObject(userColumns?.[columnManagementID], localStorageSettings?.[columnManagementID])
+      !isEqualObject(
+        userColumns?.[columnManagementID],
+        localStorageSettings?.[columnManagementID],
+      ) &&
+      columnManagementID === KUBEVIRT_V1_VIRTUALMACHINE
     ) {
       setLocalStorageSettings({
         ...localStorageSettings,
--- a/src/utils/constants/constants.ts
+++ b/src/utils/constants/constants.ts
@@ -12,3 +12,5 @@ export const VENDOR_LABEL = 'instancetype.kubevirt.io/vendor';
 export const AUTO_COMPUTE_CPU_LIMITS_LABEL = 'cpu.autocompute.kubevirt.io=true';
 
 export const ROOTDISK = 'rootdisk';
+
+export const KUBEVIRT_V1_VIRTUALMACHINE = 'kubevirt.io~v1~VirtualMachine';
--- a/src/utils/tests/resource.test.ts
+++ b/src/utils/tests/resource.test.ts
@@ -1,5 +1,5 @@
 import VirtualMachineModel from '@kubevirt-ui/kubevirt-api/console/models/VirtualMachineModel';
-import { DEFAULT_NAMESPACE } from '@kubevirt-utils/constants/constants';
+import { DEFAULT_NAMESPACE, KUBEVIRT_V1_VIRTUALMACHINE } from '@kubevirt-utils/constants/constants';
 import { getResourceUrl } from '@kubevirt-utils/resources/shared';
 
 jest.mock('@openshift-console/dynamic-plugin-sdk', () => ({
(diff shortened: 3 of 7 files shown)


hstastna:

src/utils/hooks/useKubevirtUserSettings/useKubevirtUserSettingsTableColumns.ts
```diff
@@ -35,7 +36,11 @@ const useKubevirtUserSettingsTableColumns: UseKubevirtUserSettingsTableColumnsTy
     if (!loaded || error) return;
 
     if (
-      !isEqualObject(userColumns?.[columnManagementID], localStorageSettings?.[columnManagementID])
+      !isEqualObject(
+        userColumns?.[columnManagementID],
+        localStorageSettings?.[columnManagementID],
+      ) &&
+      columnManagementID === KUBEVIRT_V1_VIRTUALMACHINE
```
metalice, line 43: Please explain. How does this fix? Seems like a selective case inside agnostic hook. Why we need it?

hstastna: I knew this question is gonna appear here because it looks like an ugly quick fix. The answer is because the column management works differently for different pages, unfortunately. Because we use custom ListPageFilter component in some places. The more custom and different stuff we are gonna have for each page, the more conditions we have to put in some hook we wanna use in all those places...

The better is to create different hooks for different places and to share some code between them which is the same => lots of refactoring and potential bugs we cannot afford now (also because lack of unit tests...). I am not so brave to do this all by myself, the logic around saving user settings + column management got too complicated for one person to do this, while it is more understandable for those who have implemented that and pretty know every variable's meaning 😏 For now, this adjustment of the existing condition is gonna make column management work perfectly in almost all the places it didn't work (just one place is even more specific, needs different fix).

metalice: Im sorry but it isnt seems reasonable, this hook is agnostic, not related to any custom other components that use it, it simply has a bug in it.
i played with it and I think I have a solution which isn't just making a special case of something please test this.
`import { useCallback, useEffect, useRef } from 'react';

import { isEqualObject } from '@kubevirt-utils/components/NodeSelectorModal/utils/helpers';
import { TableColumn, useActiveColumns } from '@openshift-console/dynamic-plugin-sdk';
import { useUserSettings } from '@openshift-console/dynamic-plugin-sdk-internal';

import useKubevirtUserSettings from './useKubevirtUserSettings';

type UseKubevirtUserSettingsTableColumnsType = <T>(input: {
  columnManagementID;
  columns;
}) => [
  activeColumns: TableColumn<T>[],
  setActiveColumns: (val: any) => void,
  loaded: boolean,
  error: Error,
];

const useKubevirtUserSettingsTableColumns: UseKubevirtUserSettingsTableColumnsType = <T>({
  columnManagementID,
  columns,
}) => {
  const [userColumns, setUserColumns, loaded, error] = useKubevirtUserSettings('columns');
  const firstTimeLoading = useRef<boolean>(true);
  const [localStorageSettings, setLocalStorageSettings] = useUserSettings<{
    [key: string]: string;
  }>('console.tableColumns');

  const [activeColumns] = useActiveColumns<T>({
    columnManagementID,
    columns,
    showNamespaceOverride: false,
  });

  useEffect(() => {
    if (!loaded || error) return;

    if (
      userColumns?.[columnManagementID] &&
      !isEqualObject(userColumns?.[columnManagementID], localStorageSettings?.[columnManagementID])
    ) {
      setLocalStorageSettings({
        ...localStorageSettings,
        [columnManagementID]: userColumns?.[columnManagementID],
      });
      return;
    }

    if (firstTimeLoading.current && activeColumns?.length > 0) {
      firstTimeLoading.current = false;
      setUserColumns?.({
        ...userColumns,
        [columnManagementID]: activeColumns.map((col) => col?.id),
      });
    }
  }, [
    activeColumns,
    columnManagementID,
    localStorageSettings,
    setLocalStorageSettings,
    setUserColumns,
    userColumns,
    loaded,
    error,
  ]);

  const setActiveColumns = useCallback(
    (columnIds: string[]) =>
      setUserColumns?.({
        ...userColumns,
        [columnManagementID]: columnIds,
      }),
    [columnManagementID, setUserColumns, userColumns],
  );

  return [activeColumns, setActiveColumns, loaded, error];
};

export default useKubevirtUserSettingsTableColumns;`

metalice: btw, there is other places such as bootablevolulellust that use the same DataSourceModelRef for a few list, which is also a bug here as name should be unique

hstastna: So I've tested your suggested change and it does not work, it works only for VMs list. My solution worked for every list page, except Preferences list, where it didn't work anyway. I suggest to use my solution, at least temporary, until we find a better solution.

metalice: This is just a sample of a use case of non existing entry in config map, for this scenario its works.
Hilda, this addition that you added isn't fixing, u are canceling the mechanism of updating, its just like using the useActiveColumns hook directly, go over the code try to understand what is going on and what is the update scenario, it should have a real fix and it wont be that simple. if u have any question about a piece of function or code better to ask, this solution that u provided isn't a solution.

hstastna: Thank you for the comment and provided info. Feels like I wasn't going thru the code at all and have no idea what I am doing. I'm closing the PR for now so better ones could be opened soon.

PR #1780, opened 2024-02-05
CNV-37857: Enable SSH secret selection on quick create flow

avivtur:
## 📝 Description

Allowing to select a secret for SSH on VM quick create flow.

## 🎥 Demo

After:

ssh-quick-create2
ssh-quick-create


src/views/catalog/utils/quick-create-vm.ts
```diff
@@ -82,17 +78,13 @@ export const quickCreateVM: QuickCreateVMType = async ({
...
-    ? addSecretToVM(overridedVM, authorizedSSHKey)
-    : overridedVM;
-
-  const { objects } = replaceTemplateVM(processedTemplate, vmToCreate);
+  const { objects } = replaceTemplateVM(processedTemplate, overridedVM);
 
   const createdObjects = await createMultipleResources(objects, models, namespace);
 
-  const createdVM = createdObjects.find(
+  const vmToCreate = createdObjects.find(
```
metalice, line 85: why changing the name? i think the VM is already created in this step

PR #1791, opened 2024-02-12
CNV-37813: upgrading to react router v6

metalice:
## 📝 Description

Replacing react-router v5 in v6 version.

## 🎥 Demo

Internal changes only, no difference in view.


metalice:
until this https://github.com/openshift/release/pull/48740 is merged it wont pass test, our ci is using 4.14 and we need 4.15

PR #1794, opened 2024-02-14
CNV-34180: fetch user accessible hyperconverge config resource

upalatucci:
## 📝 Description

Regular users cannot access Hyperconverge configuration. 
in order to fix this bug we have to fetch the new Kubevirt Hyperconverge configuration made exactly for that

## 🎥 Demo

Screenshot from 2024-02-14 12-01-51


src/utils/hooks/useKubevirtHyperconvergeConfiguration.ts/index.ts
```diff
@@ -0,0 +1,22 @@
...
+
+import { KUBEVIRT_HC_GROUP_VERSION_KIND, KUBEVIRT_HC_NAME } from './constants';
+
+export type KubevirtHyperconverged = K8sResourceCommon & {
+  spec: {
+    configuration: V1KubeVirtConfiguration;
+  };
+};
+
+const useKubevirtHyperconvergeConfiguration = () => {
```
metalice, line 14: Only 1 component uses it? 
if the answer is yes, there is no need for this hook as it does no more the k8swatch, no logic, so use the k8swatch hook
if the answer is no, please amend it a bit to be nicer. Don't simply return the k8swatch hook itself; do the same as we do in all our other hooks.

PR #1807, opened 2024-02-20
Fix crash when sshDetails is null

upalatucci:
## 📝 Description

when `sshDetails` is `null`, the catalog drawer crash.

## 🎥 Demo

**Before**

image

**After**

Screenshot from 2024-02-20 13-40-14

### DIFF
--- a/src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/AuthorizedSSHKey.tsx
+++ b/src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/AuthorizedSSHKey.tsx
@@ -71,14 +71,14 @@ const AuthorizedSSHKey: FC<AuthorizedSSHKeyProps> = ({ authorizedSSHKey, namespa
 
       if (
         secretOption === SecretSelectionOption.none &&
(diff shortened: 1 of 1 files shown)


i18n-scripts/set-english-defaults.js
```diff
@@ -48,11 +49,14 @@ function updateFile(fileName) {
       switch (determineRule(keys[i])) {
         case 0:
           [originalKey] = keys[i].split('_other');
-          updatedFile[keys[i]] = `{{count}} ${pluralize(originalKey)}`;
+
+          wordWithNoCount = originalKey.replace('{{count}}', '').trim()
```
hstastna, line 53: Are these changes in this file intentional? Feels like those don't belong to this PR...

upalatucci: No i pushed in the wrong branch

upalatucci: fixing

hstastna (approved):

avivtur:
/hold
Holding until upgrade to PF5 is completed

PR #1812, opened 2024-02-21
CNV-38577: Fix initial load of serial console

metalice:
## 📝 Description

The initial setting of the serial console wasn't loaded into the features config map.
More checks were added to tell if the VM needs to restart after enabling the setting in cluster scope.

## 🎥 Demo

Before:
as can be seen in the photos in the bug report https://issues.redhat.com/browse/CNV-38577

After:
image

### DIFF
--- a/src/views/virtualmachines/details/tabs/diagnostic/VirtualMachineLogViewer/VirtualMachineLogViewer.tsx
+++ b/src/views/virtualmachines/details/tabs/diagnostic/VirtualMachineLogViewer/VirtualMachineLogViewer.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 
 import { V1Devices } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import Loading from '@kubevirt-utils/components/Loading/Loading';
 import { getChangedGuestSystemAccessLog } from '@kubevirt-utils/components/PendingChanges/utils/helpers';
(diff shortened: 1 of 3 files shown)


avivtur (commented):
/lgtm

src/views/virtualmachines/details/tabs/diagnostic/VirtualMachineLogViewer/VirtualMachineLogViewer.tsx
```diff
@@ -15,31 +16,44 @@ import VirtualMachineBasicLogViewer from './VirtualMachineBasicLogViewer/Virtual
...
+  if (!isRunning(vm)) {
+    return <Bullseye>{t('Virtual machine is not running')}</Bullseye>;
   }
 
   if (isClusterDisabledGuestSystemLogs) {
     return <Bullseye>{t('Guest system logs are disabled at cluster')}</Bullseye>;
   }
 
+  if (isNeededRestart || !isPodLogContainerExist) {
+    return <Bullseye>{t('Guest system access logs not ready. a restart required')}</Bullseye>;
```
hstastna, line 54: Why are you adding this string here? Just recently it was suggested to me to use "Restart required" instead of "a restart required". And the sentence should start with a big letter.. Maybe you've just accidentally copied here the older version of this string?

metalice: Probably beause moved the IF check and it took the string before. anyway changed

hstastna (approved):

PR #1817, opened 2024-03-03
CNV-37815: Refactor SSH modal

pcbailey:
## 📝 Description

jira: https://issues.redhat.com/browse/CNV-37815


src/utils/components/SSHSecretModal/components/SSHOptionUseExisting/SSHOptionUseExisting.tsx
```diff
@@ -102,7 +95,13 @@ const SSHOptionUseExisting: FC<SSHOptionUseExistingProps> = ({
     }));
   };
 
-  return !isEmpty(projects) ? (
+  const showNewSecretNameField = namespace
+    ? selectedProject !== namespace
+    : selectedProject !== sshDetails?.sshSecretNamespace;
+
+  return isEmpty(projects) ? (
```
metalice, line 102: can u do isEmpty... return; instead of terenry?

src/utils/hooks/useProjects.ts
```diff
@@ -0,0 +1,15 @@
+import { modelToGroupVersionKind, ProjectModel } from '@kubevirt-ui/kubevirt-api/console';
+import { K8sResourceCommon, useK8sWatchResource } from '@openshift-console/dynamic-plugin-sdk';
+
+type UseProjects = () => string[];
+
+const useProjects: UseProjects = () => {
```
metalice, line 6: please check if we dont have this kind of hook

src/utils/components/SSHSecretModal/hooks/useSecretsSaveBtnDisabled.ts
```diff
@@ -0,0 +1,45 @@
+import { useMemo } from 'react';
```
metalice, line 1: It seems like this is a hook because of useMemo. Do we really need it, or can it be a regular function?

src/utils/hooks/useProjects.ts
pcbailey: We have `useProjectsAndPVCs` and `useProjectsAndDataSources`, but we don't have a hook for just projects. I suggest we keep this `useProjects` and use it inside of the other hooks.

src/utils/components/SSHSecretModal/components/SSHOptionUseExisting/SSHOptionUseExisting.tsx
pcbailey: Done!

src/utils/components/SSHSecretModal/hooks/useSecretsSaveBtnDisabled.ts
pcbailey: I'll check.

pcbailey: Done!

PR #1828, opened 2024-03-07
CNV-38331: Changing snapshot actions and modals text

metalice:
## 📝 Description

Changing text of snapshot actions and modals
## 🎥 Demo

After:
image

image

image

image

Before:
image

image

image

image


avivtur (commented):
Looks good! just some small comments :)

PR #1837, opened 2024-03-08
 CNV-38902: Display MigrationPolicies page for nonprivileged user

hstastna:
## 📝 Description

**Fixes:**
https://issues.redhat.com/browse/CNV-38902

Display the page also for the users without privileges to get a bit more familiar with the page, add popover for disabled _Create MigrationPolicy_ button to inform the user what to do to be able to see and create MigrationPolicies.

## 🎥 Screenshots
**Before:**
No MigrationPolicies menu item, nor the paqe displayed at all:
mig_beforee

**After?:**
mig_after

### DIFF
--- a/src/views/clusteroverview/MigrationsTab/hooks/useMigrationCardData.ts
+++ b/src/views/clusteroverview/MigrationsTab/hooks/useMigrationCardData.ts
@@ -1,14 +1,13 @@
 import {
-  MigrationPolicyModelGroupVersionKind,
   VirtualMachineInstanceMigrationModelGroupVersionKind,
   VirtualMachineInstanceModelGroupVersionKind,
 } from '@kubevirt-ui/kubevirt-api/console';
 import {
-  V1alpha1MigrationPolicy,
   V1VirtualMachineInstance,
   V1VirtualMachineInstanceMigration,
 } from '@kubevirt-ui/kubevirt-api/kubevirt';
 import { ALL_NAMESPACES_SESSION_KEY } from '@kubevirt-utils/hooks/constants';
+import useMigrationPolicies from '@kubevirt-utils/hooks/useMigrationPolicies';
 import {
   OnFilterChange,
   RowFilter,
@@ -39,13 +38,6 @@ export type UseMigrationCardDataAndFiltersValues = {
 };
 
 type UseMigrationCardDataAndFilters = (duration: string) => UseMigrationCardDataAndFiltersValues;
-
-export const useMigrationPolicies = () =>
-  useK8sWatchResource<V1alpha1MigrationPolicy[]>({
-    groupVersionKind: MigrationPolicyModelGroupVersionKind,
-    isList: true,
-    namespaced: false,
-  });
 
 const useMigrationCardDataAndFilters: UseMigrationCardDataAndFilters = (duration: string) => {
   const migrationsDefaultConfigurations = useHCMigrations();
--- a/src/views/clusteroverview/MigrationsTab/hooks/useMigrationCardData.ts
+++ b/src/views/clusteroverview/MigrationsTab/hooks/useMigrationCardData.ts
@@ -40,6 +40,13 @@ export type UseMigrationCardDataAndFiltersValues = {
 
 type UseMigrationCardDataAndFilters = (duration: string) => UseMigrationCardDataAndFiltersValues;
 
+export const useMigrationPolicies = () =>
+  useK8sWatchResource<V1alpha1MigrationPolicy[]>({
+    groupVersionKind: MigrationPolicyModelGroupVersionKind,
+    isList: true,
+    namespaced: false,
+  });
+
 const useMigrationCardDataAndFilters: UseMigrationCardDataAndFilters = (duration: string) => {
   const migrationsDefaultConfigurations = useHCMigrations();
   const [activeNamespace] = useActiveNamespace();
@@ -59,10 +66,7 @@ const useMigrationCardDataAndFilters: UseMigrationCardDataAndFilters = (duration
     namespace,
   });
 
-  const [mps] = useK8sWatchResource<V1alpha1MigrationPolicy[]>({
-    groupVersionKind: MigrationPolicyModelGroupVersionKind,
-    isList: true,
-  });
+  const [mps] = useMigrationPolicies();
 
   const migrationsData = getMigrationsTableData(
     vmims,
--- a/src/views/migrationpolicies/list/MigrationPoliciesList.tsx
+++ b/src/views/migrationpolicies/list/MigrationPoliciesList.tsx
@@ -1,10 +1,10 @@
 import React, { FC } from 'react';
-import { useMigrationPolicies } from 'src/views/clusteroverview/MigrationsTab/hooks/useMigrationCardData';
 
 import { MigrationPolicyModelRef } from '@kubevirt-ui/kubevirt-api/console';
 import { V1alpha1MigrationPolicy } from '@kubevirt-ui/kubevirt-api/kubevirt';
 import ListPageFilter from '@kubevirt-utils/components/ListPageFilter/ListPageFilter';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
+import useMigrationPolicies from '@kubevirt-utils/hooks/useMigrationPolicies';
 import { isEmpty } from '@kubevirt-utils/utils/utils';
 import {
   ListPageBody,
@@ -47,18 +47,15 @@ const MigrationPoliciesList: FC = () => {
           loaded={loaded && loadedColumns}
           onFilterChange={onFilterChange}
         />
-        {isEmpty(mps) && <MigrationPoliciesEmptyState loadError={!loaded && loadError} />}
-        {loaded && (
-          <VirtualizedTable<V1alpha1MigrationPolicy>
-            columns={activeColumns}
-            data={data}
-            EmptyMsg={() => <></>}
-            loaded={loaded && loadedColumns}
-            loadError={loadError}
-            Row={MigrationPoliciesRow}
-            unfilteredData={unfilteredData}
-          />
-        )}
+        <VirtualizedTable<V1alpha1MigrationPolicy>
+          columns={activeColumns}
+          data={data}
+          EmptyMsg={() => <MigrationPoliciesEmptyState />}
+          loaded={loaded && loadedColumns}
+          loadError={loadError}
+          Row={MigrationPoliciesRow}
+          unfilteredData={unfilteredData}
+        />
       </ListPageBody>
     </>
   );
--- a/src/views/migrationpolicies/list/MigrationPoliciesList.tsx
+++ b/src/views/migrationpolicies/list/MigrationPoliciesList.tsx
@@ -1,9 +1,7 @@
 import React, { FC } from 'react';
+import { useMigrationPolicies } from 'src/views/clusteroverview/MigrationsTab/hooks/useMigrationCardData';
 
-import {
-  MigrationPolicyModelGroupVersionKind,
-  MigrationPolicyModelRef,
-} from '@kubevirt-ui/kubevirt-api/console';
+import { MigrationPolicyModelRef } from '@kubevirt-ui/kubevirt-api/console';
 import { V1alpha1MigrationPolicy } from '@kubevirt-ui/kubevirt-api/kubevirt';
 import ListPageFilter from '@kubevirt-utils/components/ListPageFilter/ListPageFilter';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
@@ -11,7 +9,6 @@ import { isEmpty } from '@kubevirt-utils/utils/utils';
 import {
   ListPageBody,
   ListPageHeader,
-  useK8sWatchResource,
   useListPageFilter,
   VirtualizedTable,
 } from '@openshift-console/dynamic-plugin-sdk';
@@ -23,11 +20,7 @@ import useMigrationPoliciesListColumns from './hooks/useMigrationPoliciesListCol
 
 const MigrationPoliciesList: FC = () => {
   const { t } = useKubevirtTranslation();
-  const [mps, loaded, loadError] = useK8sWatchResource<V1alpha1MigrationPolicy[]>({
-    groupVersionKind: MigrationPolicyModelGroupVersionKind,
-    isList: true,
-    namespaced: false,
-  });
+  const [mps, loaded, loadError] = useMigrationPolicies();
 
   const [columns, activeColumns, loadedColumns] = useMigrationPoliciesListColumns();
   const [unfilteredData, data, onFilterChange] = useListPageFilter(mps);
@@ -54,16 +47,18 @@ const MigrationPoliciesList: FC = () => {
           loaded={loaded && loadedColumns}
           onFilterChange={onFilterChange}
         />
-        {loaded && isEmpty(mps) && <MigrationPoliciesEmptyState />}
-        <VirtualizedTable<V1alpha1MigrationPolicy>
-          columns={activeColumns}
-          data={data}
-          EmptyMsg={() => <></>}
-          loaded={loaded && loadedColumns}
-          loadError={loadError}
-          Row={MigrationPoliciesRow}
-          unfilteredData={unfilteredData}
-        />
+        {isEmpty(mps) && <MigrationPoliciesEmptyState loadError={!loaded && loadError} />}
+        {loaded && (
+          <VirtualizedTable<V1alpha1MigrationPolicy>
+            columns={activeColumns}
+            data={data}
+            EmptyMsg={() => <></>}
+            loaded={loaded && loadedColumns}
+            loadError={loadError}
+            Row={MigrationPoliciesRow}
+            unfilteredData={unfilteredData}
+          />
+        )}
       </ListPageBody>
     </>
   );
--- a/src/views/migrationpolicies/list/components/MigrationPoliciesCreateButton/MigrationPoliciesCreateButton.tsx
+++ b/src/views/migrationpolicies/list/components/MigrationPoliciesCreateButton/MigrationPoliciesCreateButton.tsx
@@ -2,8 +2,14 @@ import React, { FC } from 'react';
 import { useNavigate } from 'react-router-dom-v5-compat';
 
 import { MigrationPolicyModelGroupVersionKind } from '@kubevirt-ui/kubevirt-api/console';
+import MigrationPolicyModel from '@kubevirt-ui/kubevirt-api/console/models/MigrationPolicyModel';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
-import { ListPageCreateDropdown } from '@openshift-console/dynamic-plugin-sdk';
+import {
+  K8sVerb,
+  ListPageCreateDropdown,
+  useAccessReview,
+} from '@openshift-console/dynamic-plugin-sdk';
+import { Button, Tooltip } from '@patternfly/react-core';
 
 import { createItems, migrationPoliciesPageBaseURL } from '../../utils/constants';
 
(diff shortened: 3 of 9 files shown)


hstastna:
// @avivtur @metalice @pcbailey @upalatucci @yzamir @vojtechszocs please review

hstastna:

src/views/migrationpolicies/list/components/MigrationPoliciesEmptyState/MigrationPoliciesEmptyState.tsx
```diff
@@ -39,7 +46,11 @@ const MigrationPoliciesEmptyState: FC = () => {
       </EmptyStateBody>
       <EmptyStateFooter>
         <EmptyStateActions>
-          <MigrationPoliciesCreateButton />
+          {isUnprivilegedUser ? (
+            <MigrationPoliciesCreateButtonDisabled />
+          ) : (
+            <MigrationPoliciesCreateButton />
+          )}
```
metalice, line 53: why not using 1 component for the button and let this component check for privileges?

src/views/migrationpolicies/list/MigrationPoliciesList.tsx
```diff
@@ -28,6 +28,9 @@ const MigrationPoliciesList: FC = () => {
     isList: true,
     namespaced: false,
   });
+  const isUnprivilegedUser =
+    !loaded &&
+    loadError?.toString()?.includes('migrationpolicies.migrations.kubevirt.io is forbidden');
```
metalice, line 33: please use access check and not trying to parse an error msg, the error can be changed without any notice.
u can find examples in the permissions hooks that we have

src/views/migrationpolicies/list/components/MigrationPoliciesEmptyState/MigrationPoliciesEmptyState.tsx
hstastna: Because code is more readable now and this can be changed very soon, too, as the discussions about future of MigrationPolicies is still happening and different approaches occur, so it would be much easier to delete the unnecessary part of the code, once another change will be proposed (it is probable).

src/views/migrationpolicies/list/MigrationPoliciesList.tsx
hstastna: Can you please provide more details about what you mean? Haven't found anything related yet.. Thanks.

metalice: for example https://github.com/kubevirt-ui/kubevirt-plugin/blob/main/src/utils/hooks/usePermissions/usePermissions.ts

src/views/migrationpolicies/list/components/MigrationPoliciesEmptyState/MigrationPoliciesEmptyState.tsx
metalice: It seems to me like a ternary inside a return of a component is much less readily and increases file size; this can be combined into a single component of MigrationPoliciesCreateButton

hstastna: So I will have to put ternary operator inside another component :D I don't see any difference. Also I will have to propagate another prop there, unnecessary.

metalice: well no, if you build it correctly you won't need the ternary

metalice: and the component check the the privileges so no need for another prop

hstastna: Well in this file the component does not check exactly what we need. We need to check the error message, unfortunately. Also I am building the logic on top of what was already built there (not ideal) and according to actual conditions (changing decisions of various people here and there and too often). Maybe I should have added a comment to the code, to make it more clear and provide more info about why this way.

metalice: > Well I am still not convinced. In this file the component does not check exactly what we need. We need to check the error message specifically, unfortunately. Also I am building the logic on top of what was already built there (not ideal) and according to actual conditions (changing decisions of various people here and there and too often). Maybe I should have added a comment to the code, to make it more clear and provide more info about why this way.

1. Yes, u need to add it.
2. We don't need to check error messages, and we don't do it anywhere else in our code base, as messages can be changed. They aren't part of the API contract; they are simply strings.
3. Don't just add logic and make it less maintainable, readable, and scalable. Every touch in code should improve, not just add things because it's the assignment.
4. What do u mean by changing the decisions of various people? what was implemented before won't mean it's what we want now. the product is getting bigger, more requirements are added all the time, and we need to scale and build accordingly; things are always changing and will continue to change; what was right a month ago may not be right two months ahead, according to the road map. it's ok to change the code that others wrote in order to get the task done, but we need to make the code better, not harder to maintain and read.

If you still don't want to make the right changes and would like to continue to discuss this, you can reach me privately.

hstastna: Of course I do want to make the right changes. But it seems we have very different opinion on what is right actually, and it seems you don't understand many reasons why and what I had to face to make the decisions I made. Will talk to you later.

src/views/migrationpolicies/list/MigrationPoliciesList.tsx
hstastna: This hook seems to be for totally a different use!

metalice: This hook is an example of checking permissions for the user; it's exactly what is needed to know if the button should be disabled or not based on permissions.

hstastna:

src/views/migrationpolicies/list/MigrationPoliciesList.tsx
```diff
@@ -54,16 +47,18 @@ const MigrationPoliciesList: FC = () => {
...
+          <VirtualizedTable<V1alpha1MigrationPolicy>
+            columns={activeColumns}
+            data={data}
+            EmptyMsg={() => <></>}
+            loaded={loaded && loadedColumns}
+            loadError={loadError}
+            Row={MigrationPoliciesRow}
+            unfilteredData={unfilteredData}
+          />
+        )}
```
avivtur, line 61:
```suggestion
        <VirtualizedTable<V1alpha1MigrationPolicy>
          columns={activeColumns}
          data={data}
          EmptyMsg={() => <MigrationPoliciesEmptyState />}
          loaded={loaded && loadedColumns}
          loadError={loadError}
          Row={MigrationPoliciesRow}
          unfilteredData={unfilteredData}
        />
```

src/views/migrationpolicies/list/components/MigrationPoliciesEmptyState/MigrationPoliciesEmptyState.scss
```diff
@@ -3,3 +3,12 @@
...
 }
+
+.MigrationPoliciesEmptyState {
+  &__button {
+    .pf-v5-c-button.pf-m-primary {
+      color: var(--pf-v5-c-button--disabled--Color);
+      background-color: var(--pf-v5-c-button--disabled--BackgroundColor);
+    }
+  }
+}
```
avivtur, line 14: Not needed, we can get this from PF

src/views/migrationpolicies/list/components/MigrationPoliciesEmptyState/MigrationPoliciesEmptyState.tsx
```diff
@@ -39,6 +48,21 @@ const MigrationPoliciesEmptyState: FC = () => {
...
+                )
+              }
+              aria-label={'Help'}
+              position={PopoverPosition.right}
+            >
+              <div className="MigrationPoliciesEmptyState__button">
+                <Button>{t('Create MigrationPolicy')}</Button>
+              </div>
+            </Popover>
+          )}
```
avivtur, line 65: We should add these changes to the MigrationPoliciesCreateButton as it's also can be rendered if we have items on the list

src/views/migrationpolicies/list/components/MigrationPoliciesEmptyState/MigrationPoliciesEmptyState.tsx
```diff
@@ -5,20 +5,29 @@ import migrationPoliciesEmptyState from 'images/migrationPoliciesEmptyState.svg'
...
+  Popover,
+  PopoverPosition,
 } from '@patternfly/react-core';
 
 import MigrationPoliciesCreateButton from '../MigrationPoliciesCreateButton/MigrationPoliciesCreateButton';
 
 import './MigrationPoliciesEmptyState.scss';
 
-const MigrationPoliciesEmptyState: FC = () => {
+export type MigrationPoliciesEmptyStateProps = {
```
avivtur, line 24: why export this type? I don't see it used anywhere outside this file

avivtur (changes requested):
Hi Hilda :) 
Nice work
to complete my comments I also recommend changing the MigrationPoliciesCreateButton.tsx to this:
```
import React, { FC } from 'react';
import { useNavigate } from 'react-router-dom-v5-compat';

import { MigrationPolicyModelGroupVersionKind } from '@kubevirt-ui/kubevirt-api/console';
import {
  K8sVerb,
  ListPageCreateDropdown,
  useAccessReview,
} from '@openshift-console/dynamic-plugin-sdk';

import { createItems, migrationPoliciesPageBaseURL } from '../../utils/constants';
import MigrationPolicyModel from '@kubevirt-ui/kubevirt-api/console/models/MigrationPolicyModel';
import { Button, Tooltip } from '@patternfly/react-core';

const MigrationPoliciesCreateButton: FC = () => {

  const [canCreateMigrationPolicy] = useAccessReview({
    group: MigrationPolicyModel.apiGroup,
    resource: MigrationPolicyModel.plural,
    verb: 'create' as K8sVerb,
  });

  const onCreate = (type: string) => {
    return type === 'form'
      ? navigate(`${migrationPoliciesPageBaseURL}/form`)
      : navigate(`${migrationPoliciesPageBaseURL}/~new`);
  };

  if (!canCreateMigrationPolicy) {
    return (
      <Tooltip
        content={t(
          'To perform this action you must get permission from your Organization Administrator.',
        )}
      >
        <Button isAriaDisabled>{t('Create MigrationPolicy')}</Button>
      </Tooltip>
    );
  }

  return (
    <ListPageCreateDropdown
      createAccessReview={{ groupVersionKind: MigrationPolicyModelGroupVersionKind }}
      items={createItems}
      onClick={onCreate}
    >
      {t('Create MigrationPolicy')}
    </ListPageCreateDropdown>
  );
};

export default MigrationPoliciesCreateButton;
```

This will cover both checking for create access review per user, and it applies to both occurrences we have for this button.
If you have any questions I'd be happy to answer

src/views/migrationpolicies/list/components/MigrationPoliciesEmptyState/MigrationPoliciesEmptyState.scss
hstastna: We need it. We need to keep the button clickable while having the style of not clickable one. Because of the design. To achieve this, we need to style the button. Otherwise it won't be clickable and no any popover could be rendered. Same for hovering on the button, unfortunately.

src/views/migrationpolicies/list/components/MigrationPoliciesEmptyState/MigrationPoliciesEmptyState.tsx
hstastna: This is problematic again, because the button won't be rendered at all (because of how PF implemented that) in case of the user without permissions, while we do need to render it (because of the design). I would have to add another prop to MigrationPoliciesCreateButton, but Matan told me I won't need that, but then we are not able to achieve the expected look, if I won't put this code in here. Because we need to check loaded or loadError at least. And this seems to me better (while checking the permissions once) than checking permissions of the same thing again and again.

src/views/migrationpolicies/list/MigrationPoliciesList.tsx
avivtur: This shouldn't keep loading and loading since the loaded is taking care of inside the VirtualizedTable component, and conditionally rendering according to loaded alone will miss if you have loadError as no permissions

src/views/clusteroverview/MigrationsTab/hooks/useMigrationCardData.ts
```diff
@@ -40,6 +40,13 @@ export type UseMigrationCardDataAndFiltersValues = {
 
 type UseMigrationCardDataAndFilters = (duration: string) => UseMigrationCardDataAndFiltersValues;
 
+export const useMigrationPolicies = () =>
+  useK8sWatchResource<V1alpha1MigrationPolicy[]>({
+    groupVersionKind: MigrationPolicyModelGroupVersionKind,
+    isList: true,
+    namespaced: false,
+  });
+
```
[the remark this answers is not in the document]
hstastna: Yes, right!

src/views/migrationpolicies/list/components/MigrationPoliciesEmptyState/MigrationPoliciesEmptyState.scss
avivtur: This can be achieved by using isAriaDisabled prop as I suggested in the code for MigrationPoliciesCreateButton.tsx

src/views/migrationpolicies/list/components/MigrationPoliciesEmptyState/MigrationPoliciesEmptyState.tsx
avivtur: Have you tried this? All changes I suggested here was tested locally and works as expected.
Please try this on your branch

src/views/migrationpolicies/list/MigrationPoliciesList.tsx
hstastna: But we don't want to display the permission error in the UI anymore, it is in the design. Without the condition we cannot achieve the expected result. Or can we? Did you test it? How did it look?

avivtur: I saw the design and it's a mistake in my opinion.
Showing the list is about if a user have 'list' permission while create a new policy is checking for the 'create' permissions.
I'm discussing this with Yifat from UX, please keep following the Jira ticket, thanks :)

src/views/migrationpolicies/list/components/MigrationPoliciesEmptyState/MigrationPoliciesEmptyState.scss
hstastna: Will try it out, thanks a lot, Aviv! :)

hstastna:
@avivtur Do u think the following picture is what we want to see? I am not sure, the actual design is quite far from this.
migg
Anyway, to achieve what we can see above in the picture, can be done by just one simple code change:
https://github.com/kubevirt-ui/kubevirt-plugin/pull/1837/files#diff-2d79046ce9b0f8fda82dd631ce959a36f6c350960d54cf8fdb1579f3c718e83aL533-L535
But there was more in the design I've received, and it seemed folks were quite sure about that...

avivtur:
> @avivtur Do u think the following picture is what we want to see? I am not sure, the actual design is quite far from this. migg Anyway, to achieve what we can see above in the picture, can be done by just one simple code change: https://github.com/kubevirt-ui/kubevirt-plugin/pull/1837/files#diff-2d79046ce9b0f8fda82dd631ce959a36f6c350960d54cf8fdb1579f3c718e83aL533-L535 But there was more in the design I've received, and it seemed folks were quite sure about that...

Great so the cluster you're using the test user is missing list permissions (and probably also create permissions as well)
To get only list permissions to test user for the migrationpolicies resource you can add this YAML:
```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: allow-migration-policies
rules:
  - apiGroups: ["migrations.kubevirt.io"]
    resources: ["migrationpolicies"]
    verbs: ["list", "get"]
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: allow-migration-policies-binding
subjects:
  - kind: Group
    name: system:authenticated
    apiGroup: rbac.authorization.k8s.io
roleRef:
  kind: ClusterRole
  name: allow-migration-policies
  apiGroup: rbac.authorization.k8s.io
```

avivtur:
> What are we trying to achieve by this? I am confused. Thanks.

When logging in as admin to a cluster, you can assign 'list' RBAC permission to the user 'test', this achieves removing the error page for a user, and you are able to see what a cluster looks like for a regular user, that an admin assigned him with permission to 'list' resources with the kind of 'MigrationPolicy'

src/views/migrationpolicies/list/components/MigrationPoliciesEmptyState/MigrationPoliciesEmptyState.tsx
hstastna: Depends on what we think about how the expected result looks like...

hstastna:
I am moving this PR to WIP as discussions about how the UI should look like, are still in progress.

src/views/clusteroverview/MigrationsTab/hooks/useMigrationPolicies.ts
avivtur, line 1: This hook is used in views `cluster overview` and in `migration policies`, It would be more appropriate to move this hook to under `src/utils/hooks`

avivtur (commented):
Looks great, Great job!
Just one small fix and this is good to go

src/views/migrationpolicies/list/components/MigrationPoliciesCreateButton/MigrationPoliciesCreateButton.tsx
```diff
@@ -2,21 +2,45 @@ import React, { FC } from 'react';
...
   };
 
+  if (!canCreateMigrationPolicy) {
+    return (
+      <Tooltip
+        content={t(
+          'To perform this action you must get permission from your Organization Administrator.',
+        )}
+      >
+        <Button isAriaDisabled>{t('Create MigrationPolicy')}</Button>
```
metalice, line 39: what is isAriaDisabled? we need it?

src/views/migrationpolicies/list/MigrationPoliciesList.tsx
```diff
@@ -1,17 +1,14 @@
 import React, { FC } from 'react';
+import useMigrationPolicies from 'src/views/clusteroverview/MigrationsTab/hooks/useMigrationPolicies';
```
metalice, line 2: please fix import

src/views/clusteroverview/MigrationsTab/hooks/useMigrationPolicies.ts
metalice: still here

src/views/migrationpolicies/list/components/MigrationPoliciesCreateButton/MigrationPoliciesCreateButton.tsx
hstastna: https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaDisabled

Using `isAriaDisabled` here improves accessibility the way that the button is simply there but read as disabled. Without using it may act the button is not there at all (for the reader).
Why didn't I use also `disabled`? Because when using that attribute for that PF `Button` component, we cannot display popover by clicking on the button at all. We need not to have it disabled while looking like it is disabled.. according to the design I received 🤷🏼‍♀️

PR #1864, opened 2024-04-04
CNV-39531: Align Getting started card with console

hstastna:
## 📝 Description

**Fixes:**
https://issues.redhat.com/browse/CNV-39531

Display _Getting started_ card as the expandable section, same as the new format in the console. Also use the `Tooltip` PF React component instead of popover to show help text on hovering the help icon in the title of the card and hence fix the incorrect cursor when hovering on the "?" icon.

_Important note:_
Please, recognize the difference between tooltip and popover components: tooltip is displayed on hover vs. popover on click. In this PR, I am using tooltip, which provides better user experience and is also aligned with console (even when the style for that help text in console looks like PF Popover component and not tooltip, which is incorrect).

## 🎥 Demo
**Before:**

https://github.com/kubevirt-ui/kubevirt-plugin/assets/13417815/84767c06-c074-4dfe-8dcc-1bc71ba3e28c

**After:**

https://github.com/kubevirt-ui/kubevirt-plugin/assets/13417815/bc4bcac9-bbea-43bf-9a1b-2d5d19f285e8

### DIFF
--- a/src/views/clusteroverview/OverviewTab/getting-started-card/utils/getting-started-grid/GettingStartedGrid.scss
+++ b/src/views/clusteroverview/OverviewTab/getting-started-card/utils/getting-started-grid/GettingStartedGrid.scss
@@ -5,12 +5,11 @@
   &__header.pf-v5-c-card__header {
     // Use padding sm instead of lg to fix alignment of the KebabToggle action button.
     padding-right: var(--pf-global--spacer--sm);
+    padding-block-start: 0;
   }
-  &__tooltip {
-    white-space: pre-line;
-  }
+
   &__tooltip-icon {
-    padding-left: var(--pf-global--spacer--sm);
+    margin-left: var(--pf-global--spacer--sm);
   }
 
   &__action-dropdown {
@@ -42,6 +41,18 @@
       padding-bottom: calc(var(--pf-v5-c-card--child--PaddingBottom) / 2);
       padding-left: var(--pf-v5-c-card--child--PaddingLeft);
       padding-right: var(--pf-v5-c-card--child--PaddingRight);
+    }
+  }
+
+  &__tooltip {
+    background: var(--pf-c-page__main-section--m-light--BackgroundColor);
+  }
+
+  &__expandable {
+    background: var(--pf-c-page__main-section--m-light--BackgroundColor);
+
+    .pf-v5-c-expandable-section__toggle-text {
+      color: var(--pf-v5-global--Color--dark-100);
     }
   }
 }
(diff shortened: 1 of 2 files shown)


hstastna:

src/views/clusteroverview/OverviewTab/getting-started-card/utils/getting-started-grid/GettingStartedGrid.scss
```diff
@@ -44,4 +43,17 @@
       padding-right: var(--pf-v5-c-card--child--PaddingRight);
     }
   }
+
+  &__tooltip {
+    inset: auto auto 0px 5px !important;
```
pcbailey, line 48: Did you try increasing specificity before adding `!important `?

hstastna: Yes, I did.

hstastna: @pcbailey @metalice, this is the rare occasion. Tried anything to avoid using `!important` (of course!!!), but nothing worked as expected. The problem lies in how PF components' styles are implemented (not well for our UI designs) and how we constantly need to override them and adjust many things around.

Please, provide more specific suggestions or try something, too, otherwise this is not very helpful. Maybe you'll find some way I haven't found. I'd be happy to learn something or to get help with this.

Note that I must have used `inset` together with `!important` to override exactly this prop and its values to achieve the expected look. Any other position-like prop set by me here didn't work, it was overridden at the end.

Thanks.

metalice: @hstastna , u should better explore all the options before asking us to do it for u.
the issue u are facing caused by __tooltip-icon that is using padding instead of margin, simply change to
&__tooltip-icon {
    margin-left: var(--pf-global--spacer--sm);
  }
and you can remove the inset which is complelty not needed.
BTW u have a bug, when clicking on '?' the tooltip also open the expand section which is unwanted beahiver. thinking popover is more fit for '?' as its expecting a click while tooltip is simply hover. across our app when we have '?' its a popover with a click.

Before:
image

After:
image

hstastna: @metalice I am sorry I asked for help before exploring all the options (while there can be so many options for exploring for many days... the new unseen option to solve something can be found anytime). I'll never do such a mistake again. Next time I'll better be stuck for weeks before asking for help.

PR #1873, opened 2024-04-08
CNV-38061: Missing SR-IOV binding to VM spec

avivtur:
## 📝 Description

we try to bind sr-iov to the VM interface when added but should be sriov, which causing to not add the sriov interface properly

## 🎥 Demo

Before:

https://github.com/kubevirt-ui/kubevirt-plugin/assets/67270715/e68be0b1-a08d-4365-9aae-472f2cca8831

After:

https://github.com/kubevirt-ui/kubevirt-plugin/assets/67270715/41920848-833b-4924-86fb-5dbf2b136558

### DIFF
--- a/src/utils/components/NetworkInterfaceModal/utils/helpers.ts
+++ b/src/utils/components/NetworkInterfaceModal/utils/helpers.ts
@@ -96,7 +96,7 @@ export const createInterface = (
   interfaceType: string,
 ): V1Interface => {
   return {
-    [interfaceType.toLowerCase()]: {},
+    [interfaceType]: {},
     macAddress: interfaceMACAddress,
     model: interfaceModel,
     name: nicName,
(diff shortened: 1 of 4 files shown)


src/utils/components/NetworkInterfaceModal/utils/helpers.ts
```diff
@@ -96,7 +96,7 @@ export const createInterface = (
   interfaceType: string,
 ): V1Interface => {
   return {
-    [interfaceType.toLowerCase()]: {},
+    [interfaceType.replace('-', '').toLowerCase()]: {},
```
ormergi, line 99: How about defining a mapping between the UI interface Type name (e.g.: `SR-IOV`) and Kubevirt interface API option (e.g: `sriov`)?
It"ll help avoid issues like this PR fixes, also easier to extend in case new type of interface is introduced.

avivtur: Hi @ormergi 
You're right, we already have a mapping but it wasn't used properly, now it should work as you suggested

PR #1878, opened 2024-04-10
CNV-40059: Align VMs and Templates lists empty pages with the new format

hstastna:
## 📝 Description

_This PR is one of the PRs related to:_
https://issues.redhat.com/browse/CNV-40059

_Design doc:_
https://docs.google.com/document/d/1q1J7QJPbklWNV02hqBIIi6uNwHKaipUFZLA3RJF5A9E

Align VMs and Templates lists empty pages with the new format for more consistent look across the UI, add the appropriate icons and text, links, remove unnecessary create buttons, filtering and pagination. Also the link to get to know more about creating VMs from Templates from the doc was updated, in this PR. I've made sure that the lists look as expected in case of not being empty, too.

Additionally, in case when there are item in the list but after filtering we get zero results, prevent displaying unnecessary pagination showing zero results and pages - for VMs list.

## 🎥 Screenshots
**Before:**
VMs list empty page:
vm_before
VM Templates list empty page:
vm_templates_before
VMs list page with zero results after filtering:
vv_before

**After:**
VMs list empty page:
vm_after
VM Templates list empty page:
templates_afterr
VMs list page with zero results after filtering:
vv_after


src/views/templates/list/VirtualMachineTemplatesList.tsx
```diff
@@ -46,36 +48,38 @@ const VirtualMachineTemplatesList: FC<VirtualMachineTemplatesListProps> = ({ nam
...
-                id: modelToRef(TemplateModel),
-                selectedColumns: new Set(activeColumns?.map((col) => col?.id)),
-                type: t('Template'),
-              }}
-              data={data}
-              loaded={templatesLoaded && loadedColumns}
-              onFilterChange={onFilterChange}
-              rowFilters={filters}
-            />
+            {!isEmpty(templates) && (
```
metalice, line 62: why 2 isEmpty and not 1 over the stack component?

hstastna: because it gets us a bit different result

src/views/virtualmachines/list/VirtualMachinesList.tsx
```diff
@@ -153,46 +153,48 @@ const VirtualMachinesList: FC<VirtualMachinesListProps> = ({ kind, namespace })
...
-                page: 1,
-                startIndex: 0,
-              }));
-            }}
-            data={unfilteredData}
-            loaded={loaded}
-            rowFilters={filters}
-            searchFilters={searchFilters}
-          />
           {!isEmpty(vms) && (
```
metalice, line 156: istead of doint !isEmpty u can take the code from line 200 
{loaded && isEmpty(data) && (
          <VirtualMachineEmptyState catalogURL={catalogURL} namespace={namespace} />
        )}
and do a 
if (!loaded) return <Loader />
if (loaded && isEmpty(data)) {
      return    <VirtualMachineEmptyState catalogURL={catalogURL} namespace={namespace} />
        }
above the return of this component.

then we can get rid of all those && checks

metalice: same goes to templates

src/views/templates/list/VirtualMachineTemplatesList.tsx
hstastna: @metalice I looked closer at this and realized you are talking about "2 isEmpty" conditions, but I can see only one in this file:
`templatesLoaded && isEmpty(filteredData)`
So now I got confused about what you really meant by your comment. However if you meant that I've added more of "!isEmpty(templates)" here, your comment would make much more sense and I could answer like this:
There are more conditions like this because we want to render different components in different situations. In the 2nd StackItem under the ListPageBody, there are more components: ListPageFilter, VirtualMachineTemplatesEmptyState, VirtualizedTable. Each of them needs to be rendered in a slightly different conditions:
- ListPageFilter makes sense to be rendered if there are some templates, some unfiltered data
- VirtualMachineTemplatesEmptyState - for this component, a kinda different situation occurs: we want to render this if there are no templates or .. or if there are some but we get zero results after filtering
- VirtualizedTable - we want to let this render always, it takes care by itself about itself if there is no data, no templates
Hence I don't think we can optimize this. Or can we? I can't see it. Each mentioned component is kinda different (different circumstances when we want to render it). There is some intersection but I think not enough to make it better.

src/views/virtualmachines/list/VirtualMachinesList.tsx
hstastna: I don't know. I can see here the same pattern and same situation as for templates. Hence I think we cannot get rid of those checks. Only if we decided to render those components in different situations than we render now. Or maybe I just don't understand what you've suggested?

src/views/templates/list/VirtualMachineTemplatesList.tsx
metalice: I meant u have one is empty on line 52 and another on 56 u can use only one

hstastna: I can see. Yes, I have them there. And if you look closer, you will find one appears under `ListPageHeader` while the 2nd one appears under `ListPageBody`. I cannot unify them simply. We want to render both `ListPageHeader` and `ListPageBody` no matter if those child components should be rendered or not. Of course we could code it a different way to achieve the same result but I've found only much worse possibilities than the actual one, which is nicely readable, IMO, and hence easily changeable in the future (for example if it was decided to display components in different circumstances comparing to now).

metalice: I looked on the code even deeper and you can make it much more clear.
just add to VirtualMachineTemplatesEmptyState a ` <ListPageHeader title={t('VirtualMachine Templates')} />`
and in template list just add
```
if (templatesLoaded && isEmpty(filteredData)) {
    return <VirtualMachineTemplatesEmptyState namespace={namespace} />;
  }
```
before the return. same thing can be applied to all empty states u are doing. i will be much more clear and could avoid all those && in return statements.

hstastna: Now I better understand what you were suggesting. Thanks! Will update the PR asap..

hstastna: Hi @metalice, I've taken another look and now it's not quite clear to me again. Or.. it is but it looks kinda problematic to me. I understand it would be much nicer to get rid of those conditions. Trust me I don't like it too. But I haven't found any nicer solution to achieve the expected behavior (yet?) while keeping some reasonable readability of the code.

The suggestion you've provided feels to me that it's missing something, or it simply doesn't take some things into account: for example that we need to distinguish between filtered vs unfiltered data. This matters because we want to display various components in 3 different situations:
- always
- only when there is some (unfiltered) data
- only when there is some **filtered** data (subset of unfiltered data).

If I did what you suggest, we could end up in situation when we'd missing some components in the UI in some situations. For example if there was some unfiltered data but zero results after filtering, we'd see `VirtualMachineTemplatesCreateButton` there displayed always, if I'd remove that `!isEmpty(templates)` from there. and that's not what's considered as the expected behavior (or the current behavior). Similar for other components in this file. I think we'd make it just more complicated and possibly also buggy, comparing to the actual expected behavior if we'd try to get rid of all those conditions this way. So I don't know. We will need to figure out something a bit different to improve this code or to leave it as it is. Or to propose a different expected behavior.

metalice: the only case to show this empty state that u are creating is when data is empty, u need to ignore the filteredData, for filteredData the emptystate of table should handle. this components u are creating is for no data in ns/all-ns at all.

hstastna: I must not agree. We cannot ignore filtered vs unfiltered data. Why? Because ignoring this difference leads to hiding filtering component even in cases when it should not be hidden. Imagine the scenario there is some data and you want to filter the data. You will add some string for example to filter the data and if result is zero, you will end up in the UI where filtering component suddenly disappears and you have no chance to edit what you are looking for or to clear filtering. And that is totally incorrect behavior.

metalice: I'm guessing u don't understand so I'll try again. The component u are creating the empty state is only for situation of no data in ns or all ns scope. In case that data is not empty it won't show. In casedata is NOT empty but filtered data is empty the prop empty state component on the list is handling it. U shouldn't remove the prop. Test it first.

metalice: BTW I made exactly that for network plugin. U can have a look here  https://github.com/openshift/networking-console-plugin/pull/38

hstastna: Seems that this information was missing: that expected result has changed. Anyway I must make sure about it. Thank you for pointing this out.

hstastna: Hi @metalice thank you for your help. So I've tried to implement changes you've suggested, checked your PR, and here is the buggy result I was talking about earlier: in case of zero results when filtering, we are no longer able to display the original list of templates, nor clear the filter we wanted to apply because we will be missing (not only) filtering component (because of missing conditions comparing to my original/actual solution):

https://github.com/kubevirt-ui/kubevirt-plugin/assets/13417815/e2ae1a7f-3877-49bb-b28d-503ded8fffe1

And that's pretty bad, IMHO. Hence we do need to distinguish between the filtered and unfiltered data. Hence I insist on my solution until we find a better and working one.

metalice: why did the expected result change? the expected result is a unified component for no data situation.

hstastna: I don't know, I am not aware of any changes of the expected result. It just seems to me from your words. Anyway, the suggested changes in the code lead to an ugly buggy scenario (see the video in the previous comment) I wanted to avoid already from the beginning. No matter how the clear expected result is formulated, we should not implement such buggy scenarios to the UI. I was trying to fix that, of course, and what I've ended up in was my original solution with all those "!isEmpty" conditions, unfortunately.

metalice: > Hi @metalice thank you for your help. So I've tried to implement changes you've suggested, checked your PR, and here is the buggy result I was talking about earlier: in case of zero results when filtering, we are no longer able to display the original list of templates, nor clear the filter we wanted to apply because we will be missing (not only) filtering component (because of missing conditions comparing to my original/actual solution):
> 
>  buggy_if_no_results_found_when_filter.mp4 
> And that's pretty bad, IMHO. Hence we do need to distinguish between the filtered and unfiltered data. Hence I insist on my solution until we find a better and working one.

What is the condition to render empty list component u created? its should be on data and not filteredData ...

metalice: so u could adjust accordingly as we speaking here for a lot of comments already on filter vs data. There is no need to simply copy a code; you need to adjust it if needed. Btw you also have this example https://github.com/kubevirt-ui/kubevirt-plugin/pull/1878#discussion_r1560836123

hstastna: Of course there is no need to simply copy a code. However some details are important to talk about them precisely and directly, otherwise misunderstandings and other problems might occur. Anyway, no worries, I'm sure we'll figure out the best solution asap.

hstastna:
@metalice PTAL about the changes I've done (a separate commit just for review purpose, later I can squash, of course). Thanks!

src/views/virtualmachines/list/VirtualMachinesList.tsx
```diff
@@ -178,27 +182,28 @@ const VirtualMachinesList: FC<VirtualMachinesListProps> = ({ kind, namespace })
...
+            className="list-managment-group__pagination"
+            isLastFullPageShown
+            itemCount={data?.length}
+            page={pagination?.page}
+            perPage={pagination?.perPage}
+            perPageOptions={paginationDefaultValues}
+          />
         </div>
-        {loaded && isEmpty(data) && (
+        {/* {loaded && isEmpty(data) && (
```
metalice, line 200: please remove

PR #1879, opened 2024-04-11
UI settings should not be stored in default namespace

avivtur:
## 📝 Description

Moving UI setting ConfigMap to `openshift-cnv` namespace.
In case of CM existing on default namespace, we copy it to openshift-cnv, but it will NOT be deleted on default namespace. (should remove this mechanism in a future version).

## 🎥 Demo

After:
user-settings-cm-openshift-cnv


src/utils/hooks/useKubevirtUserSettings/utils/utils.ts
```diff
@@ -29,3 +41,41 @@ export const patchUserConfigMap = async (
     model: ConfigMapModel,
     resource: userConfigMap,
   });
+
+const createSCData = (data: IoK8sApiCoreV1ConfigMap['data']) => ({
```
metalice, line 45: please pick a better name

src/utils/hooks/useKubevirtUserSettings/utils/utils.ts
```diff
@@ -29,3 +41,41 @@ export const patchUserConfigMap = async (
...
+  data,
+  metadata: {
+    name: KUBEVIRT_USER_SETTINGS_CONFIG_MAP_NAME,
+    namespace: OPENSHIFT_CNV,
+  },
+});
+
+const createBindings = async () => {
```
metalice, line 53: between what?

metalice: naming :)

src/utils/hooks/useKubevirtUserSettings/utils/utils.ts
```diff
@@ -29,3 +41,41 @@ export const patchUserConfigMap = async (
...
+export const moveUserSettingsToOpenshiftCNV = async (userSCDefaultNS: IoK8sApiCoreV1ConfigMap) => {
+  await k8sCreate<IoK8sApiCoreV1ConfigMap>({
+    data: createSCData(userSCDefaultNS?.data),
+    model: ConfigMapModel,
+  });
+
+  createBindings();
+};
+
+export const createResources = async (userName: string) => {
```
metalice, line 74: naming

src/utils/hooks/useKubevirtUserSettings/utils/utils.ts
```diff
@@ -29,3 +41,41 @@ export const patchUserConfigMap = async (
...
+    model: RoleModel,
+  });
+
+  await k8sCreate<IoK8sApiRbacV1RoleBinding>({
+    data: userSettingsRoleBinding,
+    model: RoleBindingModel,
+  });
+};
+
+export const moveUserSettingsToOpenshiftCNV = async (userSCDefaultNS: IoK8sApiCoreV1ConfigMap) => {
```
metalice, line 65: this is a good name! :)

PR #1884, opened 2024-04-18
CNV-40598: Adding default valeu to storage checkup timeout

metalice:
## 📝 Description

Adding default value of 10m to storage checkup from

## 🎥 Demo

After:
image

Before:
image

### DIFF
--- a/src/views/checkups/storage/components/form/CheckupsStorageForm.tsx
+++ b/src/views/checkups/storage/components/form/CheckupsStorageForm.tsx
@@ -23,7 +23,7 @@ import './checkups-storage-form.scss';
 const CheckupsStorageForm = () => {
   const { t } = useKubevirtTranslation();
   const [name, setName] = useState<string>(generatePrettyName('kubevirt-storage-checkup'));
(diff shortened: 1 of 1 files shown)


src/views/checkups/storage/components/form/CheckupsStorageForm.tsx
```diff
@@ -23,7 +23,7 @@ import './checkups-storage-form.scss';
 const CheckupsStorageForm = () => {
   const { t } = useKubevirtTranslation();
   const [name, setName] = useState<string>(generatePrettyName('kubevirt-storage-checkup'));
-  const [timeOut, setTimeOut] = useState<string>();
+  const [timeOut, setTimeOut] = useState<string>('10');
```
avivtur, line 26: please add const with meaningful name like TEN_MINUTES

metalice: No need this is a single use

avivtur (approved):

PR #1893, opened 2024-04-21
CNV-38284: UI settings should not be stored in default namespace

avivtur:
## 📝 Description

We are moving the UI settings ConfigMap to `openshift-cnv` namespace.
If an existing CM is on the default namespace, we ignore it and create a new one to openshift-cnv namespace.

## 🎥 Demo

After:
user-settings-cm-openshift-cnv


upalatucci (approved):

metalice:
@avivtur any update here?

PR #1898, opened 2024-04-25
CNV-32699: Customize instance type part 1

metalice:
## 📝 Description

This part 1 includes the infrastructure for signals and adding and editing details tabs functions/files/naming in customized IT flow.
next pr's will include the other tabs.

## 🎥 Demo

Before:
image

After:
image


src/views/catalog/CustomizeInstanceType/tabs/configuration/utils/tabs/CustomizeInstanceTypeDetailsTab.tsx
```diff
@@ -0,0 +1,180 @@
...
+import { asAccessReview, getAnnotation, getName } from '@kubevirt-utils/resources/shared';
+import { DESCRIPTION_ANNOTATION, getDevices, getHostname } from '@kubevirt-utils/resources/vm';
+import { updateCustomizeInstanceType, vmSignal } from '@kubevirt-utils/store/customizeInstanceType';
+import { K8sVerb, useAccessReview } from '@openshift-console/dynamic-plugin-sdk';
+import { DescriptionList, Grid, GridItem, Switch, Title } from '@patternfly/react-core';
+import DetailsSectionBoot from '@virtualmachines/details/tabs/configuration/details/components/DetailsSectionBoot';
+import DetailsSectionHardware from '@virtualmachines/details/tabs/configuration/details/components/DetailsSectionHardware';
+
+const CustomizeInstanceTypeDetailsTab = () => {
+  //   const store = useInstanceTypeVMStore();
```
metalice, line 28: i will adjust in next pr's

src/utils/store/customizeInstanceType.ts
```diff
@@ -0,0 +1,38 @@
...
+import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import { isEmpty } from '@kubevirt-utils/utils/utils';
+import { effect, signal } from '@preact/signals-react';
+
+import {
+  getCustomizeInstanceTypeSessionStorage,
+  saveCustomizeInstanceTypeSessionStorage,
+} from './customizeInstanceType/utils/utils';
+
+export const vmSignal = signal<V1VirtualMachine>(null);
```
pcbailey, line 12: In the comments of the video you recommended we watch I saw warnings about using signals because it patches React internals. I saw one comment that said that a second blog post from Preact noted that they are no longer doing that, but I didn't see that when I read the second blog post. Can we be absolutely sure that signals won't cause unexpected problems down the line?

src/views/catalog/CreateFromInstanceTypes/components/CreateVMFooter/CreateVMFooter.tsx
```diff
@@ -118,6 +119,13 @@ const CreateVMFooter: FC = () => {
           autoUpdateEnabled,
         ),
       );
+      vmSignal.value = generateVM(
```
pcbailey, line 122: Toward the end of the video the video creator said that the recommended way to use signals in components is to inject them as props in order to make testing easier. Is there a reason why you didn't do that here and in the other place the signal is used below?

metalice: What do u mean inject them as prop? 
1. for testing purpose, while we don't do unit testing, there isn't any difference as e2e will work out of the box.
2. this line is we we load the the initial state to vmSignal

src/utils/store/customizeInstanceType.ts
metalice: This is a great question! 

You are correct that older versions changed the internal. While this is fine on current React versions as it's thoroughly tested and works with them (for regular React setup), it may cause issues with the next React releases.
But as you can see in signal-react, releases https://github.com/preactjs/signals/releases/tag/%40preact%2Fsignals-react%402.0.0, they are removing the internal changes because they had issues that were hard to debug, and values weren't updated (mostly in next. js—from reading posts). By the way, we are using 2.0.1. 

Anyway, if any bugs surface while using signals, there is only one single file for the store that is consumed, it will be easy to replace with any other store mechanism if we encounter any problems. It's very modular, and there is a separation between logic and rendering.

src/views/catalog/CreateFromInstanceTypes/components/CreateVMFooter/CreateVMFooter.tsx
metalice: Seems like the only advantage is unit testing. I prefer much on direct consuming as a much bigger advantage of avoiding prop drilling abusing. Which is the biggest advantage of store. In any case if in the future unit testing will be added one mock file can solve store issue to all related test.

PR #1901, opened 2024-04-29
CNV-39556: Change pending changes message for hot-unplug VF

hstastna:
## 📝 Description

**Fixes:**
https://issues.redhat.com/browse/CNV-39556

We have no support to hot-unplug VF, hence we need to update the message accordingly to:
_"The following areas have pending changes that will be applied when this VirtualMachine is restarted."_
as the VM live migration will have no effect on it.

## 🎥 Screenshots
**Before:**
pending_before

**After:**
pending_after


hstastna:

PR #1904, opened 2024-04-29
CNV-39941: Load balancer bonding alert

upalatucci:
## 📝 Description

If there is no IP, it's because the load balancer is going to be bonded but it's not ready yet. 

Waiting for UX/UI approve 

**New**

Screenshot from 2024-04-30 14-55-34
image

**Old**

Screenshot from 2024-04-29 16-46-54


yfrimanm:
@upalatucci Re the alert, I don’t think we need an alert for this. We just need to reflect the status. Let’s just make the dropdown input much shorter and add “pending” next to it. Once it’s ok we can switch it to a green check mark to indicate it is ready.
Thank you.

metalice (approved):

PR #1922, opened 2024-05-07
CNV-41522: Fix customize IT

metalice:
## 📝 Description

Fix the flow of customize IT
 

Edit disk
Edit nic
Cancel redirect
Adding SSH
Boot managment

## 🎥 Demo

### DIFF
--- a/src/utils/store/customizeInstanceType.ts
+++ b/src/utils/store/customizeInstanceType.ts
@@ -37,12 +37,14 @@ export const updateCustomizeInstanceType: UpdateCustomizeInstanceType = (
   updateValues,
 ): V1VirtualMachine => {
   let vm = vmSignal.value;
+
   updateValues.forEach(({ data, merge = false, path }) => {
     //replace complete vm obj
     if (isEmpty(path)) {
       vm = data;
       return;
     }
+
     vm = produce(vm, (vmDraft) => {
       const pathParts = path.split('.');
       let obj = vmDraft;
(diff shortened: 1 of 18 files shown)


src/views/catalog/CustomizeInstanceType/components/CustomizeITVMFooter.tsx
```diff
@@ -28,6 +28,7 @@ import './CustomizeITVMFooter.scss';
 const CustomizeITVMFooter: FC = () => {
   const { t } = useKubevirtTranslation();
   const navigate = useNavigate();
+  const [activeNamespace] = useActiveNamespace();
```
upalatucci, line 31: Can this page be rendered on 'all-namespaces' ? In this case activeNamespace will be `ALL_NAMESPACES_SESSION_KEY`

src/utils/store/customizeInstanceType.ts
```diff
@@ -37,12 +37,14 @@ export const updateCustomizeInstanceType: UpdateCustomizeInstanceType = (
   updateValues,
 ): V1VirtualMachine => {
   let vm = vmSignal.value;
+
```
pcbailey, line 40: 🎉

src/views/catalog/wizard/tabs/network/components/list/NetworkInterfaceActions.tsx
```diff
@@ -8,23 +8,27 @@ import TabModal from '@kubevirt-utils/components/TabModal/TabModal';
...
   nicName,
   nicPresentation,
+  onUpdateVM,
 }) => {
   const { t } = useKubevirtTranslation();
-  const { updateVM, vm } = useWizardVMContext();
+  const { updateVM, vm: vmContext } = useWizardVMContext();
   const { createModal } = useModal();
-
+  const vm = vmSignal.value || vmContext;
```
pcbailey, line 30: Is this a temporary solution or will we always use context as a fallback? If it's not temporary, what's the value in having two store solutions?

metalice: in the end i would like to remove all use of context and stay only with signals so its temp until a wider refactor

src/views/catalog/CustomizeInstanceType/components/CustomizeITVMFooter.tsx
metalice: should be fine, as this is only for the cancel button, but a wider check and adjustment will be followed so changes will be made again

PR #1939, opened 2024-05-13
CNV-40335: fix search preferences and isntancetype by label

upalatucci:
## 📝 Description

When clicking on labels in the resource details page, the page gets redirected to the search page with the resource kind and label selected. Just like in photo. In those cases, console load the resource list page declared in the console-extension file with some props. 

- make sure to hide tabs (Cluster / User ) in the preference and instancetype page and just show the list when the list is rendered in the search page (using the useIsSearchPage hook) 
- make sure the component use the props passed from search like fieldSelector, selector and nameFilter to filter the resources in the list

We should do the same with the other resources that we support 

Screenshot from 2024-05-13 14-45-19
Screenshot from 2024-05-13 14-45-12


metalice:
lets talk about it, there is a significant design change here

PR #1943, opened 2024-05-14
CNV-39558: Add tooltip for disabled SSH over NodePort option

hstastna:
## 📝 Description

**Fixes:**
https://issues.redhat.com/browse/CNV-39558

Add tooltip to provide more info to the user in case when "SSH over NodePort" service is disabled, in VM's SSH settings.
The final text: _NodePort service is disabled. Ask your cluster admin to enable it in cluster settings._

## 🎥 Screenshots
**Before:**
node_before

**After:**
node_after

### DIFF
--- a/src/utils/components/SSHAccess/components/SSHServiceSelect.tsx
+++ b/src/utils/components/SSHAccess/components/SSHServiceSelect.tsx
@@ -65,8 +65,15 @@ const SSHServiceSelect: FC<SSHServiceSelectProps> = ({
             'Opens a specific port on all Nodes in the cluster. If the Node is publicly accessible, any traffic sent to this port is forwarded to the Service.',
           )}
           id={SERVICE_TYPES.NODE_PORT}
-          isDisabled={!nodePortEnabled}
+          isAriaDisabled={!nodePortEnabled}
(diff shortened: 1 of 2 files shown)


hstastna:

src/utils/components/SSHAccess/components/SSHServiceSelect.tsx
```diff
@@ -65,8 +65,15 @@ const SSHServiceSelect: FC<SSHServiceSelectProps> = ({
             'Opens a specific port on all Nodes in the cluster. If the Node is publicly accessible, any traffic sent to this port is forwarded to the Service.',
           )}
           id={SERVICE_TYPES.NODE_PORT}
-          isDisabled={!nodePortEnabled}
+          isAriaDisabled={!nodePortEnabled}
```
metalice, line 68: why changing this?

hstastna: Because with this change, we achieve the expected behavior and look. Using `isDisabled` does not allow us to display the tooltip while option being disabled, unfortunately.

metalice (approved):

PR #1945, opened 2024-05-15
CNV-41633: Fix VM list EmptyState on filtering with pod proxy

upalatucci:
## 📝 Description

When cluster has proxy pod enabled, the `unfilteredData` is empty as the proxy pod return empty array from the fetch even if the user just filtered data. 

In that case, just watch the selected filters. If we are using the pod for fetching data and the user has selected a filter, don't show the empty state. 

Show the empty state if :

unfiltered data is empty if the cluster is not using the pod
if we are using the pod and unfiltered data is empty, check if the user has selected some filters from the query params (no other way to do it )

### DIFF
--- a/src/views/virtualmachines/list/VirtualMachinesList.tsx
+++ b/src/views/virtualmachines/list/VirtualMachinesList.tsx
@@ -40,6 +40,7 @@ import { useVMListFilters } from '../utils';
 import VirtualMachineEmptyState from './components/VirtualMachineEmptyState/VirtualMachineEmptyState';
 import VirtualMachineRow from './components/VirtualMachineRow/VirtualMachineRow';
 import VirtualMachinesCreateButton from './components/VirtualMachinesCreateButton/VirtualMachinesCreateButton';
+import useSelectedFilters from './hooks/useSelectedFilters';
 import useVirtualMachineColumns from './hooks/useVirtualMachineColumns';
 
 import '@kubevirt-utils/styles/list-managment-group.scss';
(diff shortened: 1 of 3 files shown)


src/views/virtualmachines/list/hooks/useSelectedFilters.ts
```diff
@@ -0,0 +1,31 @@
...
+    );
+
+    selectedFilters.push(
+      ...(searchFilters || []).map((filter) =>
+        searchParams.get(filter.type) ? filter.type : null,
+      ),
+    );
+
+    searchParams.get('name') && selectedFilters.push('name');
+    searchParams.get('labels') && selectedFilters.push('labels');
```
avivtur, line 25: Can you please add consts for name and labels?

avivtur (commented):
looks very good :+1:  just a small comment, and also maybe we want to locate this hook in a general location so other pages use this hook too?

src/views/virtualmachines/list/hooks/useSelectedFilters.ts
pcbailey: +1

src/views/virtualmachines/list/VirtualMachinesList.tsx
```diff
@@ -145,7 +148,9 @@ const VirtualMachinesList: FC<VirtualMachinesListProps> = ({ kind, namespace })
     !loadingFeatureProxy &&
     loadedColumns;
 
-  if (loaded && isEmpty(unfilteredData)) {
+  const vmsFilteredWithProxy = isProxyPodAlive && selectedFilters.length > 0;
+
+  if (loaded && isEmpty(unfilteredData) && !vmsFilteredWithProxy) {
```
pcbailey, line 153: Suggestion: Create a variable with a descriptive name for these conditions to make it more readable.

src/views/virtualmachines/list/hooks/useSelectedFilters.ts
upalatucci: fixed

src/views/virtualmachines/list/VirtualMachinesList.tsx
upalatucci: added a const with a readable name

avivtur (approved):

PR #1963, opened 2024-05-21
CNV-41205: Add checkboxes for each disk to delete

upalatucci:
## 📝 Description

Show a checkbox list for each volume that we can delete.
This modal removes the VM owner reference to all volumes and removes the dataVolumes from the dataVolumeTemplates array from the VM. 

Nit: fix the grace period question mark space 

## 🎥 Demo

Screenshot from 2024-05-21 14-18-13

### DIFF
--- a/src/views/virtualmachines/actions/components/DeleteVMModal/DeleteVMModal.tsx
+++ b/src/views/virtualmachines/actions/components/DeleteVMModal/DeleteVMModal.tsx
@@ -1,23 +1,25 @@
 import React, { FC, useState } from 'react';
 import { useNavigate } from 'react-router-dom-v5-compat';
 
-import { PersistentVolumeClaimModel } from '@kubevirt-ui/kubevirt-api/console';
 import DataVolumeModel from '@kubevirt-ui/kubevirt-api/console/models/DataVolumeModel';
 import VirtualMachineModel, {
   VirtualMachineModelRef,
 } from '@kubevirt-ui/kubevirt-api/console/models/VirtualMachineModel';
+import { V1beta1DataVolume } from '@kubevirt-ui/kubevirt-api/containerized-data-importer/models';
+import { IoK8sApiCoreV1PersistentVolumeClaim } from '@kubevirt-ui/kubevirt-api/kubernetes';
 import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
 import ConfirmActionMessage from '@kubevirt-utils/components/ConfirmActionMessage/ConfirmActionMessage';
 import { GracePeriodInput } from '@kubevirt-utils/components/GracePeriodInput/GracePeriodInput';
 import TabModal from '@kubevirt-utils/components/TabModal/TabModal';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
 import { useLastNamespacePath } from '@kubevirt-utils/hooks/useLastNamespacePath';
-import { buildOwnerReference, compareOwnerReferences } from '@kubevirt-utils/resources/shared';
-import { k8sDelete, k8sPatch } from '@openshift-console/dynamic-plugin-sdk';
+import { buildOwnerReference } from '@kubevirt-utils/resources/shared';
+import { k8sDelete } from '@openshift-console/dynamic-plugin-sdk';
 import { ButtonVariant, Stack, StackItem } from '@patternfly/react-core';
 
 import DeleteOwnedResourcesMessage from './components/DeleteOwnedResourcesMessage';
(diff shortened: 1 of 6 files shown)


src/views/virtualmachines/actions/components/DeleteVMModal/DeleteVMModal.tsx
```diff
@@ -29,65 +31,27 @@ type DeleteVMModalProps = {
...
-        });
-      });
+    await removeDataVolumeTemplatesToVM(
+      vm,
+      volumesToSave.filter((volume) => volume.kind === DataVolumeModel.kind) as V1beta1DataVolume[],
+    );
 
-      await Promise.allSettled([...pvcPromises, ...dvPromises]);
-    }
+    await Promise.allSettled(updateVolumeResources(volumesToSave, vmOwnerRef));
```
metalice, line 54: i dont thinks allSettled is needed anymore. no array here

src/views/virtualmachines/actions/components/DeleteVMModal/components/DeleteOwnedResourcesMessage.tsx
```diff
@@ -36,7 +41,7 @@ const DeleteOwnedResourcesMessage: FC<DeleteOwnedResourcesMessageProps> = ({
     );
   }
 
-  const pvcsWithNoDataVolumes = pvcs?.filter((pvc) => !findPVCOwner(pvc, dataVolumes));
+  const pvcsWithNoDataVolumes = pvcs?.filter((pvc) => !findPVCOwner(pvc, dataVolumes)) || [];
```
metalice, line 44: why filtering disk with datavolume? we have a GC for datavolumes once PVC is bound and ready, what happen if disks in error stage and we want to delete the VM and disks? the PVC still has DV and should be deleted as well

src/views/virtualmachines/actions/components/DeleteVMModal/utils/helpers.ts
```diff
@@ -38,3 +41,27 @@ export const findPVCOwner = (
...
+) => {
+  const dataVolumeTemplates = getDataVolumeTemplates(vm);
+
+  const dvIndexes = dataVolumesToSave
+    .map((dataVolume) =>
+      dataVolumeTemplates.findIndex((template) => getName(template) === getName(dataVolume)),
+    )
+    .filter((index) => index !== -1);
+
+  if (isEmpty(dvIndexes)) return Promise.resolve();
```
metalice, line 57: why promise resolve?

src/views/virtualmachines/actions/components/DeleteVMModal/components/DeleteVolumeCheckbox.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+const DeleteVolumeCheckbox: FC<DeleteVolumeCheckboxProps> = ({
+  resource,
+  setVolumesToSave,
+  volumesToSave,
+}) => {
+  const { t } = useKubevirtTranslation();
+  const resourceName = getName(resource);
+
+  const saveVolume = () => setVolumesToSave((prevVolumes) => prevVolumes.concat([resource]));
```
metalice, line 28: [...prev, resource]

src/views/virtualmachines/actions/components/DeleteVMModal/components/DeleteVolumeCheckbox.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  const deleteVolume = () =>
+    setVolumesToSave((prevVolumes) =>
+      prevVolumes.filter(
+        (volume) => !(volume.kind === resource.kind && getName(volume) === resourceName),
```
metalice, line 33: same as line 42, can be in a func

src/views/virtualmachines/actions/components/DeleteVMModal/components/DeleteOwnedResourcesMessage.tsx
upalatucci: This is just to show the right amount of disks. It would be confusing to show two disks with the same name one for the DV and another for the PVC. 

The real logic behind the delete disks logic is to delete the VM owner reference for the resources that the user doesn't want to delete. THe other resources will be deleted because of owner reference deletion policy

src/views/virtualmachines/actions/components/DeleteVMModal/utils/helpers.ts
upalatucci: fixed

metalice (approved):

PR #1964, opened 2024-05-22
CNV-42021: fix vm list for search page

upalatucci:
## 📝 Description

- Label filters not working for the list
- VM list appearance not adequate for the search page (name filter show up on the VM list and should appear only on the top of the search page, the VirtualMachine title should not show up) 

## 🎥 Demo

**Before**

Screenshot from 2024-05-22 10-01-20

**How other resource lists looks like in this page**

Screenshot from 2024-05-22 10-01-49

**After**

Screenshot from 2024-05-22 09-54-58

### DIFF
--- a/src/views/virtualmachines/list/VirtualMachinesList.tsx
+++ b/src/views/virtualmachines/list/VirtualMachinesList.tsx
@@ -23,6 +23,7 @@ import {
   paginationInitialState,
 } from '@kubevirt-utils/hooks/usePagination/utils/constants';
 import useSingleNodeCluster from '@kubevirt-utils/hooks/useSingleNodeCluster';
+import { ListPageProps } from '@kubevirt-utils/utils/types';
 import { isEmpty } from '@kubevirt-utils/utils/utils';
 import {
   K8sResourceCommon,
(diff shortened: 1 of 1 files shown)


src/views/virtualmachines/list/VirtualMachinesList.tsx
```diff
@@ -45,24 +46,32 @@ import useVirtualMachineColumns from './hooks/useVirtualMachineColumns';
...
+}) => {
   const { t } = useKubevirtTranslation();
   const catalogURL = `/k8s/ns/${namespace || DEFAULT_NAMESPACE}/catalog`;
   const { featureEnabled, loading: loadingFeatureProxy } = useFeatures(KUBEVIRT_APISERVER_PROXY);
   const isProxyPodAlive = useKubevirtDataPodHealth();
   const query = useQuery();
+
   const [vms, vmLoaded, loadError] = useKubevirtWatchResource<V1VirtualMachine[]>(
     {
+      fieldSelector,
```
metalice, line 68: You need to check first with the proxy. Its probably working before merging; u need to run the proxy locally.

upalatucci: In searchpage labels filtering is not working for VM list

PR #1967, opened 2024-05-22
CNV-40419: fix template undefined on permissions error

upalatucci:
## 📝 Description

In some cases the template drawer was crashing on non permission users


hstastna (commented):
The code looks good. Could you please, add some video to see how it works now? As I was not able to reproduce the bug, it is kinda harder to test if this really has helped to fix the problem. But the code looks reasonable. Thank you, Ugo!

upalatucci:
To reproduce the bug you have to use the test user. The admin user has all permissions. You need a user that does not have permissions

metalice (approved):

PR #1968, opened 2024-05-22
CNV-39940: UI does not show installed version

avivtur:
## 📝 Description

The UI assumes that the subscription that holds the version metadata has a unique name.
Since the name of the subscription CR is arbitrary, we now check if the subscription is from the package of `kubevirt-hyperconverged` or in some community upstream cases the package named `community-kubevirt-hyperconverged`.

## 🎥 Demo

Before:
version-missing

After:
version-missing-after

### DIFF
--- a/src/views/clusteroverview/utils/hooks/useKubevirtCSVDetails.ts
+++ b/src/views/clusteroverview/utils/hooks/useKubevirtCSVDetails.ts
@@ -9,7 +9,6 @@ import { isUpstream } from '@kubevirt-utils/utils/utils';
 import { useK8sWatchResource } from '@openshift-console/dynamic-plugin-sdk';
 
 import {
-  HCO_OPERATORHUB_NAME,
   KUBEVIRT_HYPERCONVERGED,
   OPENSHIFT_CNV,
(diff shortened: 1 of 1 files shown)


src/views/clusteroverview/utils/hooks/useKubevirtCSVDetails.ts
```diff
@@ -33,12 +32,19 @@ type UseKubevirtCSVDetails = {
...
+  const [subscriptions, loadedSubscription, loadSubscriptionError] = useK8sWatchResource<
+    SubscriptionKind[]
+  >({
+    groupVersionKind: SubscriptionModelGroupVersionKind,
+    isList: true,
+    namespace: isUpstream ? KUBEVIRT_HYPERCONVERGED : OPENSHIFT_CNV,
+  });
+
+  const subscription = useMemo(
+    () => subscriptions?.find((sub) => sub?.spec?.name === KUBEVIRT_HYPERCONVERGED),
```
orenc1, line 44: `community-kubevirt-hyperconverged` in case of `isUpstream == true`

hstastna (approved):

PR #1976, opened 2024-05-27
CNV-41205: fix sameVolume func

upalatucci:
## 📝 Description

small fix


src/views/virtualmachines/actions/components/DeleteVMModal/utils/helpers.ts
```diff
@@ -69,4 +69,4 @@ export const removeDataVolumeTemplatesToVM = (
 export const sameVolume = (
   volumeA: IoK8sApiCoreV1PersistentVolumeClaim | V1beta1DataVolume,
   volumeB: IoK8sApiCoreV1PersistentVolumeClaim | V1beta1DataVolume,
-) => volumeA.kind === volumeB.kind && getName(volumeA) === volumeB;
+) => volumeA.kind === volumeB.kind && getName(volumeA) === getName(volumeB);
```
avivtur, line 72: We should also compare namespaces here

avivtur: And can't we use isEqualObject function here? why is this function needed?

upalatucci: I would not compare them with isEqualObject as a small change can result in a false negative. Maybe one is an older version of the other or something like that

upalatucci: done

metalice (approved):

PR #1977, opened 2024-05-27
CNV-42157: Fix disk modal template

upalatucci:
## 📝 Description

- use VM parameter to create dataVolumes and PVC names. For each vms we'll have different datavolumes


src/utils/components/DiskModal/utils/helpers.ts
```diff
@@ -43,13 +41,6 @@ import { DYNAMIC } from '../components/utils/constants';
 import { getInitialStateDiskForm, mapSourceTypeToVolumeType } from './constants';
 import { DiskFormState, DiskModalProps, SourceTypes, VolumeTypes } from './types';
 
-const nameWithoutParameter = (name: string, defaultValue?) => {
```
metalice, line 46: why we used it in the first place?

upalatucci: Me and aviv have no idea.

metalice (approved):

PR #1981, opened 2024-05-27
CNV-40344: Make expanded sections indented as expected

hstastna:
## 📝 Description

**Fixes:**
https://issues.redhat.com/browse/CNV-40344

Our console folks broke our css by using `!important`, which is not a good practice, exactly here:
https://github.com/hstastna/console/blob/master/frontend/packages/console-shared/src/components/getting-started/GettingStartedExpandableGrid.scss#L58

Hence, to fix broken expandable sections in our UI, we must use `!important`, too. 

## 🎥 Screenshots
**Before:**
Sections in Overview settings not indented at all:
expand_before

**After:**
expand_after


hstastna:

metalice (approved):

PR #1997, opened 2024-06-04
CNV-41586: Missing functions for sysprep customizr IT flow

metalice:
## 📝 Description

Fix customize IT flow sysprep - missing functions

## 🎥 Demo


src/utils/store/customizeInstanceType/utils/utils.ts
```diff
@@ -19,3 +19,9 @@ export const getCustomizeInstanceTypeSessionStorage = (): V1VirtualMachine => {
   }
   return null;
 };
+
+export const mergeData = (seedData, appendData) => {
```
avivtur, line 23: Can you add types here?

PR #1998, opened 2024-06-04
CNV-41590: sysprep volume missing on creation of a VM with Sysprep

avivtur:
## 📝 Description

On create VM with IT flow (quick create) the sysprep was done after the VM was created, hence it wasn't added to the VM.

## 🎥 Demo

> Please add a video or an image of the behavior/changes

PR #2002, opened 2024-06-05
CNV-42087: Improve sysprep section when the answer file is added

avivtur:
## 🎥 Demo

Before:
sysprep-display-b4

After:

sysprep-display2-gutter


avivtur:
/hold 
Until #1998 is merged

metalice (approved):

PR #2049, opened 2024-06-20
CNV-43114: fix getBootDisk for instanceType vms

upalatucci:
## 📝 Description

vms created with instancetype do not have the rootdisk in the yaml but it actually exists.
So in that case just return the usual rootdisk object.

### DIFF
--- a/src/utils/resources/vm/utils/selectors.ts
+++ b/src/utils/resources/vm/utils/selectors.ts
@@ -12,6 +12,7 @@ import {
   V1VirtualMachineCondition,
 } from '@kubevirt-ui/kubevirt-api/kubevirt';
 import { DYNAMIC_CREDENTIALS_SUPPORT } from '@kubevirt-utils/components/DynamicSSHKeyInjection/constants/constants';
+import { ROOTDISK } from '@kubevirt-utils/constants/constants';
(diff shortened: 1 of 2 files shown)


src/utils/resources/vm/utils/selectors.ts
```diff
@@ -130,12 +131,20 @@ export const getAffinity = (vm: V1VirtualMachine) => vm?.spec?.template?.spec?.a
...
-  (getDisks(vm) || [])
+export const getBootDisk = (vm: V1VirtualMachine): V1Disk => {
+  const isInstanceTypeVM = Boolean(getInstanceTypeMatcher(vm));
+
+  const rootDiskVolume = (getVolumes(vm) || []).find((volume) => volume.name === ROOTDISK);
+
+  const defaultRootDisk =
+    isInstanceTypeVM && Boolean(rootDiskVolume) ? { name: ROOTDISK } : getDisks(vm)?.[0];
+
+  return (getDisks(vm) || [])
```
metalice, line 142: u are using getDisks(vm) 3 times, save it to a const once

upalatucci: Solved

PR #2054, opened 2024-06-24
CNV-43245: use runStrategy if defined

upalatucci:
## 📝 Description

Documentation: https://docs.openshift.com/container-platform/4.11/virt/virtual_machines/virt-create-vms.html#virt-about-runstrategies-vms_virt-create-vms

`spec.running` and `spec.runStrategy` are mutually exclusive. `spec.runStrategy` specified by user and we add `spec.running`

Solution: do not set `spec.running` if `spec.runStrategy` is already defined by the template. Show in the 'start vm' checkbox label the runStrategy if defined. runStrategy `always` or `runonfailure` are special cases of running: true so the checkbox will be checked. 

If the user changes the checkbox status, remove `spec.runStrategy` and use `spec.running`. 

Note: I removed the `startVM` from all the contexts as this is something that is defined in the VM itself. So instead of having another state in the context, I decided to change the VM property `spec.running` directly

## 🎥 Demo

**Catalog Drawer**
https://github.com/kubevirt-ui/kubevirt-plugin/assets/29160323/f3ab2080-0069-4139-ba71-af2d040a0927

**Wizard**
https://github.com/kubevirt-ui/kubevirt-plugin/assets/29160323/8ca13d13-60fc-4e2a-af95-6c4fef036989

### DIFF
--- a/src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/hooks/useCreateDrawerForm.tsx
+++ b/src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/hooks/useCreateDrawerForm.tsx
@@ -60,6 +60,7 @@ const useCreateDrawerForm = (
     cdFile,
     diskFile,
     isBootSourceAvailable,
+    setVM,
     sshDetails,
     storageClassName,
     storageClassRequired,
@@ -71,7 +72,6 @@ const useCreateDrawerForm = (
 
   const { nameField, onVMNameChange } = useCreateVMName();
 
-  const [startVM, setStartVM] = useState(true);
   const [isQuickCreating, setIsQuickCreating] = useState(false);
   const [isCustomizing, setIsCustomizing] = useState(false);
   const [createError, setCreateError] = useState(undefined);
@@ -130,7 +130,6 @@ const useCreateDrawerForm = (
           isDisabledGuestSystemLogs,
           name: nameField,
           namespace,
-          startVM,
           subscriptionData,
         },
         template: templateToProcess,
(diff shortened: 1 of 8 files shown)


src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/hooks/useCreateDrawerForm.tsx
```diff
@@ -235,12 +234,19 @@ const useCreateDrawerForm = (
...
-    });
+    setVM(
+      produce(vm, (draftVM) => {
+        delete draftVM.spec.runStrategy;
+        draftVM.spec.running = checked;
+      }),
+    );
   };
 
+  useEffect(() => {
```
metalice, line 245: u removed the startVM state , but added this 1 time run hook with disabled liners, I don't like this practice.
I don't think this is needed at all, u trying to achieve that if no runStrategy mark  the start VM checkbox in the end?

what will happen if you put in line 269 startVM: vm?.spec?.running && !VM?.spec?.runStrayrgy or some combination of this? can this hook be removed?

upalatucci: okay. I was trying to set running true at the beginning just like before with the useState

upalatucci: So i removed the useEffect. This means that the checkbox will reflect the template configuration. If the template has a VM with running: false, the checkbox at the beginning will be false. Is it okay?

metalice (approved):

PR #2059, opened 2024-06-26
CNV-43176: Allow user to select namespace in Add volume modal

pcbailey:
## 📝 Description

These changes in this PR:

- allow the user to select the project the bootable volume will be created in directly from the Add bootable volume modal
- change the default namespace in the bootable volume list to All projects
- fix a bug where the DataSource wasn't deleted when the DataVolume is deleted

jira: https://issues.redhat.com/browse/CNV-43176

### Before

Selection_259

### After

select-bootable-volume-destination-namespace--AFTER--2024-07-12 10-54


src/utils/components/AddBootableVolumeModal/AddBootableVolumeModal.tsx
```diff
@@ -52,16 +50,17 @@ const AddBootableVolumeModal: FC<AddBootableVolumeModalProps> = ({
 }) => {
   const { t } = useKubevirtTranslation();
   const [activeNamespace] = useActiveNamespace();
-  const selectedNamespace =
```
metalice, line 55: why did. removed it? if i remember correctly we added it for a reason, but I cant remember the reason :)

src/utils/components/ProjectDropdown/utils/utils.ts
```diff
@@ -0,0 +1,29 @@
...
+      return {
+        children: name,
+        groupVersionKind: modelToGroupVersionKind(ProjectModel),
+        value: name,
+      };
+    });
+
+  const allProjects = includeAllProjects
+    ? [
+        {
```
metalice, line 20: why do we need all projects?

pcbailey: I included it because this is meant to be a general use component. You can decide whether to show it or not using a prop in the component.

src/utils/components/AddBootableVolumeModal/AddBootableVolumeModal.tsx
pcbailey: I didn't fully remove it. I just moved it to the BootableVolume object. This allows the namespace to be pulled from it since it's already passed around.

PR #2061, opened 2024-06-27
Move disk source outside the add disk modal dialog

avivtur:
## 📝 Description

Changing the `Add disk` button to a flyout-dropdown to select the disk source from outside the add disk dialog.

## 🎥 Demo

After:

https://github.com/kubevirt-ui/kubevirt-plugin/assets/67270715/111f800c-0bf6-42f5-a455-a35e078ed253


src/utils/components/DiskSourceFlyoutMenu/DiskSourceFlyoutMenu.tsx
```diff
@@ -0,0 +1,99 @@
...
+  MenuList,
+  MenuToggle,
+} from '@patternfly/react-core';
+
+import { getDiskSourceOptionGroups } from '../DiskModal/components/DiskSourceSelect/utils/utils';
+import { SourceTypes } from '../DiskModal/utils/types';
+
+import FlyoutItem from './FlyoutItem/FlyoutItem';
+
+export type DiskSourceFlyoutMenuProps = {
```
metalice, line 18: if u export it please move to a type file

src/utils/components/DiskSourceFlyoutMenu/DiskSourceFlyoutMenu.tsx
```diff
@@ -0,0 +1,99 @@
...
+  onSelect,
+}) => {
+  const { t } = useKubevirtTranslation();
+  const [isOpen, setIsOpen] = useState<boolean>(false);
+  const menuRef = useRef<HTMLDivElement>(null);
+  const toggleRef = useRef<HTMLButtonElement>(null);
+
+  const options = getDiskSourceOptionGroups(isTemplate);
+
+  const onToggleClick = () => {
```
metalice, line 36: no need for that is only used once

src/utils/components/DiskSourceFlyoutMenu/DiskSourceFlyoutMenu.tsx
```diff
@@ -0,0 +1,99 @@
...
+      {t('Add disk')}
+    </MenuToggle>
+  );
+
+  const menu = (
+    <Menu containsFlyout ref={menuRef}>
+      <MenuContent>
+        <MenuList>
+          {options?.map(({ description, groupLabel, items }) => {
+            const isFlyout = items?.length > 1;
```
metalice, line 62: no need its only used once

src/utils/components/DiskSourceFlyoutMenu/DiskSourceFlyoutMenu.tsx
```diff
@@ -0,0 +1,99 @@
...
+                {items?.[0]?.label}
+              </MenuItem>
+            );
+          })}
+        </MenuList>
+      </MenuContent>
+    </Menu>
+  );
+
+  return (
```
metalice, line 86: i understand why menu and toggle are here, as nested components, as they are deeply coupled, but still consider exporting them to new files and giving them the right Capital naming.

avivtur: Extracting those to separated components is not working well for the MenuContainer as it expects a ReactElement type which have children:
Screenshot - 2024-06-30T153533 466

src/utils/components/DiskSourceFlyoutMenu/DiskSourceFlyoutMenu.tsx
avivtur: I can do this, I thought it would be more readable like this when the condition has context to it

metalice (approved):

PR #2062, opened 2024-06-27
CNV-38351: set autoattachPodInterface to false by default

upalatucci:
## 📝 Description

- set `autoattachPodInterface` to false for instanceType vms and explicitly defined the default network interface. Explicitly defining the interface, instead of relying on the default autoAttach, let the user delete all the network interfaces from the VM on customize or after creation.
- When the user deletes the last interface from existing vms, set the flag to false, so that the user will delete all the interfaces. Currently, If the user deletes all interfaces, the VM will always keep the default one. 
- In the network list, show the label 'pending' when the user try to remove one network from a running VM. Currently we show the pending label only when the user add the network and when it remove it, we just show the pending alert in top . 
- The first interface is marked as the default one and the default interface cannot be set to absent. In that case instead of mark as absent, just remove the interface

## 🎥 Demo

**Before**

As you can see here, the default network is implicitly declared by the autoAttachPodInterface flag (default: true). So there is no way in the UI to delete that. Even after creation, you cannot delete the default interface and there is no UI change so it seems that something is not working.

https://github.com/kubevirt-ui/kubevirt-plugin/assets/29160323/e7a06ec6-4300-42bd-906c-6fc6920df00b

Here I added a network during the 'starting' status and there is no pending label. The pending label show up only on running, but the VM can be even stuck in some states like volume binding or some errors, and I'll never see the label

https://github.com/kubevirt-ui/kubevirt-plugin/assets/29160323/79ace729-6197-415a-9d60-f69d87ef0121

Here I can't delete the default interface for this error. Cant' set absent because is the default interface. 

https://github.com/kubevirt-ui/kubevirt-plugin/assets/29160323/f94e80af-c044-4b4e-95aa-0629b4ce89c7

**After**

Create VM with no networks.
When adding a new network, you can see the pending even when the VM is not in the running state. This is because the vmi is already and does not have the interface yey, you'll see the pending label.

https://github.com/kubevirt-ui/kubevirt-plugin/assets/29160323/9b2e0b47-0afe-4b25-85a7-df4adbc0d818

When you delete the network from a running VM, you can see the pending label as the deletion is not reflected to the vmi. 
Note: as described before, here we cannot mark this interface as absent because it's the default one. 

https://github.com/kubevirt-ui/kubevirt-plugin/assets/29160323/1d2ca8a5-69ab-405d-9764-50774a486c41

But the pending label will be gone when the vmi reflect the interfaces change

https://github.com/kubevirt-ui/kubevirt-plugin/assets/29160323/ea57d9c1-c891-492d-822f-b5ff72f513fb

### DIFF
--- a/src/utils/components/NetworkInterfaceModal/utils/helpers.ts
+++ b/src/utils/components/NetworkInterfaceModal/utils/helpers.ts
@@ -4,12 +4,16 @@ import VirtualMachineModel from '@kubevirt-ui/kubevirt-api/console/models/Virtua
 import { V1Interface, V1Network, V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
 import { DEFAULT_NAMESPACE } from '@kubevirt-utils/constants/constants';
 import { t } from '@kubevirt-utils/hooks/useKubevirtTranslation';
-import { getInterfaces, getNetworks } from '@kubevirt-utils/resources/vm';
+import {
+  getAutoAttachPodInterface,
+  getInterfaces,
+  getNetworks,
+} from '@kubevirt-utils/resources/vm';
 import {
   interfacesTypes,
   NetworkPresentation,
 } from '@kubevirt-utils/resources/vm/utils/network/constants';
-import { kubevirtConsole } from '@kubevirt-utils/utils/utils';
+import { isEmpty, kubevirtConsole } from '@kubevirt-utils/utils/utils';
 import { k8sPatch } from '@openshift-console/dynamic-plugin-sdk';
 import { ABSENT } from '@virtualmachines/details/tabs/configuration/network/utils/constants';
 import { isStopped } from '@virtualmachines/utils';
@@ -46,6 +50,15 @@ export const updateVMNetworkInterfaces = (
         path: '/spec/template/spec/domain/devices/interfaces',
         value: updatedInterfaces,
       },
+      ...(isEmpty(updatedInterfaces)
+        ? [
+            {
+              op: getAutoAttachPodInterface(vm) === undefined ? 'add' : 'replace',
+              path: '/spec/template/spec/domain/devices/autoattachPodInterface',
+              value: false,
+            },
+          ]
+        : []),
     ],
     model: VirtualMachineModel,
     resource: vm,
@@ -74,11 +87,11 @@ const removeInterfaceToBeDeleted = (nicName: string, vm: V1VirtualMachine): V1In
   getInterfaces(vm)?.filter(({ name }) => name !== nicName);
 
 export const updateInterfacesForDeletion = (
-  isHotPlug: boolean,
   nicName: string,
   vm: V1VirtualMachine,
+  canBeMarkedAbsent: boolean,
 ): V1Interface[] => {
-  return isHotPlug && !isStopped(vm)
+  return canBeMarkedAbsent
     ? markInterfaceAbsent(getInterfaces(vm), nicName)
     : removeInterfaceToBeDeleted(nicName, vm);
 };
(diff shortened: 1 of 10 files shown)


src/views/catalog/wizard/tabs/network/components/list/NetworkInterfaceActions.tsx
```diff
@@ -50,6 +51,11 @@ const NetworkInterfaceActions: FC<NetworkInterfaceActionsProps> = ({
 
   const onDelete = useCallback(() => {
     const updatedVM = produceVMNetworks(vm, (draftVM) => {
+      if (isEmpty(draftVM.spec.template.spec.domain.devices.interfaces)) {
```
pcbailey, line 54: Can you use the `getInterfaces` selector here?

src/views/catalog/wizard/tabs/network/components/list/NetworkInterfaceActions.tsx
```diff
@@ -50,6 +51,11 @@ const NetworkInterfaceActions: FC<NetworkInterfaceActionsProps> = ({
...
+        draftVM.spec.template.spec.domain.devices.autoattachPodInterface = false;
+        return;
+      }
+
       draftVM.spec.template.spec.networks = draftVM.spec.template.spec.networks.filter(
```
pcbailey, line 59: I know you didn't change this, but you can you use the `getNetworks` selector her to make it `getNetworks(draftVM).filter...`?

src/views/virtualmachines/details/tabs/configuration/network/utils/utils.ts
```diff
@@ -16,3 +23,27 @@ export const isPendingHotPlugNIC = (
...
+export const isPendingRemoval = (
+  vm: V1VirtualMachine,
+  vmi: V1VirtualMachineInstance,
+  nicName: string,
+): boolean => {
+  if (!vmi || isStopped(vm)) return false;
+
+  const isVMRunning = isRunning(vm);
+
+  const autoAttach = getAutoAttachPodInterface(vm) !== false;
```
pcbailey, line 39: Would you mind changing this to `autoAttachPodInterface` just to make it a little clearer?

src/utils/components/PendingChanges/utils/helpers.ts
```diff
@@ -150,11 +155,13 @@ export const getChangedNICs = (vm: V1VirtualMachine, vmi: V1VirtualMachineInstan
   const vmNICsNames = vmInterfaces?.map((nic) => nic?.name) || [];
   const vmiNICsNames = vmiInterfaces?.map((nic) => nic?.name) || [];
 
+  const autoAttach = getAutoAttachPodInterface(vm) !== false;
```
pcbailey, line 158: I'm sorry. I missed this one. Could you please change this one to `autoAttachPodInterface`, too?

src/utils/components/NetworkInterfaceModal/utils/helpers.ts
```diff
@@ -46,6 +50,15 @@ export const updateVMNetworkInterfaces = (
         path: '/spec/template/spec/domain/devices/interfaces',
         value: updatedInterfaces,
       },
+      ...(isEmpty(updatedInterfaces)
+        ? [
+            {
+              op: getAutoAttachPodInterface(vm) === undefined ? 'add' : 'replace',
```
metalice, line 56: there are new K8S ops const for ad, replace , @pcbailey please show him where

src/views/catalog/wizard/tabs/network/components/list/NetworkInterfaceActions.tsx
```diff
@@ -50,11 +52,18 @@ const NetworkInterfaceActions: FC<NetworkInterfaceActionsProps> = ({
 
   const onDelete = useCallback(() => {
     const updatedVM = produceVMNetworks(vm, (draftVM) => {
-      draftVM.spec.template.spec.networks = draftVM.spec.template.spec.networks.filter(
+      const vmInterfaces = getInterfaces(draftVM);
+      if (isEmpty(vmInterfaces)) {
```
metalice, line 56: I don't see any distinction between IT and Template. Shouldn't there be one? as this action of network effect all VM's

upalatucci: In Template mode you will not end up in this case. 
In template mode the default interface is specified in the yaml. 
You'll end up here only when you are trying to remove the default interface but the VM has nothing in the array because that interface is implicitly declared . Hope this make sense

src/utils/components/NetworkInterfaceModal/utils/helpers.ts
upalatucci: Ah yes right. I'll fix that

PR #2063, opened 2024-06-27
CNV-42619: Modifications to disk size on Templates details page don't persist

pcbailey:
## 📝 Description

This PR fixes a bug where modifications to the size of a disk in the Edit disk modal on the Templates details page don't persist.

jira: https://issues.redhat.com/browse/CNV-42619

### Before

modify-template-disk-size--BEFORE--2024-06-27 08-57

### After

modify-template-disk-size--AFTER--2024-06-27 08-56

### DIFF
--- a/src/utils/components/DiskModal/utils/helpers.ts
+++ b/src/utils/components/DiskModal/utils/helpers.ts
@@ -17,7 +17,7 @@ import {
 import { buildOwnerReference, getName } from '@kubevirt-utils/resources/shared';
 import { hasTemplateParameter } from '@kubevirt-utils/resources/template';
 import { getBootDisk, getDataVolumeTemplates, getVolumes } from '@kubevirt-utils/resources/vm';
-import { ensurePath, getRandomChars } from '@kubevirt-utils/utils/utils';
+import { ensurePath, generatePrettyName } from '@kubevirt-utils/utils/utils';
 import { k8sCreate } from '@openshift-console/dynamic-plugin-sdk';
 
 import {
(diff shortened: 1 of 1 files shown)


src/utils/components/DiskModal/utils/helpers.ts
```diff
@@ -155,12 +155,12 @@ export const getDataVolumeFromState = ({
   const dvName =
     resultVolume?.dataVolume?.name ||
     resultVolume?.persistentVolumeClaim?.claimName ||
-    `${vm?.metadata?.name}-${diskState.diskName}`;
+    nameWithoutParameter(
```
metalice, line 158: do we need it at all?

pcbailey: Yes. It's part of a bug fix: https://issues.redhat.com/browse/CNV-28231.

upalatucci: i removed nameWithoutParameter in another pr : https://github.com/kubevirt-ui/kubevirt-plugin/pull/1977

metalice: @upalatucci, I remember you removing it, as you couldn't find the reason why it was added in the first place. Now that it's removed, do we have a regression with the bug link @pcbailey mentioned?

upalatucci: When the user tries to add a new disk on a template, we need to keep the parameter in the name.

Without a parameter ${NAME}, we can use the template to create different vms, but all those vms will use the same shared disk. This is not the result that the user intend to have when creates a disk in the template. 

I suggest to create a PVC from the uploaded file with a random string without any parameters (uploadData is the method responsible for that) and use it as a source for the datavolume. The data volume has to have the parameter in its name. With this mechanism we'll clone the same PVC to each VM created with that template.

upalatucci:
/lgtm

The pr is against 4.15 so the issue that I was pointing out during review is not present here. The name, during upload,is verified and if contains parameters it is randomly created so we are fine

PR #2068, opened 2024-07-02
CNV-36085: Moving loading permissions

metalice:
## 📝 Description

Moved the permissions loading, to disabled and loading, as currently its only used once and will be ignored later 

## 🎥 Demo


src/views/checkups/storage/utils/utils.ts
```diff
@@ -217,13 +217,12 @@ export const installOrRemoveCheckupsStoragePermissions = (
   namespace: string,
   isPermitted: boolean,
   clusterRoleBinding: IoK8sApiRbacV1ClusterRoleBinding,
-): Promise<Awaited<void>> => {
+): Promise<void> => {
```
upalatucci, line 220: This function is not void. It returns error right?

upalatucci (approved):

PR #2078, opened 2024-07-09
CNV-40088: add conditions to dashboard alert popup

upalatucci:
## 📝 Description

Conditions 
Go to HCO resource page on clicking the condition

## 🎥 Demo

Screenshot 2024-07-09 at 10 57 55

### DIFF
--- a/src/views/dashboard-extensions/KubevirtHealthPopup/components/Conditions/Conditions.tsx
+++ b/src/views/dashboard-extensions/KubevirtHealthPopup/components/Conditions/Conditions.tsx
@@ -1,0 +1,51 @@
+import React, { FC } from 'react';
+import { Link } from 'react-router-dom-v5-compat';
+
+import HyperConvergedModel from '@kubevirt-ui/kubevirt-api/console/models/HyperConvergedModel';
+import useHyperConvergeConfiguration from '@kubevirt-utils/hooks/useHyperConvergeConfiguration';
+import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
+import { getResourceUrl } from '@kubevirt-utils/resources/shared';
+import { PrometheusEndpoint, usePrometheusPoll } from '@openshift-console/dynamic-plugin-sdk';
+import { Skeleton, StackItem } from '@patternfly/react-core';
+
+import ConditionIcon from './ConditionIcon';
+import { HCO_HEALTH_METRIC, VALUE_TO_LABLE } from './constants';
+
+const Conditions: FC = () => {
+  const { t } = useKubevirtTranslation();
+  const [hyperConverge, hyperLoaded] = useHyperConvergeConfiguration();
+
+  const [queryData, loaded] = usePrometheusPoll({
+    endpoint: PrometheusEndpoint.QUERY,
+    query: HCO_HEALTH_METRIC,
+  });
+
+  const condition = queryData?.data?.result?.[0]?.value?.[1];
+
+  const label = VALUE_TO_LABLE[condition];
+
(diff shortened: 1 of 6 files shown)


src/views/dashboard-extensions/KubevirtHealthPopup/components/Conditions/utils.ts
```diff
@@ -0,0 +1,10 @@
+import { t } from '@kubevirt-utils/hooks/useKubevirtTranslation';
+
+// t('plugin__kubevirt-plugin~Healthy')
+// t('plugin__kubevirt-plugin~Warning')
+// t('plugin__kubevirt-plugin~Error')
```
avivtur, line 5: this is not needed once you use the t function

src/views/dashboard-extensions/KubevirtHealthPopup/components/Conditions/Conditions.tsx
```diff
@@ -0,0 +1,52 @@
...
+const Conditions: FC = () => {
+  const { t } = useKubevirtTranslation();
+  const [hyperConverge, hyperLoaded] = useHyperConvergeConfiguration();
+
+  const [queryData, loaded] = usePrometheusPoll({
+    endpoint: PrometheusEndpoint.QUERY,
+    query: HCO_HEALTH_METRIC,
+  });
+
+  const condition = queryData?.data?.result?.[0]?.value?.[1] as unknown as number;
```
avivtur, line 24: why do we go to this specific element? and why does this need 2 `as`?

src/views/dashboard-extensions/KubevirtHealthPopup/components/Conditions/utils.ts
```diff
@@ -0,0 +1,10 @@
...
+export const VALUE_TO_LABLE = {
+  0: t('Healthy'),
+  1: t('Warning'),
+  2: t('Error'),
+};
```
avivtur, line 10: You can move this to constants.ts file and remove this file

src/views/dashboard-extensions/KubevirtHealthPopup/components/Conditions/Conditions.tsx
upalatucci: This query give us just one value that represent the health of the cluster in this moment.

We need two `as` because in typescript this value is a string, but the returned value from the query is a number. Typescript do not permit converting string to number directly

avivtur: That doesn't make sense, you can convert strings to numbers in few ways:
1. add + sign before the string: `const number = +str;`
2. convert with Number constructor `const number = Number(str);`

I took this code and run it locally, and saw that kind of conversion directly is causing a bug since the value can be undefined and converting this as number will get us a NaN and unexpected behaviors.
you can either solve this with changing the keys to be strings e.g. '0' , '1', '2' instead of 0,1,2
OR keep doing as you do now. (I'm sure there's also another way but I think one of these will be enough)

upalatucci: Even tho typescript says that the value should be a string, it's actually a number. I have to check that behavior again. But this was the reason of the cast in typescript.

upalatucci: The conversion is not done on runtime. So there is no way that you get NaN. The conversion is just to quite typescript errors down. But it's not because we are doing something weird. It's because the query do not return what typescript says it return

PR #2079, opened 2024-07-09
CNV-43096: edit disk size on instancetype

upalatucci:
## 📝 Description

Add disk size on instancetype page

TODO: in a different pr I have to add the disk size editing in the customize storage tab

## 🎥 Demo

Screenshot 2024-07-09 at 13 18 28

### DIFF
--- a/src/views/catalog/CreateFromInstanceTypes/components/VMDetailsSection/components/DiskSize.tsx
+++ b/src/views/catalog/CreateFromInstanceTypes/components/VMDetailsSection/components/DiskSize.tsx
@@ -1,0 +1,22 @@
+import React, { FC } from 'react';
+
+import { useInstanceTypeVMStore } from '@catalog/CreateFromInstanceTypes/state/useInstanceTypeVMStore';
+import CapacityInput from '@kubevirt-utils/components/CapacityInput/CapacityInput';
+import { getVolumeSnapshotSize } from '@kubevirt-utils/resources/bootableresources/selectors';
+import { convertToBaseValue, humanizeBinaryBytes } from '@kubevirt-utils/utils/humanize.js';
+
+const DiskSize: FC = () => {
+  const { instanceTypeVMState, setCustomDiskSize } = useInstanceTypeVMStore();
+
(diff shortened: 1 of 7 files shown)


src/views/catalog/CreateFromInstanceTypes/state/utils/state.ts
```diff
@@ -8,6 +8,7 @@ import {
 import { InstanceTypeVMState, InstanceTypeVMStoreState } from './types';
 
 export const instanceTypeVMInitialState: InstanceTypeVMState = {
+  customDiskSize: '',
```
avivtur, line 11: shouldn't this have a default size 30Gi like in template flow?

src/views/catalog/CreateFromInstanceTypes/components/VMDetailsSection/components/DiskSize.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { formatBytes } from '@kubevirt-utils/resources/vm/utils/disk/size';
+
+const DiskSize: FC = () => {
+  const { instanceTypeVMState, setCustomDiskSize } = useInstanceTypeVMStore();
+
+  const { customDiskSize, pvcSource, volumeSnapshotSource } = instanceTypeVMState;
+
+  const pvcDiskSize =
+    pvcSource?.spec?.resources?.requests?.storage || getVolumeSnapshotSize(volumeSnapshotSource);
+  const sizeData = formatBytes(pvcDiskSize);
```
avivtur, line 15: This can cause a bug on certain edge-case, please look at https://github.com/kubevirt-ui/kubevirt-plugin/pull/1999/files

avivtur (commented):
Looking good :) 
just some small comments

src/views/catalog/CreateFromInstanceTypes/state/utils/state.ts
upalatucci: I want to ensure the disk size is empty when the user does not customize it. With that, I'll add the `spec.storage.resources.requests.storage` details only when the user customize the size.

avivtur (approved):

PR #2084, opened 2024-07-14
CNV-37371: New setting for change bootable volumes default ns

metalice:
## 📝 Description

Now bootable volumes by RH, can be moved between namespaces, the default namespace is openshift-virtualization-os-images

## 🎥 Demo

image


src/utils/components/FilterSelect/InlineFilterSelect.tsx
```diff
@@ -49,7 +49,6 @@ const InlineFilterSelect: FC<InlineFilterSelectProps> = ({
 
   const onSelect = (_event: React.MouseEvent<Element, MouseEvent> | undefined, value: string) => {
     if (value && value !== NO_RESULTS) {
-      setSelected(value);
```
avivtur, line 52: why removed?

PR #2092, opened 2024-07-22
CNV-44838: Use VirtualMachine dynamic extension pages

upalatucci:
## 📝 Description

Add obj as prop to be compliant with the SDK horizontal navbar that call extension components giving the obj prop

## 🎥 Demo

**Adding test tab**

Screenshot 2024-07-22 at 18 16 08

Here as you can see Overview and Test tabs are 'active' 
I solved this issue 

Screenshot 2024-07-22 at 18 26 47

### DIFF
--- a/src/utils/components/HorizontalNavbar/utils/useDynamicPages.ts
+++ b/src/utils/components/HorizontalNavbar/utils/useDynamicPages.ts
@@ -1,0 +1,33 @@
+import { useMemo } from 'react';
+
+import {
+  HorizontalNavTab,
+  isHorizontalNavTab,
+  K8sModel,
+  useResolvedExtensions,
+} from '@openshift-console/dynamic-plugin-sdk';
+
+const useDynamicPages = (model: K8sModel) => {
+  const [dynamicNavTabExtensions, navTabExtentionsResolved] =
+    useResolvedExtensions<HorizontalNavTab>(isHorizontalNavTab);
+
+  return useMemo(
+    () =>
+      navTabExtentionsResolved
+        ? dynamicNavTabExtensions
+            .filter(
+              (tab) =>
+                tab.properties.model.group === model.apiGroup &&
+                tab.properties.model.version === model.apiVersion &&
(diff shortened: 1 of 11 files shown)


src/utils/components/HorizontalNavbar/utils/useDynamicPages.ts
```diff
@@ -0,0 +1,33 @@
+import { useMemo } from 'react';
+import { VirtualMachineModel } from 'src/views/dashboard-extensions/utils';
+
+import {
+  HorizontalNavTab,
+  isHorizontalNavTab,
+  useResolvedExtensions,
+} from '@openshift-console/dynamic-plugin-sdk';
+
+const useDynamicPages = () => {
```
metalice, line 10: please add return type

metalice: although the name is useDynamicPages, it's very specific for VMs, let's rename it or add an argument(VirtualMachineModel) that this hook will filter by, and let's make this hook more generic

src/views/virtualmachines/details/utils/types.ts
```diff
@@ -2,5 +2,6 @@ import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
 
 export type NavPageComponentProps = {
   instanceTypeExpandedSpec?: V1VirtualMachine;
+  obj: V1VirtualMachine;
```
metalice, line 5: Except for obj, is no other prop guaranteed? Obj is guaranteed?

upalatucci: Other than obj we get name, ns, params, plural.

upalatucci: But there is no type specified in typescript sdk

upalatucci: added those params

metalice (approved):

PR #2103, opened 2024-07-31
CNV-42505: set bus depends on window

upalatucci:
## 📝 Description

Default bus is 'virtio' but windows do not support it if the VM has no drivers installed so switch to sata.

### DIFF
--- a/src/utils/components/EnvironmentEditor/hooks/useEnvironments.ts
+++ b/src/utils/components/EnvironmentEditor/hooks/useEnvironments.ts
@@ -2,8 +2,10 @@ import { useCallback, useEffect, useMemo, useState } from 'react';
 import { Updater } from 'use-immer';
 
 import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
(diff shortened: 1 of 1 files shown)


src/utils/components/EnvironmentEditor/hooks/useEnvironments.ts
```diff
@@ -64,15 +66,17 @@ const useEnvironments = (
 
       const diskName = `environment-disk-${getRandomChars()}`;
       getDisks(draftVM).push({
-        disk: {},
+        disk: {
+          bus: isWindowsVM ? 'sata' : 'virtio',
```
avivtur, line 70: please change to InterfaceTypes.SATA and InterfaceTypes.VIRTIO

metalice (approved):

MKletz:
Will scoping it to virtio or sata causes issues mounting env disks as CDROM? That's the only way I've successfully gotten the OS to recognize it.

MKletz:
@upalatucci please see question above

PR #2121, opened 2024-08-12
CNV-38571: Monitor user interaction in InstanceTypes and Templates flow

pcbailey:
## 📝 Description

This PR adds monitoring for events in the InstanceType and Template VM creation flows.

Jira: https://issues.redhat.com/browse/CNV-38571

### DIFF
--- a/src/utils/extensions/telemetry/telemetry.ts
+++ b/src/utils/extensions/telemetry/telemetry.ts
@@ -124,7 +124,7 @@ export const logITFlowEvent = (
   key: string,
   vm?: V1VirtualMachine,
   properties?: Record<string, any>,
-) => logEventWithName(key, vm, { ...properties, flow: createVMFlowTypes.IT });
+) => logEventWithName(key, vm, { ...properties, flow: createVMFlowTypes.InstanceTypes });
 
 export const logTemplateFlowEvent = (
   key: string,
--- a/src/utils/extensions/telemetry/telemetry.ts
+++ b/src/utils/extensions/telemetry/telemetry.ts
@@ -1,0 +1,138 @@
+import { V1Template } from '@kubevirt-ui/kubevirt-api/console';
+import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import { createVMFlowTypes } from '@kubevirt-utils/extensions/telemetry/utils/constants';
+import { getName } from '@kubevirt-utils/resources/shared';
+import { kubevirtConsole } from '@kubevirt-utils/utils/utils';
+
+const SEGMENT_KEY =
+  (window as any).SERVER_FLAGS?.telemetry?.DEVSANDBOX_SEGMENT_API_KEY ||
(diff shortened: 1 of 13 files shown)


src/utils/extensions/telemetry/telemetry.ts
```diff
@@ -0,0 +1,106 @@
+import { kubevirtConsole as kvConsole } from '@kubevirt-utils/utils/utils';
```
metalice, line 1: why changing its name?

src/utils/extensions/telemetry/telemetry.ts
```diff
@@ -0,0 +1,106 @@
...
+
+const SEGMENT_KEY =
```
metalice, line 3: Am I guessing all these files are from the Console? or some kind of demo?

src/utils/extensions/telemetry/telemetry.ts
```diff
@@ -0,0 +1,106 @@
...
+      analytics.load(SEGMENT_KEY);
+    }
+  }
+};
+
+initSegment();
+
+export const eventMonitor = async (eventType: string, properties?: any) => {
+  // eslint-disable-next-line no-console
+  console.log('Demo Plugin received telemetry event: ', eventType, properties);
```
metalice, line 76: can be removed

src/views/catalog/CustomizeInstanceType/components/CustomizeITVMFooter.tsx
```diff
@@ -75,6 +82,10 @@ const CustomizeITVMFooter: FC = () => {
                       data: vmSignal.value || vm,
                       model: VirtualMachineModel,
                     });
+                    eventMonitor(CUSTOMIZE_VM_SUCCEEDED, {
```
metalice, line 85: i can see in the rest of the PR that u always send a vmName and most of the time a flow, why not creating 3 funcs,

eventMonitorWithName , eventMonitorITFlow , eventMonitorTemplateFlow , the 2 last will use eventMonitorWithName, and u simply sending the VM and the key, all the rest of logic of adding a flow and a name will be under those 3 funcs, u wont need to import getName on each file, and simpler to work with

src/utils/extensions/telemetry/telemetry.ts
pcbailey: I don't remember why I did that. I changed it back.

src/utils/extensions/telemetry/telemetry.ts
pcbailey: Yes, this is how the API key is retrieved in Console.

src/utils/extensions/telemetry/telemetry.ts
pcbailey: Done!

src/views/catalog/CustomizeInstanceType/components/CustomizeITVMFooter.tsx
pcbailey: Done! There were some places where the VM wasn't accessible, so I had to work around those.

PR #2128, opened 2024-08-13
CNV-44365: New Virtualization Perspective

upalatucci:
## 📝 Description

Add `Virtualization` perspective with just virtualization, networking, storage and observe sections
Remove console-extension.json to split file into different dynamic-plugin.ts files

## 🎥 Demo

Screenshot 2024-08-14 at 12 03 42

Screenshot 2024-08-13 at 17 21 18


package.json
```diff
@@ -1,6 +1,7 @@
 {
   "dependencies": {
     "@novnc/novnc": "1.3.0",
+    "@openshift/dynamic-plugin-sdk": "^5.0.1",
```
avivtur, line 4: is this necessary? As a dynamic plugin we should stick to `@openshift-console`, but not sure about this, @vojtechszocs can you enlighten us on this?

vojtechszocs: @avivtur Hi, I've talked about this with @upalatucci and for now, we can import `EncodedExtension` from package `@openshift/dynamic-plugin-sdk-webpack` since `@openshift-console/dynamic-plugin-sdk-webpack` depends on it.

Aviv is right that we should not import directly from OpenShift plugin SDK (instead of from Console plugin SDK) unless we have a very good reason to do so.

I'll post a Console PR to expose `EncodedExtension` type via Console plugin SDK directly to address this gap.

avivtur: @vojtechszocs Thank you for the explanation :)

vojtechszocs: https://github.com/openshift/console/pull/14167 should address this issue.

PR #2132, opened 2024-08-17
CNV-45860: Deprecate CentOS 7 and CentOS Stream 8

pcbailey:
## 📝 Description

This PR adds deprecated badges to CentOS 7 and CentOS Stream 8 resources in the bootable volumes page, bootable volumes list on the InstanceTypes page, and the catalog Templates tab. It also adds filters to hide deprecated resources.

It also moves all badge components to a single folder.

Note: Selecting the filters by default will come in a follow-up PR.

Jira: https://issues.redhat.com/browse/CNV-45860

### Bootable volumes list

deprecate-centos7-8-bootable-volumes--AFTER--2024-08-31 21-39

### Create VM from InstanceTypes form

deprecate-centos7-8-instance-types--AFTER--2024-08-31 17-38

### Templates list

deprecate-centos7-8-templates-list--AFTER--2024-08-31 22-06

### Create VM from Templates form 

deprecate-centos7-8-templates-catalog--AFTER--2024-08-31 16-54


metalice:
any update here?

src/views/bootablevolumes/list/hooks/useBootableVolumesFilters.ts
```diff
@@ -14,6 +18,19 @@ const useBootableVolumesFilters = (): RowFilter<BootableResource>[] => {
...
+        availableResourceNames?.selected?.length === 0 || !isDeprecated(getName(obj)),
+      filterGroupName: ' ',
+      items: [
+        {
+          id: t('Hide deprecated bootable volumes'),
+          title: t('Hide deprecated bootable volumes'),
+        },
+      ],
+      reducer: (obj) => isDeprecated(getName(obj)) && 'Hide deprecated bootable volumes',
+      type: 'Hide deprecated bootable volumes',
```
upalatucci, line 32: This should be something like an id

src/views/catalog/templatescatalog/components/TemplatesCatalogTile.tsx
```diff
@@ -47,10 +53,13 @@ export const TemplateTile: React.FC<TemplateTileProps> = React.memo(
...
       <div onClick={() => onClick(template)}>
         <CatalogTile
-          badges={
+          badges={[
             bootSourcesLoaded
               ? isBootSourceAvailable && [
-                  <Badge key="available-boot">{t('Source available')}</Badge>,
+                  <Stack className="badge-stack" key="badge-stack">
+                    <Badge key="available-boot">{t('Source available')}</Badge>
+                    {isDeprecated ? <DeprecatedBadge className="deprecated-template" /> : null}
```
upalatucci, line 61: With that the deprecated badge will be added only if the source is available. I think you should add the badge to the array only if the template is deprecated

src/views/bootablevolumes/list/hooks/useBootableVolumesFilters.ts
pcbailey: Fixed!

src/views/catalog/templatescatalog/components/TemplatesCatalogTile.tsx
pcbailey: Good catch, thanks! Fixed!

upalatucci (approved):

PR #2149, opened 2024-09-02
CNV-46825: Editing Instancetype series and size

metalice:
## 📝 Description

Adding the ability to edit IT series and size

*while developing I noticed an issue with start-console script that run monitor plugin so added a line to solve it
*same for CPU missing default obj if not existing.

## 🎥 Demo

editing-it


src/utils/components/InstanceTypeModal/InstanceTypeModal.tsx
```diff
@@ -0,0 +1,164 @@
...
+      onClose();
+    } catch (e) {
+      setError(e.message);
+    } finally {
+      setUpdateInProcess(false);
+    }
+  };
+
+  return (
+    <Modal
```
avivtur, line 86: Can you switch to our TabModal? this would save the actions and the handleSubmit from repeating

src/views/virtualmachines/details/tabs/configuration/details/DetailsSection.tsx
```diff
@@ -133,24 +136,38 @@ const DetailsSection: FC<DetailsSectionProps> = ({ instanceTypeVM, vm, vmi }) =>
               />
             )}
             <VirtualMachineDescriptionItem
-              messageOnDisabled={t(
-                'CPU and Memory can not be edited if the VirtualMachine is created from InstanceType',
-              )}
+              descriptionHeader={
+                <SearchItem id="cpu-memory">
+                  {vm?.spec?.instancetype ? t('InstanceType') : t('CPU | Memory')}
```
avivtur, line 141: I think !isEmpty(vm?.spec?.instancetype) would be better. you can also put it in const since you're using this 2 times in this file :)

avivtur (approved):

PR #2175, opened 2024-09-16
CNV-46660: show disk pvc not the original source

upalatucci:
## 📝 Description

**Before**

VM disk list show the original PVC where data source clone the data for the disk

VM -> DataSource -> PVC (source cloned)

The PVC that you see here is not the PVC that the VM will use but what the datavolume will clone. This will probably create confusion to the user.

Screenshot 2024-09-16 at 08 56 47

**After**

Vm disk list show the actual PVC used by the VM 

VM -> DataVolume -> Actual PVC used (The one that we have to resize, not the source cloned)

Screenshot 2024-09-16 at 08 55 57

**Why we did that?**

Because in the instancetype we want to show in the disk list the source size that will be cloned. In order to do that, we have to set the `storage.resources.requests.storage` property

### DIFF
--- a/src/utils/resources/vm/hooks/disk/useDisksSources.ts
+++ b/src/utils/resources/vm/hooks/disk/useDisksSources.ts
@@ -1,29 +1,14 @@
 import { useMemo } from 'react';
 
-import { V1beta1DataSource } from '@kubevirt-ui/kubevirt-api/containerized-data-importer/models';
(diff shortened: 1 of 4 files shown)


metalice (approved):

metalice:
@upalatucci  i dont see SC and PVC link in the after screen shot?

PR #2182, opened 2024-09-16
CNV-46660: Enable pvc resize on vm disks

upalatucci:
## 📝 Description

User can edit the PVC size from the disk modal.

Let the user only *expand* the size if the PVC is created. 

If the PVC is not created, we can change that from the data volume structure just like before.

I deleted the `isCreated` prop from Diskmodal and created a new one called `createdPVCName`. If this prop exists, the PVC already exists, so we can't change the source.   

## 🎥 Demo

**Before**

Screenshot 2024-09-16 at 15 09 53

**After**

https://github.com/user-attachments/assets/090f9c12-6b43-4313-912c-08c82116f2ec


metalice:
@upalatucci isnt it easier and more clear instead of doing in many failed isCreated? expand : diskinputsize 

to add this check into diskinputsize?

src/utils/components/DiskModal/BlankDiskModal.tsx
```diff
@@ -32,6 +34,9 @@ const BlankDiskModal: FC<V1DiskModalProps> = ({
   const isVMRunning = isRunning(vm);
 
   const isEditDisk = !isEmpty(editDiskName);
+  const isCreated = !isEmpty(createdPVCName);
+  const namespace = getNamespace(vm);
+  const [pvc] = usePVCDiskSource(createdPVCName, namespace);
```
metalice, line 39: why calling this in every modal and not just send it as prop? 1 call and not every modal change, which is option change another call..

src/utils/components/DiskModal/RegistryDiskModal.tsx
```diff
@@ -51,19 +56,21 @@ const RegistryDiskModal: FC<V1DiskModalProps> = ({
   return (
     <FormProvider {...methods}>
       <TabModal
+        onSubmit={() =>
+          handleSubmit(async (data) => submit(data, vm, editDiskName, onSubmit, pvc))()
```
metalice, line 60: better make submit receive a single obj , and not args, then order isn't important

avivtur (approved):

PR #2183, opened 2024-09-16
CNV-41496: Change template details CPU/memory modal to new design

pcbailey:
## 📝 Description

This PR addresses an bug in the template details page where the number of CPUs displayed on the details page doesn't align with the number of CPUs in the CPU and memory modal. 

Cause: The details page display was reporting vCPU count (cores * sockets * threads) whereas the modal value was displaying the number of CPU cores.

Jira: https://issues.redhat.com/browse/CNV-41496

### Before

template-cpu-doubled-unexpectedly--BEFORE--2024-09-16_16-27

### After

template-cpu-doubled-unexpectedly--AFTER--2024-09-16_16-25


src/utils/components/CPUMemoryModal/CPUMemoryModal.tsx
```diff
@@ -138,7 +138,7 @@ const CPUMemoryModal: FC<CPUMemoryModalProps> = ({
         variant={AlertVariant.info}
       />
       <div className="inputs">
-        <CPUInput cpu={cpu} setCPU={setCPU} />
+        <CPUInput cpu={cpu} setCPU={setCPU} vmCPU={getCPU(vm)} />
```
avivtur, line 141: It's hard to understand why we need both cpu and vmCPU props, is one for the input and the second for current VM's CPU? Maybe we need better names to make it more intuitive

pcbailey: `cpu` is the state variable to track what the user has entered. `vmCPU` is what's currently set on the VM/template VM spec. It's used to determine the defaults for the 'Restore defaults' button. Maybe I should rename it to `defaultCPU`?

avivtur: defaultCPU does sound better, I thought maybe cpu could be called `cpuInput` and you can keep `vmCPU` or rename it to `defaultCPU` whichever sounds better to you

avivtur (approved):

PR #2215, opened 2024-10-04
OCPBUGS-42769: Get user default project

upalatucci:
## 📝 Description

`DEFAULT_NAMESPACE` is not always a good namespace to use. 
Users may not have access to the `default` namespace.
So I added the logic:

Use `default` if it's listed in the accessible namespaces. If not, use the first non-system namespace. Otherwise, take the first that you have. What do you think? But the important thing is that we do not show this info anywhere when the user is in `all-namespaces` 

## 🎥 Demo

**Before**

https://github.com/user-attachments/assets/08922ede-508e-49c8-ab5e-7de30b5bb24e

**After**

https://github.com/user-attachments/assets/32ac35be-b1ee-4b92-b3c7-4f7199d5a183

### DIFF
--- a/src/utils/resources/namespaces/constants.ts
+++ b/src/utils/resources/namespaces/constants.ts
@@ -1,0 +1,3 @@
+export const SYSTEM_NAMESPACES_PREFIX = ['kube-', 'openshift-', 'kubernetes-'];
+
+export const SYSTEM_NAMESPACES = ['default', 'openshift'];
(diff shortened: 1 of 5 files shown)


pcbailey:
I think we should discuss the desired behavior with Ronen and Yifat. At this point, I think it's expected for a resource to be created in the default namespace when 'All Projects' is selected (for those with access to the default namespace), but we don't have that same expectation for those that don't have access to it. It may be better to force them to select an actual namespace in that situation.

PR #2217, opened 2024-10-07
CNV-49299: use runStrategy on cloning

upalatucci:
## 📝 Description

`spec.running` and `spec.runStrategy` are mutual exclusive. YAML example now use `spec.runStrategy` so patching the VM with `spec.running: true` does not work. So if the VM use `spec.running` patch that spec attribute otherwise use `spec.runStrategy`

### DIFF
--- a/src/utils/components/CloneVMModal/CloneVMModal.tsx
+++ b/src/utils/components/CloneVMModal/CloneVMModal.tsx
@@ -1,4 +1,4 @@
-import React, { FC, useEffect, useState } from 'react';
+import React, { FC, useEffect, useMemo, useState } from 'react';
 import { useNavigate } from 'react-router-dom-v5-compat';
 
 import { VirtualMachineModelRef } from '@kubevirt-ui/kubevirt-api/console';
@@ -41,6 +41,13 @@ const CloneVMModal: FC<CloneVMModalProps> = ({ headerText, isOpen, onClose, sour
     ),
   );
 
+  const vmUseRunning = useMemo(
+    () =>
+      (source as V1VirtualMachine)?.spec?.running !== undefined &&
+      (source as V1VirtualMachine)?.spec?.running !== null,
+    [source],
+  );
+
   const [startCloneVM, setStartCloneVM] = useState(false);
 
   const [initialCloneRequest, setInitialCloneRequest] = useState<V1alpha1VirtualMachineClone>();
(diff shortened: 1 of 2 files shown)


metalice:
@upalatucci, why not just change the example yaml? Its ours isn't it? or you are thinking about a case of user yaml with runstrategy ?

upalatucci:
`spec.runStrategy` is the modern way to define the VM behavior. We should always use that instead of running. It gives the user more flexibility on how the VM should behave in case of failures. This is why the example yaml was changed from `spec.running` to `spec.runStrategy`. 
Other than that, I think our code should be flexible and be able to start vms that use both mechanisms

src/utils/components/CloneVMModal/CloneVMModal.tsx
```diff
@@ -41,6 +41,13 @@ const CloneVMModal: FC<CloneVMModalProps> = ({ headerText, isOpen, onClose, sour
     ),
   );
 
+  const vmUseRunning = useMemo(
+    () =>
+      (source as V1VirtualMachine)?.spec?.running !== undefined &&
+      (source as V1VirtualMachine)?.spec?.running !== null,
```
avivtur, line 47: What about when running can be `false`? there are also cases where someone would like to pass a different runStrategy the `Always`?

upalatucci: If r`spec.running` is false, it means that the user used this attribute to run/stop the VM so we'll use the same s[attribute changing the `spec.running` . If it's not in the vm spec it means that we'll use `runStrategy` 

I can at this point call the vm API action to run the vm, so the backend will choose what strategy use.

upalatucci: runStrategies that start vms are `Always` and `RerunOnFailure`

src/utils/components/CloneVMModal/utils/helpers.tsx
```diff
@@ -64,14 +64,20 @@ export const cloneVM = (
...
+      useRunning
+        ? {
+            op: 'replace',
+            path: '/spec/running',
+            value: true,
+          }
+        : {
+            op: 'replace',
+            path: '/spec/runStrategy',
+            value: 'Always',
```
metalice, line 79: once oren pr is merged, please change this value to use the new consts

upalatucci: done

metalice (approved):

PR #2218, opened 2024-10-07
CNV-47477: fix switching project from all-namespaces

upalatucci:
## 📝 Description

Remove the `useEffect` that was causing a bug switching projects. Now the activeKey is not a state but we get it directly from the pathname. 

`useVirtualMachineInstanceTypes` hook now has a third parameter that we need to use.

## 🎥 Demo

**Before**

https://github.com/user-attachments/assets/35d5985b-d2b9-4e88-8d92-89a7c4032a94

**After**

https://github.com/user-attachments/assets/2869fc22-d564-4b50-871a-7a0b3793352e

### DIFF
--- a/src/views/instancetypes/list/InstanceTypePage.tsx
+++ b/src/views/instancetypes/list/InstanceTypePage.tsx
@@ -1,4 +1,4 @@
-import React, { FC, useEffect, useMemo, useState } from 'react';
+import React, { FC, useMemo } from 'react';
 import { useLocation, useNavigate } from 'react-router-dom-v5-compat';
 import classNames from 'classnames';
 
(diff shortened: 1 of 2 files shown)


src/views/instancetypes/list/InstanceTypePage.tsx
```diff
@@ -29,12 +29,14 @@ const InstanceTypePage: FC<ListPageProps> = (props) => {
...
+    () => (location?.pathname.includes(VirtualMachineClusterInstancetypeModel.kind) ? 0 : 1),
+    [location?.pathname],
   );
   const [instanceTypes, loaded, loadError] = useVirtualMachineInstanceTypes(
     props?.fieldSelector,
     props?.selector,
+    true,
   );
 
   const urlUserPreference = useMemo(
```
avivtur, line 42: this should be renamed to urlUserInstancetypes

avivtur (approved):

PR #2221, opened 2024-10-09
CNV-41403: localize loading

upalatucci:
## 📝 Description

disk list loading is just for fetching PVC so we don't have to show the list loading
for environment list the loading is just for selecting new items, so we can show the loading only on the select if visible. 
(We can't show loading inside the select dropdown for a patternfly bug)

## 🎥 Demo

**Before**

https://issues.redhat.com/secure/attachment/13183584/storage_reload_times.mp4

**After**

https://github.com/user-attachments/assets/bd364d89-4b04-4ce1-8b41-14f89f59bcc4

### DIFF
--- a/src/utils/components/EnvironmentEditor/components/EnvironmentEditor.tsx
+++ b/src/utils/components/EnvironmentEditor/components/EnvironmentEditor.tsx
@@ -1,151 +1,54 @@
-import * as React from 'react';
+import React, { FC } from 'react';
 
-import {
-  IoK8sApiCoreV1ConfigMap,
-  IoK8sApiCoreV1Secret,
-  IoK8sApiCoreV1ServiceAccount,
-} from '@kubevirt-ui/kubevirt-api/kubernetes/models';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
-import { Button, Divider, TextInput, Tooltip } from '@patternfly/react-core';
-import { Select, SelectGroup, SelectVariant } from '@patternfly/react-core/deprecated';
+import { Button, TextInput, Tooltip } from '@patternfly/react-core';
 import { MinusCircleIcon } from '@patternfly/react-icons';
 
-import { EnvironmentKind, MapKindToAbbr } from '../constants';
-import { EnvironmentOption } from '../utils';
+import { EnvironmentKind } from '../constants';
 
-import EnvironmentSelectOption from './EnvironmentSelectOption';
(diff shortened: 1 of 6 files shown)


src/utils/components/EnvironmentEditor/components/EnvironmentEditor.tsx
```diff
@@ -89,63 +90,78 @@ const EnvironmentEditor: React.FC<EnvironmentEditorProps> = ({
...
-                isDisabled={environmentNamesSelected?.includes(serviceAccount.metadata.name)}
-                key={serviceAccount.metadata.name}
-                kind={EnvironmentKind.serviceAccount}
-                name={serviceAccount.metadata.name}
-              />
-            ))}
-          </SelectGroup>
-        </Select>
+        {loaded ? (
+          <Select
```
avivtur, line 94: Can you please extract this Select to a different component or if it's too tricky with prop drilling any other way to keep the file shorter?

src/utils/components/EnvironmentEditor/components/EnvironmentEditor.tsx
```diff
@@ -1,48 +1,49 @@
 import * as React from 'react';
```
avivtur, line 1: Please change this import and remove all React. occurrences in this file

src/views/virtualmachines/details/tabs/configuration/storage/components/tables/disk/DiskRow.tsx
```diff
@@ -72,15 +82,17 @@ const DiskRow: FC<
       </TableData>
 
       <TableData activeColumnIDs={activeColumnIDs} id="source">
-        {isPVCSource(obj) ? (
+        {sourcesLoaded && hasPVC && (
```
avivtur, line 85: Is sourcesLoaded needed because it affects the hasPVC const?

src/utils/components/EnvironmentEditor/components/EnvironmentEditor.tsx
upalatucci: yes sure

src/views/virtualmachines/details/tabs/configuration/storage/components/tables/disk/DiskRow.tsx
upalatucci: Why does it affect the const?

avivtur (approved):

PR #2208, opened 2024-10-01
OCPONZ-4464: docker image support for s390x

sanjayy-ibm:
## 📝 Description

> Add a brief description

## 🎥 Demo

> Please add a video or an image of the behavior/changes

PR #2222, opened 2024-10-09
CNV-49906: fix docker image builder

upalatucci:
## 📝 Description

[here](https://github.com/kubevirt-ui/kubevirt-plugin/pull/2208) we added docker support for s390x archs.
Github on creating images cross-arch is giving network errors.
This is because the network is very slow on those cross-arch docker builds. So just make sure to have github be patient :) 

## 🎥 Demo


metalice:
@upalatucci please open a bug for it and connect it

Issue #2072, opened 2024-07-03
Deployment instructions on Kubernetes cluster


legalgig:
I'm trying to figure out how to deploy this plugin to Kubernetes cluster (K3s). I've already had deployed OKD-Console and it's working fine by itself. I tried to deploy kubevirt plugin using the Container image from quay but all it returns is 404, after a little troubleshooting I figured out that the Dockerfile copies the html files to wrong direction, nginx expects them in `/opt/app-root/src` but Dockerfile copies it to `/usr/share/nginx/html`. After copying the files from `/usr/share/nginx/html` to `/opt/app-root/src` the Virtualization dropdown shows up on the OKD-Console and it semi-works.

After testing a little I found out: 
- Listing anything (VMs, InstanceTypes) doesn't work, it just loads indefinitely
- Creating VMs from YAML works, it redirects me to newly created VM and I can start it and see the output in VNC console
- Templating is dependent on openshift CRDs so it shouldn't be listed on normal kubernetes clusters

Here are my deployment files for both okd-console and kubevirt-plugin
kubevirt-ui-plugin
```
---
apiVersion: v1
kind: Service
metadata:
  name: kubevirt-ui-plugin
  namespace: kubevirt
  labels:    
    app.kubernetes.io/name: kubevirt-ui-plugin
    app.kubernetes.io/instance: kubevirt-ui-plugin
  annotations:
spec:
  type: ClusterIP
  ports:
    - name: http
      port: 8080
      targetPort: http
      protocol: TCP
  selector:    
    app.kubernetes.io/name: kubevirt-ui-plugin
    app.kubernetes.io/instance: kubevirt-ui-plugin
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: kubevirt-ui-plugin
  namespace: kubevirt
  labels:    
    app.kubernetes.io/name: kubevirt-ui-plugin
    app.kubernetes.io/instance: kubevirt-ui-plugin
  annotations:
    reloader.stakater.com/auto: "true"
spec:
  revisionHistoryLimit: 3
  replicas: 1
  strategy:
    type: RollingUpdate
    rollingUpdate:
      maxUnavailable: 1
  selector:
    matchLabels:      
      app.kubernetes.io/name: kubevirt-ui-plugin
      app.kubernetes.io/instance: kubevirt-ui-plugin
  template:
    metadata:      
      annotations:
      labels:  
        app.kubernetes.io/name: kubevirt-ui-plugin
[...]

PR #2227, opened 2024-10-11
Make kubevirt works with okd

upalatucci:
## 📝 Description

Using a kinD cluster i was able to install kubevirt, okd console, and make things work.

use `HYPERCONVERGED_NAMESPACE` to retrieve the right hyperconverged namespace. `kubevirt-hyperconverged` or `openshift-cnv` depending on okd or openshift. 

Also one other thing to fix for infinite loading:
in case of auth disabled mode (only development) we get an error on fetching the user (there is no user)
so if there is an error with the user in `useKubevirtUserSettings`, just stop loading. 
We'll have this error only in development but because we are just fetching the user settings, it's better to not have infinite loading.

solving the #2072  issue . Because kubevirt support plain k8s, it would be a nice thing to have also the UI plugin support this

## 🎥 Demo

**Before**

Screenshot from 2024-10-11 16-48-13
Screenshot from 2024-10-11 16-48-09
Screenshot from 2024-10-11 16-48-03

**After**

Screenshot from 2024-10-11 16-37-56
Screenshot from 2024-10-11 16-37-52
Screenshot from 2024-10-11 16-37-48

### DIFF
--- a/src/utils/utils/utils.ts
+++ b/src/utils/utils/utils.ts
@@ -1,7 +1,11 @@
 import { animals, colors, NumberDictionary, uniqueNamesGenerator } from 'unique-names-generator';
 
 import { IoK8sApiCoreV1Service } from '@kubevirt-ui/kubevirt-api/kubernetes';
-import { DEFAULT_NAMESPACE } from '@kubevirt-utils/constants/constants';
+import {
+  DEFAULT_NAMESPACE,
+  KUBEVIRT_HYPERCONVERGED,
+  OPENSHIFT_CNV,
+} from '@kubevirt-utils/constants/constants';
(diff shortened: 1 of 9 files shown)


src/utils/utils/utils.ts
```diff
@@ -32,6 +36,9 @@ export const get = (obj: unknown, path: string | string[], defaultValue = undefi
 
 export const isUpstream = (window as any).SERVER_FLAGS?.branding === 'okd';
 
+export const getHyperconvergedNamespace = () =>
```
metalice, line 39: this name isnt right, this func differ between cnv to hyperconverge , maybe change it to getDefaulktOperatonNS or something else?

metalice (approved):

PR #2240, opened 2024-10-29
CNV-43052: Select multiple vms at a time

upalatucci:
## 📝 Description

Let the user select multiple vms to run actions in parallel

1. Should we have a loading state somewhere?
2. Currently we cannot have the selection for all vms for SDK VirtualizedList limitations. Are we sure to let the user select ALL vms in the namespace or cluster? Can this have issues with running the action for 100 or 1000 vms? 

## 🎥 Demo

https://github.com/user-attachments/assets/195a4d32-921b-40f4-a272-3eef5ad6617e

### DIFF
--- a/src/views/virtualmachines/actions/hooks/useMultipleVirtualMachineActions.ts
+++ b/src/views/virtualmachines/actions/hooks/useMultipleVirtualMachineActions.ts
@@ -1,0 +1,42 @@
+import { useMemo } from 'react';
+
+import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import { ActionDropdownItemType } from '@kubevirt-utils/components/ActionsDropdown/constants';
+import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
+
+import { restartVM, startVM, stopVM, unpauseVM } from '../actions';
+
+type UseMultipleVirtualMachineActions = (vms: V1VirtualMachine[]) => ActionDropdownItemType[];
+
+const useMultipleVirtualMachineActions: UseMultipleVirtualMachineActions = (vms) => {
+  const { t } = useKubevirtTranslation();
+
+  const actions: ActionDropdownItemType[] = useMemo(() => {
+    return [
+      {
+        cta: () => vms.forEach((vm) => startVM(vm)),
+        id: 'vm-action-start',
+        label: t('Start'),
+      },
+      {
+        cta: () => vms.forEach((vm) => restartVM(vm)),
+        id: 'vm-action-restart',
+        label: t('Restart'),
+      },
+      {
+        cta: () => vms.forEach((vm) => stopVM(vm)),
+        id: 'vm-action-stop',
+        label: t('Stop'),
+      },
+      {
+        cta: () => vms.forEach((vm) => unpauseVM(vm)),
+        id: 'vm-action-unpause',
+        label: t('Unpause'),
+      },
+    ];
+  }, [t, vms]);
+
+  return actions;
+};
+
+export default useMultipleVirtualMachineActions;
--- a/src/views/virtualmachines/list/VirtualMachinesList.scss
+++ b/src/views/virtualmachines/list/VirtualMachinesList.scss
@@ -3,3 +3,15 @@
     height: fit-content;
   }
 }
+
+.vm-listpagebody .pf-v5-c-table {
+  .pf-v5-c-table__th.pf-v5-c-table__check {
+    width: 50px !important;
+    max-width: 50px;
+  }
+
+  .selection-column {
+    width: 50px;
+    max-width: 50px;
+  }
+}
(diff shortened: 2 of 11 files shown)


src/views/virtualmachines/list/hooks/useSelectVirtualMachines.ts
```diff
@@ -0,0 +1,41 @@
...
+  selectVM: (vm: V1VirtualMachine) => void;
+};
+
+const findVM = (selectedVMs, vm) =>
+  selectedVMs.find(
+    (selectedVM) =>
+      getName(selectedVM) === getName(vm) && getNamespace(selectedVM) === getNamespace(vm),
+  );
+
+const useSelectedVirtualMachines = create<UseSelectedVirtualMachinesType>()((set, get) => ({
```
metalice, line 21: why picking zustand and not signals?

src/views/virtualmachines/actions/hooks/useMultipleVirtualMachineActions.ts
```diff
@@ -0,0 +1,72 @@
...
+
+import { restartVM, startVM, stopVM, unpauseVM } from '../actions';
+
+type UseMultipleVirtualMachineActions = (vms: V1VirtualMachine[]) => ActionDropdownItemType[];
+
+const useMultipleVirtualMachineActions: UseMultipleVirtualMachineActions = (vms) => {
+  const { t } = useKubevirtTranslation();
+  const namespaces = Array.from(new Set(vms?.map((vm) => getNamespace(vm))));
+
+  const namespace = namespaces.length > 1 ? undefined : namespaces[0];
```
metalice, line 18: you can bind both in useMemo and no need for Array.from , set has size property

metalice: to get first element if u need u can use
const [first] = set;

src/views/virtualmachines/list/VirtualMachinesList.scss
```diff
@@ -3,3 +3,15 @@
     height: fit-content;
   }
 }
+
+.vm-listpagebody .pf-v5-c-table {
+  .pf-v5-c-table__th.pf-v5-c-table__check {
+    width: 50px !important;
```
metalice, line 9: we need the important? can add a more nested selector?

src/views/virtualmachines/list/VirtualMachinesList.tsx
```diff
@@ -152,6 +156,8 @@ const VirtualMachinesList: FC<VirtualMachinesListProps> = ({ kind, namespace })
 
   const noVMs = isEmpty(unfilteredData) && !vmsFilteredWithProxy;
 
+  const allVMsSelected = data?.length === selectedVMs.length;
```
metalice, line 159: that will trigger a lot of rerenders, did u checked? compared with useMemo and without?

upalatucci: I can use useMemo, yeah but I would suggest investigating deeply as in this component there are many things that trigger re-render. Starting from rowData and columnLayout. Those components get created on the fly at every rendering so they alone trigger a children re-rendering

src/views/virtualmachines/actions/hooks/useMultipleVirtualMachineActions.ts
upalatucci: i would remove the accessReview. Thinking more about that. 
If the user selects vms into two different namespaces, we will ask if he has access to all namespaces. 
If the user is not admin it will not be able to perform actions in this case so I would make the action already available. If the user does not have permissions to perform the action, it will just not happen

upalatucci: Example:
I have two vms into 'default' and 'my-test-namespace', I select both vms and try to make an action.
The UI should not verify if I have access to all namespaces. It should verify that I have access to only those two namespaces. This means that we have to make two different fetch calls. One to verify that I can patch vms in the 'default' namespace and one for 'my-test-namespace'.  This is not scalable for multiple namespaces in case of a user that select all vms in all available namespaces.

upalatucci: I will just not verify anything at this point. If the user does not have permissions its action will not do anything

src/views/virtualmachines/list/VirtualMachinesList.scss
upalatucci: Result without important

Screenshot 2024-11-08 at 10 35 43

src/views/virtualmachines/list/hooks/useSelectVirtualMachines.ts
upalatucci: i can convert it to signals.

metalice (approved):

PR #2245, opened 2024-11-04
CNV-46447: Export modal part 1

upalatucci:
## 📝 Description

Initial part of Export modal

upload to a registry URL the volume. 

what we are missing here:
- pod progress in the modal
- save the registry info like username password and so on. 
- make sure to manage registry infos from the user settings

## 🎥 Demo

**Export Modal**

https://github.com/user-attachments/assets/1917ad06-8010-4dc8-9df2-3d8bd81d5e99

**Pod Created**

Screenshot 2024-11-04 at 12 10 48


src/utils/components/ExportModal/utils.ts
```diff
@@ -0,0 +1,151 @@
...
+        namespace,
+      },
+      type: 'Opaque',
+    },
+    model: SecretModel,
+    ns: namespace,
+  });
+
+export const createServiceAccount = (namespace: string) =>
+  Promise.all([
```
metalice, line 42: why not async await ?

src/utils/components/ExportModal/ExportModal.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const [registryName, setRegistryName] = useState(() => `registry-${getRandomChars()}`);
+  const [username, setUsername] = useState('');
+  const [password, setPassword] = useState('');
+  const [destination, setDestination] = useState('');
+
+  return (
+    <TabModal
+      onSubmit={async () => {
+        const secretName = `registry-secret-${getRandomChars()}`;
+        // eslint-disable-next-line no-console
```
metalice, line 33: ?

src/utils/components/ExportModal/ExportModal.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+  const namespace = getNamespace(vm);
+  const [registryName, setRegistryName] = useState(() => `registry-${getRandomChars()}`);
+  const [username, setUsername] = useState('');
+  const [password, setPassword] = useState('');
+  const [destination, setDestination] = useState('');
+
+  return (
+    <TabModal
+      onSubmit={async () => {
```
metalice, line 31: no need for try catc here?

src/utils/components/ExportModal/utils.ts
upalatucci: sure

src/utils/components/ExportModal/ExportModal.tsx
upalatucci: No the tabmodal handle the try catch

src/utils/components/ExportModal/constants.ts
avivtur, line 1: Please use constants in this file instead of hard-coded values wherever is appropriate like instead of kind, names, K8sVerbs etc.

PR #2248, opened 2024-11-05
CNV-45822: Tree view part 1

avivtur:
## 📝 Description

Adding a tree view display to VM perspective

NOTE: The transition from a single VM to a folder view which filters the VMs by folder is not working for now.

## 🎥 Demo
Admin:
option1-tree-view
op

Non-admin: (all projects does not appear)

option1-tree-view-closed-non-admin
option1-tree-view-non-admin


src/views/virtualmachines/tree/VirtualMachineTreeView.tsx
```diff
@@ -0,0 +1,120 @@
...
+  const { setOrRemoveQueryArgument } = useQueryParamsMethods();
+  const pathname = location.pathname;
+  const isAdmin = useIsAdmin();
+  const [activeNamespace, setActiveNamespace] = useActiveNamespace();
+
+  const [vms, vmsLoaded, vmsError] = useKubevirtWatchResource<V1VirtualMachine[]>({
+    groupVersionKind: VirtualMachineModelGroupVersionKind,
+    isList: true,
+    limit: OBJECTS_FETCHING_LIMIT,
+  });
```
upalatucci, line 47: Does non-admin users will see the threeview?Do we plan to do something different for them here?
This will generate an error as they can't list resources in all namespaces.

avivtur: You're right, I've added a new hook named `useTreeViewData` to get all project, and than for each project we get a watch for the vms.
I'm a bit afraid from this hook since I'm not sure how it will work out at large clusters

start-console.sh
```diff
@@ -49,8 +49,8 @@ BRIDGE_K8S_MODE="off-cluster"
 BRIDGE_K8S_AUTH="bearer-token"
 BRIDGE_K8S_MODE_OFF_CLUSTER_SKIP_VERIFY_TLS=true
 BRIDGE_K8S_MODE_OFF_CLUSTER_ENDPOINT=$(oc whoami --show-server)
-BRIDGE_K8S_MODE_OFF_CLUSTER_THANOS=$(oc -n openshift-config-managed get configmap monitoring-shared-config -o jsonpath='{.data.thanosPublicURL}')
-BRIDGE_K8S_MODE_OFF_CLUSTER_ALERTMANAGER=$(oc -n openshift-config-managed get configmap monitoring-shared-config -o jsonpath='{.data.alertmanagerPublicURL}')
+# BRIDGE_K8S_MODE_OFF_CLUSTER_THANOS=$(oc -n openshift-config-managed get configmap monitoring-shared-config -o jsonpath='{.data.thanosPublicURL}')
```
metalice, line 52: please remove

src/views/virtualmachines/tree/VirtualMachineTreeView.tsx
upalatucci: Yeah right. If the user is admin, we can continue watching vms like this. With just one request. Maybe we can fetch vms by projects only if the user is non admin?

src/views/virtualmachines/tree/hooks/useTreeViewData.ts
```diff
@@ -0,0 +1,58 @@
...
+  );
+
+  const memoizedVMs = useMemo(
+    () => (isAdmin ? allVMs : Object.values(allowedResources).flatMap((resource) => resource.data)),
+    [allVMs, allowedResources, isAdmin],
+  );
+
+  return {
+    isAdmin,
+    loaded: projectNamesLoaded && (isAdmin ? allVMsLoaded : !isEmpty(memoizedVMs)),
```
upalatucci, line 53: i would not use `!isEmpty(memoizedVMs)` as it will keep loading in case of non-admin user with no vms. I think we should use 
`Object.values(allowedResources).every((resource) => resource.loaded)` or something like that

avivtur: The reason I'm doing like this is because a user can see a project in the project select, but he can not list vms on that namespace while there are other namespaces he can list vms in, the way you suggest while not show the tree view if there's one namespace that gets an error.
I can better the condition if I do `Object.values(allowedResources).some((resource) => resource.loaded)` wdyt?

upalatucci: Yes i like your solution

upalatucci (approved):

metalice:
/hold

i would like to have glimpse also :)

src/views/virtualmachines/tree/hooks/useSyncClicksEffects.ts
```diff
@@ -0,0 +1,52 @@
...
+      if (pathname.startsWith(`/k8s/${ALL_NAMESPACES}`)) {
+        setSelectedTreeItem(dataMap[ALL_NAMESPACES_SESSION_KEY]);
+        return;
+      }
+
+      if (
+        pathname.startsWith('/k8s/ns/') &&
+        (pathname.endsWith(VirtualMachineModelRef) ||
+          pathname.endsWith(`${VirtualMachineModelRef}/`))
+      ) {
```
metalice, line 31: is this chcek needed? i mean if its not all namespaces it not a must that it will be with ns ?

src/views/virtualmachines/tree/hooks/useSyncClicksEffects.ts
```diff
@@ -0,0 +1,52 @@
...
+        if (folderFilterName) {
+          setSelectedTreeItem(
+            dataMap[`${FOLDER_SELECTOR_PREFIX}/${activeNamespace}/${folderFilterName}`],
```
metalice, line 34: why folder selector prefix is required for the data structure? ns/folder is not enough?

src/views/virtualmachines/tree/hooks/useSyncClicksEffects.ts
```diff
@@ -0,0 +1,52 @@
...
+          );
+          return;
+        }
+
+        setSelectedTreeItem(dataMap[`${PROJECT_SELECTOR_PREFIX}/${activeNamespace}`]);
```
metalice, line 39: project isnt same as ns?

src/views/virtualmachines/tree/hooks/useSyncClicksEffects.ts
```diff
@@ -0,0 +1,52 @@
...
+        return;
+      }
+
+      const vmName = pathname.split('/')[5];
+      if (vmName && pathname.includes(vmName)) {
```
metalice, line 44: u are extracting vmNAme from path and checking its there?

src/views/virtualmachines/tree/VirtualMachineTreeView.tsx
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    if (treeViewItem.id.startsWith(PROJECT_SELECTOR_PREFIX)) {
+      setActiveNamespace(treeItemName);
+      return navigate(
+        getResourceUrl({
+          activeNamespace: treeItemName,
+          model: VirtualMachineModel,
+        }),
+      );
```
metalice, line 84: seems redundant those 2 actions, as change in navigate is changing URL , so activeNampace should be change also.. please check

src/views/virtualmachines/tree/VirtualMachineTreeView.tsx
```diff
@@ -0,0 +1,113 @@
...
+      getResourceUrl({
+        activeNamespace: vmNamespace,
+        model: VirtualMachineModel,
+        resource: vmsMapper?.[vmNamespace]?.[vmName],
+      }),
+    );
+  };
+
+  return (
+    <Grid style={{ height: document.getElementById('content-scrollable')?.offsetHeight || 0 }}>
```
metalice, line 98: why ?

src/views/virtualmachines/tree/hooks/useTreeViewData.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  const [allVMs, allVMsLoaded] = useK8sWatchResource<V1VirtualMachine[]>({
+    groupVersionKind: VirtualMachineModelGroupVersionKind,
+    isList: true,
+    limit: OBJECTS_FETCHING_LIMIT,
+  });
+
+  // user has limited access, so we can only get vms from allowed namespaces
+  const allowedResources = useK8sWatchResources<{ [key: string]: V1VirtualMachine[] }>(
+    Object.fromEntries(
```
metalice, line 31: why u need fromEntries here?

src/views/virtualmachines/tree/hooks/useTreeViewData.ts
```diff
@@ -0,0 +1,61 @@
...
+      projectNamesLoaded && !isAdmin
+        ? (projectNames || []).map((namespace) => [
+            namespace,
+            {
+              groupVersionKind: VirtualMachineModelGroupVersionKind,
+              isList: true,
+              namespace,
+            },
```
metalice, line 39: why not simply return 
```
{
              groupVersionKind: VirtualMachineModelGroupVersionKind,
              isList: true,
              namespace,
            } 
```

why u need to return array with [ns,obj] ?

src/views/virtualmachines/tree/utils/utils.tsx
```diff
@@ -0,0 +1,125 @@
...
+    if (!projectMap[vmNamespace]) {
+      projectMap[vmNamespace] = { folders: {}, ungrouped: [] };
+    }
+
+    if (folder) {
+      if (!projectMap[vmNamespace].folders[folder]) {
+        projectMap[vmNamespace].folders[folder] = [];
+      }
+      projectMap[vmNamespace].folders[folder].push(vmTreeItem);
+    } else {
```
metalice, line 64: no need for else u can return

avivtur: changed

src/views/virtualmachines/tree/VirtualMachineTreeView.tsx
avivtur: removed

src/views/virtualmachines/tree/hooks/useSyncClicksEffects.ts
```diff
@@ -0,0 +1,52 @@
...
+        setSelectedTreeItem(dataMap[`${activeNamespace}/${vmName}`]);
+        return;
+      }
+
+      return;
+    }
+  }, [activeNamespace, loaded, location.search, location.pathname]);
```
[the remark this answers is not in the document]
avivtur: I wanted to sync between the clicks in the VM list view and the project select (which will now be gone) to the tree-view, If you have a better idea I'm open to changes. 
The reason I did this in an effect is that I needed to listen to changes of active namespace and the URL path

src/views/virtualmachines/tree/hooks/useSyncClicksEffects.ts
avivtur: you're right, changed to just check if the name exists

src/views/virtualmachines/tree/hooks/useSyncClicksEffects.ts
avivtur: what project are you talking about?

src/views/virtualmachines/tree/hooks/useSyncClicksEffects.ts
avivtur: I need to know it's a folder, by ns/folder I can't distinguish if the name at the end is a VM name or a folder name, this prefix helps me identify the folder items

src/views/virtualmachines/tree/hooks/useSyncClicksEffects.ts
avivtur: it's needed since if you have a VM name at the end, you can have a filter in the URL and both include the VM ref in their URL, this condition ensures that I'm really at the NS perspective

src/views/virtualmachines/tree/VirtualMachineTreeView.tsx
```diff
@@ -0,0 +1,136 @@
...
+  };
+
+  return (
+    <Drawer isExpanded isInline position="start">
+      <DrawerContent
+        panelContent={
+          <DrawerPanelContent
+            style={{
+              height: document.getElementById('content-scrollable').offsetHeight || 0,
+              overflowX: 'hidden',
```
metalice, line 109: u can spread a class here, i understand the issue with height, but other values can come from class

src/views/virtualmachines/tree/VirtualMachineTreeView.tsx
```diff
@@ -0,0 +1,136 @@
...
+  const vmsMapper = useMemo(() => convertResourceArrayToMap(vms, true), [vms]);
+
+  const treeData = useMemo(
+    () => createTreeViewData(projectNames, vms, activeNamespace, isAdmin, location.pathname),
+    [projectNames, vms, activeNamespace, isAdmin, location.pathname],
+  );
+
+  const { filteredItems, onSearch, setShowAll, showAll } = useTreeViewSearch(treeData);
+
+  useSyncClicksEffects(activeNamespace, loaded, location);
```
metalice, line 59: do we still need it? even after hiding the ns bar? activeNamespace isn't enough?

avivtur: This is still needed to be in sync with vm list click not from treeview

metalice (approved):

PR #2253, opened 2024-11-07
CNV-49572: fix configuration layout

upalatucci:
## 📝 Description

the tabs have `width: fit-content;` and `min-width: max-content;` but no `max-width`.

## 🎥 Demo

**Before**

Screenshot 2024-11-07 at 09 21 15

**After**

Screenshot 2024-11-07 at 09 21 55

### DIFF
--- a/src/views/virtualmachines/details/tabs/configuration/virtual-machine-configuration-tab.scss
+++ b/src/views/virtualmachines/details/tabs/configuration/virtual-machine-configuration-tab.scss
@@ -4,6 +4,7 @@
     margin: var(--pf-global--spacer--sm);
     width: fit-content;
     min-width: max-content;
+    max-width: 100px;
   }
   &--content {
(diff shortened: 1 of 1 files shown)


src/views/virtualmachines/details/tabs/configuration/virtual-machine-configuration-tab.scss
```diff
@@ -4,6 +4,7 @@
     margin: var(--pf-global--spacer--sm);
     width: fit-content;
     min-width: max-content;
+    max-width: 100px;
```
metalice, line 7: have u tested on various resolutions? u can using chrome dev tools

upalatucci: Yes. It's working with this

PR #2259, opened 2024-11-08
CNV-43245: Use template runstrategy

upalatucci:
## 📝 Description

We want in the drawer to make the VM starting by default. 

Instead of using `Always` as a starting default, lets use `RerunOnFailures`. If the template already have something else defined like `Always` or `Once` do not modify that. Keep that.

### DIFF
--- a/src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/hooks/useDrawerContext.tsx
+++ b/src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/hooks/useDrawerContext.tsx
@@ -13,8 +13,12 @@ import { Updater, useImmer } from 'use-immer';
 import { V1Template } from '@kubevirt-ui/kubevirt-api/console';
 import { V1beta1DataVolumeSpec, V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
 import { SSHSecretDetails } from '@kubevirt-utils/components/SSHSecretModal/utils/types';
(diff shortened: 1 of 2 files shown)


src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/hooks/useDrawerContext.tsx
```diff
@@ -103,7 +107,11 @@ const useDrawer = (template: V1Template) => {
     const templateWithRunning = produce(templateWithGeneratedParams, (draftTemplate) => {
       const draftVM = getTemplateVirtualMachineObject(draftTemplate);
 
-      if (isEmpty(draftVM?.spec?.running)) draftVM.spec.runStrategy = RUNSTRATEGY_ALWAYS;
+      if (
+        isEmpty(draftVM?.spec?.running) &&
+        [RUNSTRATEGY_HALTED, RUNSTRATEGY_MANUAL].includes(draftVM?.spec?.runStrategy)
+      )
```
avivtur, line 113: Something is not logical here, it can't be both with empty value and include one of the values

upalatucci: we are verifying `spec.running` and `spec.runStrategy`.

avivtur: excuse me I thought I saw something different

PR #2261, opened 2024-11-12
CNV-50914: Support for ephemeral nic

metalice:
## 📝 Description

Suuport for ephemeral nic. This nic is added by the machine and removed on restart. I added a label of ephemeral. Currently, this type of nic is causing the UI to crash.

## 🎥 Demo

Before: Crash , oh no something went wrong

After:

image

### DIFF
--- a/src/views/virtualmachines/details/tabs/configuration/network/components/list/NetworkInterfaceRow.tsx
+++ b/src/views/virtualmachines/details/tabs/configuration/network/components/list/NetworkInterfaceRow.tsx
@@ -1,12 +1,19 @@
 import React, { FC } from 'react';
 
-import { V1Network, V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import {
+  V1Network,
+  V1VirtualMachine,
+  V1VirtualMachineInstanceNetworkInterface,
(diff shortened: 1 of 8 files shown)


src/views/virtualmachines/details/tabs/configuration/network/components/list/NetworkInterfaceRow.tsx
```diff
@@ -17,36 +24,45 @@ export type NetworkInterfaceRowProps = {
...
     NetworkPresentation,
-    { isPending: (network: V1Network) => boolean; vm: V1VirtualMachine }
+    {
+      isInterfaceEphemeral: (
+        network: V1Network,
+        iface: V1VirtualMachineInstanceNetworkInterface,
+      ) => undefined | V1VirtualMachineInstanceNetworkInterface;
+      isPending: (network: V1Network) => boolean;
+      vm: V1VirtualMachine;
+    }
```
upalatucci, line 34: Why passing this function in the RowProps? I think you changed idea in the middle of the work don't you?

metalice: just to keep the schema , to be same as isPending

PR #2275, opened 2024-11-19
CNV-50977: CNV-51361: fix cve packages

upalatucci:
## 📝 Description

upgrade css-loader to have min 8.4.31 in postcss. With this upgrade we'll use 8.4.49
upgrade cypress, eslint and resolve-url-loader to have cross-spawn min 7.0.5. With the upgrade, we'll use 7.0.6


metalice:
i see only yarn.lock, where is the package.json ?

metalice (approved):

PR #2276, opened 2024-11-20
check more pages

gouyang:
## 📝 Description

1. Check all tabs can be loaded under Virtualiztion
2. Check all the example yaml works under Virtualiztion

## 🎥 Demo


cypress/views/tab.ts
```diff
@@ -1,19 +1,24 @@
...
   NetworkInterfaces = 'horizontal-link-Network interfaces',
+  // vm tabs
   Overview = 'horizontal-link-Overview',
-  Parameters = 'horizontal-link-Parameters',
+  Parameters = 'horizontal-link-Parameters', // template tab
   Scheduling = 'horizontal-link-Scheduling',
   Scripts = 'horizontal-link-Scripts',
+  Settings = 'horizontal-link-Settings',
   Snapshots = 'horizontal-link-Snapshots',
+  // virt overview tabs
```
avivtur, line 20: I think it will be more readable to split the tabs to multiple enums each representing a different group of tabs like detailsTabs, configurationSubTabs, overviewTabs and remove the comments that split those tabs

cypress/views/selector.ts
avivtur, line 1: I think we should choose either `data-test` or `data-test-id` and stick with only one, do we have a standard for this?

cypress/views/tab.ts
```diff
@@ -29,51 +34,77 @@ export const tab = {
   navigateToConsole: () => {
     navigateToTab(tabs.Console);
   },
+  // vm configuration sub tabs
+  navigateToDetails: () => {
+    navigateToTab(tabs.Configuration);
+    cy.byButtonText('Details').click();
```
avivtur, line 40: I believe it would be better to have cypress find elements by some other property like `data-test` or some other property if missing rather than text found on screen as it has more chance to be changed in the future, WDYT?

locales/en/plugin__kubevirt-plugin.json
avivtur, line 1: is this file needed?

cypress/tests/e2e/overview.cy.ts
```diff
@@ -0,0 +1,48 @@
...
+      }
+    });
+
+    cy.visitOverview();
+    cy.wait(10000);
+    cy.get('body').then(($body) => {
+      if ($body.text().includes('Do not show this again')) {
+        cy.get(welcomeCheckbox).check();
+      }
+    });
```
avivtur, line 25: This part repeats twice, can we encapsulate this into a function?

cypress/support/index.ts
```diff
@@ -20,3 +20,12 @@ Cypress.on('uncaught:exception', () => {
 });
 
 Cypress.Cookies.debug(true);
+
+// ignore fetch in command log
```
avivtur, line 24: Can you please explain to me why we need this?

avivtur (commented):
Great work :)
Left you a few comments

cypress/views/selector.ts
gouyang: If developers add unique `data-test` or `data-test-id` for every elements, I'd be very happy to use it.
Eg: For items under `Virtualization`, it used `data-test="nav"` for all items, so use `data-test-id` for them.
For button like `Create`, it only has `data-test`, so use `data-test` for it.

cypress/views/tab.ts
gouyang: Again, the `data-test` or `data-test-id` is missing on these elements.

cypress/support/index.ts
gouyang: the fetch command is filling up the screen and make it harder to debug while checking errors on screenshots.

locales/en/plugin__kubevirt-plugin.json
gouyang: it's added by mistake, will remove it.

cypress/views/tab.ts
gouyang: split them into `vmTabs` and `overviewTabs`.

cypress/tests/e2e/overview.cy.ts
gouyang: done

cypress/views/selector.ts
avivtur: @gouyang can you open a ticket on where `data-test` is missing? I think it's worth adding, wdyt? 
I think it would help the tests eventually be more clean if we would have a standard for this. 
let's go with this for now and improve it once we add the `data-test` on the code side.

cypress/views/tab.ts
avivtur: see comment above :)

gouyang: I opened the issue https://issues.redhat.com/browse/CNV-52246 and assigned it to you.

PR #2285, opened 2024-11-25
WIP: CNV-51570: Redirect from 'All projects' to accessible project for non-admin user

pcbailey:
## 📝 Description

Non-admin users shouldn't be shown 'All projects'. This PR redirects these users to an accessible project instead. This isn't a complete solution. This should happen automatically on the console side. I will create a PR on console as well.

Jira: https://issues.redhat.com/browse/CNV-51570

### DIFF
--- a/src/utils/hooks/useSetDefaultNonAdminUserProject/useSetDefaultNonAdminUserProject.ts
+++ b/src/utils/hooks/useSetDefaultNonAdminUserProject/useSetDefaultNonAdminUserProject.ts
@@ -1,0 +1,24 @@
+import { DEFAULT_NAMESPACE } from '@kubevirt-utils/constants/constants';
+import { useIsAdmin } from '@kubevirt-utils/hooks/useIsAdmin';
+import useProjects from '@kubevirt-utils/hooks/useProjects';
+import { isSystemNamespace } from '@kubevirt-utils/hooks/useSetDefaultNonAdminUserProject/utils/utils';
+import { useActiveNamespace } from '@openshift-console/dynamic-plugin-sdk';
+
+type UseSetDefaultNonAdminUserProject = () => void;
+
+const useSetDefaultNonAdminUserProject: UseSetDefaultNonAdminUserProject = () => {
+  const isAdmin = useIsAdmin();
+  const [activeNamespace, setActiveNamespace] = useActiveNamespace();
+  const [projects] = useProjects();
+
+  const nonDefaultUserProjects = projects.reduce((acc, project) => {
+    if (!isSystemNamespace(project)) acc.push(project);
+    return acc;
+  }, []);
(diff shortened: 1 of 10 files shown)


src/utils/hooks/useSetDefaultNonAdminUserProject/useSetDefaultNonAdminUserProject.ts
```diff
@@ -0,0 +1,24 @@
...
+  const isAdmin = useIsAdmin();
+  const [activeNamespace, setActiveNamespace] = useActiveNamespace();
+  const [projects] = useProjects();
+
+  const nonDefaultUserProjects = projects.reduce((acc, project) => {
+    if (!isSystemNamespace(project)) acc.push(project);
+    return acc;
+  }, []);
+
+  if (isAdmin || projects.includes(activeNamespace)) return;
```
avivtur, line 19: Moving this before the reduce method will save a loop run

src/views/clusteroverview/ClusterOverviewPage.tsx
```diff
@@ -17,6 +17,7 @@ import TopConsumersTab from './TopConsumersTab/TopConsumersTab';
 const ClusterOverviewPage: FC = () => {
   const { t } = useKubevirtTranslation();
   const isAdmin = useIsAdmin();
+  // useSetDefaultNonAdminUserProject();
```
avivtur, line 20: on purpose?

src/views/instancetypes/list/InstanceTypePage.tsx
```diff
@@ -23,16 +25,21 @@ import '@kubevirt-utils/styles/list-managment-group.scss';
...
-  const activeNamespace = props?.namespace ?? ALL_NAMESPACES;
+  const [activeNamespace] = useActiveNamespace();
+  const currentNamespace = props?.namespace ?? activeNamespace ?? ALL_NAMESPACES;
 
   const isSearchPage = useIsSearchPage();
 
   const activeTabKey = useMemo(
-    () => (location?.pathname.includes(VirtualMachineClusterInstancetypeModel.kind) ? 0 : 1),
+    () =>
+      location?.pathname.includes(VirtualMachineClusterInstancetypeModel.kind) && isAdmin ? 0 : 1,
```
avivtur, line 38: why do we need the isAdmin here too? won't this prevent from non-admin users to reach the VM create flow with IT?

avivtur (commented):
This issue sounds more like a problem with console rather than our plugin issue, can explain to me the logic for us to remove manually the all-namespaces?

pcbailey:
This bug has been moved to console: https://issues.redhat.com/browse/OCPBUGS-45944.

PR #2292, opened 2024-11-29
Add networking-console-plugin on start-console

upalatucci:
## 📝 Description

With this we'll be able to run networking-console-plugin with our `start-console.sh` script .

Some changes have been required to make sure in networking-console-plugin we can do the exact same thing for kubevirt-plugin 

the `--port` doesn't really work with the `yarn start` command so I used the env variable `PORT` but I left the flag in order to support both ways

needs https://github.com/openshift/networking-console-plugin/pull/147 to work

we'll be able to run the plugin within kubevirt repo with:
`yarn start-console networking-console-plugin` 

or 

`yarn start-console-with-networking`

## 🎥 Demo

Screenshot 2024-11-29 at 18 22 59


avivtur:
@upalatucci so in order for this to work, we will need to run yarn start-console only in one project right?
If there are new steps for this to work please add them to the description of the PR.
Is will still be possible to run kubevirt plugin with the networking plugin right?

upalatucci:
yes just like running start-console with the `monitoring-plugin`. 
With `yarn start-console networking-console-plugin`

PR #2299, opened 2024-12-04
CNV-49533: Add Memory, CPU and Network usage on vm list

upalatucci:
## 📝 Description

Adding Network, Memory and CPU Usage in VM list

Note: 
- right now the columns are not sortable but we can add that later
-  For network we use the total usage (received + transferred)

## 🎥 Demo

Screenshot 2024-12-04 at 18 01 40

### DIFF
--- a/src/views/virtualmachines/list/components/VirtualMachineRow/VirtualMachineRowLayout.tsx
+++ b/src/views/virtualmachines/list/components/VirtualMachineRow/VirtualMachineRowLayout.tsx
@@ -1,4 +1,4 @@
-import * as React from 'react';
+import React, { FC, useMemo } from 'react';
 
 import { VirtualMachineModelGroupVersionKind } from '@kubevirt-ui/kubevirt-api/console';
 import {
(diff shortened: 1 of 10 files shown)


src/views/virtualmachines/list/components/VirtualMachineRow/components/NetworkUsage.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+  const totalTransferred = xbytes(networkUsage || 0, {
+    fixed: 0,
+    iec: true,
+  });
+
+  return <div>{totalTransferred}ps</div>;
+};
+
+export default NetworkUsage;
```
metalice, line 25: here u dont use memo?

src/views/virtualmachines/list/components/VirtualMachineRow/VirtualMachineRowLayout.tsx
```diff
@@ -32,6 +36,9 @@ const VirtualMachineRowLayout: React.FC<
 > = ({ activeColumnIDs, obj, rowData: { ips, isSingleNodeCluster, node, vmim } }) => {
   const selected = isVMSelected(obj);
 
+  const vmName = getName(obj);
```
metalice, line 39: what about usememo here? as if ips, node, vmim, isSingle.. getting values it will run

src/views/virtualmachines/list/hooks/useVMMetrics.ts
```diff
@@ -0,0 +1,132 @@
...
+  setVMMemoryUsage,
+  setVMNetworkUsage,
+} from '../metrics';
+
+import { getVMListQueries, VMListQueries } from './constants';
+import { getVMNamesFromPodsNames } from './utils';
+
+const useVMMetrics = () => {
+  const [activeNamespace] = useActiveNamespace();
+  const allNamespace = activeNamespace === ALL_NAMESPACES_SESSION_KEY;
```
metalice, line 27: please check if adding it to useMemo reduce renders

PR #2305, opened 2024-12-09
CNV-52273: Virtualization icon is bigger than others in perspective switcher

lkladnit:
## 📝 Resize Virt icon to same size as the rest of switcher icons

This fixes [CNV-52273: Virtualization icon is bigger than others in perspective switcher](https://issues.redhat.com/browse/CNV-52273) bug.

**Before**
image

**After**
Screenshot 2024-12-09 at 21 53 53

### DIFF
--- a/src/perspective/virtualization-icon.tsx
+++ b/src/perspective/virtualization-icon.tsx
@@ -5,10 +5,10 @@ export default () => (
     <svg
       aria-hidden="true"
       fill="currentColor"
-      height="2em"
+      height="1em"
       role="img"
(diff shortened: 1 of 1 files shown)


avivtur:
@lkladnit the size are better now, it's not aligned properly and that it's smaller the icon is just a blur to me, maybe making the others big as this was?

lkladnit:
> @lkladnit the size are better now, it's not aligned properly and that it's smaller the icon is just a blur to me, maybe making the others big as this was?

Agree, IMO looks way better
image

PR #2307, opened 2024-12-10
CNV-51530: hide virtctl on managed udn namespaces

upalatucci:
## 📝 Description

Hide the virtctl command for VMs on udn managed namespaces
Hide the `Internal FQDN` interfaces

## 🎥 Demo

**Before**
Screenshot 2024-12-10 at 12 22 11

**After**
Screenshot 2024-12-10 at 14 47 24

Screenshot 2024-12-10 at 12 21 41

### DIFF
--- a/src/utils/components/SSHAccess/components/ConsoleOverVirtctl.tsx
+++ b/src/utils/components/SSHAccess/components/ConsoleOverVirtctl.tsx
@@ -1,11 +1,19 @@
 import React, { FC } from 'react';
 import { Trans } from 'react-i18next';
-import { Link } from 'react-router-dom-v5-compat';
 
+import UserDefinedNetworkModel from '@kubevirt-ui/kubevirt-api/console/models/UserDefinedNetworkModel';
 import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import ExternalLink from '@kubevirt-utils/components/ExternalLink/ExternalLink';
 import { getConsoleVirtctlCommand } from '@kubevirt-utils/components/SSHAccess/utils';
 import { documentationURL } from '@kubevirt-utils/constants/documentation';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
+import {
+  ClusterUserDefinedNetworkModelGroupVersionKind,
+  UserDefinedNetworkModelGroupVersionKind,
+} from '@kubevirt-utils/models';
+import { getName, getNamespace } from '@kubevirt-utils/resources/shared';
+import useNamespaceUDN from '@kubevirt-utils/resources/udn/hooks/useNamespaceUDN';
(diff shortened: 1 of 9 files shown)


src/utils/components/SSHAccess/components/ConsoleOverVirtctl.tsx
```diff
@@ -27,43 +36,65 @@ type ConsoleOverVirtctlProps = {
 const ConsoleOverVirtctl: FC<ConsoleOverVirtctlProps> = ({ vm }) => {
   const { t } = useKubevirtTranslation();
 
+  const [udn, clusterUDN] = useNamespaceUDN(getNamespace(vm));
+
+  const namespaceManagedByUDN = !isEmpty(udn) || !isEmpty(clusterUDN);
```
metalice, line 41: why not return this also from the hook? u use it also in another file

src/utils/utils/matchSelector.ts
```diff
@@ -0,0 +1,54 @@
+import {
```
metalice, line 1: please explain this file

src/utils/components/SSHAccess/components/ConsoleOverVirtctl.tsx
```diff
@@ -27,43 +36,65 @@ type ConsoleOverVirtctlProps = {
...
+              <>
+                <Trans t={t}>
+                  <div>
+                    Open an SSH connection with the VM using the cluster API server. You must be
+                    able to access the API server and have virtctl command line tool installed.
+                  </div>
+                  <br />
+                  <div>
+                    For more details, see{' '}
+                    <Link to={documentationURL.VIRTCTL_CLI}>Installing virtctl</Link> in Getting
```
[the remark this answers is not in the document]
upalatucci: Sure

src/utils/utils/matchSelector.ts
upalatucci: To understand if ClusterUDN is managing a namespace, I have to go through its `matchLabels` and `matchExpressions` and verify that the project labels verify both. 

The matchExpression that I use with ClusterUDN in the networking plugin is

```
{key: 'kubernetes.io/metadata.name', operator: 'In', values: [projectA, projectB]}
```

But in general ClusterUDN can also have `matchLabels` and different kind of `matchExpressions` so to support ClusterUDN created through command line or yaml I wrote this function 


I created this file in the utils as this kind of validation is [common in k8s](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#resources-that-support-set-based-requirements) for other things so I think it can be helpful in the future.

upalatucci: The k8s field that has `matchLabels` and `matchExpressions` is called commonly `Selector` so this is why I named the file `matchSelector`.

The flow is:

`matchLabels` requirements can be transformed to expressions simply by using the `Equal` operator. So I did that to simplify the function. 

After that I just had to transform operators to functions like `includes` and `! includes`

PR #2328, opened 2024-12-18
CNV-53430: make examples compatible with s390x arch clusters

upalatucci:
## 📝 Description

This piece is not compatible with s390x arch clusters.

```
            features:
              acpi: {}
              smm:
                enabled: true
            firmware:
              bootloader:
                efi: {}
            machine:
              type: q35
```

window.SERVER_FLAGS.nodeArchitectures has info on the arch in the clusters

### DIFF
--- a/src/templates/vm-yaml.ts
+++ b/src/templates/vm-yaml.ts
@@ -1,71 +1,113 @@
 import VirtualMachineModel from '@kubevirt-ui/kubevirt-api/console/models/VirtualMachineModel';
+import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import { isClusterOnS390X, safeDump } from '@kubevirt-utils/utils/utils';
 
-export const defaultVMYamlTemplate = `
-apiVersion: ${VirtualMachineModel.apiGroup}/${VirtualMachineModel.apiVersion}
-kind: ${VirtualMachineModel.kind}
-metadata:
-  name: example
-  annotations:
-    description: VM example
-  labels:
-    app: example
-    os.template.kubevirt.io/fedora: 'true'
-spec:
-  runStrategy: Halted
-  template:
-    metadata:
-      annotations:
-        vm.kubevirt.io/flavor: small
-        vm.kubevirt.io/os: fedora
-        vm.kubevirt.io/workload: server
-      labels:
-        kubevirt.io/domain: example
-        kubevirt.io/size: small
-    spec:
-      domain:
-        cpu:
-          cores: 1
-          sockets: 1
-          threads: 1
-        devices:
-          disks:
-            - disk:
-                bus: virtio
-              name: rootdisk
-            - disk:
-                bus: virtio
-              name: cloudinitdisk
-          interfaces:
-            - masquerade: {}
-              model: virtio
-              name: default
-          networkInterfaceMultiqueue: true
-          rng: {}
-        features:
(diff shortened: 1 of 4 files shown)


src/utils/utils/utils.ts
```diff
@@ -157,3 +157,9 @@ export const removeDockerPrefix = (image: string) => image?.replace(DOCKER_PREFI
 
 export const isAllNamespaces = (namespace: string) =>
   !namespace || namespace === ALL_NAMESPACES || namespace === ALL_NAMESPACES_SESSION_KEY;
+
+const S390X_ARCH = 's390x';
```
metalice, line 161: move to a general const file please

src/templates/vm-yaml.ts
```diff
@@ -1,71 +1,119 @@
...
+      },
+    },
+  },
+};
+
+if (isClusterOnS390X) {
+  delete defaultVM.spec.template.spec.domain.machine;
+  delete defaultVM.spec.template.spec.domain.firmware;
+  delete defaultVM.spec.template.spec.domain.features;
+}
```
metalice, line 117: why delete and not use optional spreading? ...(isSomething && {something: something}) , they all 3 fields inside domain

src/utils/utils/utils.ts
upalatucci: done

src/templates/vm-yaml.ts
upalatucci: fixed

src/utils/utils/utils.ts
```diff
@@ -157,3 +159,17 @@ export const removeDockerPrefix = (image: string) => image?.replace(DOCKER_PREFI
 
 export const isAllNamespaces = (namespace: string) =>
   !namespace || namespace === ALL_NAMESPACES || namespace === ALL_NAMESPACES_SESSION_KEY;
+
+export const isClusterOnS390X = !isEmpty(
+  (window as any).SERVER_FLAGS?.nodeArchitectures?.find((arch) => arch === S390X_ARCH),
```
spadgett, line 164: It's worth noting that this isn't a public API and could change. I wouldn't anticipate it changing, but it could. I'd suggest opening a management console issue to provide a supported API (and you can fallback to `SERVER_FLAGS` for older console versions).

@jhadvig @TheRealJon FYI, the kubevirt plugin is using `window.SERVER_FLAGS.nodeArchitectures`, so we should avoid removing or changing that if possible.

TheRealJon: Wouldn't this issue apply to any SERVER_FLAG value? These are just available globally at runtime and could be used by any plugin, so making breaking changes to them is not very safe.

spadgett: Yes. `SERVER_FLAGS` are internal APIs. Any use by plugins could break in a future version.

We mention that CSS classes aren't APIs in the docs. We probably need to call out JavaScript globals as well, and perhaps anything in Redux.

metalice (approved):

upalatucci:
@metalice @spadgett @TheRealJon  i talked with multiple people and observed the templates in `s390x` and `amd64` clusters. Templates do not have in their structure   `features` `firmware` and `machine` fields, so we can remove that without logic

upalatucci:
closing in favor of: https://github.com/kubevirt-ui/kubevirt-plugin/pull/2329

PR #2332, opened 2024-12-25
CNV-54208: The dropdown menu of "Actions" is not closed by clicking somewhere else

avivtur:
## 📝 Description

This regression was introduced when we changed the actions dropdown to a Popper component that needs to add an effect to make this behavior happen (which comes out of the box when using Select component with onOpenChange prop).
Enhancing the useClickOutside hook that we already have for this cases to get more than one ref (the menu and the toggle components refs) to solve this issue.

## 🎥 Demo

Before:

https://github.com/user-attachments/assets/6c44b35f-0351-4e42-991b-cc86e14ba2a4

After:

https://github.com/user-attachments/assets/cfb3bc18-0853-4ccc-afa4-79cfedc185a1


.gitignore
```diff
@@ -21,4 +21,5 @@ console-extensions.md
 .idea/
 scripts/ca.crt
 scripts/console-client-secret
-*.json-*
\ No newline at end of file
+*.json-*
+.codegpt
```
metalice, line 25: ?

metalice (approved):

PR #2336, opened 2024-12-31
CNV-53452: VirtualMachines page crashes for nonpriv user when user hasn't created a project

avivtur:
## 📝 Description

Tree view crashed to non priv users when they have no projects, fixing this and disabling the switch on the toolbar when no projects (system projects can be found but they are filtered by default as they shouldn't appear at all).

## 🎥 Demo
before:

https://github.com/user-attachments/assets/2b24e0ed-40d0-4e26-b550-c04be58b143a

after:

https://github.com/user-attachments/assets/2b6af05e-9f1d-4dc1-ac6a-fa606f496912

### DIFF
--- a/src/views/virtualmachines/tree/components/TreeViewToolbar.tsx
+++ b/src/views/virtualmachines/tree/components/TreeViewToolbar.tsx
@@ -19,10 +19,15 @@ import { HIDE, SHOW, SHOW_EMPTY_PROJECTS_KEY, TREE_VIEW_SEARCH_ID } from '../uti
 
 type TreeViewToolbarProps = {
   closeComponent: ReactNode;
+  isSwitchDisabled: boolean;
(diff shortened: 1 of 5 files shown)


src/views/virtualmachines/tree/components/TreeViewToolbar.tsx
```diff
@@ -51,8 +56,9 @@ const TreeViewToolbar: FC<TreeViewToolbarProps> = ({ closeComponent, onSearch })
               </SplitItem>
               <SplitItem isFilled />
               <Switch
-                checked={showEmptyProjects === HIDE}
+                checked={isSwitchDisabled ? false : showEmptyProjects === HIDE}
```
metalice, line 59: !isSwitchDisabled && showEmptyProjects === HIDE

metalice (approved):

PR #2337, opened 2024-12-31
[wip] Avoid blinking tree view

avivtur:
## 📝 Description

Wrapping both VM list page and VM details page components with a tree-view component gives 2 different instances, which causes the tree-view unmount and mount every time we move from ns item to VM item and vice versa.
Manipulating the routes at our side is help to wrap both components with just one instance of tree-view.
NOTE: the ~new path to create a VM from YAML is not working well as `:name` identifier is catching it as a single VM name.


avivtur:
/hold
Changes causing tests failures as yaml create VM is missing.

metalice:
lgtm by me but i want another set of eyes - @upalatucci could you ptal

PR #2339, opened 2025-01-06
CNV-53248: Add registry credentials to the template wizard flow

pcbailey:
## 📝 Description

This PR fixes an issue where the registry credentials entered in the quick create template flow don't persist in the customize template flow. It also triggers deletion of the created Secret when the VM is deleted.

Jira: https://issues.redhat.com/browse/CNV-53248

## 🎥 Demo

[registry-credentials-persist-to-customize-flow--AFTER--2025-01-06 08-50.webm](https://github.com/user-attachments/assets/4c668f29-0949-4c34-ace2-a33b0484e699)

### DIFF
--- a/src/utils/components/DiskModal/DiskModal.tsx
+++ b/src/utils/components/DiskModal/DiskModal.tsx
@@ -12,6 +12,7 @@ import { V1DiskModalProps } from './utils/types';
 const DiskModal: FC<V1DiskModalProps> = ({
   createDiskSource,
   createdPVCName,
+  defaultFormValues,
   editDiskName,
   isOpen,
   onClose,
@@ -34,6 +35,7 @@ const DiskModal: FC<V1DiskModalProps> = ({
   return (
     <Modal
       createDiskSource={createDiskSource}
+      defaultFormValues={defaultFormValues}
       editDiskName={editDiskName}
       isCreated={!isEmpty(createdPVCName)}
       isOpen={isOpen}
--- a/src/utils/components/DiskModal/components/DiskSourceSelect/components/DiskSourceContainer/DiskSourceContainer.tsx
+++ b/src/utils/components/DiskModal/components/DiskSourceSelect/components/DiskSourceContainer/DiskSourceContainer.tsx
@@ -3,49 +3,109 @@ import { FieldPath, useFormContext } from 'react-hook-form';
 
 import { V1DiskFormState } from '@kubevirt-utils/components/DiskModal/utils/types';
 import FormGroupHelperText from '@kubevirt-utils/components/FormGroupHelperText/FormGroupHelperText';
+import { FormPasswordInput } from '@kubevirt-utils/components/FormPasswordInput/FormPasswordInput';
+import { FormTextInput } from '@kubevirt-utils/components/FormTextInput/FormTextInput';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
 import { OS_NAME_TYPES } from '@kubevirt-utils/resources/template';
 import { isUpstream } from '@kubevirt-utils/utils/utils';
-import { FormGroup, TextInput, ValidatedOptions } from '@patternfly/react-core';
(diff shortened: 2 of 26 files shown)


src/utils/components/DiskModal/DiskModal.tsx
```diff
@@ -32,17 +33,19 @@ const DiskModal: FC<V1DiskModalProps> = ({
...
-      editDiskName={editDiskName}
-      isCreated={!isEmpty(createdPVCName)}
-      isOpen={isOpen}
-      onClose={onClose}
-      onSubmit={onSubmit}
-      onUploadedDataVolume={onUploadedDataVolume}
-      pvc={pvc}
-      vm={vm}
-    />
+    <WizardVMContextProvider>
```
avivtur, line 36: I believe the disk modal is general, why adding the wizard context to it?

src/utils/components/DiskModal/components/DiskSourceSelect/components/DiskSourceContainer/DiskSourceContainer.tsx
```diff
@@ -3,49 +3,109 @@ import { FieldPath, useFormContext } from 'react-hook-form';
...
+        </FormGroupHelperText>
+      </FormGroup>
+      {!isEphemeralDiskSource && (
+        <>
+          <FormGroup
+            className="disk-source-form-group"
+            fieldId={diskSourceUsernameFieldID}
+            label={t('Username')}
+          >
+            <FormTextInput
```
avivtur, line 75: Here you are using FormTextInput and above there's FormInput, what is the difference between them? can we use just one of them to remove one import?

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/hooks/useCreateDrawerForm.tsx
```diff
@@ -289,6 +289,12 @@ const useCreateDrawerForm = (
           template,
           ANNOTATIONS.displayName,
         );
+
+        ensurePath(tabsDataDraft, 'disks.disks.rootDiskRegistryCredentials');
```
avivtur, line 293: the path you ensure here doesn't match the path you assign below

avivtur (commented):
Great job!

src/utils/components/DiskModal/DiskModal.tsx
pcbailey: I need access to the registry credentials in `RegistryDiskModal`. I have the `useRegistryCredentials` hook for that, but it requires access to the context. It was either this, prop drill the credentials, or create a special component and muddy the modal selection code.

I'm open to other suggestions if you have them.

src/utils/components/DiskModal/components/DiskSourceSelect/components/DiskSourceContainer/DiskSourceContainer.tsx
pcbailey: The comment in the FormTextInput file says: "PatternFly changes the signature of the 'onChange' handler for input elements. This causes issues with React Hook Form as it expects the default signature for an input element. So we have to create this wrapper component that takes care of converting these signatures for us."

Fixed!

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/hooks/useCreateDrawerForm.tsx
pcbailey: Fixed!

src/utils/components/DiskModal/DiskModal.tsx
upalatucci: The Diskmodal has `onSubmit` callback that you can use to update the credentials. Is this a good suggestion?

upalatucci:
/hold

we have to add ownerReferece to the secret in order to delete the secret when the VM is deleted. Right now the secret does not have it and we manually do that in the pr

upalatucci (approved):

PR #2340, opened 2025-01-06
CNV-54179: Improve folder select

avivtur:
## 📝 Description

Show create option when at all times, and fix text of create option.

## 🎥 Demo

After:
fixed-folder-select
fixed-folder-select2


src/utils/components/SelectTypeahead/SelectTypeahead.tsx
```diff
@@ -43,35 +44,40 @@ const SelectTypeahead: FC<SelectTypeaheadProps> = ({
...
-
-      // If no option matches the filter exactly, display creation option
-      if (!initialOptions?.some((option) => option.value === filterValue) && canCreate) {
-        newSelectOptions = [...newSelectOptions, getCreateOption(filterValue, canCreate)];
-      }
+    if (
+      !canCreate &&
+      newSelectOptions.length === 0 &&
+      filterValue &&
+      !newSelectOptions.some((option) => option.value === NOT_FOUND)
```
pcbailey, line 74: IIUC this option wouldn't never be reached if the second condition `newSelectOptions.length === 0` is true. Am I missing something?

pcbailey:
It's just my opinion, but 'Create folder ""' looks weird in the screenshot.

avivtur:
> It's just my opinion, but 'Create folder ""' looks weird in the screenshot.

Suggestions?

src/utils/components/SelectTypeahead/SelectTypeahead.tsx
avivtur: you're right, fixed

src/utils/components/SelectTypeahead/SelectTypeahead.tsx
upalatucci, line 1: i can see there is an occurrency like 

```
      if (value === CREATE_NEW) {
```

Should we put 
```
   if ([CREATE_NEW, NOT_FOUND].includes(value)
```

??

avivtur: where this should go?

metalice: @avivtur - please go over this comment and adjust if needed in a followup pr

metalice (approved):

PR #2349, opened 2025-01-09
Secrets no dots

upalatucci:
## 📝 Description

> Add a brief description

## 🎥 Demo

Screenshot 2025-01-09 at 15 45 00


locales/en/plugin__kubevirt-plugin.json
```diff
@@ -1067,6 +1067,7 @@
   "seconds": "seconds",
   "Secret name": "Secret name",
   "Secret name must be unique in this namespace.": "Secret name must be unique in this namespace.",
+  "Secret name must not contain dots": "Secret name must not contain dots",
```
pcbailey, line 1070: The proper name for the character is "period", so I think we should change "dots" to "periods" here.

metalice: @upalatucci

upalatucci: @metalice done

metalice (approved):

PR #2353, opened 2025-01-13
CNV-49026: Change 'hide deprecated bootable volumes' filter to 'show deprecated bootable volumes'

pcbailey:
## 📝 Description

This PR changes the "hide deprecated bootable volumes" filter in the Bootable volumes list and on the create from InstanceTypes page to "show deprecated bootable volumes" and hides the deprecated volumes by default.

Jira: https://issues.redhat.com/browse/CNV-49026

## 🎥 Demo

**Note**: RHEL8 is NOT deprecated. There were no deprecated volumes on this cluster, so I hard-coded RHEL8 to be considered as deprecated for the demo.

[show-deprecated-boot-volumes--AFTER--2025-01-13 07-25.webm](https://github.com/user-attachments/assets/efee1453-084c-4f17-82ed-ebcb56f9db57)


src/utils/resources/bootableresources/hooks/useHideDeprecatedBootableVolumes.ts
```diff
@@ -17,22 +13,15 @@ const useHideDeprecatedBootableVolumes: UseHideDeprecatedBootableVolumes = (
...
-  const { setParam } = useURLParams();
-
   useEffect(() => {
     if (currentTab !== CREATE_VM_TAB.INSTANCE_TYPES && currentTab !== undefined) return null;
 
-    // This is to select the 'Hide deprecated bootable volumes' filter by default
-    onFilterChange(HIDE_DEPRECATED_BOOTABLE_VOLUMES, {
-      all: [HIDE_DEPRECATED_BOOTABLE_VOLUMES],
-      selected: [HIDE_DEPRECATED_BOOTABLE_VOLUMES],
+    // Calculate the filtered bootable volumes once causing deprecated volumes to be hidden by default
```
upalatucci, line 19: I think we don't need this hook anymore right ?

pcbailey: In my testing the deprecated volumes show on load without the hook. If you select and then deselect the "Show deprecated bootable volumes" filter it hides them as expected. The hook triggers the filtering mechanism, which allows them to be hidden by default.

PR #2356, opened 2025-01-13
CNV-51653: Fix bootable volume column sorting on create VM from IT page

pcbailey:
## 📝 Description

This PR fixes the sorting of bootable volumes table columns on the create VM from InstanceTypes page. The sorting algorithm wasn't accounting for the namespace column which is present when All Projects is selected, which threw off which column was being sorted by one index.

Jira: https://issues.redhat.com/browse/CNV-51653

### Before

[bootable-volumes-sort--BEFORE--2025-01-13 06-55.webm](https://github.com/user-attachments/assets/e0d43311-59d6-4822-ad18-2ea7a463a0f8)

### After

[bootable-volumes-sort--AFTER--2025-01-13 06-52.webm](https://github.com/user-attachments/assets/8ea78e70-51a5-49b0-83b6-d7f5505d287c)

### DIFF
--- a/src/views/catalog/CreateFromInstanceTypes/components/BootableVolumeList/BootableVolumeList.tsx
+++ b/src/views/catalog/CreateFromInstanceTypes/components/BootableVolumeList/BootableVolumeList.tsx
@@ -91,6 +91,7 @@ const BootableVolumeList: FC<BootableVolumeListProps> = ({
     pvcSources,
     volumeSnapshotSources,
     pagination,
+    volumeListNamespace === ALL_PROJECTS,
   );
 
   useEffect(() => {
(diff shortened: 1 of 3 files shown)


metalice (approved):

src/views/catalog/CreateFromInstanceTypes/components/BootableVolumeList/BootableVolumeList.tsx
```diff
@@ -91,6 +91,7 @@ const BootableVolumeList: FC<BootableVolumeListProps> = ({
     pvcSources,
     volumeSnapshotSources,
     pagination,
+    volumeListNamespace === ALL_PROJECTS,
```
upalatucci, line 94: Shouldn't this be !==?

PR #2368, opened 2025-01-16
CNV-55113: rollback migration

upalatucci:
## 📝 Description

Make sure to edit back the VM when storageclass migration is cancelled.

VM has a status of the ongoing storageclass migration reporting the original and the destination PVC name. 

Make sure to replace the datavolume templates names and pvcs claim names from the destination back to the original ones 

Make sure to distinguish 'Cancel storage migration' and 'Cancel compute migration' as we have to do two different actions. 

With the storage migration we don't have to remove the VMIM as it's not created by us but by the virt controller. Rollbacking the VM status to the previous one, the virt controller will cancel the migration

Screenshot 2025-01-20 at 10 14 04
Screenshot 2025-01-20 at 10 14 43


src/views/virtualmachines/actions/components/VirtualMachineMigration/utils.ts
```diff
@@ -139,6 +138,13 @@ const createPatchData = (
         path: `/spec/dataVolumeTemplates/${dataVolumeIndex}/metadata/name`,
         value: destinationDataVolumeName,
       });
+
+      if (destinationDV?.spec?.storage?.storageClassName)
```
pcbailey, line 142: Do we have selectors for DVs? If not, maybe we should create some and use them.

pcbailey (commented):
Looks good! I just had a few small comments.

src/views/virtualmachines/actions/utils.ts
```diff
@@ -0,0 +1,95 @@
...
+          path: `/metadata/annotations/${originalClaimAnnotation.replace('/', '~1')}`,
+        },
+        {
+          op: 'replace',
+          path: `/spec/template/spec/volumes/${volumeIndex}/persistentVolumeClaim/claimName`,
+          value: originalClaimStorageMigration,
+        },
+      ];
+    })
+    .filter(Boolean)
```
metalice, line 91: why u need filter here? seems like it always be true

src/views/virtualmachines/actions/utils.ts
```diff
@@ -0,0 +1,95 @@
...
+          value: originalClaimStorageMigration,
+        },
+        {
+          op: 'replace',
+          path: `/spec/dataVolumeTemplates/${dvIndex}/metadata/name`,
+          value: originalClaimStorageMigration,
+        },
+      ];
+    })
+    .filter(Boolean)
```
metalice, line 65: same as comment below

src/views/virtualmachines/actions/utils.ts
upalatucci: fixed with `reduce`

metalice (approved):

PR #2372, opened 2025-01-18
Adding cypress infrastructure for all tests

lkladnit:
## 📝 Description

This PR adds all cypress related strings, enums, functions and objects that will be used in testing.
No tests were added, but the existing tests were updated to use imported functions, objects and selectors
All existing tests passed locally after updating.

Screenshot 2025-01-19 at 19 22 29


.cleanup
metalice, line 1: why naming it .cleanup , usually bash script are {name}.sh

lkladnit: changed names to setup.sh and cleanup.sh

PR #2389, opened 2025-01-27
CNV-53271: select one namespace in catalog page

upalatucci:
## 📝 Description

In catalog, non-admin users do not see the `Create VirtualMachine` button enabled when `All namespaces` is selected. Don't really know what to do to enable that, hey have to select a namespace. 

Using `default` as a default namespace does not work as non-admin users may not have permissions to that namespace as it happens in `https://console.redhat.com/openshift/sandbox`

Solution:

select a default namespace. If the user have access to the default namespace, select it, if not, select the first non-system namespace and if not found, select the first namespace that we can find

## 🎥 Demo

**Before**

Template catalog
Screenshot 2025-01-27 at 10 14 08

Instance Type
https://github.com/user-attachments/assets/1a69cbc0-c5b5-4d0f-8a0f-09d87ca0fe74

**After**

https://github.com/user-attachments/assets/6dbfab8b-c3ad-4e56-9746-207893dc2a4b

### DIFF
--- a/src/views/catalog/Catalog.tsx
+++ b/src/views/catalog/Catalog.tsx
@@ -5,11 +5,15 @@ import { useSignals } from '@preact/signals-react/runtime';
 
 import CreateVMHorizontalNav from './CreateVMHorizontalNav/CreateVMHorizontalNav';
 import CustomizeInstanceTypeVirtualMachine from './CustomizeInstanceType/CustomizeInstanceTypeVirtualMachine';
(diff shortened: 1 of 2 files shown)


avivtur:
@upalatucci the bug is talking about giving a proper error message, enforcing a project is not a good idea, what if it's a new cluster and the user has not created a namespace yet?
Let's think of a new approach since the error is coming from console

PR #2408, opened 2025-01-31
CNV-55107: loading on vm change

upalatucci:
## 📝 Description

On VM change, we wait only for the VM to be fetched. We should also wait for the instancetype spec expansion request
Changing VMs without waiting will make the screen flickering between the previous instance and the other 
## 🎥 Demo

**Before**

https://github.com/user-attachments/assets/fa80e07b-923b-4708-b33c-994a44d58886

**After**

https://github.com/user-attachments/assets/a752b893-9d51-4a88-ba03-31f26f4e2b31

### DIFF
--- a/src/utils/components/HorizontalNavbar/HorizontalNavbar.tsx
+++ b/src/utils/components/HorizontalNavbar/HorizontalNavbar.tsx
@@ -7,18 +7,26 @@ import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
 import { isEmpty } from '@kubevirt-utils/utils/utils';
 import { Nav, NavList } from '@patternfly/react-core';
 
+import StateHandler from '../StateHandler/StateHandler';
+
 import useDynamicPages from './utils/useDynamicPages';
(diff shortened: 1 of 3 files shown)


src/utils/components/HorizontalNavbar/HorizontalNavbar.tsx
```diff
@@ -73,12 +81,14 @@ const HorizontalNavbar: FC<HorizontalNavbarProps> = ({ instanceTypeExpandedSpec,
           return (
             <Route
               Component={(props) => (
-                <Component
-                  instanceTypeExpandedSpec={instanceTypeExpandedSpec}
-                  obj={vm}
-                  params={params}
-                  {...props}
-                />
+                <StateHandler loaded={!expandedSpecLoading} withBullseye>
```
avivtur, line 84: how this works for VMs created with templates or YML?

upalatucci: works great. THe expandedSpecLoading is `false` from the start with template vms

PR #2413, opened 2025-02-03
fix: use v4.18.0-27 uploader image

codingben:
Use image from unreleased registry that
will contain disk-uploader binary.

### DIFF
--- a/src/utils/components/ExportModal/constants.ts
+++ b/src/utils/components/ExportModal/constants.ts
@@ -92,4 +92,4 @@ export const ALREADY_CREATED_ERROR_CODE = 409;
 
 export const UPSTREAM_UPLOADER_IMAGE = 'quay.io/kubevirt/tekton-tasks:v0.23.0';
 export const DOWNSTREAM_UPLOADER_IMAGE =
-  'registry.redhat.io/container-native-virtualization/kubevirt-tekton-tasks-create-datavolume-rhel9:v4.17';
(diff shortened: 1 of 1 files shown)


codingben:
@upalatucci Can you please review it?

src/utils/components/ExportModal/constants.ts
```diff
@@ -92,4 +92,4 @@ export const ALREADY_CREATED_ERROR_CODE = 409;
 
 export const UPSTREAM_UPLOADER_IMAGE = 'quay.io/kubevirt/tekton-tasks:v0.23.0';
 export const DOWNSTREAM_UPLOADER_IMAGE =
-  'registry.redhat.io/container-native-virtualization/kubevirt-tekton-tasks-create-datavolume-rhel9:v4.17';
+  'registry-proxy.engineering.redhat.com/rh-osbs/container-native-virtualization-kubevirt-tekton-tasks-create-datavolume-rhel9:v4.18.0-27';
```
dominikholler, line 95: What is the reason of lock to a fixed build instead of using the flating tag `v4.18` ?

codingben: I'd prefer to stick to a specific image build that will work, instead of allowing v4.18 that will pull any build image that may be broken.

metalice (approved):

src/utils/components/ExportModal/constants.ts
dominikholler: - How can we know, if a specific build will be published at all?

- An externally available image has to be used

PR #2419, opened 2025-02-04
CNV-55889: SSH over LoadBalancer creates Service with incorrect selector

avivtur:
## 📝 Description

Replacing load balancer service selector according to: https://issues.redhat.com/browse/CNV-55889 

## 🎥 Demo

Before:
selector-service-b4

After:
correct-selector-service

### DIFF
--- a/src/utils/components/SSHAccess/SSHAccess.tsx
+++ b/src/utils/components/SSHAccess/SSHAccess.tsx
@@ -1,7 +1,7 @@
 import React, { FC } from 'react';
 
 import { IoK8sApiCoreV1Service } from '@kubevirt-ui/kubevirt-api/kubernetes';
-import { V1VirtualMachine, V1VirtualMachineInstance } from '@kubevirt-ui/kubevirt-api/kubevirt';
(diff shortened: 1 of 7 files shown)


upalatucci (approved):

sarahbx:
@avivtur when using `vm.kubevirt.io/name` the VM must be selected instead of the VMI

avivtur:
> @avivtur when using `vm.kubevirt.io/name` the VM must be selected instead of the VMI

@sarahbx what do you mean?

sarahbx:
> > @avivtur when using `vm.kubevirt.io/name` the VM must be selected instead of the VMI
> 
> @sarahbx what do you mean?

In your example above, the VMI is referenced, not the VM.
VMI name: `fedora-sapphire-piranha-33-{random-suffix}`
VM name: `fedora-sapphire-piranha-33`

So the Pod selector would be:
`vm.kubevirt.io/name: fedora-sapphire-piranha-33`

upalatucci (approved):

PR #2429, opened 2025-02-05
CNV-55086: persist vm tab on treeview vm switch

upalatucci:
## 📝 Description

When switching between vms, persist the tab

During the switch make sure to visualize all the tabs and have the loading inside them.
Previously there were two StateHandler for the loading. As you can see from the demo, there was a loading before displaying the tabs and another loading for the tab content. Now we display the tabs and the loading inside so the switch between vms does not have so much shifts

## 🎥 Demo

**Before**

https://github.com/user-attachments/assets/61ff03c4-5974-41af-a05b-6b369de4c674

**After**

https://github.com/user-attachments/assets/afe75262-3184-4fc3-b6db-85ea16e81767

### DIFF
--- a/src/utils/components/HorizontalNavbar/HorizontalNavbar.tsx
+++ b/src/utils/components/HorizontalNavbar/HorizontalNavbar.tsx
@@ -15,15 +15,17 @@ import { NavPageKubevirt, trimLastHistoryPath } from './utils/utils';
 import './horizontal-nav-bar.scss';
 
 type HorizontalNavbarProps = {
-  expandedSpecLoading?: boolean;
+  error?: any;
   instanceTypeExpandedSpec?: V1VirtualMachine;
+  loaded?: boolean;
   pages: NavPageKubevirt[];
   vm?: V1VirtualMachine;
 };
 
 const HorizontalNavbar: FC<HorizontalNavbarProps> = ({
-  expandedSpecLoading,
+  error,
   instanceTypeExpandedSpec,
+  loaded,
   pages,
   vm,
 }) => {
@@ -81,7 +83,7 @@ const HorizontalNavbar: FC<HorizontalNavbarProps> = ({
           return (
             <Route
               Component={(props) => (
-                <StateHandler loaded={!expandedSpecLoading} withBullseye>
+                <StateHandler error={error} loaded={loaded} withBullseye>
                   <Component
                     instanceTypeExpandedSpec={instanceTypeExpandedSpec}
                     obj={vm}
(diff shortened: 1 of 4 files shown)


src/views/virtualmachines/details/VirtualMachineNavPage.tsx
```diff
@@ -42,16 +41,15 @@ const VirtualMachineNavPage: React.FC<VirtualMachineDetailsPageProps> = ({
...
-        <div className="VirtualMachineNavPage--tabs__main">
-          <HorizontalNavbar
-            expandedSpecLoading={expandedSpecLoading}
-            instanceTypeExpandedSpec={instanceTypeExpandedSpec}
-            pages={pages}
-            vm={vm}
-          />
-        </div>
-      </StateHandler>
+      <div className="VirtualMachineNavPage--tabs__main">
```
metalice, line 44: why removing state handler?

src/views/virtualmachines/tree/utils/utils.tsx
```diff
@@ -175,26 +176,29 @@ const createAllNamespacesTreeItem = (
   return allNamespacesTreeItem;
 };
 
+const getVMInfoFromPathname = (pathname: string) => {
+  const splittedPathname = pathname.split('/');
+  const currentVMTab = splittedPathname?.[6] || '';
+  const vmName = splittedPathname[5];
```
metalice, line 182: add ?.

src/views/virtualmachines/tree/utils/utils.tsx
```diff
@@ -175,26 +176,29 @@ const createAllNamespacesTreeItem = (
...
+  const vmNamespace = splittedPathname[3];
```
metalice, line 183: ?.

src/views/virtualmachines/tree/utils/utils.tsx
```diff
@@ -175,26 +176,29 @@ const createAllNamespacesTreeItem = (
   return allNamespacesTreeItem;
 };
 
+const getVMInfoFromPathname = (pathname: string) => {
+  const splittedPathname = pathname.split('/');
```
pcbailey, line 180: nitpick: 'splitted' is an archaic form. Use 'split' instead: `splitPathname`

src/utils/components/HorizontalNavbar/HorizontalNavbar.tsx
```diff
@@ -15,15 +15,17 @@ import { NavPageKubevirt, trimLastHistoryPath } from './utils/utils';
 import './horizontal-nav-bar.scss';
 
 type HorizontalNavbarProps = {
-  expandedSpecLoading?: boolean;
+  error?: any;
```
pcbailey, line 18: Is there any way to narrow this type?

upalatucci: no. The k8sWatch hook use any for the error type so nothing to do about it.

PR #2436, opened 2025-02-06
CNV-54461: [RFE] Add link to VM status on VM list page

avivtur:
## 🎥 Demo
Before:
No popover
vm-status-popover-b4

After:
vm-status-popover


src/views/virtualmachines/list/components/VirtualMachineRow/VirtualMachineRow.tsx
```diff
@@ -26,10 +29,19 @@ const VirtualMachineRow: FC<
...
   const vmNamespace = getNamespace(vm);
   const vmi = getVmi(vmNamespace, vmName);
+  const status = (
+    <>
+      <StatusWithPopover vm={vm} vmi={vmi} />
+      {vm?.status?.printableStatus === printableVMStatus.Migrating && (
+        <VirtualMachineMigrationPercentage vm={vm} />
+      )}
+    </>
+  );
```
upalatucci, line 39: Can't we just have it in the VirtualMachineRowLayout? Why do we need to inject from here to there?

avivtur: I wanted to avoid drilling VMI object to the row, but we can have either way

upalatucci (approved):

PR #2460, opened 2025-02-14
CNV-55963: right click action in the treeview

upalatucci:
## 📝 Description

Add right click action menu in the treeview

I had to fetch the vmim list in order to have an action menu appearing without any loading. 

Can this hit performance? Not really as we already fetch them anyways in the VM list and VM details.
 
Why fetching it again if we fetch it in another place? I had to do that as treeview is in tech preview and can be disabled with a flag. 
When the flag will be removed, maybe we can think of having one place to fetch them one time only. 
I used signals to not have props drilling and in the future we can use the same method to spread everything in the VM list components 

## 🎥 Demo

https://github.com/user-attachments/assets/14967c7a-f7b5-405d-a726-e006ce4b0c77


src/views/virtualmachines/tree/VirtualMachineTreeView.scss
```diff
@@ -1,6 +1,10 @@
 .vms-tree-view {
   overflow-x: hidden;
 
+  .pf-v5-c-tree-view__list-item {
+    position: relative;
```
metalice, line 5: this is the default do we need it? are they setting it differently?

src/views/virtualmachines/tree/components/TreeViewRightClickAction.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+  const [actions] = useVirtualMachineActionsProvider(vm, vmim, isSingleNodeCluster);
+
+  if (!contextMenuTrigger) return null;
+
+  return (
+    <>
+      <Popper
+        popper={
+          <Menu containsFlyout style={treeViewActionMenuStyle}>
```
metalice, line 38: why style and not CN?

src/views/virtualmachines/tree/hooks/useContextMenuListener.tsx
```diff
@@ -0,0 +1,34 @@
+import { useCallback, useLayoutEffect, useState } from 'react';
+
+import { TreeViewDataItem } from '@patternfly/react-core';
+
+type UseContextMenuListenerType = (
+  treeData: TreeViewDataItem[],
+) => [triggerElement: HTMLElement | null, hide: () => void];
+
+const useContextMenuListener: UseContextMenuListenerType = (treeData) => {
```
metalice, line 9: contextMenu is the event, but a better name will be right click?

src/views/virtualmachines/tree/hooks/useTreeViewData.ts
```diff
@@ -49,11 +57,49 @@ export const useTreeViewData = (): UseTreeViewData => {
...
+  );
+
+  vmimMapperSignal.value = memoizedVMIMs;
+
   const memoizedVMs = useMemo(
     () => (isAdmin ? allVMs : Object.values(allowedResources).flatMap((resource) => resource.data)),
     [allVMs, allowedResources, isAdmin],
   );
 
+  vmsSignal.value = memoizedVMs;
```
metalice, line 101: what are signals used here for ?

src/views/virtualmachines/tree/VirtualMachineTreeView.scss
upalatucci: No, the tree items are `position: static`.

src/views/virtualmachines/tree/components/TreeViewRightClickAction.tsx
upalatucci: uh i didn't change it back when I was testing

src/views/virtualmachines/tree/hooks/useContextMenuListener.tsx
upalatucci: okay good

src/views/virtualmachines/tree/hooks/useTreeViewData.ts
upalatucci: Those signals are used in the `TreeViewRightClickAction` component.
When the user right click on the item, from the id element I get VM namespace and name.
With those I search from the `vmsSignal` and `vmimMapperSignal` the VM and relative VMIM to build the action menu.

src/views/virtualmachines/tree/hooks/useTreeviewItemRightClick.tsx
```diff
@@ -0,0 +1,34 @@
+import { useCallback, useLayoutEffect, useState } from 'react';
+
+import { TreeViewDataItem } from '@patternfly/react-core';
+
+type UseTreeviewItemRightClickType = (
```
pcbailey, line 5: nitpick: `UseTreeviewItemRightClickType` -> `UseTreeViewItemRightClick`

It looks the majority of hook types in the code base don't have 'Type' at the end, so I would drop it here to remain consistent.

src/views/virtualmachines/tree/components/TreeViewRightClickAction.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { getVMIMFromMapper } from '@virtualmachines/utils/mappers';
+
+import useTreeviewItemRightClick from '../hooks/useTreeviewItemRightClick';
+import { vmimMapperSignal, vmsSignal } from '../utils/signals';
+
+type TreeViewRightClickActionProps = {
+  treeData: TreeViewDataItem[];
+};
+
+const TreeViewRightClickAction: FC<TreeViewRightClickActionProps> = ({ treeData }) => {
```
pcbailey, line 17: Suggestion: `TreeViewRightClickActionMenu`

src/views/virtualmachines/utils/mappers.ts
```diff
@@ -14,3 +14,6 @@ export type VmimMapper = { [key: string]: { [key: string]: V1VirtualMachineInsta
 
 export const getVMIFromMapper = (vmiMapper: VmiMapper, vm: V1VirtualMachine) =>
   vmiMapper?.mapper?.[getNamespace(vm)]?.[getName(vm)];
+
+export const getVMIMFromMapper = (vmimMapper: VmimMapper, name: string, namespace: string) =>
```
pcbailey, line 18: If it's not too much trouble, would you mind renaming `VmiMapper` and `VmimMapper` in this file to `VMIMapper` and `VMIMMapper` so they match our conventions?

pcbailey (commented):
This is looking great! I just have some minor comments/suggestions.

src/views/virtualmachines/tree/hooks/useTreeviewItemRightClick.tsx
```diff
@@ -0,0 +1,39 @@
+import { useCallback, useLayoutEffect, useState } from 'react';
+
+import { TreeViewDataItem } from '@patternfly/react-core';
+
+type UseTreeViewItemRightClick = (treeData: TreeViewDataItem[]) => {
```
pcbailey, line 5: The file name should also be renamed. Sorry, I didn't mention that.

PR #2475, opened 2025-02-19
CNV-56219: fix guided tour

upalatucci:
## 📝 Description

The state handling on top of the VM Configuration Tab component was in a infinite loading

AND 
the gouded tour was using the all-namespaces URL 

I changed the VM URL link removing /all-namespaces/ and use /ns/default (the tour VM has default namespace)

## 🎥 Demo

**Before**

https://github.com/user-attachments/assets/5647669b-0d5b-4efa-95fc-516fbb743774

**Fix**

https://github.com/user-attachments/assets/762f4f3f-9bf2-40d5-9b2f-908a051870a1

### DIFF
--- a/src/utils/components/GuidedTour/utils/constants.tsx
+++ b/src/utils/components/GuidedTour/utils/constants.tsx
@@ -89,7 +89,7 @@ export const tourSteps: Step[] = [
       'On the Configuration tab on the VirtualMachine page, you can search for and edit any configurable item using the search box.',
     ),
     data: {
-      route: '/k8s/all-namespaces/kubevirt.io~v1~VirtualMachine/rhel9-tour-guide/configuration',
+      route: '/k8s/ns/default/kubevirt.io~v1~VirtualMachine/rhel9-tour-guide/configuration',
     },
     disableBeacon: true,
     placement: 'bottom',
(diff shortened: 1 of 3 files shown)


avivtur (approved):

src/utils/components/GuidedTour/utils/constants.tsx
```diff
@@ -68,7 +68,7 @@ export const tourSteps: Step[] = [
   },
```
metalice, line 68: is it working the same for non admin users?

upalatucci: Ah.. this is why all-namespaces right ?

upalatucci: okay now fixed. all-namespaces url are all supported and the VM link is 
/k8s/ns/default/kubevirt.io~v1~VirtualMachine/rhel9-tour-guide/configuration
but because of my change in the [VirtualMachineNavPage.tsx](https://github.com/kubevirt-ui/kubevirt-plugin/pull/2475/files#diff-e4680557f12a577766937e346ef2012e2014529efa4fa62c429ba629ac82cfc3) , everything works for the non-admin user

metalice (approved):

PR #2479, opened 2025-02-20
CNV-56822: add treeview to guided tour

upalatucci:
## 📝 Description

Add a new step for the review in the guided tour.

The tree view will be visible during the guided tour even if the user have disabled the feature.

in DetailsSection there is a small fix to not have an infinite loading configuration tab during the guided tour (the VM seems to be a little bit different from the usual VM that we create through instance type) 

## 🎥 Demo

*TourVM used here to show the treeview (could be empty for new users)*

Screenshot 2025-02-20 at 15 00 51


src/views/virtualmachines/tree/VirtualMachineTreeView.tsx
```diff
@@ -44,14 +45,16 @@ const VirtualMachineTreeView: FC<VirtualMachineTreeViewProps> = ({
   const [drawerWidth, setDrawerWidth] = useLocalStorage(TREE_VIEW_LAST_WIDTH, OPEN_DRAWER_SIZE);
   const [drawerOpen, setDrawerOpen] = useLocalStorage(SHOW_TREE_VIEW, SHOW);
 
-  const { featureEnabled: treeViewEnabled, loading } = useFeatures(TREE_VIEW);
+  const { featureEnabled: treeViewFeatureEnabled, loading } = useFeatures(TREE_VIEW);
+
+  const hideTreeview = !treeViewFeatureEnabled && !runningTourSignal.value;
```
avivtur, line 50: if the treeview feature is enabled but the tour is not running, won't this mean the tree view won't be shown? why we need this condition?

locales/en/plugin__kubevirt-plugin.json
```diff
@@ -1474,6 +1475,7 @@
   "VirtualMachines per resource": "VirtualMachines per resource",
   "VirtualMachines statuses": "VirtualMachines statuses",
   "Visit the blog": "Visit the blog",
+  "Visualize all your virtual machines and easily navigate between them using the treeview. You can also see their details with a status summary.": "Visualize all your virtual machines and easily navigate between them using the treeview. You can also see their details with a status summary.",
```
avivtur, line 1478: VirtualMachines and I think treeview should be tree view

avivtur (commented):
Looking at the screenshot and seeing a non-populated tree view, doesn't really give the feel we introduce the tree, can we create a mock tree for the guided tour? we also `tourGuideVM` const just for that purpose

src/views/virtualmachines/tree/VirtualMachineTreeView.tsx
upalatucci: Maybe the condition is a little bit confusing but its correct.
THe treeview will be hided when the feature is disabled and its not running a guided tour. 
The inverse would be to show the treeview if the feature is enabled and is running a guided tour. In all those two cases

PR #2482, opened 2025-02-24
CNV-56288: update deprecated Selects

adamviktora:
## 📝 Description

First part of [CNV-56288](https://issues.redhat.com/browse/CNV-56288).
Replaces deprecated Select usage in AffinityModal and in EnvironmentEditor in order to update from PF5 to PF6 more easily in another PR that will follow.

- AffinityModal 
  - the first Select was replaced with SimpleSelect from the Patternfly templates package
  - the other Select was a typeahead with multiple chips, it has been replaced with a combination of LabelGroup and InputGroup

- EnvironmentEditor
  - the Select with filter and groups has been replaced with an existing `InlineFilterSelect` component (some minor changes to that component were necessary too)
  - Patternfly has removed the SelectOptionObject, `value` now has to be a string type - this creates some problems with fitting both the environmentName and environmentKind to each option value, I created a combined string, but then we can filter by the whole string value, not just the visible option name itself. It might not be bad - we can type in "secret" and it filters out all secrets, but it can be an unexpected behavior. 
    - Question: is every environmentName guaranteed to be unique? If so, we might just use the environmentName as a value.
  - Note: before finding out there is an `InlineFilterSelect` component, I created a solution using a custom menu, which might be better, because it keeps the filter input sticky upon scrolling, shows the Dividers (which were there before) and the filtering works as expected (by the option name only), [here is a link to that commit](https://github.com/kubevirt-ui/kubevirt-plugin/commit/3e2dc362ff092895168fd1fa84cbf26e118748ad). Let me know which approach is preferable.

## 🎥 Demo

AffinityModal before:

https://github.com/user-attachments/assets/762e9c2d-0b69-4681-ac83-c155561cb44d

AffinityModal after:

https://github.com/user-attachments/assets/f4a370b5-9149-4f40-9ff6-c222b1db3eb3

EnvironmentEditor before:
 
https://github.com/user-attachments/assets/4217b23a-da9b-4c0d-9cdc-dac72ae78604

EnvironmentEditor after - using `InlineFilterSelect`:

https://github.com/user-attachments/assets/9451213b-2448-44d1-ac7b-0994c0189f6d

EnvironmentEditor after - using custom menu:

https://github.com/user-attachments/assets/fbe2b791-dfbc-4044-bc5d-3ee90d440f0e


src/utils/components/AffinityModal/components/AffinityEditModal/AffinityForm/components/AffinityEditRow.tsx
```diff
@@ -55,41 +53,26 @@ const AffinityExpressionRow: FC<AffinityExpressionRowProps> = ({
...
-          placeholderText={enableValueField ? t('Enter value') : ''}
-          selections={enableValueField ? values : []}
-          typeAheadAriaLabel={t('Enter value')}
-          variant={SelectVariant.typeaheadMulti}
-        >
-          {values?.map((option) => (
-            <SelectOption isDisabled={false} key={option} value={option} />
-          ))}
-        </Select>
+        {enableValueField && (
```
metalice, line 69: Why not pass prop of isDisabled as before?

src/utils/components/AffinityModal/components/AffinityEditModal/AffinityForm/components/AffinityEditRowValues.tsx
```diff
@@ -0,0 +1,73 @@
...
+  values: string[];
+};
+
+export const AffinityEditRowValues: React.FC<AffinityEditRowValuesProps> = ({
+  onClear,
+  onSelect,
+  values,
+}) => {
+  const { t } = useKubevirtTranslation();
+  const [inputValue, setInputValue] = React.useState('');
```
metalice, line 26: u can, namely import useState, no need for React.useState

src/utils/components/AffinityModal/components/AffinityEditModal/AffinityForm/components/AffinityEditRowValues.tsx
```diff
@@ -0,0 +1,73 @@
...
+    onSelect(null, inputValue);
+    setInputValue('');
+  };
+
+  const addingIsDisabled =
+    !inputValue || values.some((val) => val.toLowerCase() === inputValue.toLowerCase());
+
+  return (
+    <Stack hasGutter>
+      <LabelGroup isClosable numLabels={999} onClick={onClear}>
```
metalice, line 42: 999?

src/utils/components/AffinityModal/components/AffinityEditModal/AffinityForm/components/AffinityEditRowValues.tsx
```diff
@@ -0,0 +1,73 @@
...
+  TextInput,
+} from '@patternfly/react-core';
+
+type AffinityEditRowValuesProps = {
+  onClear?: () => void;
+  onSelect?: (event, selection) => void;
+  values: string[];
+};
+
+export const AffinityEditRowValues: React.FC<AffinityEditRowValuesProps> = ({
```
metalice, line 20: u can, namely import FC, no need for React.FC

src/utils/components/AffinityModal/components/AffinityEditModal/AffinityForm/components/AffinityEditRowValues.tsx
```diff
@@ -0,0 +1,73 @@
...
+      <InputGroup>
+        <InputGroupItem isFill>
+          <TextInput
+            onKeyDown={(event) => {
+              if (event.key === 'Enter' && !addingIsDisabled) {
+                event.preventDefault();
+                onAdd();
+              }
+            }}
+            aria-label={t('Enter value')}
```
metalice, line 58: u can save this as const as u also use it below

src/utils/components/AffinityModal/components/AffinityEditModal/AffinityForm/components/AffinityEditRowValues.tsx
```diff
@@ -0,0 +1,73 @@
...
+            onChange={(_event, value) => setInputValue(value)}
+            placeholder={t('Enter value')}
+            type="text"
+            value={inputValue}
+          />
+        </InputGroupItem>
+        <InputGroupItem>
+          <Button isDisabled={addingIsDisabled} onClick={onAdd} variant="control">
+            Add
```
metalice, line 67: translation is missing

src/utils/components/EnvironmentEditor/components/EnvironmentSelectResource.tsx
```diff
@@ -96,61 +57,59 @@ const EnvironmentSelectResource: FC<EnvironmentSelectResourceProps> = ({
...
+
+  const selectedValue = getEnvironmentOptionValue(environmentName, kind);
+
+  const getEnhancedSelectOptionProps = (
+    optionName: string,
+    optionKind: EnvironmentKind,
+  ): EnhancedSelectOptionProps => ({
+    children: (
+      <>
+        <span className="sr-only">{optionKind}</span>
```
metalice, line 72: what is this class name? from PF?

metalice:
Hey @adamviktora - great first PR! left some comments :)

src/utils/components/AffinityModal/components/AffinityEditModal/AffinityForm/components/AffinityEditRowValues.tsx
adamviktora: Oops, I actually commited it like this.
So, the default numLabels is I think 5 and there is this bug when clicking on the "N more" overflow label, it closes the modal and reloads the screen:

https://github.com/user-attachments/assets/77817e99-a907-4f6e-9e3a-e0d542fb89c4

I did not investigate yet why is it happening. As a workaround I added this 999 so the overflow label never displays. How many those labels will there usually be? Is it ok to show them all and not have the overflow label? Or should I rather resolve the bug?

src/utils/components/EnvironmentEditor/components/EnvironmentSelectResource.tsx
adamviktora: It was there before, so I just kept it. But yes, it is from patternfly, it hides the element and keeps it in the DOM for screen readers:

Screenshot 2025-02-25 at 9 20 34

src/utils/components/AffinityModal/components/AffinityEditModal/AffinityForm/components/AffinityEditRow.tsx
adamviktora: We can, depends on what is better:

disabling:
Screenshot 2025-02-25 at 9 32 35

hiding:
Screenshot 2025-02-25 at 9 28 22

Do you prefer the disabling variant?

src/utils/components/AffinityModal/components/AffinityEditModal/AffinityForm/components/AffinityEditRowValues.tsx
metalice: Is this bug from PF, or how are we using the component? please check .

src/utils/components/AffinityModal/components/AffinityEditModal/AffinityForm/components/AffinityEditRow.tsx
metalice: u can add a prop for the component as hide and let the component handle the situation by himself instead of controlling from outside.

metalice:
image

I see an overlap , is it adjustable or we need to open a bug for PF?

src/utils/components/AffinityModal/components/AffinityEditModal/AffinityForm/components/AffinityEditRowValues.tsx
adamviktora: Yes, it was a Patternfly bug, the button doesn't have `type="button"` so it submits the form. It is resolved in newer PF 5.x versions and in PF 6.

I'll keep the `numLabels={Number.MAX_VALUE}` for now and return it back to default in the update to PF6 PR

adamviktora:
In PF6 it is resolved:

Screenshot 2025-02-25 at 11 48 35

If this PR will not be released and it will be merged only after updating to PF6, do we care about the PF5 bug?

src/utils/components/AffinityModal/components/AffinityEditModal/AffinityForm/components/AffinityEditRowValues.tsx
```diff
@@ -0,0 +1,78 @@
...
+    setInputValue('');
+  };
+
+  const addingIsDisabled =
+    !inputValue || values.some((val) => val.toLowerCase() === inputValue.toLowerCase());
+
+  const textInputPlaceholder = t('Enter value');
+  const labelsToShowCount = Number.MAX_VALUE; // always show all labels
+
+  return isHidden ? null : (
```
avivtur, line 45: A ternary operator is ok, a more readable code would be something like:
```
if (isHidden) return null;

return ...
```

src/utils/components/AffinityModal/components/AffinityEditModal/AffinityForm/components/AffinityEditRowValues.tsx
```diff
@@ -0,0 +1,78 @@
...
+          <Label id={value} key={value} onClose={() => onClose(value)} variant="outline">
+            {value}
+          </Label>
+        ))}
+      </LabelGroup>
+      <InputGroup>
+        <InputGroupItem isFill>
+          <TextInput
+            onKeyDown={(event) => {
+              if (event.key === 'Enter' && !addingIsDisabled) {
```
avivtur, line 58: Please check if we have an existing const for 'Enter', If we don't please create one

avivtur (commented):
Looking great @adamviktora :rocket: 
Just a couple of nitpicks

src/utils/components/AffinityModal/components/AffinityEditModal/AffinityForm/components/AffinityEditRowValues.tsx
adamviktora: I remembered that Patternfly has a list of constants for keys, so I used that

avivtur: Even better :)

PR #2485, opened 2025-02-25
SSH key/secret related tests added

lkladnit:
#### This PR adds two SSH-related tests to upstream.
This sums up to **40** tests, **2** of which are currently disabled
Closes https://issues.redhat.com/browse/CNV-57083

### DIFF
--- a/cypress/tests/e2e/b-overview.cy.ts
+++ b/cypress/tests/e2e/b-overview.cy.ts
@@ -1,4 +1,11 @@
+import { TEST_NS } from '../../utils/const/index';
+import { authSSHKey } from '../../utils/const/string';
+import { tabModal } from '../../views/actions';
+import { addNew, manageKeysText, secretName, uploadSecret } from '../../views/selector-catalog';
+import { userButtonTxt } from '../../views/selector-instance';
 import { tab } from '../../views/tab';
+
+const testSecret = 'test-secret';
 
 describe('Test Virtualization Overview', () => {
   before(() => {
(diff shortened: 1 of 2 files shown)


metalice:
please open a jira ticket and connect it here

lkladnit:
https://issues.redhat.com/browse/CNV-57083 updated

metalice:
please add the ticket to the tile as all other prs and then run "/jira refresh" - it will connect the integration

cypress/tests/e2e/post-action.cy.ts
```diff
@@ -0,0 +1,7 @@
+import { MINUTE } from '../../utils/const/index';
+
+describe('Holding cluster', () => {
+  it('wait 100 minutes', () => {
```
metalice, line 4: ?

cypress/tests/e2e/b-overview.cy.ts
```diff
@@ -1,5 +1,12 @@
+import { TEST_NS } from '../../utils/const/index';
+import { authSSHKey } from '../../utils/const/string';
+import { tabModal } from '../../views/actions';
+import { addNew, manageKeysText, secretName, uploadSecret } from '../../views/selector-catalog';
+import { userButtonTxt } from '../../views/selector-instance';
 import { tab } from '../../views/tab';
 
+const testSecret = 'test-secret';
```
metalice, line 8: const file?

gouyang:
two general cooments:
1. don't add something that won't run, like the post-action
2. don't check existing stuff in upstream test as the test environment is fresh every time, such check the exists secret in the test.

cypress/tests/all.cy.ts
```diff
@@ -2,3 +2,4 @@ import './e2e/a-setup.cy.ts';
 import './e2e/b-overview.cy.ts';
 import './e2e/check-tab-yaml.cy.ts';
 import './e2e/create-vm.cy.ts';
+// import './e2e/post-action.cy.ts';
```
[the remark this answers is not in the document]
lkladnit: This is a test that keeps the testing cluster alive for debugging. It's not used normally, so it's commented out

cypress/tests/e2e/b-overview.cy.ts
lkladnit: it's just a name

lkladnit:
> two general suggestions:
> 
> 1. don't add something that won't run, like the post-action
> 2. don't check existing stuff in upstream test as the test environment is fresh every time, such check the exists secret in the test.

i removed post-action line from `all.cy.ts`

PR #2489, opened 2025-02-26
Fix Form Submission Behavior for Virtual Machine Creation

fallmo:
Screenshot 2025-02-26 at 4 36 01 PM

## 📝 Description

PR ensures that pressing `Enter` key in the VM creation form correctly triggers the Quick Create VirtualMachine action (which is expected by users) instead of causing an unintended full-page reload (default browser behavior). 

## Changes

- Added a form submit handler (onSubmitForm) to prevent default submission behavior and trigger `onQuickCreate`.
- Removed `onClick={onQuickCreate}` from the Quick Create VirtualMachine button since form submission now handles it, and `button[type=submit]` automatically submits the form when clicked (so no need for an onClick handler).
- Remove `e: MouseEvent` from onQuickCreate parameters, as default browser behavior is already prevented on the form submission (`e.preventDefault()`)

## 🎥 Demo

Current behavior upon pressing `Enter` key.

https://github.com/user-attachments/assets/2a0893e3-dc4d-4e23-af4a-16ede9bd53cb

### DIFF
--- a/src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
+++ b/src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
@@ -70,7 +70,7 @@ export const TemplatesCatalogDrawerCreateForm: FC<TemplatesCatalogDrawerCreateFo
 
     const error = templateLoadingError || createError;
     return (
-      <form className="template-catalog-drawer-form" id="quick-create-form">
+      <div className="template-catalog-drawer-form" id="quick-create-form">
         <Stack hasGutter>
           <>
             <StackItem>
@@ -80,6 +80,7 @@ export const TemplatesCatalogDrawerCreateForm: FC<TemplatesCatalogDrawerCreateFo
                     <TextInput
                       aria-label="virtualmachine name"
                       data-test-id="template-catalog-vm-name-input"
+                      form={DRAWER_FORM_ID}
                       isDisabled={Boolean(templateLoadingError)}
                       isRequired
                       name="vmname"
@@ -198,7 +199,7 @@ export const TemplatesCatalogDrawerCreateForm: FC<TemplatesCatalogDrawerCreateFo
             </Split>
           </StackItem>
         </Stack>
-      </form>
+      </div>
     );
   },
 );


fallmo:
Hi @adamviktora , have you had a chance to look at this? Let me know if any changes are needed. Thanks!

adamviktora (changes requested):
Hi @fallmo. It looks like after the changes the virtual machine won't be created upon clicking the button. Pressing enter doesn't work for me (unless the button is focused with tab, but that should trigger the onClick anyway) - using Firefox 135.

Maybe another approach could be to add an onKeyDown event to the TextInput, check if key equals to "Enter" and then call onQuickCreate. 

Also I am not sure if such behavior would always be expected by users, so maybe a review from a UX person would also help

fallmo:
> Hi @fallmo. It looks like after the changes the virtual machine won't be created upon clicking the button. Pressing enter doesn't work for me (unless the button is focused with tab, but that should trigger the onClick anyway) - using Firefox 135.
> 
> Maybe another approach could be to add an onKeyDown event to the TextInput, check if key equals to "Enter" and then call onQuickCreate.
> 
> Also I am not sure if such behavior would always be expected by users, so maybe a review from a UX person would also help

Hey @adamviktora thanks for the review! I revisited the changes based on your feedback:

- Virtual Machine not being created upon clicking the button:
You're right, normally, a type="submit" button triggers form submission automatically. However, in this case, the form= attribute on the button is different from the actual form's ID, which prevented it from working as expected. I've addressed this by restoring the previous approach of manually triggering form submission via onClick. (Why use onSubmitForm instead of calling onQuickCreate directly? It just saves one line of code.)

-  Pressing enter doesn't work unless the button is focused:
Enter only works when focused on a text input. As you suggested another option would indeed be to use onKeyDown on the text input to listen for an enter key. This should work, but let me know if you prefer the other option. That said, the main goal of this PR is to prevent the default browser behavior (page reload on Enter)

I look forward to a review from a UX person. I think it's a common expectation that pressing Enter submits a form (or triggers the primary action, which in this case is Quick Create). I submitted this PR because I noticed users—including myself—typing a VM name, pressing Enter, and expecting it to submit, only to have the page reload.

Looking forward to your thoughts.

upalatucci:
Hei guys.
I think the issue is different. 
We have two form elements in the drawer and this is wrong. 
One for the actual inputs and one for the buttons in the footer.
This is because we are using patternfly modal that split the footer from the body. 

To solve this we created buttons with `form={DRAWER_FORM_ID}` but we also have to remove the second `form` element in the `TemplatesCatalogDrawerCreateForm` and use `form={DRAWER_FORM_ID}` also on the `TextInput` in this same component. 

Thanks @fallmo  for pointing this issue and even providing a solution. Your solution is correct actually but I think the main issue is that we have two form elements for the same form and I would love to merge your pr with this fix

adamviktora (commented):
Hey @fallmo, thanks for updating this!

Pressing enter when focused on the text input still doesn't trigger the submit for me, unless `form={DRAWER_FORM_ID}` is added to the TextInput as @upalatucci suggests.

fallmo:
Hey @upalatucci  and @adamviktora, I see what you are saying. At this point I think there are two possible solutions. The quick one to keep the current setup with the two forms, and (1) add an `onKeyDown` to trigger the QuickCreate when the enter key is pressed (as @adamviktora suggested) (2) and a submit handler for the form to `e.preventDefault()` as a preventative measure. 
The second option is to rewrite the current setup so that there is only one form with one `onSubmit` handler handling submission; this option would be more complex as it would require modifications from `TemplatesCatalogDrawer.tsx` component all the way down.  What do you suggest?

upalatucci:
No don't worry @fallmo  just remove the form (2 option) and use the `form={DRAWER_FORM_ID}`  in the textinput and that's it .You don't have to edit `TemplatesCatalogDrawer` file

fallmo:
> No don't worry @fallmo just remove the form (2 option) and use the `form={DRAWER_FORM_ID}` in the textinput and that's it .You don't have to edit `TemplatesCatalogDrawer` file

The thing is that the original form does not have an onSubmit handler so the page reload will persist

fallmo:
@upalatucci  you were right. All that was needed was a `form={DRAWER_FORM_ID}` on the input.

adamviktora (approved):
LGTM 👍

PR #2495, opened 2025-03-03
CNV-57354: add test for top consumers tab

gouyang:
## 📝 Description

> Add a brief description

## 🎥 Demo

### DIFF
--- a/cypress/tests/overview/top-consumers-tab.cy.ts
+++ b/cypress/tests/overview/top-consumers-tab.cy.ts
@@ -1,0 +1,34 @@
+import { topConsumerCard } from '../../views/overview';
+import { Perspective, switchPerspective } from '../../views/perspective';
+import { tab } from '../../views/tab';
+
+const enum Card {
+  CPU = 0,
+  Memory,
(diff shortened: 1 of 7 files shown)


cypress/tests/overview/top-consumers-tab.cy.ts
```diff
@@ -0,0 +1,25 @@
...
+describe('Test Virtualization Top consumer tab', () => {
+  before(() => {
+    cy.visit('');
+    switchPerspective(Perspective.Administrator);
+  });
+
+  it('ID(CNV-9295) check data available for cpu, memory and storage', () => {
+    cy.visitOverview();
+    tab.navigateToTopConsumers();
+    cy.get(topConsumerCard).eq(0).should('not.contain', 'No data available');
```
avivtur, line 14: suggestion: since .eq(0) eq(1) etc are not readable for someone who didn't wrote this test, I want to suggest creating an enum to have more understanding which exact part we testing
e.g if 0 is the title or body or something else of the top consumer card.

upalatucci (approved):

PR #2496, opened 2025-03-03
CNV-56225: Use fixed 4/3 aspect ratio for VNC connect component

rszwajko:
## 📝 Description

Use the same layout for connecting state.

### After

https://github.com/user-attachments/assets/15ffda6a-8541-4f73-b171-557ba864c8e8


adamviktora (commented):
Great update! I really like the refactoring too.

One thing I am noticing is that when VM is stopped, the elements are not aligned as expected:
Screenshot 2025-03-07 at 10 02 51

This can be resolved by changing `virtual-machines-overview-tab-details.scss` -> `.console-overview` -> `justify-content` to `flex-start`

adamviktora (approved):
🎉

PR #2500, opened 2025-03-04
CNV-57355: add test for cluster general settings

gouyang:
## 📝 Description

Test Overview -> Settings -> Cluster -> General settings


cypress/tests/overview/settings-cluster-general.cy.ts
```diff
@@ -0,0 +1,96 @@
...
+  describe('Test Memory density', () => {
+    it('ID(CNV-11225) enable memory density', () => {
+      cy.contains('Memory density').click();
+      cy.get('[id="memory-density-feature"]').find('input[type="checkbox"]').check({ force: true });
+    });
+
+    it('verify higherWorkloadDensity in HCO', () => {
+      const percentage = '.spec.higherWorkloadDensity.memoryOvercommitPercentage';
+      cy.checkHCOSpec(percentage, '150', true);
+      cy.wait(10000);
```
rszwajko, line 51: 1. why we need to wait after a check? 
2. where is the source for  `checkHCOSpec()`?

cypress/tests/overview/settings-cluster-general.cy.ts
```diff
@@ -0,0 +1,96 @@
...
+    it('ID(CNV-11223) disable centos-stream-9-image-cron', () => {
+      cy.contains('Automatic images download').click();
+      cy.get('#centos-stream9-image-cron-auto-image-download-switch')
+        .find('input.pf-v5-c-switch__input')
+        .uncheck({ force: true });
+      cy.wait(10000);
+    });
+
+    it('delete centos-stream9 volumesnapshot', () => {
+      const cmd =
```
rszwajko, line 76: we have no assertions here (and below) that action was successful

cypress/tests/overview/settings-cluster-general.cy.ts
```diff
@@ -0,0 +1,96 @@
...
+    it('enable centos-stream9-image-cron again', () => {
+      cy.get('#centos-stream9-image-cron-auto-image-download-switch')
+        .find('input.pf-v5-c-switch__input')
+        .check({ force: true });
+      cy.wait(60000);
+    });
+
+    it('check centos-stream9 volumesnapshot is downloaded again', () => {
+      cy.wait(60000);
+      const cmd =
```
rszwajko, line 91: instead of one long timeout we could check the condition in a loop with a smaller timout

rszwajko (commented):
Interesting stuff!

cypress/tests/overview/settings-cluster-general.cy.ts
```diff
@@ -0,0 +1,96 @@
...
+    cy.visit('');
+    cy.wait(15000);
+    cy.switchToVirt();
+    cy.visitVirtPerspectiveOverview();
+    tab.navigateToSettings();
+  });
+
+  describe('Check installed version appears', () => {
+    it('check version', () => {
+      cy.get('[data-test="detail-item-value"]').contains('4');
```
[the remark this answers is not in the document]
gouyang: The test point is not verify the version is correct or not as it's reading from backend, it's not predicatable. The test is checking the presenting of the versions. Keep it simple is enough and it's a safety test.

Also the version looks different on clusters:
Eg: the last tag is v4.18.0 in the repo, but on 4.19 cluster, it's '4.19.0' and on 4.18 production cluster is '4.18.0-848'.

cypress/tests/overview/settings-cluster-general.cy.ts
```diff
@@ -0,0 +1,96 @@
...
+    });
+
+    it('verify live migration limits in HCO', () => {
+      const maxPerCluster = '.spec.liveMigrationConfig.parallelMigrationsPerCluster';
+      const maxPerNode = '.spec.liveMigrationConfig.parallelOutboundMigrationsPerNode';
+      const spec = '.spec.liveMigrationConfig';
+      const cmd = `oc get -n openshift-cnv hyperconverged kubevirt-hyperconverged -o jsonpath='{${spec}}'`;
+      cy.exec(cmd).then((res) => {
+        cy.task('log', res.stdout);
+      });
```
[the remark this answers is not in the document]
gouyang: we can do it if it becomes a real problem for test.

cypress/tests/overview/settings-cluster-general.cy.ts
gouyang: 1. the wait should be placed after check the option, not here. The purpose is waiting for the backend complete the updating before uncheck the option, remove flaky for the test.
2. add it  now.

cypress/tests/overview/settings-cluster-general.cy.ts
gouyang: put the block into retry

cypress/tests/overview/settings-cluster-general.cy.ts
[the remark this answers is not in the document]
rszwajko: > the test is checking the presenting of the versions
>
OK that makes sense. We indirectly verify that the version string is not empty.

cypress/support/commands.ts
```diff
@@ -0,0 +1,21 @@
...
+    interface Chainable {
+      checkHCOSpec(spec: string, matchString: string, include: boolean): void;
+    }
+  }
+}
+
+Cypress.Commands.add('checkHCOSpec', (spec: string, matchString: string, include: boolean) => {
+  cy.exec(
+    `oc get -n kubevirt-hyperconverged hyperconverged kubevirt-hyperconverged -o jsonpath='{${spec}}'`,
+  ).then((result) => {
```
rszwajko, line 14: In case of error the method will skip the `expect()` checks. We need fail the check in `catch()`.

cypress/tests/overview/settings-cluster-general.cy.ts
```diff
@@ -0,0 +1,107 @@
+import { SECOND } from '../../utils/const/index';
+import { tab } from '../../views/tab';
+
+describe('Test Cluster General settings', () => {
+  before(() => {
+    cy.login();
+    cy.visit('');
+    cy.wait(10000);
```
rszwajko, line 8: IMHO the `wait()` here (and in different places too) gives a false sense of security - we should embed the timeout in `cy.get()` calls. 
i.e. a good example is ` switchPerspective()` where wait up to 5 minutes (but will proceed as soon as possible)
https://github.com/kubevirt-ui/kubevirt-plugin/blob/67d8f730d8c3c49347209683c8fa8667fb172a09/cypress/views/perspective.ts#L19
not so good example is a `wait()` below (line 20)  which should be embedded (and increased) in the next `get()` call i.e.
``
 cy.get(toggle, { timeout: 5 * MINUTE }).click()`
```

File: cypress/tests/overview/settings-cluster-general.cy.ts (hunk 1 above)
Comment (rszwajko) [reply cypress/tests/overview/settings-cluster-general.cy.ts]: checkout how @sjd78 implemented active checking in a loop:
https://github.com/konveyor/tackle2-ui/blob/991b02aa1001bf05eb53e868e48b8bb42091e48f/hack/setup-operator.sh#L43
Usage:
```
  echo "Waiting for the Tackle CRD to exist"
  retry_command 10 10 \
    $KUBECTL get customresourcedefinitions.apiextensions.k8s.io tackles.tackle.konveyor.io

```
This way you can provide a big maximal timeout i.e. 10min but at the same time be quick if thing are going smoothly.

File: cypress/tests/overview/settings-cluster-general.cy.ts (hunk 3 above)
Comment (rszwajko) [reply cypress/tests/overview/settings-cluster-general.cy.ts]: In this case re-try on the entire test case is OK but in many cases you need it for a single step.
Checkout how @sjd78 implemented active checking in a loop:
https://github.com/konveyor/tackle2-ui/blob/991b02aa1001bf05eb53e868e48b8bb42091e48f/hack/setup-operator.sh#L43
Usage:
```
  echo "Waiting for the Tackle CRD to exist"
  retry_command 10 10 \
    $KUBECTL get customresourcedefinitions.apiextensions.k8s.io tackles.tackle.konveyor.io

```
This way you can provide a big maximal timeout i.e. 10min but at the same time be quick if thing are going smoothly.

cypress/tests/overview/settings-cluster-general.cy.ts
gouyang: The timeout outside of `cy.get` is what I learned from the test execution, there is flakiness that the page will refresh/flash when try to open a dropdown menu and it brokes a follow-up click.

We have lots of similar case in D/S tests and it's necessary.

cypress/tests/overview/settings-cluster-general.cy.ts
gouyang: cypress will report failure if the command is failed to execute.

cypress/support/commands.ts
gouyang: Again, the error handle is leaving to cypress, if the command failed to execute, qe need to fix the comand.

cypress/tests/overview/settings-cluster-general.cy.ts
gouyang: it runs via cy.exec, the wrapped command is hard to use there. it reports some error based on my experience in the past.

rszwajko (approved):

upalatucci (approved):

lkladnit (approved):

PR #2503, opened 2025-03-06
CNV-49608:  Default to raw VM in VirtualMachineNavPageTitle

rszwajko:
## 📝 Description

Before the fix, VirtualMachineNavPageTitle used the result of
expand-spec instead of a raw VM object. The approach failed
in scenarios where:
1. isInstanceTypeVM() was used as this information is by design
   erased by the backend.
2. the VM was sent to the backend - as expanded VM is considered a new
   VM (with references to old VM removed)
The main section distinguished both objects and passed them separately
as 'instanceTypeExpandedSpec' and 'vm' (both of type V1VirtualMachine).

After the fix the raw VM object will be used by default in
VirtualMachineNavPageTitle and the expanded VM will be passed as
'instanceTypeExpandedSpec'.

Key points:
1. hasOwnProperty() checks used for 'autoattachGraphicsDevice' were
   refactored to use isHeadlessMode() helper
3. created an independent version of InstanceTypeModal that retrieves
   all available instance types - the extra query will be fired only
   when the modal is opened
4. retrieve the instance type name used by VirtualMachineInstance
   from annotations

Reference-Url: https://github.com/kubevirt-ui/kubevirt-plugin/pull/2414
Reference-Url: https://github.com/kubevirt/kubevirt/blob/47d69dd7dbf57932cbc70c59bab6e930d26cda3d/staging/src/kubevirt.io/api/core/v1/types.go#L1060
Reference-Url: https://github.com/kubevirt/kubevirt/blob/4c2642b9101d9b734a906ccc2dd157ca6276cf69/pkg/instancetype/expand/expand.go#L114

### Before
Screenshot From 2025-03-06 21-04-38
### After
Screenshot From 2025-03-06 21-35-13
Screenshot From 2025-03-17 19-37-54

### DIFF
--- a/src/utils/components/HardwareDevices/HardwareDevicesHeadlessMode.tsx
+++ b/src/utils/components/HardwareDevices/HardwareDevicesHeadlessMode.tsx
@@ -3,6 +3,7 @@ import produce from 'immer';
 
 import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
+import { isHeadlessMode } from '@kubevirt-utils/resources/vm';
 import { ensurePath } from '@kubevirt-utils/utils/utils';
 import { Flex, FlexItem, Switch } from '@patternfly/react-core';
 
@@ -15,10 +16,7 @@ type HardwareDevicesHeadlessModeProps = {
 const HardwareDevicesHeadlessMode: FC<HardwareDevicesHeadlessModeProps> = ({ onSubmit, vm }) => {
   const { t } = useKubevirtTranslation();
 
-  const devices = vm?.spec?.template?.spec?.domain?.devices;
-  const [isChecked, setIsChecked] = useState<boolean>(
-    devices?.hasOwnProperty('autoattachGraphicsDevice') && !devices?.autoattachGraphicsDevice,
-  );
+  const [isChecked, setIsChecked] = useState<boolean>(isHeadlessMode(vm));
 
   const updateHeadlessMode = (checked: boolean) => {
     const updatedVM = produce<V1VirtualMachine>(vm, (vmDraft: V1VirtualMachine) => {
--- a/src/utils/components/HardwareDevices/modal/HardwareDevicesHeadlessModeModal.tsx
+++ b/src/utils/components/HardwareDevices/modal/HardwareDevicesHeadlessModeModal.tsx
@@ -6,6 +6,7 @@ import FormGroupHelperText from '@kubevirt-utils/components/FormGroupHelperText/
 import ModalPendingChangesAlert from '@kubevirt-utils/components/PendingChanges/ModalPendingChangesAlert/ModalPendingChangesAlert';
 import TabModal from '@kubevirt-utils/components/TabModal/TabModal';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
+import { isHeadlessMode } from '@kubevirt-utils/resources/vm';
 import { ensurePath } from '@kubevirt-utils/utils/utils';
 import { Checkbox, Form, FormGroup } from '@patternfly/react-core';
 
@@ -24,11 +25,8 @@ const HardwareDevicesHeadlessModeModal: FC<HardwareDevicesHeadlessModeModalProps
   vm,
   vmi,
 }) => {
-  const devices = vm?.spec?.template?.spec?.domain?.devices;
   const { t } = useKubevirtTranslation();
-  const [checked, setChecked] = useState<boolean>(
-    devices?.hasOwnProperty('autoattachGraphicsDevice') && !devices?.autoattachGraphicsDevice,
-  );
+  const [checked, setChecked] = useState<boolean>(isHeadlessMode(vm));
 
   const updatedVirtualMachine = useMemo(() => {
     const updatedVM = produce<V1VirtualMachine>(vm, (vmDraft: V1VirtualMachine) => {
--- a/src/utils/components/HeadlessMode/HeadlessMode.tsx
+++ b/src/utils/components/HeadlessMode/HeadlessMode.tsx
@@ -1,6 +1,7 @@
 import React, { FC, useState } from 'react';
 
 import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import { isHeadlessMode } from '@kubevirt-utils/resources/vm';
 import { Switch } from '@patternfly/react-core';
 
 type HeadlessModeProps = {
@@ -9,11 +10,7 @@ type HeadlessModeProps = {
 };
 
 const HeadlessMode: FC<HeadlessModeProps> = ({ updateHeadlessMode, vm }) => {
-  const devices = vm?.spec?.template?.spec?.domain?.devices;
-  const [isChecked, setIsChecked] = useState<boolean>(
-    devices?.hasOwnProperty('autoattachGraphicsDevice') &&
-      devices?.autoattachGraphicsDevice === false,
-  );
+  const [isChecked, setIsChecked] = useState<boolean>(isHeadlessMode(vm));
   return (
     <Switch
       onChange={(_event, checked: boolean) => {
--- a/src/utils/components/InstanceTypeModal/InstanceTypeModal.tsx
+++ b/src/utils/components/InstanceTypeModal/InstanceTypeModal.tsx
@@ -1,6 +1,5 @@
 import React, { FC, useMemo, useState } from 'react';
 
-import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
 import { Content, Flex, FlexItem, SelectList, SelectOption } from '@patternfly/react-core';
 import { InstanceTypeUnion } from '@virtualmachines/details/tabs/configuration/utils/types';
@@ -8,6 +7,7 @@ import { InstanceTypeUnion } from '@virtualmachines/details/tabs/configuration/u
 import FormPFSelect from '../FormPFSelect/FormPFSelect';
 import TabModal from '../TabModal/TabModal';
 
+import { InstanceTypeModalProps } from './utils/types';
 import {
   getInstanceTypeFromSeriesAndSize,
   getInstanceTypeSeriesAndSize,
@@ -17,25 +17,13 @@ import {
   mappedInstanceTypesToSelectOptions,
 } from './utils/util';
 
-type InstanceTypeModalProps = {
-  allInstanceTypes: InstanceTypeUnion[];
-  instanceType: InstanceTypeUnion;
-  instanceTypeVM: V1VirtualMachine;
-  isOpen: boolean;
-  onClose: () => void;
-  onSubmit: (
-    updatedVM: V1VirtualMachine,
-    instanceType: InstanceTypeUnion,
-  ) => Promise<V1VirtualMachine>;
-};
-
 const InstanceTypeModal: FC<InstanceTypeModalProps> = ({
   allInstanceTypes,
   instanceType,
-  instanceTypeVM,
   isOpen,
   onClose,
   onSubmit,
+  vm,
 }) => {
   const { t } = useKubevirtTranslation();
   const mappedInstanceTypes = useMemo(
@@ -47,20 +35,20 @@ const InstanceTypeModal: FC<InstanceTypeModalProps> = ({
     [instanceType],
   );
 
-  const [series, setSeries] = useState<string>(
+  const [series, setSeries] = useState<string | undefined>(
     getInstanceTypeSeriesDisplayName(mappedInstanceTypes, instanceTypeSeries),
   );
 
-  const [size, setSize] = useState<string>(
+  const [size, setSize] = useState<string | undefined>(
     getInstanceTypesPrettyDisplaySize(mappedInstanceTypes, instanceTypeSeries, instanceTypeSize),
   );
 
   const handleSubmit = (selectedInstanceType: InstanceTypeUnion) =>
-    onSubmit(instanceTypeVM, selectedInstanceType);
+    onSubmit(vm, selectedInstanceType);
 
   return (
     <TabModal
-      headerText={t('Edit Instancetype')}
+      headerText={t('Edit Instance Type')}
       isOpen={isOpen}
       obj={getInstanceTypeFromSeriesAndSize(mappedInstanceTypes, series, size)}
       onClose={onClose}
--- a/src/utils/components/InstanceTypeModal/StandaloneInstanceTypeModal.tsx
+++ b/src/utils/components/InstanceTypeModal/StandaloneInstanceTypeModal.tsx
@@ -1,0 +1,40 @@
+import React, { FC, useMemo } from 'react';
+
+import useInstanceTypesAndPreferences from '@catalog/CreateFromInstanceTypes/state/hooks/useInstanceTypesAndPreferences';
+
+import ErrorAlert from '../ErrorAlert/ErrorAlert';
+import Loading from '../Loading/Loading';
+
+import { InstanceTypeModalProps } from './utils/types';
+import InstanceTypeModal from './InstanceTypeModal';
+
+const StandaloneInstanceTypeModal: FC<
+  Pick<InstanceTypeModalProps, 'isOpen' | 'onClose' | 'onSubmit' | 'vm'>
+> = ({ isOpen, onClose, onSubmit, vm }) => {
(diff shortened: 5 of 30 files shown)


src/utils/components/HardwareDevices/HardwareDevicesHeadlessMode.tsx
```diff
@@ -16,9 +16,7 @@ const HardwareDevicesHeadlessMode: FC<HardwareDevicesHeadlessModeProps> = ({ onS
   const { t } = useKubevirtTranslation();
 
   const devices = vm?.spec?.template?.spec?.domain?.devices;
-  const [isChecked, setIsChecked] = useState<boolean>(
-    devices?.hasOwnProperty('autoattachGraphicsDevice') && !devices?.autoattachGraphicsDevice,
-  );
+  const [isChecked, setIsChecked] = useState<boolean>(devices?.autoattachGraphicsDevice === false);
```
avivtur, line 19: I think isHeadlessModeVMI(vm?.spec?.template) should work the same, can you check?

src/utils/components/HeadlessMode/HeadlessMode.tsx
```diff
@@ -10,10 +10,7 @@ type HeadlessModeProps = {
 
 const HeadlessMode: FC<HeadlessModeProps> = ({ updateHeadlessMode, vm }) => {
   const devices = vm?.spec?.template?.spec?.domain?.devices;
-  const [isChecked, setIsChecked] = useState<boolean>(
-    devices?.hasOwnProperty('autoattachGraphicsDevice') &&
-      devices?.autoattachGraphicsDevice === false,
-  );
+  const [isChecked, setIsChecked] = useState<boolean>(devices?.autoattachGraphicsDevice === false);
```
avivtur, line 13: same here if isHeadlessModeVMI(vm?.spec?.template) works the same please change to utilize the util function :)

src/utils/components/InstanceTypeModal/InstanceTypeModal.tsx
```diff
@@ -119,4 +122,26 @@ const InstanceTypeModal: FC<InstanceTypeModalProps> = ({
   );
 };
 
+export const ConnectedInstanceTypeModal: FC<
```
avivtur, line 125: we don't put 2 components into one file, please move this to a different file

src/utils/components/InstanceTypeModal/InstanceTypeModal.tsx
```diff
@@ -119,4 +122,26 @@ const InstanceTypeModal: FC<InstanceTypeModalProps> = ({
...
+  Pick<InstanceTypeModalProps, 'instanceTypeVM' | 'isOpen' | 'onClose' | 'onSubmit'>
+> = (props) => {
```
avivtur, line 127: if props are used in the component and not just passed on, we should destruct if for readability

src/utils/resources/vm/utils/instanceTypes.ts
```diff
@@ -1,5 +1,19 @@
...
+import { getAnnotation } from '@kubevirt-utils/resources/shared';
 import { isEmpty } from '@kubevirt-utils/utils/utils';
 
 export const isInstanceTypeVM = (vm: V1VirtualMachine): boolean =>
   !isEmpty(vm?.spec?.instancetype) || !isEmpty(vm?.spec?.preference);
+
+const NAMESPACED_INSTANCE_TYPE_NAME_ANNOTATION = 'kubevirt.io/instancetype-name';
+const CLUSTER_INSTANCE_TYPE_NAME_ANNOTATION = 'kubevirt.io/cluster-instancetype-name';
+const NAMESPACED_PREFERENCE_NAME_ANNOTATION = 'kubevirt.io/preference-name';
+const CLUSTER_PREFERENCE_NAME_ANNOTATION = 'kubevirt.io/cluster-preference-name';
```
avivtur, line 11: Please use separated files for constants and utils

src/utils/components/HardwareDevices/HardwareDevicesHeadlessMode.tsx
metalice: u can use isHeadlessModeVMI ?

src/utils/components/HardwareDevices/modal/HardwareDevicesHeadlessModeModal.tsx
```diff
@@ -26,9 +26,7 @@ const HardwareDevicesHeadlessModeModal: FC<HardwareDevicesHeadlessModeModalProps
 }) => {
   const devices = vm?.spec?.template?.spec?.domain?.devices;
   const { t } = useKubevirtTranslation();
-  const [checked, setChecked] = useState<boolean>(
-    devices?.hasOwnProperty('autoattachGraphicsDevice') && !devices?.autoattachGraphicsDevice,
-  );
+  const [checked, setChecked] = useState<boolean>(devices?.autoattachGraphicsDevice === false);
```
metalice, line 29: u can use isHeadlessModeVMI ? expand it to also include vm

src/views/virtualmachines/details/tabs/configuration/details/components/DetailsSectionBoot.tsx
```diff
@@ -49,7 +47,7 @@ const DetailsSectionBoot: FC<DetailsSectionBootProps> = ({
 
   useEffect(() => {
     expandURLHash(getSearchItemsIds(getDetailsTabBootIds(vm)), location?.hash, setIsExpanded);
-  }, [vm, location?.hash]);
+  }, [vm, location?.hash, setIsExpanded]);
```
metalice, line 50: no need to include setState into deps - do u have extensions that announce if some deps are missing?

src/utils/components/PendingChanges/utils/helpers.ts
```diff
@@ -444,17 +455,12 @@ export const getChangedHeadlessMode = (
   if (isEmpty(vm) || isEmpty(vmi)) {
     return false;
   }
-  const vmDevices = vm?.spec?.template?.spec?.domain?.devices;
-  const vmiDevices = vmi?.spec?.domain?.devices;
 
-  const vmHeadless = !!vmDevices?.autoattachGraphicsDevice;
-  const vmiHeadless = !!vmiDevices?.autoattachGraphicsDevice;
+  // assume headless only if the property is explicitly false
+  const vmHasGraphic = vm?.spec?.template?.spec?.domain?.devices?.autoattachGraphicsDevice ?? true;
```
metalice, line 460: isHeadlessModeVMI can be used here?

src/utils/components/HardwareDevices/HardwareDevicesHeadlessMode.tsx
rszwajko: The helper has been made more generic and moved to `selectors.ts`

src/utils/resources/vm/utils/instanceTypes.ts
rszwajko: Used existing (empty) `src/utils/resources/instancetype/constants.ts` file.

src/utils/resources/vm/hooks/useInstanceTypeExpandSpec.ts
```diff
@@ -35,7 +35,16 @@ const useInstanceTypeExpandSpec: UseInstanceTypeExpandSpec = (vm) => {
...
         const response = await consoleFetch(url);
         const json = await response.json();
-        setInstanceTypeExpandedSpec(json);
+        const expandedVm = V1VirtualMachineFromJSON(json);
+        expandedVm.spec = {
+          ...expandedVm.spec,
+          // re-establish the reference to the instance type
+          // expand-spec removes it as the output should allow creating an independent VM
+          instancetype: innerVM.spec.instancetype,
+          preference: innerVM.spec.preference,
```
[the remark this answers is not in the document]
rszwajko: Switched to annotations.

src/views/virtualmachines/details/tabs/configuration/details/components/DetailsSectionBoot.tsx
rszwajko: The method comes from our custom hook `useToggle()`  - if it's not added to the list I get warnings in IDE

rszwajko:
@metalice @avivtur 
Reverted back to passing expanded VM and raw VM separately. 
Note that I've also dropped  the usage of ` V1VirtualMachineFromJSON(json)`.
Unfortunately the YAML editor is not supporting properties set to `undefined`.
image

src/utils/components/InstanceTypeModal/ConnectedInstanceTypeModal.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const instanceType = allInstanceTypes.find(
+    (it) => it.metadata.name === instanceTypeVM?.spec?.instancetype?.name,
+  );
+  return (
+    <>
+      {!loaded && <Loading />}
+      {loaded && loadError && <ErrorAlert error={loadError} />}
+      {loaded && !loadError && (
+        <InstanceTypeModal
+          {...otherProps}
```
metalice, line 23: as a rule of thumb , using {...otherProps} , should be avoided.
For example, lets say other props contiant 3 props.
if InstanceTypeModal , has useMemo, useCallback, that listen to only 1 prop from otherProps, it will recalc even if the 2 other props are getting new values, as {...otherProps} will always create a new object, new reference so useMemo and useCallback will fail compare and recalc. better to explicit pass props.

src/utils/components/InstanceTypeModal/ConnectedInstanceTypeModal.tsx
```diff
@@ -0,0 +1,33 @@
+import React, { FC } from 'react';
+
+import useInstanceTypesAndPreferences from '@catalog/CreateFromInstanceTypes/state/hooks/useInstanceTypesAndPreferences';
+
+import ErrorAlert from '../ErrorAlert/ErrorAlert';
+import Loading from '../Loading/Loading';
+
+import InstanceTypeModal, { InstanceTypeModalProps } from './InstanceTypeModal';
+
+const ConnectedInstanceTypeModal: FC<
```
metalice, line 10: The name of this compent isnt clear, what does it mean connected? Please try rename to something clearer

src/utils/components/InstanceTypeModal/ConnectedInstanceTypeModal.tsx
```diff
@@ -0,0 +1,33 @@
...
+> = ({ instanceTypeVM, ...otherProps }) => {
+  const { allInstanceTypes, loaded, loadError } = useInstanceTypesAndPreferences();
+  const instanceType = allInstanceTypes.find(
+    (it) => it.metadata.name === instanceTypeVM?.spec?.instancetype?.name,
+  );
+  return (
+    <>
+      {!loaded && <Loading />}
+      {loaded && loadError && <ErrorAlert error={loadError} />}
+      {loaded && !loadError && (
```
metalice, line 21: i think it more clear to work with returns

if (!loaded) return Loaidng
...

src/utils/components/PendingChanges/hooks/usePendingChanges.tsx
```diff
@@ -80,6 +84,7 @@ export const usePendingChanges = (
     isList: true,
   });
 
+  const instanceTypeChanged = checkInstanceTypeChanged(vm, vmi);
```
metalice, line 87: You introduce new IT change, what happen if u change IT? We are getting the message or IT change and memory and CPU? both of them?

rszwajko: > what happen if u change IT?
>
notification is displayed - only for the instance type (as visible on the screenshot)

src/utils/components/InstanceTypeModal/ConnectedInstanceTypeModal.tsx
rszwajko: I will check it but we should be OK here. The component receives individual props which have the same references.

src/utils/components/InstanceTypeModal/ConnectedInstanceTypeModal.tsx
rszwajko: The naming comes from  React Redux - in this context it's just that (some) data is connected to the component.
We can change it - what do you propose?

src/utils/components/InstanceTypeModal/InstanceTypeModal.tsx
```diff
@@ -17,7 +17,7 @@ import {
   mappedInstanceTypesToSelectOptions,
 } from './utils/util';
 
-type InstanceTypeModalProps = {
+export type InstanceTypeModalProps = {
```
metalice, line 20: if u are exporting type, please extract it to a util/type file

src/utils/components/InstanceTypeModal/ConnectedInstanceTypeModal.tsx
metalice: I will call this InstanceTypeModal and the inner component InstanceTypeModalContent or some variation

src/utils/components/InstanceTypeModal/ConnectedInstanceTypeModal.tsx
rszwajko: Works for me.

src/utils/components/InstanceTypeModal/ConnectedInstanceTypeModal.tsx
rszwajko: In the end I've used `StandaloneInstanceTypeModal`. The `InstanceTypeModalContent` was not accurate because both components are used directly.

src/utils/components/InstanceTypeModal/StandaloneInstanceTypeModal.tsx
```diff
@@ -0,0 +1,36 @@
...
+import Loading from '../Loading/Loading';
+
+import { InstanceTypeModalProps } from './utils/types';
+import InstanceTypeModal from './InstanceTypeModal';
+
+const StandaloneInstanceTypeModal: FC<
+  Pick<InstanceTypeModalProps, 'isOpen' | 'onClose' | 'onSubmit' | 'vm'>
+> = ({ vm, ...otherProps }) => {
+  const { allInstanceTypes, loaded, loadError } = useInstanceTypesAndPreferences();
+  const instanceType = allInstanceTypes.find(
```
metalice, line 15: shouldnt this be on useMemo? as u are passing all items each rendering

src/utils/components/InstanceTypeModal/utils/util.ts
```diff
@@ -80,34 +83,42 @@ const sortInstanceTypeSizes = (a: InstanceTypeRecord, b: InstanceTypeRecord) =>
...
 
   const matchedSeries = instanceTypesSeries.find(
-    (series) => series.displayNameSeries === instanceTypeSeries,
+    (series) => series?.displayNameSeries === instanceTypeSeries,
   );
-  const matchedSize = Object.values(matchedSeries?.sizes).find(
-    (size) => size.prettyDisplaySize === instanceTypeSize,
+
+  if (!matchedSeries) {
+    return undefined;
```
metalice, line 117: undefined isn't used for return values; undefined in JS usually means uninitialized var. Better return null

rszwajko: This is consistent with what we return currently - note that optional chaining operator `?.` and  `find()` return `undefined`.  IMHO having both in one method is overkill.

src/utils/components/InstanceTypeModal/StandaloneInstanceTypeModal.tsx
rszwajko: IMHO it makes no difference. It could be even slower  - `allInstanceTypes` is less then 20 items. We can add it if you prefer.

src/utils/components/InstanceTypeModal/utils/util.ts
metalice: ?. make sense to return undefined, https://262.ecma-international.org/6.0/#sec-terms-and-definitions-undefined-type , u can also commit the return value and simply use return , u will get same value

src/utils/components/InstanceTypeModal/StandaloneInstanceTypeModal.tsx
metalice: They are going to introduce new families of IT - so better include it as potential this number can be much larger

src/utils/components/InstanceTypeModal/StandaloneInstanceTypeModal.tsx
```diff
@@ -0,0 +1,36 @@
...
+  if (!loaded) {
+    return <Loading />;
+  }
+  if (loadError) {
+    return <ErrorAlert error={loadError} />;
+  }
+
+  return (
+    <InstanceTypeModal
+      {...otherProps}
```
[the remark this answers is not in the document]
rszwajko: the opinion is not commonly shared:)

PR #2505, opened 2025-03-07
[release-4.12] CNV-50618: Update dompurify from 2.3.10 to 2.5.8

rszwajko:
## 📝 Description

Library dompurify is an indirect dependency of @patternfly/quickstarts. The updated version matches the requested pattern ^2.2.6. Accidental downgrade is prevented by enforcing the minimum version with the 'resolutions` section.

## 🎥 Demo

none

### DIFF
--- a/package.json
+++ b/package.json
@@ -159,6 +159,7 @@
   },
   "resolutions": {
     "webpack": "^5.68.0",
-    "@types/react": "17.0.40"
+    "@types/react": "17.0.40",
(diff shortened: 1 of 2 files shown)


avivtur:
@rszwajko I don't see dompurify on main downwards, shouldn't we handle this from main branch to 4.18 going down?
cc: @metalice

avivtur (approved):

PR #2521, opened 2025-03-19
test: reenable and fix e2e tests

gouyang:
## 📝 Description
1. enable e2e test
2. use custom namespace instead of "default"
3. skip test of "creating VM", will enable it in a separate PR


cypress/tests/e2e/a-setup.cy.ts
```diff
@@ -25,6 +34,11 @@ xdescribe('Prepare the cluster for test', () => {
     });
   });
 
+  it('close welcome modal if it appears', () => {
+    cy.wait(15000);
```
rszwajko, line 38: do we need a separate test/action? it seems we are embedding `tickWelcomeModal()` in the tests anyway

gouyang: I'm trying to get the test pass, the welcome modal is appearing in later test and it's odd.

rszwajko (approved):

upalatucci (approved):

PR #2522, opened 2025-03-19
CNV-55202: TreeView folders/projects right click actions

upalatucci:
## 📝 Description

Add folders and projects right-click actions in the Treeview

## 🎥 Demo

https://github.com/user-attachments/assets/f40110c1-79da-45db-9b0a-1df1f8f4593e

### DIFF
--- a/src/views/virtualmachines/tree/components/TreeViewRightClickActionMenu/utils.ts
+++ b/src/views/virtualmachines/tree/components/TreeViewRightClickActionMenu/utils.ts
@@ -1,0 +1,44 @@
+import { NavigateFunction } from 'react-router-dom-v5-compat';
+
+import { ActionDropdownItemType } from '@kubevirt-utils/components/ActionsDropdown/constants';
+import { t } from '@kubevirt-utils/hooks/useKubevirtTranslation';
+import { getLabel, getNamespace } from '@kubevirt-utils/resources/shared';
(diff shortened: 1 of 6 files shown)


src/views/virtualmachines/tree/components/TreeViewRightClickActionMenu/index.tsx
```diff
@@ -0,0 +1,36 @@
+import React, { FC } from 'react';
```
pcbailey, line 1: Why is this file named `index.tsx` instead of the component name?

src/views/virtualmachines/tree/components/TreeViewRightClickActionMenu/utils.ts
```diff
@@ -0,0 +1,29 @@
...
+): ActionDropdownItemType => ({
+  cta: () => navigate(`/k8s/ns/${namespace}/catalog`),
+  id: 'create-vm',
+  label: t('Create VirtualMachine'),
+});
+
+export const getVMsTrigger = (triggerElement: HTMLElement | null) => {
+  if (!triggerElement) return [];
+
+  const [prefix, namespace, folderName] = triggerElement?.id?.split('/') || ['', '', ''];
```
pcbailey, line 21: `triggerElement?.id?.split('/')` appears to be used a few times. Consider making it a function with a default value argument:

`const [prefix, namespace, folderName] = getTriggerElementComponents(triggerElement, ['', '', '']);`

PR #2524, opened 2025-03-19
CNV-55151: Choose access and volume mode based on recommendations from StorageProfile

rszwajko:
## 📝 Description

Based on the feedback in [CNV-55151](https://issues.redhat.com//browse/CNV-55151) and in [CNV-55147](https://issues.redhat.com/browse/CNV-55147) we should change the current behavior:
1. instead of "Apply optimized StorageProfile settings" section - expose all the options send via storage profile API. The first ([the most recommended](https://github.com/kubevirt/containerized-data-importer/blob/main/doc/storageprofile.md)) should be the default but the user should be allowed to change and even override the recommendation.
2. drop the section based on static data
3. the list of available modes visible will be enhanced with:
      1. label indicating the recommendation status( 3 levels): most recommended, recommended or not recommended
      2.  tooltip providing more information - including recommendation rank
4. the volume mode change triggers changes in the access mode recommendations - due to this fact the section was moved to the left side.

Note that validation based on static (and likely outdated) data was not removed in [UploadPVCFormAccessMode.tsx](https://github.com/kubevirt-ui/kubevirt-plugin/blob/4b4e63d0b1fa329927b98be9d07510b2c5fc6fbc/src/views/cdi-upload-provider/upload-pvc-form/UploadPVCFormAccessMode.tsx#L39) and [UploadPVCFormVolumeMode.tsx](https://github.com/kubevirt-ui/kubevirt-plugin/blob/4b4e63d0b1fa329927b98be9d07510b2c5fc6fbc/src/views/cdi-upload-provider/upload-pvc-form/UploadPVCFormVolumeMode.tsx#L32). 

### Before
Assuming the following  list of `claimPropertySets`:
```
[
  {"accessModes": ["ReadWriteMany"],"volumeMode":"Block"},
  {"accessModes": ["ReadWriteOnce"],"volumeMode": "Block"},
  {"accessModes": ["ReadWriteOnce"],"volumeMode": "Filesystem"}
]

```
On the screen, user has 2 mutually exclusive choices
1. "Apply optimized StorageProfile settings" - here the first item from the `claimPropertySets` is displayed. The actual values get cleared and (eventually) server side will apply them . Note (on the screenshot below) that user cannot chose between options.
image

6. "manual" - the list is based on (and validated against) static data - which seems an outdated version of [storagecapabilities.go](https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/storagecapabilities/storagecapabilities.go).  Note (on the screenshot below) that the options don't match those send via StorageProfile API
image

#### Layout
Screenshot From 2025-04-03 20-57-00

#### All tooltips 
storage_profile_tooltipsV2

#### Screencast
https://github.com/user-attachments/assets/767dd68f-a732-4840-9b4f-7147d498b327


src/utils/components/DiskModal/components/StorageProfileSettings/ApplyStorageProfileSettings.tsx
```diff
@@ -1,79 +1,57 @@
...
+  const [storage] = watch([DATAVOLUME_TEMPLATE_STORAGE]);
+  const { storageClassName } = storage;
 
-  const { accessModes, storageClassName, volumeMode } = storage;
+  const { claimPropertySets, error, loaded } = useStorageProfileClaimPropertySets(storageClassName);
 
-  const { claimPropertySets, loaded: storageProfileLoaded } =
-    useStorageProfileClaimPropertySets(storageClassName);
+  if (!loaded) {
+    return <Skeleton screenreaderText={t('Loading Storage Profile')} />;
```
avivtur, line 28: in CNV we have a convention that a CR names like StorageProfile as PascalCase meaning Storage Profile should be presented as StorageProfile.

src/utils/components/DiskModal/components/StorageProfileSettings/ApplyStorageProfileSettings.tsx
```diff
@@ -1,79 +1,57 @@
...
-      <ApplyStorageProfileSettingsCheckbox
-        claimPropertySets={claimPropertySets}
-        disabled={!storageProfileLoaded || !claimPropertySets || isEmpty(claimPropertySets)}
-        handleChange={handleApplyOptimizedSettingsChange}
-        isChecked={storageProfileSettingsApplied}
+    <>
+      <Checkbox
+        id="allow-all-storage-profile-options"
+        isChecked={allowAllModes}
+        label={t('Enable options not recommended by Storage Profile API')}
```
avivtur, line 40: here too please change to StorageProfile

src/utils/components/DiskModal/components/StorageProfileSettings/RecommendationLabel.tsx
```diff
@@ -0,0 +1,40 @@
+import React, { FC } from 'react';
+
+import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
+import { Label, Tooltip } from '@patternfly/react-core';
+
+const RecommendationLabel: FC<{ priority: number }> = ({ priority }) => {
+  const { t } = useKubevirtTranslation();
+  const labelsForRecommendations = [
+    <Tooltip content={t('The most recommended option according to Storage Profile API')} key="1st">
```
avivtur, line 9: here too please change to StorageProfile on all text

src/utils/components/DiskModal/components/StorageProfileSettings/RecommendationLabel.tsx
```diff
@@ -0,0 +1,40 @@
+import React, { FC } from 'react';
+
+import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
+import { Label, Tooltip } from '@patternfly/react-core';
+
+const RecommendationLabel: FC<{ priority: number }> = ({ priority }) => {
```
avivtur, line 6: please add type for component props above component

src/utils/components/DiskModal/components/StorageProfileSettings/RecommendationLabel.tsx
avivtur, line 1: The component here is reaping the <Tooltip><Label></Tooltip> component with different values.
Consider extracting the contents to a constant and render to component appropriately according to the priority

src/utils/components/DiskModal/components/StorageProfileSettings/VolumeMode.tsx
```diff
@@ -1,58 +1,69 @@
-import React, { FC, useEffect, useMemo } from 'react';
-import { Controller, useFormContext } from 'react-hook-form';
+import React, { FC, useEffect } from 'react';
+import { useFormContext } from 'react-hook-form';
+import { uniq } from 'lodash';
```
avivtur, line 3: Since when have we entered loadsh in the project? @metalice @upalatucci @pcbailey 
I thought we wanted to avoid this package as it's big and not needed
https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore, please find the the replacement here and and it to same location as isEmpty is found

src/utils/components/DiskModal/components/utils/modesMapping.ts
```diff
@@ -1,8 +1,11 @@
+import { uniq } from 'lodash';
```
avivtur, line 1: here too please switch the uniq func with our own creation

src/utils/components/DiskModal/components/StorageProfileSettings/VolumeMode.tsx
```diff
@@ -1,58 +1,69 @@
...
-
-  const allowedVolumeModes = useMemo(
-    () => getVolumeModeForProvisioner(storageClassProvisioner, accessModes?.[0]),
-    [accessModes, storageClassProvisioner],
+  const recommendedVolumeModes = uniq(
+    claimPropertySets
+      .map((it) => it.volumeMode)
+      .filter(Boolean)
+      .map((mode): V1beta1StorageSpecVolumeModeEnum => V1beta1StorageSpecVolumeModeEnum[mode])
+      .filter(Boolean),
```
avivtur, line 32: add optional chaining in case claimPropertySets is undefined

src/utils/components/DiskModal/components/utils/modesMapping.ts
```diff
@@ -252,3 +255,13 @@ export const getVolumeModeForProvisioner = (
 
   return initialVolumeModes;
 };
+
+export const getAccessModesForVolume = (volumeMode: string, claimPropertySets: ClaimPropertySets) =>
```
avivtur, line 259: add optional chaining in case claimPropertySets is undefined

avivtur (commented):
@rszwajko Hi :) 
the screenshot suggest that there's no spacing between the text to the label.
and I'm not sure if it's just me, but having a checkmark at both 1st and 2nd labels but only one is selected that's confusing.
Also consider changing the 1st, 2nd, etc labels to have isCompact prop, I think it will have a nicer look in the modal.

src/utils/components/DiskModal/components/StorageProfileSettings/VolumeMode.tsx
rszwajko: It seems we need to discuss the strategy:) I've used it because it's already there. Note that there are modern alternatives like [radash](https://radash-docs.vercel.app/docs/getting-started)

src/utils/components/DiskModal/components/StorageProfileSettings/VolumeMode.tsx
rszwajko: At the component level it should be well defined (may be empty). The typing reflects that. The fallback it's handled at the parent level i.e.
```
<AccessMode allowAllModes={allowAllModes} claimPropertySets={claimPropertySets ?? []} />
```

src/utils/components/DiskModal/components/StorageProfileSettings/RecommendationLabel.tsx
rszwajko: I'll look into it. Note that component approach allows flexible styling and structure.

src/utils/components/DiskModal/components/StorageProfileSettings/VolumeMode.tsx
rszwajko: OK. I've just realized we are compiling with `strict: false`.

src/utils/components/DiskModal/components/StorageProfileSettings/ApplyStorageProfileSettings.tsx
rszwajko: OK. I had a similar case with InstanceType and I was told to change it back to "Instance Type".
CC: @metalice

src/utils/components/DiskModal/components/StorageProfileSettings/VolumeMode.tsx
rszwajko: @avivtur the approach in https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore?tab=readme-ov-file#_uniq cannot be used as we need to keep the initial order (most recommended come first).

src/utils/components/DiskModal/components/StorageProfileSettings/RecommendationLabel.tsx
rszwajko: The labels may differ in structure - I would leave them in one component (without template).

src/utils/components/DiskModal/components/StorageProfileSettings/VolumeMode.tsx
metalice: how do u know the recommendation?

metalice:
@rszwajko im missing description for the pr to understand whats going on.
About those thumbs up icons, did you checked with Yifat?

src/utils/components/DiskModal/components/utils/modesMapping.ts
```diff
@@ -252,3 +255,16 @@ export const getVolumeModeForProvisioner = (
...
+  claimPropertySets: ClaimPropertySets,
+  volumeMode?: string,
+) =>
+  uniq(
+    claimPropertySets
+      .filter((it) => it.volumeMode === volumeMode)
+      .flatMap((it) => it.accessModes)
+      .filter(Boolean)
+      .map((mode) => V1beta1StorageSpecAccessModesEnum[mode])
+      .filter(Boolean),
```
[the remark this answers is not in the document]
rszwajko: the current flow seemed the most readable for me - what would  you suggest instead?

rszwajko:
No functional changes in the last force push - the PR was rebased on top of current main and commit message was provided.

src/utils/components/DiskModal/components/utils/modesMapping.ts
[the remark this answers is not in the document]
metalice: `[].reduce((acc, it) => {
    if (it.volumeMode === volumeMode && it.accessModes) {
      const modes = it.accessModes.flatMap((mode) => V1beta1StorageSpecAccessModesEnum[mode]);
      acc.push(...modes);
    }
  }, []);`

something like that may work

metalice:
@rszwajko did you contact Yifat and/or Ronen about replacing the icons with a label?

src/utils/components/DiskModal/components/StorageProfileSettings/RecommendationLabel.tsx
rszwajko: The icons were removed in the newest iteration and the structure was similar enough to use one component.

src/utils/components/DiskModal/components/utils/modesMapping.ts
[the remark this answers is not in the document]
rszwajko: if you have no strong opinion I would stick to the existing code

src/utils/components/DiskModal/components/StorageProfileSettings/AccessMode.tsx
```diff
@@ -1,56 +1,56 @@
...
   const { t } = useKubevirtTranslation();
 
   const { setValue, watch } = useFormContext<V1DiskFormState>();
 
-  const [storageClassProvisioner, accessModes, volumeMode] = watch([
-    STORAGE_CLASS_PROVIDER_FIELD,
-    ACCESS_MODE_FIELD,
-    VOLUME_MODE_FIELD,
-  ]);
+  const [accessModes, volumeMode] = watch([ACCESS_MODE_FIELD, VOLUME_MODE_FIELD]);
```
metalice, line 25: better to useWatch hook, it will cause less rendering then watch()

adamviktora (commented):
Hello @rszwajko, I was looking at the updates from user's perspective from now, I will add a code review later.

If I understand the issue https://issues.redhat.com/browse/CNV-55147 correctly, it may happen that: "DV creation happens forever, status is CloneInProgress" if user sets up the volume/access mode incorrectly. 

If that is the case, I think it shouldn't be even possible to set it up this way and click "Save" from the UI.

I assume that choosing this option:
Screenshot 2025-04-03 at 13 47 32 copy
will lead to that bug.

If all of the above is true (it is possible that I just misunderstood it), then I don't see a reason why is the option enabled. And adding "Not recommended" without a warning indication is not enough to prevent a bug IMHO.

I see that in a previous version you had the problematic options disabled:
Screenshot 2025-04-03 at 14 09 08
I think that is very good. What was a reason for changing that?

Also I think that using colors to distinguish between the Most recommended / Recommended / Not recommended labels is a good idea - because right now the difference between Most recommended and Not recommended (which may lead to a bug) is just two letters. And I don't see a reason for having the middle Recommended label.

Status labels can be used: https://www.patternfly.org/components/label#outlined-labels, success for most recommended (maybe without the icon) and danger for Not recommended - or rather disable it entirely if possible. 

I am not a UX person, this is just how I'd feel about it as a user using the product.

rszwajko:
@adamviktora 
>it may happen that: "DV creation happens forever, status is CloneInProgress" if user sets up the volume/access mode incorrectly.
> If that is the case, I think it shouldn't be even possible to set it up this way and click "Save" from the UI.
>
It might be the right choice i.e. if user knows that this mode will be supported (but was not configured yet).
But user should really know what he's doing.

> I see that in a previous version you had the problematic options disabled:
> I think that is very good. What was a reason for changing that?
>
I share your opinion here. The change was requested in  review by @metalice and confirmed by UXD.

> Also I think that using colors to distinguish between the Most recommended / Recommended / Not recommended labels is a good idea - because right now the difference between Most recommended and Not recommended (which may lead to a bug) is just two letters. 
>
+1

> And I don't see a reason for having the middle Recommended label.
> 
Use may chose less optimal option for some reason. In theory all options may be available in the recommendation but with different priority/rank.

rszwajko:
@metalice @adamviktora 
Latest push contains the new messages suggested by the Doc team and colors for the labels suggested by @adamviktora

src/utils/components/DiskModal/components/StorageProfileSettings/RecommendationLabel.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  if (priority > 0 && priority < recommendationCount) {
+    return {
+      label: t('Recommended'),
+      tooltip: tooltipForOtherRecommended(priority, t),
+    };
+  }
+
+  return {
+    className: 'pf-m-warning',
```
adamviktora, line 41: Can we add the warning icon here too? With `status="warning"`, or even `status="danger"`, I think that just the warning color is not enough

src/utils/components/DiskModal/components/StorageProfileSettings/ApplyStorageProfileSettings.scss
```diff
@@ -3,4 +3,12 @@
     margin-left: var(--pf-t--global--spacer--md);
     margin-top: var(--pf-t--global--spacer--sm);
   }
+  &--labelWithGap {
+    display: flex;
+    gap: 5px;
+    align-items: anchor-center;
+  }
+  &--rowPadding {
+    padding-bottom: 5px;
```
adamviktora, line 12: Instead of this class, can we add `isStack` to the FormGroup which is wrapping the Radios?

src/utils/components/DiskModal/components/StorageProfileSettings/ApplyStorageProfileSettings.scss
```diff
@@ -3,4 +3,12 @@
     margin-left: var(--pf-t--global--spacer--md);
     margin-top: var(--pf-t--global--spacer--sm);
   }
+  &--labelWithGap {
+    display: flex;
+    gap: 5px;
```
adamviktora, line 8: I'd avoid using hardcoded pixel values and use `var(--pf-t--global--spacer--sm)` instead

adamviktora (commented):
@rszwajko Thanks for pushing the update, this is looking good!

About the "Recommended" labels, my point was that it feels like there are too many labels. If the point is to tell users via the tooltip if the option is 2nd / 3rd (or possibly 4th, but currently there are not that many options to choose from), then the tooltip could be on the option itself and it wouldn't be so visually overloaded (but then it wouldn't be consistent with the options that have the label on the tooltip). So I'm OK with keeping it like this :)

Few comments below regarding some style updates

src/utils/components/DiskModal/components/StorageProfileSettings/RecommendationLabel.tsx
rszwajko: I agree. Actually the icon was there in the prev iteration but I was asked to remove it.  That's why I'm using CSS classes instead - if you use `status` you need to use the icon (place is reserved inside the label)

rszwajko:
@adamviktora 
Added suggested CSS improvements in the newest force push. Visually the changes are hard to notice(see below) so I've kept the old screenshots in the description.
### After
image

adamviktora:
Thanks! Maybe a little more margin between Volume mode and Access mode would be good. I don't wanna be too nit-picky though.

rszwajko:
@adamviktora 
> Maybe a little more margin between Volume mode and Access
>
Beside proximity  we have no other indication that those options are closely related. With more spacing they may visually "drift" away. I would keep it as it is (unless I get more feedback suggesting this direction).

src/utils/components/DiskModal/components/StorageProfileSettings/VolumeMode.tsx
```diff
@@ -1,58 +1,96 @@
...
+      labelHelp={
+        <HelpTextIcon
+          bodyContent={
+            <Trans ns="plugin__kubevirt-plugin" t={t}>
+              Learn more about
+              <Link target="_blank" to={documentationURL.STORAGE_PROFILES}>
+                {' '}
+                StorageProfile
+              </Link>
+              .
```
metalice, line 67: ?

src/utils/components/DiskModal/components/StorageProfileSettings/VolumeMode.tsx
```diff
@@ -1,58 +1,96 @@
...
 
   useEffect(() => {
-    if (!allowedVolumeModes?.includes(volumeMode as V1beta1StorageSpecVolumeModeEnum)) {
-      setValue(VOLUME_MODE_FIELD, allowedVolumeModes[0]);
+    if (volumeMode || !recommendedVolumeModes?.[0]) {
+      return;
     }
-  }, [allowedVolumeModes, volumeMode, setValue]);
+    const mostRecommended = recommendedVolumeModes[0];
+    setBothModes(mostRecommended);
```
metalice, line 53: something in the loigc here is wrong, useEffect trigger setBotModes and setBothModes trigger useEffect

src/utils/components/DiskModal/components/StorageProfileSettings/VolumeMode.tsx
```diff
@@ -1,58 +1,96 @@
...
+const VolumeMode: FC<VolumeModeProps> = ({ claimPropertySets }) => {
   const { t } = useKubevirtTranslation();
 
-  const { control, setValue, watch } = useFormContext<V1DiskFormState>();
+  const { setValue, watch } = useFormContext<V1DiskFormState>();
 
   const volumeMode = watch(VOLUME_MODE_FIELD);
-  const accessModes = watch(ACCESS_MODE_FIELD) as V1beta1StorageSpecAccessModesEnum[];
-  const storageClassProvisioner = watch(STORAGE_CLASS_PROVIDER_FIELD);
+  const recommendedVolumeModes = uniq(
```
metalice, line 31: why uniq? u can use set, its string values

src/utils/components/DiskModal/components/StorageProfileSettings/RecommendationLabel.tsx
```diff
@@ -0,0 +1,60 @@
...
+    case 2:
+      return t('The 3rd most suggested option in StorageProfile.');
+    default:
+      return t('The {{priority}}th most suggested option in StorageProfile.', {
+        priority: priority + 1,
+      });
+  }
+};
+
+const fromPriorityToLabels = (priority, recommendationCount, t) => {
```
metalice, line 24: this need to be in a util file and no need to send t as arg

metalice: please add types to function

src/utils/components/DiskModal/components/StorageProfileSettings/VolumeMode.tsx
```diff
@@ -1,58 +1,96 @@
...
+import './ApplyStorageProfileSettings.scss';
+type VolumeModeProps = {
+  claimPropertySets: ClaimPropertySets;
+};
+
+const VolumeMode: FC<VolumeModeProps> = ({ claimPropertySets }) => {
   const { t } = useKubevirtTranslation();
 
-  const { control, setValue, watch } = useFormContext<V1DiskFormState>();
+  const { setValue, watch } = useFormContext<V1DiskFormState>();
```
metalice, line 28: again using watch? why not from useWatch ?

src/utils/components/DiskModal/components/StorageProfileSettings/ApplyStorageProfileSettings.tsx
```diff
@@ -1,79 +1,48 @@
...
-          </FlexItem>
-          <FlexItem>
-            <VolumeMode />
-          </FlexItem>
-        </Flex>
-      )}
-    </div>
+    <Flex
+      className="ApplyStorageProfileSettings--volume-access-section"
+      spaceItems={{ default: 'spaceItems3xl' }}
```
metalice, line 37: no need ot mix between CN and styling props, if you have both, add the styling prop to CN, no need 2 places for styling

src/utils/components/DiskModal/components/StorageProfileSettings/AccessMode.tsx
rszwajko: > where do u use setValue before initial state?
>
Already pointed above:
>However we do use setValue before registering the hook i.e. when setting [access mode based on volume mode](https://github.com/rszwajko/kubevirt-plugin/blob/bc4cdcf03cd198b7aeca7820c60b76e882bb407f/src/utils/components/DiskModal/components/StorageProfileSettings/VolumeMode.tsx#L56) or when r[esetting the form values](https://github.com/rszwajko/kubevirt-plugin/blob/bc4cdcf03cd198b7aeca7820c60b76e882bb407f/src/utils/components/DiskModal/components/StorageClassAndPreallocation/StorageClassSelect.tsx#L53) after storage class change
>

> if u have effect that setValue, before all initialize just give a default value
>
There is no static default here. The value needs to be retrieved dynamically.

src/utils/components/DiskModal/components/StorageProfileSettings/VolumeMode.tsx
rszwajko: The message was suggested and accepted by the doc team - see https://issues.redhat.com/browse/CNV-55151?focusedId=26922579&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-26922579

src/utils/components/DiskModal/components/StorageProfileSettings/VolumeMode.tsx
rszwajko: The approach follows the existing pattern - `useEffect` hooks are used to update the form based on dynamically retrieved data. We could change the pattern and/or replace react hook form but this requires follow-up issue. 
As for this particular hook: it is triggered with every volume change but early return works as expected and ensures that the setter is used only if volume is undefined (initial run and then on every storage class change)

src/utils/components/DiskModal/components/StorageProfileSettings/VolumeMode.tsx
rszwajko: We've already discussed this in https://github.com/kubevirt-ui/kubevirt-plugin/pull/2524#discussion_r2007942877
Basically we need to keep the initial order.

src/utils/components/DiskModal/components/StorageProfileSettings/RecommendationLabel.tsx
rszwajko: > no need to send t as arg
>
It depends if we want to support changing the language without reloading the page. If not then we can move the translations outside of React lifecycle.

src/utils/components/DiskModal/components/StorageProfileSettings/VolumeMode.tsx
rszwajko: as discussed above https://github.com/kubevirt-ui/kubevirt-plugin/pull/2524#discussion_r2023068263 it seems we cannot use it without re-designing the data flow in this area.

src/utils/components/DiskModal/components/StorageProfileSettings/ApplyStorageProfileSettings.tsx
rszwajko: AFAIK the component provided styling is always preferred  due to underlying logic i.e https://github.com/patternfly/patternfly-react/blob/4ea82f7af5e42834f2a906ec5c7cc1912466ca2b/packages/react-core/src/layouts/Flex/Flex.tsx#L519

metalice: how its the prefered way? form what u sent I see that CN can override as its the last

src/utils/components/DiskModal/components/StorageProfileSettings/VolumeMode.tsx
rszwajko: I have no strong opinion. The message was approved by the doc team so I think it's fine (and they do noticed missing dots in other places).

src/utils/components/DiskModal/components/StorageProfileSettings/ApplyStorageProfileSettings.tsx
metalice: "However, the manual override would start failing once we started using other provided styling options." I don't know and am not sure at all. Anyway, this link misses the point of having one place of styling instead of two.

rszwajko: The main source of styling is the component built-in styling. The `className` is only supplemental (for parts not covered by the component).

adamviktora: I agree with Radek on this one. IMHO we should use as much PatternFly default props as possible and avoid custom CSS overrides. I have already seen some bugs caused by custom CSS and they may occur again when upgrading to newer PF version in the future.

Also `spaceItems` is not the same as `gap`, it can't be replaced with just one CSS prop, it adds margin to every but last child.

metalice:
@rszwajko please rebase lets merge it

PR #2528, opened 2025-03-24
CNV-54573: fix drop vm with no label error

upalatucci:
## 📝 Description

Error on dropping VM with no labels.
Store the dragging VM instead of only namespace to make use to patch it accordingly.

Adding the /metadata/labels to a VM that already have it, it will means to wipe it every other labels else out

### DIFF
--- a/src/views/virtualmachines/tree/hooks/dragndrop.ts
+++ b/src/views/virtualmachines/tree/hooks/dragndrop.ts
@@ -1,5 +1,8 @@
 import { VirtualMachineModel } from 'src/views/dashboard-extensions/utils';
 
+import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import { getLabels, getNamespace } from '@kubevirt-utils/resources/shared';
+import { isEmpty } from '@kubevirt-utils/utils/utils';
 import { k8sPatch } from '@openshift-console/dynamic-plugin-sdk';
 import { TreeViewDataItem } from '@patternfly/react-core';
 import { VM_FOLDER_LABEL } from '@virtualmachines/tree/utils/constants';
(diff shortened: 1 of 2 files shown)


src/views/virtualmachines/tree/hooks/dragndrop.ts
```diff
@@ -1,39 +1,46 @@
...
+import { isEmpty } from '@kubevirt-utils/utils/utils';
 import { k8sPatch } from '@openshift-console/dynamic-plugin-sdk';
 import { TreeViewDataItem } from '@patternfly/react-core';
 import { VM_FOLDER_LABEL } from '@virtualmachines/tree/utils/constants';
 
 import { DRAG_N_DROP_LISTENERS, VALID_DRAG_TARGET_BACKGROUND_COLOR } from './constants';
+import { getVMFromElementID } from './utils';
 
-let draggingVMNamespace: null | string = null;
+let draggingVM: null | V1VirtualMachine = null;
```
metalice, line 13: why are we using let here?

upalatucci: becuase we change it when the user starts to drag. Should I use signals?

src/views/virtualmachines/tree/hooks/dragndrop.ts
```diff
@@ -59,18 +66,14 @@ export const dropEventListeners = (treeViewItem: TreeViewDataItem): RemoveListen
...
 
-    const [namespace, name] = vmID.split('/');
-
-    changeVMFolder(name, namespace, folderName);
+    changeVMFolder(folderName);
 
     folderHTMLElement.style.backgroundColor = 'unset';
   };
 
   const dragOverHandler = (ev) => {
```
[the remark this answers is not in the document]
upalatucci: I'll fix that on the other pr https://github.com/kubevirt-ui/kubevirt-plugin/pull/2529

pcbailey:
I'm good with this once you address Matan's `let` comment.

PR #2530, opened 2025-03-24
CNV-57968: fix selected VMs state after deletion from CLI

adamviktora:
## 📝 Description

Actually fixes https://issues.redhat.com/browse/CNV-57968. Guohua found out that the state of selected VMs does not reflect changes to the VMs list from CLI. 

I kept the previous fix https://github.com/kubevirt-ui/kubevirt-plugin/pull/2515 (which was only fixing deletion from UI) so we don't keep unnecessary deleted VMs inside `selectedVMs` signal.

## 🎥 Demo

Before:

https://github.com/user-attachments/assets/97fdfd6f-12c5-4a2b-b90f-22405154b142

After:

https://github.com/user-attachments/assets/3d2d583a-fbff-46b0-84de-850a7961aea4

### DIFF
--- a/src/views/virtualmachines/list/components/VirtualMachineSelection/VirtualMachineSelection.tsx
+++ b/src/views/virtualmachines/list/components/VirtualMachineSelection/VirtualMachineSelection.tsx
@@ -11,8 +11,9 @@ import {
   MenuToggle,
   MenuToggleElement,
 } from '@patternfly/react-core';
+import useExistingSelectedVMs from '@virtualmachines/list/hooks/useExistingSelectedVMs';
 
-import { deselectAll, selectAll, selectedVMs } from '../../selectedVMs';
+import { deselectAll, selectAll } from '../../selectedVMs';
 
 import './virtual-machine-selection.scss';
 
@@ -25,6 +26,7 @@ type VirtualMachineSelectionProps = {
 const VirtualMachineSelection: FC<VirtualMachineSelectionProps> = ({ loaded, pagination, vms }) => {
   const { t } = useKubevirtTranslation();
   const [isOpen, setIsOpen] = useState(false);
+  const existingSelectedVMs = useExistingSelectedVMs(vms);
 
   if (!loaded) return null;
 
@@ -42,8 +44,9 @@ const VirtualMachineSelection: FC<VirtualMachineSelectionProps> = ({ loaded, pag
 
   const numCurrentPageVMs = currentPageVMs.length;
 
-  const partiallyChecked = !isEmpty(selectedVMs.value) && selectedVMs.value.length !== vms.length;
-  const isChecked = selectedVMs.value.length === vms.length;
+  const selectionIsEmpty = isEmpty(existingSelectedVMs);
+  const isChecked = !selectionIsEmpty && existingSelectedVMs.length === vms.length;
+  const partiallyChecked = !selectionIsEmpty && !isChecked;
 
   return (
     <Dropdown
@@ -56,8 +59,8 @@ const VirtualMachineSelection: FC<VirtualMachineSelectionProps> = ({ loaded, pag
         >
           <Checkbox
             label={
-              !isEmpty(selectedVMs.value)
-                ? t('{{length}} selected', { length: selectedVMs.value.length })
+              !selectionIsEmpty
+                ? t('{{length}} selected', { length: existingSelectedVMs.length })
                 : null
             }
             onChange={(event, checked) => {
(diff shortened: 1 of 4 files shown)


src/views/virtualmachines/list/hooks/useExistingSelectedVMs.ts
```diff
@@ -0,0 +1,27 @@
...
+import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import { getName, getNamespace } from '@kubevirt-utils/resources/shared';
+
+import { VMIdentifier } from '../selectedVMs';
+
+const useExistingSelectedVMs = (vms: V1VirtualMachine[], selectedVMs: VMIdentifier[]) => {
+  const createName = (vmId: VMIdentifier) => `${vmId.name}-${vmId.namespace}`;
+
+  const vmNames = useMemo(() => {
+    const vmsSet = new Set<string>();
```
rszwajko, line 12: you can create Set directly from an array i.e. sth like `new Set(vms.map(createName))`

rszwajko (commented):
The hook looks good but I feel we have a general problem with `selectedVMs`:  
1. if we needed a wrapper hook in one place,  we need it in all places where it is used.  
2. it's not triggering re-renders but we use it as it would do that i.e. in VirtualMachineBulkActionButton

src/views/virtualmachines/list/hooks/useExistingSelectedVMs.ts
```diff
@@ -0,0 +1,27 @@
+import { useMemo } from 'react';
+
+import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import { getName, getNamespace } from '@kubevirt-utils/resources/shared';
+
+import { VMIdentifier } from '../selectedVMs';
+
+const useExistingSelectedVMs = (vms: V1VirtualMachine[], selectedVMs: VMIdentifier[]) => {
```
upalatucci, line 8: `selectedVMs` is a signal and can be directly imported in this file without having it as a input.
We can use this hook in the `VirtualMachineBulkActionButton` component and also in the `VirtualMachinesList`

src/views/virtualmachines/list/components/VirtualMachineSelection/VirtualMachineSelection.tsx
```diff
@@ -42,8 +44,9 @@ const VirtualMachineSelection: FC<VirtualMachineSelectionProps> = ({ loaded, pag
 
   const numCurrentPageVMs = currentPageVMs.length;
 
-  const partiallyChecked = !isEmpty(selectedVMs.value) && selectedVMs.value.length !== vms.length;
-  const isChecked = selectedVMs.value.length === vms.length;
+  const partiallyChecked =
+    !isEmpty(existingSelectedVMs) && existingSelectedVMs.length !== vms.length;
```
avivtur, line 48: why we need to check that the existingSelectedVMs.length !== vms.length ?

src/views/virtualmachines/list/selectedVMs.ts
```diff
@@ -3,7 +3,9 @@ import { getName, getNamespace } from '@kubevirt-utils/resources/shared';
 import { isEmpty } from '@kubevirt-utils/utils/utils';
 import { signal } from '@preact/signals-core';
 
-export const selectedVMs = signal<{ name: string; namespace: string }[]>([]);
+export type VMIdentifier = { name: string; namespace: string };
```
avivtur, line 6: please avoid using acronyms like VM, I suggest something like VirtualMachineIdentifier.
This also the kind of type that can actually suit all k8s CRs, so consider making it more general :)

src/views/virtualmachines/list/components/VirtualMachineSelection/VirtualMachineSelection.tsx
upalatucci: To understand if the user selected all the vms or just a subset.

The checkbox will show a check if the user selected all the vms. If selected just some of them, it will show a different state. The `-` in the screenshots

adamviktora: I rewrote this to:
```
  const isChecked = existingSelectedVMs.length === vms.length;
  const partiallyChecked = !isChecked && !isEmpty(existingSelectedVMs);
  ```
  
 So it is a bit more clear and we can use the isChecked variable

PR #2535, opened 2025-03-25
CNV-58306: Use stricter typing for InstanceTypeModal

rszwajko:
## 📝 Description

Extend safeguards and consolidate existing types.
The crash while editing custom InstanceType should be already prevented with #2503     
For non-Red Hat InstanceType and/or instance type not following the naming convention there will be no initial selection (see the screenshot below).

### After
Editing  a VM based on custom InstanceType `foointancetype` 
Screenshot From 2025-03-25 12-23-40

### DIFF
--- a/src/utils/components/InstanceTypeModal/utils/util.ts
+++ b/src/utils/components/InstanceTypeModal/utils/util.ts
@@ -1,6 +1,5 @@
 import { parseSize } from 'xbytes';
 
-import { InstanceTypeSize } from '@catalog/CreateFromInstanceTypes/components/SelectInstanceTypeSection/utils/types';
 import {
   INSTANCETYPE_CLASS_DISPLAY_NAME,
   INSTANCETYPE_DESCRIPTION_ANNOTATION,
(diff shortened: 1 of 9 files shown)


src/utils/components/InstanceTypeModal/utils/util.ts
```diff
@@ -1,28 +1,22 @@
...
-} from './types';
-
-export const getInstanceTypeItemSizePrettyDisplay = (it: InstanceTypeUnion): string =>
-  `${it?.metadata.name.split('.').pop()}: ${it?.spec?.cpu?.guest} ${t('CPUs')}, ${
-    it?.spec?.memory?.guest
-  } ${t('Memory')}`;
+import { InstanceTypeRecord, MappedInstanceTypes } from './types';
 
+export const getInstanceTypeItemSizePrettyDisplay = (it: InstanceTypeUnion): string => {
+  const name = it?.metadata?.name?.split('.').pop() ?? it?.metadata?.name ?? '';
```
pcbailey, line 17: If we add `K8sResourceCommon` to the `InstanceTypeUnion` type you can use the `getName` selector, which will make this look a little cleaner.

It would also be nice to add selectors to src/utils/resources/instancetype for the 'guest' and 'memory' properties below.

src/utils/components/InstanceTypeModal/utils/util.ts
```diff
@@ -1,28 +1,22 @@
...
-  MappedInstanceTypes,
-} from './types';
-
-export const getInstanceTypeItemSizePrettyDisplay = (it: InstanceTypeUnion): string =>
-  `${it?.metadata.name.split('.').pop()}: ${it?.spec?.cpu?.guest} ${t('CPUs')}, ${
-    it?.spec?.memory?.guest
-  } ${t('Memory')}`;
+import { InstanceTypeRecord, MappedInstanceTypes } from './types';
 
+export const getInstanceTypeItemSizePrettyDisplay = (it: InstanceTypeUnion): string => {
```
pcbailey, line 16: Suggestion: `getInstanceTypeSizePrettyDisplay` (I think the 'Item' is unnecessary)

upalatucci (approved):

PR #2552, opened 2025-04-01
CNV-58333: Enable automatic image

upalatucci:
## 📝 Description

- enable switch even when the feature is not set in the configuration. before this change, the switch was enabled only when the featuregate was manually set true or false
- use the `spec.enableCommonBootImageImport` field as `spec.featureGates.enableCommonBootImageImport` is deprecated 

## 🎥 Demo

Screenshot 2025-04-01 at 10 38 07

### DIFF
--- a/src/views/clusteroverview/SettingsTab/ClusterTab/components/GeneralSettings/TemplatesAndImagesManagement/components/AutomaticImagesDownload/AutomaticImagesDownload.tsx
+++ b/src/views/clusteroverview/SettingsTab/ClusterTab/components/GeneralSettings/TemplatesAndImagesManagement/components/AutomaticImagesDownload/AutomaticImagesDownload.tsx
@@ -29,7 +29,10 @@ const AutomaticImagesDownload: FC<AutomaticImagesDownloadProps> = ({
 
   const [hyperConverged, loaded] = hyperConvergeConfiguration;
   const isEnabledAutomaticImagesDownload =
-    hyperConverged?.spec?.featureGates?.enableCommonBootImageImport;
+    hyperConverged?.spec?.enableCommonBootImageImport !== undefined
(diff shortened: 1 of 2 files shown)


src/views/clusteroverview/SettingsTab/ClusterTab/components/GeneralSettings/TemplatesAndImagesManagement/components/AutomaticImagesDownload/AutomaticImagesDownload.tsx
```diff
@@ -28,8 +29,12 @@ const AutomaticImagesDownload: FC<AutomaticImagesDownloadProps> = ({
   const isAdmin = useIsAdmin();
 
   const [hyperConverged, loaded] = hyperConvergeConfiguration;
-  const isEnabledAutomaticImagesDownload =
-    hyperConverged?.spec?.featureGates?.enableCommonBootImageImport;
+  const isEnabledAutomaticImagesDownload = isEmpty(
+    hyperConverged?.spec?.enableCommonBootImageImport,
```
metalice, line 33: if its not feature gate but value is null? possible?

upalatucci: how should i consider null? true or false? Or unset? I think would be unset

upalatucci: `Note: spec.enableCommonBootImageImport is true by default.` so should be true if its unset okay. Changing accordingly

metalice (approved):

PR #2555, opened 2025-04-02
CNV-56841: show VirtualMachine utilization summary

adamviktora:
## 📝 Description

Adds summary of CPU, Memory and Storage usage for the current namespace.

- also fixes a bug in Overview -> StorageUtil panel, which was counting too much storage because of duplicate mount points - it would get a sum of these:
Screenshot 2025-04-02 at 9 18 24

**Question:**
For memory capacity I used a sum of rounded values for each VM, not an actual metric. For storage capacity however, I did use the metric value.
I based this on the Utilization board in VM details:
Screenshot 2025-04-02 at 12 44 10

Should I rather use a rounded value for storage too? Or the metric for memory?

## 🎥 Demo
Before:
Screenshot 2025-04-02 at 12 52 02

After:

https://github.com/user-attachments/assets/d7cdf7c5-0ca3-426a-b3a8-39cb4a329e5f

### DIFF
--- a/src/views/virtualmachines/list/hooks/useVMTotalsMetrics.ts
+++ b/src/views/virtualmachines/list/hooks/useVMTotalsMetrics.ts
@@ -1,0 +1,69 @@
+import { useMemo } from 'react';
+
+import { V1VirtualMachine, V1VirtualMachineInstance } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import { ALL_NAMESPACES_SESSION_KEY } from '@kubevirt-utils/hooks/constants';
+import { getNamespace } from '@kubevirt-utils/resources/shared';
+import {
+  PrometheusEndpoint,
+  useActiveNamespace,
+  usePrometheusPoll,
+} from '@openshift-console/dynamic-plugin-sdk';
+import { METRICS } from '@overview/OverviewTab/metric-charts-card/utils/constants';
+
+import { getCpuText, getMemoryCapacityText, getMetricText } from '../utils/processVMTotalsMetrics';
+import { getVMTotalsQueries, VMTotalsQueries } from '../utils/totalsQueries';
+
+const useVMTotalsMetrics = (vms: V1VirtualMachine[], vmis: V1VirtualMachineInstance[]) => {
+  const [activeNamespace] = useActiveNamespace();
+  const allNamespace = activeNamespace === ALL_NAMESPACES_SESSION_KEY;
+  const currentTime = useMemo<number>(() => Date.now(), []);
+
+  const namespacesList = useMemo(() => [...new Set(vms.map((vm) => getNamespace(vm)))], [vms]);
+
+  const queries = useMemo(
+    () => getVMTotalsQueries(activeNamespace, namespacesList),
+    [activeNamespace, namespacesList],
+  );
+
+  const prometheusPollProps = {
+    endpoint: PrometheusEndpoint?.QUERY,
+    endTime: currentTime,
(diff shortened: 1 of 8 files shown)


src/views/virtualmachines/list/utils/totalsQueries.ts
```diff
@@ -0,0 +1,29 @@
...
+export const getVMTotalsQueries = (namespace: string, namespacesList: string[]) => {
+  let byNamespace = ' by (namespace)';
+  let namespaceFilter = `{namespace='${namespace}'}`;
+  let namespaceFilterCpu = `namespace='${namespace}'`;
+
+  if (namespace === ALL_NAMESPACES_SESSION_KEY) {
+    byNamespace = '';
+    namespaceFilter = '';
+    namespaceFilterCpu = `namespace=~'${namespacesList.join('|')}'`;
+  }
```
upalatucci, line 20: we don't like to use let. 

Can you do something like 

```suggestion
  const byNamespace = namespace !== ALL_NAMESPACES_SESSION_KEY ? ' by (namespace)' : '';
  const namespaceFilter = namespace !== ALL_NAMESPACES_SESSION_KEY ?`{namespace='${namespace}'}` : '';
  const namespaceFilterCpu =namespace !== ALL_NAMESPACES_SESSION_KEY ? `namespace='${namespace}'` :`namespace=~'${namespacesList.join('|')}'` ;
```

What do you think?

adamviktora: yeah, I usually don't like it either but here it seemed cleaner to me than using the same ternary 3 times

src/views/virtualmachines/list/hooks/useVMTotalsMetrics.ts
```diff
@@ -0,0 +1,112 @@
...
+      })
+      .reduce((acc, cur) => acc + cur, 0);
+
+    return getValueWithUnitText(bytes, METRICS.MEMORY);
+  };
+
+  const getMetricText = (response: PrometheusResponse, metric: string) => {
+    const bytes = getRawNumber(response);
+    return getValueWithUnitText(bytes, metric);
+  };
```
metalice, line 100: seems like all of those can be in a util file

src/views/virtualmachines/list/utils/totalsQueries.ts
```diff
@@ -0,0 +1,27 @@
...
+  TOTAL_STORAGE_CAPACITY: 'TOTAL_STORAGE_CAPACITY',
+  TOTAL_STORAGE_USAGE: 'TOTAL_STORAGE_USAGE',
+};
+
+export const getVMTotalsQueries = (namespace: string, namespacesList: string[]) => {
+  const isAllNamespaces = namespace === ALL_NAMESPACES_SESSION_KEY;
+
+  const byNamespace = isAllNamespaces ? '' : ' by (namespace)';
+  const namespaceFilter = isAllNamespaces ? '' : `{namespace='${namespace}'}`;
+  const namespaceFilterCpu = isAllNamespaces
```
metalice, line 16: please adjust the names ot be more clear,
for example
orderByNamespace or aggragteByNamespace..

adamviktora:
@metalice I pushed an update.

Also what do you think about this?

> 
> For memory capacity I used a sum of rounded values for each VM, not an actual metric. For storage capacity however, I did use the metric value.
> I based this on the Utilization board in VM details:
> image
> Should I rather use a rounded value for storage too? Or the metric for memory?

upalatucci (approved):

PR #2558, opened 2025-04-03
CNV-59302: use BulkSelect for VirtualMachine selection

adamviktora:
## 📝 Description

Replaces VM selection component with a [BulkSelect](https://www.patternfly.org/component-groups/controls/bulk-select) component from `@patternfly/react-component-groups` package

## 🎥 Demo

Before:

https://github.com/user-attachments/assets/37846c82-4b84-4b9e-aff5-7733c19b8b17

https://github.com/user-attachments/assets/b547a313-521f-4133-a78a-fa5b2985aa38

After:

https://github.com/user-attachments/assets/fc06ea70-2bd8-4056-a51b-07e69d737b49


upalatucci (approved):

cypress/views/selector-common.ts
```diff
@@ -61,8 +61,8 @@ export const bootsourceTH = 'th[data-label="Boot source"]';
 export const cpuTH = 'th[data-label="CPU | Memory"]';
 export const vmCount = '.pf-v6-c-menu-toggle__text > b';
 export const kebabBtn = '.pf-v6-c-table__td.dropdown-kebab-pf.pf-v6-c-table__action';
-export const selectDropdownToggle = 'button.virtual-machine-selection';
-export const selectAllDropdownOption = '*[data-test-id="select-all"]';
+export const selectDropdownToggle = '.virtual-machine-selection button';
```
metalice, line 64: if slready changing, why not adding an id or data test id ?

src/views/virtualmachines/list/components/VirtualMachineSelection/VirtualMachineSelection.tsx
```diff
@@ -24,68 +20,38 @@ type VirtualMachineSelectionProps = {
...
+  const isPageChecked = currentPageVMs.every((vm) => existingSelectedVMs.includes(vm));
+  const isPagePartiallyChecked =
+    !isPageChecked && currentPageVMs.some((vm) => existingSelectedVMs.includes(vm));
+
+  const handleBulkSelect = (value: BulkSelectValue) => {
+    value === BulkSelectValue.none && deselectAll();
+    value === BulkSelectValue.page && selectAll(currentPageVMs);
+    value === BulkSelectValue.all && selectAll(vms);
+    value === BulkSelectValue.nonePage && selectAll(vms.filter((vm) => !currentPageVMsSet.has(vm)));
+  };
```
metalice, line 41: can this be exported to a util file?

adamviktora: @metalice What is the advantage of that?

Or should I make it a general function that we can reuse in the future for other things not only VMs?

cypress/views/selector-common.ts
adamviktora: It is not possible to add props to the button now when using the BulkSelect component, but I updated it to `data-ouia-component-id="BulkSelect-toggle"`

src/views/virtualmachines/list/components/VirtualMachineSelection/VirtualMachineSelection.tsx
metalice: The advantage is that tsx files should be as much as possible only view files; any logic we can extract to custom hooks or utils file is preferred, keeping the TSX clean and lint as possible.

upalatucci (approved):

PR #2560, opened 2025-04-04
CNV-57918: add bulk actions

upalatucci:
## 📝 Description

Add edit labels, delete and move folders.

Edit labels will only show the labels that vms have in common with the same key and value. 
The edit labels see the common labels (key, value) and generate the patch for all vms comparing the initial common and the finale produced labels

## 🎥 Demo

https://github.com/user-attachments/assets/a903bf03-675d-4fb0-a845-55b89a66ba3d


adamviktora (commented):
Hey @upalatucci, this is looking good!

Few things I am noticing from user's perspective (some of them are for sure out of scope of this PR, so we can open followup issues):

1. 
Screenshot 2025-04-04 at 13 21 34

The Delete modal (and all other `ConfirmMultipleVMActionsModal`s) say "Are you sure you want to _stop_ ..." instead of pause / restart / stop and delete, not sure if it is intentional or not but it is a little confusing.

I'd also make the delete modal more in the "danger" style: warning title icon + danger button variant. Possibly even add the confirmation TextInput (considering that it can delete multiple VMs, I think it'd be good protection).


2. It looks like whenever a folder is empty, it gets deleted (it may be a feature, but maybe not)

3. Folder name can't be a substring of `Create folder ""`, so it is not possible to e.g. create a folder called `folder`

upalatucci:
hei @adamviktora  thanks for the review

1. i'll fix that
2. yes this is because folders are not entities but just a VM label. We group vms using this label. If there are no vms with that label, we can't simply know there was a folder with that name

upalatucci:
@adamviktora  FIXED

adamviktora:
Thanks for updating it @upalatucci! 

Can we also make the Delete modal in a warning style? We can use a component from `react-component-groups` for it: https://www.patternfly.org/component-groups/error-communication/warning-modal#warning-modal-with-a-checkbox, we can even use the confirm action to delete multiple VMs - either with the provided checkbox or with a text input (https://www.patternfly.org/components/modal/design-guidelines#confirm-a-destructive-action)

src/views/virtualmachines/actions/components/ConfirmMultipleVMActionsModal/ConfirmMultipleVMActionsModal.tsx
```diff
@@ -56,7 +56,7 @@ const ConfirmMultipleVMActionsModal: FC<ConfirmMultipleVMActionsModalProps> = ({
         title={t('{{actionType}} {{numVMs}} VirtualMachines?', { actionType, numVMs })}
       />
       <ModalBody>
-        {t('Are you sure you want to stop ')}
+        {t('Are you sure you want to {{actionType}} ', { actionType: actionType.toLowerCase() })}
```
metalice, line 59: i dont see action type also translated, so we will end up with a sentence in split language

src/views/virtualmachines/actions/BulkVirtualMachineActionFactory.tsx
```diff
@@ -0,0 +1,166 @@
...
+                  data: getLabelsDiffPatch(labels, getLabels(vm)),
+                  model: VirtualMachineModel,
+                  resource: vm,
+                }),
+              ),
+            );
+          }}
+          initialLabels={commonLabels}
+          isOpen={isOpen}
+          obj={vms[0]}
```
metalice, line 52: its guranteed? or need ?.

upalatucci: I can add it anyway. Should be guaranteed

src/views/virtualmachines/actions/components/ConfirmMultipleVMActionsModal/ConfirmMultipleVMActionsModal.tsx
upalatucci: splitting the phrases?

PR #2561, opened 2025-04-04
CNV-59430: fix menu being cut off by tree view panel

adamviktora:
## 📝 Description

Fixes a bug of menu being cut off by the VM tree view panel.

This fix attaches the Menu div to the body element so it can appear over the edge of the panel.

## 🎥 Demo

Before:

Screenshot 2025-04-04 at 12 12 57

After:
Screenshot 2025-04-04 at 12 02 08


src/views/virtualmachines/tree/VirtualMachineTreeView.scss
```diff
@@ -63,6 +63,11 @@
   }
 }
 
+.right-click-action-menu--vm {
+  min-width: 0 !important;
```
adamviktora, line 67: This `!important` is necessary, because Popper adds a min-width to inline style and the component is made in PatternFly in a way, that you can't override them with a `style` property on the Popper. So `!important` ended up being the only option.

upalatucci (approved):

src/views/virtualmachines/tree/components/TreeViewRightClickActionMenu/VMRightClickActionMenu.tsx
```diff
@@ -41,9 +41,9 @@ const VMRightClickActionMenu: FC<VMRightClickActionMenuProps> = ({ hideMenu, tri
           </MenuContent>
         </Menu>
       }
-      appendTo={triggerElement}
+      distance={-6}
```
metalice, line 44: what is this magic number?

adamviktora: As I say above:

> I also added distance={-6} to keep the previous look where the menu appears a little on top of the VM tree view item. But if that was unintentional before, we can remove it.

It will end up being moved 6px to the top direction

src/views/virtualmachines/tree/components/TreeViewRightClickActionMenu/VMRightClickActionMenu.tsx
```diff
@@ -41,9 +44,9 @@ const VMRightClickActionMenu: FC<VMRightClickActionMenuProps> = ({ hideMenu, tri
           </MenuContent>
         </Menu>
       }
-      appendTo={triggerElement}
+      distance={menuDistance}
```
upalatucci, line 47: Make it a constant all in uppercase in a constant.ts file

upalatucci (approved):

PR #2562, opened 2025-04-04
CNV-58425: don't disable switch on click

adamviktora:
## 📝 Description

Doesn't disable switch while waiting for the option to change.

I am not sure if this fix doesn't break something, and whether it should rather stay disabled while loading. We could make it disabled, but not apply the disabled styles.

## 🎥 Demo

Before:

https://github.com/user-attachments/assets/8ff4d98d-ca36-40d0-b3b6-9b37d9ef492c

After:

https://github.com/user-attachments/assets/edac3da3-e58f-49dc-ba97-2f855b7e6c76

### DIFF
--- a/src/utils/styles/cursor.scss
+++ b/src/utils/styles/cursor.scss
@@ -1,0 +1,5 @@
+.kv-cursor--loading {
+  :hover {
+    cursor: wait;
+  }
+}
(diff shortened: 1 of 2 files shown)


src/views/clusteroverview/SettingsTab/ClusterTab/components/PersistentReservationSection/persistent-reservation-section.scss
```diff
@@ -0,0 +1,5 @@
+.persistent-reservation-section--loading {
+  :hover {
+    cursor: wait;
+  }
+}
```
upalatucci, line 5: can we do a kubevirt global class for it? so if we want to have other switches with this loading we don't have to create other classes ?

PR #2565, opened 2025-04-08
CNV-59566: SelectTypeahead and FolderSelect enhancements

adamviktora:
## 📝 Description

Fixes bugs and improves UX of `SelectTypeahead` and related `FolderSelect`, which uses it.

Bug fixes:
- fixes a bug, where you can't create a folder with name that is a substring of `Create folder ""`
- fixes a bug, where a folder with an invalid name could have been created

UX improvements
- menu doesn't close on clicking the input (useful e.g. when user is selecting word with a double-click)
- "create option" is not shown, if an option with that name can't be created
- invalid characters are detected immediately
- in `MoveToFolderModal`, it is visually distinguished if user chooses a folder, or just types something in and clicks outside. And it is not possible to save if nothing is selected. 
    - This behavior is not improved in other occasions of `FolderSelect` though, so maybe we could reset the input text to an empty string on clicking outside if nothing is selected?
- focus stays on the input when choosing options with arrow up/down

## 🎥 Demo

BEFORE:
Overall behavior:

https://github.com/user-attachments/assets/9499c5eb-c8fd-4780-8ff5-6ef3991f0a01

Folder is shown with invalid name in the tree-view bug:

https://github.com/user-attachments/assets/ec49f0ef-2d16-4cd3-83d7-5ef58d5ebed3

Clicking outside means no folder is created:

https://github.com/user-attachments/assets/aef2bbba-ccef-4453-b617-93cc9c62b0c4

Create option is shown, although such option exists:

https://github.com/user-attachments/assets/acebcdb4-3543-4d42-a132-20d74f602a40

AFTER:
Overall behavior:

https://github.com/user-attachments/assets/b5156095-daa3-4d96-8ab6-1bd01913820f

Clicking outside vs choosing a folder:

https://github.com/user-attachments/assets/8dbf9e59-0fe1-4dfe-b2c0-0f8f70de9b7a


src/utils/components/FolderSelect/utils/validation.tsx
```diff
@@ -0,0 +1,51 @@
...
+    if (invalidCharacter.trim() === '') {
+      return <>{t('Spaces are not allowed')}</>;
+    }
+
+    return (
+      <>
+        {t('Invalid character')}: {invalidCharacter}
+      </>
+    );
+  }
```
upalatucci, line 26: don't like the nested `if`s. Can we do 
```suggestion
  if (invalidCharacter) {
      return   <>{invalidCharacter.trim() === '' ? t('Spaces are not allowed') :  {t('Invalid character')}: {invalidCharacter}}</> : ;
    }
  }
```

or something like that

upalatucci (approved):

PR #2567, opened 2025-04-09
CNV-58549: use nad instead of udn to detect the udn in the namespace

upalatucci:
## 📝 Description

Cluster UDN cannot be fetch if the user is not an admin.

to detect if the network is under a UDN, fetch the nads and search for a primary nad  with layer 2 or 3


src/utils/resources/udn/constants.ts
```diff
@@ -0,0 +1,6 @@
+export const LAYER2_TOPOLOGY = 'layer2';
+export const LAYER3_TOPOLOGY = 'layer3';
+export const LOCALNET_TOPOLOGY = 'localnet';
+
+export const PrimaryTopologys = [LAYER2_TOPOLOGY, LAYER3_TOPOLOGY];
+export const SecondaryTopologys = [LOCALNET_TOPOLOGY];
```
adamviktora, line 6: nit: change spelling to `Topologies`

adamviktora (commented):
It is still possible to create a VM from Catalog -> InstanceTypes. Shouldn't be the create button also disabled there?

Screenshot 2025-04-09 at 11 29 17

upalatucci:
Fixed 
Screenshot 2025-04-10 at 11 00 49

adamviktora (approved):

PR #2571, opened 2025-04-10
CNV-56943: Remove pending changes and live reload alerts in add/edit NIC modal

rszwajko:
## 📝 Description

  
Remove logic for detecting required restart and rely on the information
received from the backend. If the operation changes non live-updatable
network field then RestartRequired condition is set on the VM.

Reference-Url: https://github.com/kubevirt/kubevirt/pull/14602

### Before (pending changes alert triggered via code)
image

### After
Screenshot From 2025-04-10 13-43-14


rszwajko:
@gouyang 
I've removed the live reload alert as requested in https://issues.redhat.com/browse/CNV-56943.
In the code I see there is another similar alert about pending changes so I've also removed it (both are visible on the screenshot above).

@pcbailey 
Both alerts seem integral part of the feature - can we remove them safely?

metalice:
@rszwajko what is the status here?

rszwajko:
@pcbailey 
>Has full NIC hot-plug been implemented? If yes, then it should be fine to remove it, but if there are still NIC types that aren't immediately hot-plugged after being added then we would still need to display the alert for those NICs.
>

The hotplug story does not cover all the cases (and probably won't cover in the near future).
However the main point of this change (as I understand it) is to rely on the server to provide information if restart is required. There have been recent improvements in this area - i.e. https://github.com/kubevirt/kubevirt/pull/14602
The problem with detecting required restart in the UI is that the backend often backports networking features which makes our checks incorrect.

PR #2574, opened 2025-04-10
CNV-59558: CNV-59559: fix disk modal CapacityInput

adamviktora:
## 📝 Description

- fixes a bug, where clicking ` + ` in the Number input resulted in resetting the number to the initial value ([CNV-59559](https://issues.redhat.com//browse/CNV-59559))
   - it was because strings were compared instead of numbers
- another bug which is fixed was that user can't manually type in a smaller number ([CNV-59558](https://issues.redhat.com/browse/CNV-59558)) e.g. if the capacity is 4 GiB and you wanna set it to 12 GiB, when you type in "1" it results in the reset back to 4 GiB
  - also you couldn't change the unit to e.g. MiB, you have to write a bigger value first and then change it to MiB (e.g. if you wanna move from 4 GiB to 5.5 GiB you have to use MiB, because it doesn't support decimal values)
  
- there is a bug that the size of a created disk is larger than what was specified in the form:

https://github.com/user-attachments/assets/7a198896-7b2f-462c-b42e-65ea48a57fd2

## 🎥 Demo

BEFORE:
Jira bug [CNV-59559](https://issues.redhat.com//browse/CNV-59559):

https://github.com/user-attachments/assets/727f04d6-4ad8-42b4-a160-8a413fef1996

Jira bug [CNV-59558](https://issues.redhat.com//browse/CNV-59558):

https://github.com/user-attachments/assets/ff903a13-ae50-4e49-9996-1243de1928d2

Bug where you can set a smaller size without warning:

https://github.com/user-attachments/assets/34e52567-fe7a-4dde-b879-dffe6d6ebb5f

AFTER:
- one thing I don't like is having the default value lower than the current = minimum value. We could ceil it up, but than the value is changed by default (and user might overlook it). Or we could use the exact value (with a decimal point) - but decimal points are not supported in this NumberInput - you can't manually type it in.

WDYT? What should be the default value? In the current "before" version the value is just the whole number without the numbers after decimal point. And the form value is set to null, so nothing gets updated if you keep it like that - which is quite good, but it doesn't really reflect the actual size.

https://github.com/user-attachments/assets/3a61c9ac-3f4d-4c7b-9ceb-e9c22998231a

### DIFF
--- a/src/utils/components/CapacityInput/CapacityInput.tsx
+++ b/src/utils/components/CapacityInput/CapacityInput.tsx
@@ -14,12 +14,19 @@ import { SelectOption } from '@patternfly/react-core';
 import FormGroupHelperText from '../FormGroupHelperText/FormGroupHelperText';
 import FormPFSelect from '../FormPFSelect/FormPFSelect';
 
-import { CAPACITY_UNITS, removeByteSuffix } from './utils';
+import {
+  CAPACITY_UNITS,
+  getErrorValue,
+  getUnitFromSize,
+  getValueFromSize,
+  removeByteSuffix,
+} from './utils';
 
 type CapacityInputProps = {
   isEditingCreatedDisk?: boolean;
   isMinusDisabled?: boolean;
   label?: string;
+  minValue?: number;
   onChange: (quantity: string) => void;
   size: string;
 };
@@ -28,23 +35,24 @@ const CapacityInput: FC<CapacityInputProps> = ({
   isEditingCreatedDisk,
   isMinusDisabled,
   label,
+  minValue,
   onChange,
   size,
 }) => {
   const { t } = useKubevirtTranslation();
-  const [unitValue = ''] = size?.match(/[a-zA-Z]+/g) || [];
-  const [sizeValue = 0] = size?.match(/[0-9]+/g) || [];
-  const unit = !unitValue?.endsWith('B') ? `${unitValue}B` : unitValue;
-  const value = Number(sizeValue);
+  const unit = getUnitFromSize(size);
+  const value = getValueFromSize(size);
 
   const onFormatChange = (_, newUnit: CAPACITY_UNITS) => {
     onChange(`${Number(value)}${removeByteSuffix(newUnit)}`);
   };
   const unitOptions = Object.values(CAPACITY_UNITS);
-  if (!unitOptions?.includes(unit as CAPACITY_UNITS)) unitOptions.push(unit as CAPACITY_UNITS);
+  if (!unitOptions?.includes(unit)) unitOptions.push(unit);
 
   const validated: ValidatedOptions =
-    !value || value <= 0 ? ValidatedOptions.error : ValidatedOptions.default;
+    !value || value <= 0 || (minValue && value < minValue)
+      ? ValidatedOptions.error
+      : ValidatedOptions.default;
 
   return (
     <FormGroup
@@ -91,10 +99,12 @@ const CapacityInput: FC<CapacityInputProps> = ({
         </SplitItem>
       </Split>
       <FormGroupHelperText validated={validated}>
-        {validated === ValidatedOptions.error &&
-          t('Size cannot be {{errorValue}}', {
-            errorValue: value < 0 ? 'negative' : 'zero',
-          })}
+        {validated === ValidatedOptions.error && (
+          <>
+            {t('Size cannot be')} {getErrorValue(value)}
+            {value > 0 && ` ${minValue} ${unit}`}
+          </>
+        )}
       </FormGroupHelperText>
     </FormGroup>
   );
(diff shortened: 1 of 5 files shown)


src/utils/components/CapacityInput/CapacityInput.tsx
```diff
@@ -28,23 +29,29 @@ const CapacityInput: FC<CapacityInputProps> = ({
...
-  if (!unitOptions?.includes(unit as CAPACITY_UNITS)) unitOptions.push(unit as CAPACITY_UNITS);
+  if (!unitOptions?.includes(unit)) unitOptions.push(unit);
 
   const validated: ValidatedOptions =
-    !value || value <= 0 ? ValidatedOptions.error : ValidatedOptions.default;
+    !value || value <= 0 || value < minValue ? ValidatedOptions.error : ValidatedOptions.default;
+
+  const getErrorValue = () => {
+    if (value > 0) {
+      return `less than ${minValue} ${unit}`;
```
upalatucci, line 51: this will not be translated. You can split the translations

src/utils/components/CapacityInput/CapacityInput.tsx
```diff
@@ -91,10 +98,12 @@ const CapacityInput: FC<CapacityInputProps> = ({
         </SplitItem>
       </Split>
       <FormGroupHelperText validated={validated}>
-        {validated === ValidatedOptions.error &&
-          t('Size cannot be {{errorValue}}', {
-            errorValue: value < 0 ? 'negative' : 'zero',
-          })}
+        {validated === ValidatedOptions.error && (
+          <>
+            {t('Size cannot be {{errorValue}}', { errorValue: getErrorValue() })}
```
upalatucci, line 103: No. I think you have to translate it like this
```suggestion
            {<>t('Size cannot be') {' '} {getErrorValue()}</>)}
```

and then 

```
    if (value > 0) {
      return t('less than');
    }
    return value < 0 ? t('negative') : t('zero');
```

To detect those translations.


Please if `yarn i18n` is not able to detect those translations, add 
```

      // t('less than')
    // t('negative')
    // t('zero')
    ```
    Somewhere so that it will pick it up and it will add them in the translation JSON file

File: src/utils/components/CapacityInput/CapacityInput.tsx [hunk 3]
```diff
@@ -28,23 +29,29 @@ const CapacityInput: FC<CapacityInputProps> = ({
...
   const onFormatChange = (_, newUnit: CAPACITY_UNITS) => {
     onChange(`${Number(value)}${removeByteSuffix(newUnit)}`);
   };
   const unitOptions = Object.values(CAPACITY_UNITS);
-  if (!unitOptions?.includes(unit as CAPACITY_UNITS)) unitOptions.push(unit as CAPACITY_UNITS);
+  if (!unitOptions?.includes(unit)) unitOptions.push(unit);
 
   const validated: ValidatedOptions =
-    !value || value <= 0 ? ValidatedOptions.error : ValidatedOptions.default;
+    !value || value <= 0 || value < minValue ? ValidatedOptions.error : ValidatedOptions.default;
```
metalice, line 47: minValue is optional, value < undefined?

src/utils/components/CapacityInput/CapacityInput.tsx
```diff
@@ -28,23 +29,29 @@ const CapacityInput: FC<CapacityInputProps> = ({
...
+
+  const getErrorValue = () => {
+    if (value > 0) {
```
metalice, line 50: i think u can use !isEmpty here

src/utils/components/CapacityInput/CapacityInput.tsx
```diff
@@ -91,10 +98,12 @@ const CapacityInput: FC<CapacityInputProps> = ({
...
       </Split>
       <FormGroupHelperText validated={validated}>
-        {validated === ValidatedOptions.error &&
-          t('Size cannot be {{errorValue}}', {
-            errorValue: value < 0 ? 'negative' : 'zero',
-          })}
+        {validated === ValidatedOptions.error && (
+          <>
+            {t('Size cannot be')} {getErrorValue()}
+            {value > 0 && ` ${minValue} ${unit}`}
```
metalice, line 104: i think u can use !isEmpty here

src/utils/components/DiskModal/components/DiskSizeInput/useDefaultSizes.ts
```diff
@@ -0,0 +1,23 @@
+import { useState } from 'react';
+
+import { CAPACITY_UNITS } from '@kubevirt-utils/components/CapacityInput/utils';
+import { convertToBaseValue, humanizeBinaryBytes } from '@kubevirt-utils/utils/humanize.js';
+
+const useDefaultSizes = (pvcSize: string) => {
```
metalice, line 6: usePVCDefaultSize

metalice (approved):

PR #2582, opened 2025-04-15
CNV-59218: fix migration ui from feedback

upalatucci:
## 📝 Description

Feedback from https://docs.google.com/document/d/1fck7ReNtHU6zUJmpjMtSIwNip8tD1n5lg9iPKwd2CHc/edit?tab=t.0

- Show what SC the VM is currently on for when users select to migrate selected volumes.
- In step 2 at the SC dropdown when users need to select the destination SC, add “Current” in brackets
- CANNOT BE DONE: Incompatible SCs should be displayed as disabled
- In step 3 (Review) add “Source SC”
- Change "View report" to "View log" link

## 🎥 Demo

Screenshot 2025-04-15 at 15 17 56
Screenshot 2025-04-15 at 15 18 01
Screenshot 2025-04-15 at 15 19 45
Screenshot 2025-04-15 at 15 23 41


locales/en/plugin__kubevirt-plugin.json
```diff
@@ -55,6 +56,7 @@
...
   "<0>Open an SSH connection with the VM using the cluster API server. You must be able to access the API server and have virtctl command line tool installed.</0><br/><2>For more details, see <2>Installing virtctl</2> in Getting started with OpenShift Virtualization.</2>": "<0>Open an SSH connection with the VM using the cluster API server. You must be able to access the API server and have virtctl command line tool installed.</0><br/><2>For more details, see <2>Installing virtctl</2> in Getting started with OpenShift Virtualization.</2>",
   "<0>Select the storage to migrate for <2>{{vmsCount}} VirtualMachines with {{volumesCount}} Volumes</2> from the source (current) storage class {storageClasses?.join(', ')}</0>": "<0>Select the storage to migrate for <2>{{vmsCount}} VirtualMachines with {{volumesCount}} Volumes</2> from the source (current) storage class {storageClasses?.join(', ')}</0>",
+  "<0>Select the storage to migrate for <2></2>from the source (current) storage class {storageClasses?.join(', ')}</0>": "<0>Select the storage to migrate for <2></2>from the source (current) storage class {storageClasses?.join(', ')}</0>",
```
rszwajko, line 59: source code inside translation

src/views/virtualmachines/actions/components/VirtualMachineMigration/tabs/VirtualMachineMigrationDestinationTab.tsx
```diff
@@ -41,7 +43,8 @@ const VirtualMachineMigrationDestinationTab: FC<VirtualMachineMigrationDestinati
                   linkTo={false}
                   name={storageClass}
                 />
-                {defaultStorageClassName === storageClass ? t('(default)') : ''}
+                {vmStorageClassNames.includes(storageClass) && t('(current)')}{' '}
```
rszwajko, line 46: can we use labels here?

src/views/virtualmachines/actions/components/VirtualMachineMigration/tabs/VirtualMachineMigrationReviewTab.tsx
```diff
@@ -58,6 +60,12 @@ const VirtualMachineMigrationReviewTab: FC<VirtualMachineMigrationReviewTabProps
               </Td>
               <Td>{t('VirtualMachine storage')}</Td>
             </Tr>
+            <Tr>
+              <Td width={30}>
+                <strong>{t('Source StorageClasses')}</strong>
+              </Td>
+              <Td>{vmStorageClassNames.join(', ')}</Td>
```
rszwajko, line 67: comma-delimited list is consistent with what we have here already but I would look for different solutions. How about using plain text if there is one item and an unordered list when there are more ?

rszwajko (commented):
> CANNOT BE DONE: Incompatible SCs should be displayed as disabled
>
Depends what do we mean by incompatible - we do have information from StorageProfile and can check if target has the same Voume/Access Modes
For reference check: #2524

src/views/virtualmachines/actions/components/VirtualMachineMigration/tabs/VirtualMachineMigrationReviewTab.tsx
upalatucci: there is no space for unordered list. But in general the vm should have max 1 or 2 storageclasses so there is no issue with that. maybe i can discuss this with yifat and create an enhancement in a different pr

PR #2585, opened 2025-04-16
CNV-58541: use only MTC migrations

upalatucci:
## 📝 Description

The storage migration will be available only if MTC is installed

The old method using `updateVolumesStrategy` is removed

This means that all storageclass can be used and all pvcs can be migrated

the migplan do not support `rollback` and `migrateState` specs together. only one of the two. So rollabcking means to remove the `migrateState` and put the `rollback` state

## 🎥 Demo

IF MTC IS NOT INSTALLED

Screenshot 2025-05-05 at 13 03 42
Screenshot 2025-05-05 at 13 03 51

MTC migration

Screenshot 2025-05-05 at 12 53 58

This should also resolve https://issues.redhat.com/browse/CNV-61006


src/views/virtualmachines/actions/hooks/utils.ts
```diff
@@ -0,0 +1,4 @@
+import { MigPlan } from '@kubevirt-utils/resources/migrations/constants';
+
+export const sortMigPlansByCreationTimestamp = (a: MigPlan, b: MigPlan): number =>
+  a?.metadata?.creationTimestamp?.localeCompare(b?.metadata?.creationTimestamp);
```
rszwajko, line 4: the comparator is not symetric  - it should treat the falsy values the same on both sides. Currently it's not the case:
```
["a","","b",undefined,"c", null].sort((a,b)=> a?.localeCompare(b))
Array(6) [ "", "a", "b", "c", null, undefined ]
```

src/views/virtualmachines/actions/hooks/useStorageMigrations.ts
```diff
@@ -0,0 +1,61 @@
...
+          namespace: DEFAULT_MIGRATION_NAMESPACE,
+        }
+      : null,
+  );
+
+  const latestVMMigPlan = migPlans
+    ?.filter((migPlan) =>
+      migPlan?.spec?.persistentVolumes?.find(
+        (pv) =>
+          pv?.selection?.action === 'copy' && vmPVCNames?.includes(pv?.pvc?.name?.split(':')?.[0]),
```
rszwajko, line 46: 1. a candidate for a named utility method
2. please provide an example for the parsed string (as a comment)

src/views/virtualmachines/actions/VirtualMachineActionFactory.tsx
```diff
@@ -76,23 +80,23 @@ export const VirtualMachineActionFactory = {
...
-      cta: () => {
-        cancelMigration(vmim);
-        return rollbackStorageMigration(vm);
-      },
-      disabled: isSingleNodeCluster,
+      cta: () =>
+        k8sPatch({
+          data: [
+            { op: 'replace', path: '/spec/rollback', value: true },
+            { op: 'remove', path: '/spec/migrateState' },
```
rszwajko, line 95: removing migration state seems unusual - why it's up to the UI to clean it up?

src/views/virtualmachines/actions/actions.ts
```diff
@@ -116,22 +110,3 @@ export const deleteVM = async (vm: V1VirtualMachine) => {
...
 };
-
-export const rollbackStorageMigration = async (vm: V1VirtualMachine) => {
-  if (isEmpty(getUpdateStrategy(vm))) return;
-
-  const patchData: Patch[] = createRollbackPatchData(vm);
-
-  patchData.push({
-    op: 'remove',
-    path: '/spec/updateVolumesStrategy',
```
rszwajko, line 127: it seems we are changing the rollback procedure - is it a new change in 4.19?

src/views/virtualmachines/actions/components/VirtualMachineMigration/hooks/useMigratableStorageClasses.ts
```diff
@@ -1,26 +0,0 @@
-import { V1beta1StorageSpecAccessModesEnum } from '@kubevirt-ui/kubevirt-api/kubevirt';
-import { StorageProfileModelGroupVersionKind } from '@kubevirt-utils/models';
-import { getName } from '@kubevirt-utils/resources/shared';
-import { K8sResourceCommon, useK8sWatchResource } from '@openshift-console/dynamic-plugin-sdk';
-
-const useMigratableStorageClasses = (): string[] => {
```
rszwajko, line 6: removal seems not related to the PR

rszwajko (changes requested):

src/views/virtualmachines/actions/VirtualMachineActionFactory.tsx
upalatucci: the migplan do not support rollback and migrateState specs together. only one of the two

src/views/virtualmachines/actions/actions.ts
upalatucci: Yes we need that as we don't want to use `updateVolumesStrategy` anymore. Its a storage migration method that do not work for all cases

src/views/virtualmachines/actions/components/VirtualMachineMigration/hooks/useMigratableStorageClasses.ts
upalatucci: Yes its related becuase MTC migration support all storage classes  instead of `updateVolumesStrategy`

src/views/virtualmachines/actions/hooks/utils.ts
upalatucci: oh i understand but the creationTimestamp will not be `""`. Maybe `a` or `b` are undefined so I guarded that with `?`

src/views/virtualmachines/actions/hooks/useStorageMigrations.ts
upalatucci: okay i'll do it

rszwajko (approved):
Looks good enough! We can refine in follow-ups if needed.

PR #2589, opened 2025-04-18
CNV-59371: Show metrics on running vm

upalatucci:
## 📝 Description

Hide metrics on non-running vm

(The metrics are shown as there is a delay in Prometheus  )

## 🎥 Demo

https://github.com/user-attachments/assets/20053c47-4cc3-4bdb-91de-42ad5ceae1f5

### DIFF
--- a/src/views/virtualmachines/list/components/VirtualMachineRow/VirtualMachineRowLayout.tsx
+++ b/src/views/virtualmachines/list/components/VirtualMachineRow/VirtualMachineRowLayout.tsx
@@ -86,13 +86,13 @@ const VirtualMachineRowLayout: FC<
         {ips}
       </TableData>
       <TableData activeColumnIDs={activeColumnIDs} className="vm-column" id="memory-usage">
-        <MemoryPercentage vmiMemory={vmiMemory} vmName={vmName} vmNamespace={vmNamespace} />
+        <MemoryPercentage vm={obj} vmiMemory={vmiMemory} />
       </TableData>
       <TableData activeColumnIDs={activeColumnIDs} className="vm-column" id="cpu-usage">
-        <CPUPercentage vmName={vmName} vmNamespace={vmNamespace} />
+        <CPUPercentage vm={obj} />
       </TableData>
       <TableData activeColumnIDs={activeColumnIDs} className="vm-column" id="network-usage">
-        <NetworkUsage vmName={vmName} vmNamespace={vmNamespace} />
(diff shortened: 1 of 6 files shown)


upalatucci:
Yes i can but you'll conitnue to see the metrics after the VM is stopped for another 30s @metalice

src/views/virtualmachines/list/components/VirtualMachineRow/VirtualMachineRowLayout.tsx
```diff
@@ -86,13 +89,15 @@ const VirtualMachineRowLayout: FC<
         {ips}
       </TableData>
       <TableData activeColumnIDs={activeColumnIDs} className="vm-column" id="memory-usage">
-        <MemoryPercentage vmiMemory={vmiMemory} vmName={vmName} vmNamespace={vmNamespace} />
+        {vmRunning && (
```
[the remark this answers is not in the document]
upalatucci: i would pass the vm at this point and extract name, namespaces and running state in the component. WDYT?

src/views/virtualmachines/list/hooks/constants.ts
```diff
@@ -16,8 +16,8 @@ export const getVMListQueries = (namespace: string) => {
 
   return {
     [VMListQueries.CPU_REQUESTED]: `kube_pod_resource_request{resource='cpu',${namespaceFilter}}`,
-    [VMListQueries.CPU_USAGE]: `rate(kubevirt_vmi_cpu_usage_seconds_total{${namespaceFilter}}[5m])`,
+    [VMListQueries.CPU_USAGE]: `rate(kubevirt_vmi_cpu_usage_seconds_total{${namespaceFilter}}[30s])`,
```
rszwajko, line 19: IMHO changing the window to 30s impacts functionality - we should double check if that's OK

metalice: I think 30s is right as this is the minimum promethus is getting updates , and we are showing it on the VM list, without any way to change values from the UI, so it will always show the latest updated value which is what we want , the only concern is the amount of rerender we will get.. @upalatucci is it safe for performance?

upalatucci: yes we are not changing the number of re-rendering and the pooling timeout. Just the data that we fetch. The pooling timeout will remain the same: `15s` ( the default one)

rszwajko (approved):

PR #2592, opened 2025-04-18
Make start-console script compatible with zsh

upalatucci:
## 📝 Description

```
declare -A plugins=(["monitoring-plugin"]="https://github.com/openshift/monitoring-plugin.git", ["networking-console-plugin"]="https://github.com/openshift/networking-console-plugin.git")
declare -A runningPlugins=(["podman-linux"]="kubevirt-plugin=http://localhost:9001", ["podman"]="kubevirt-plugin=http://host.containers.internal:9001", ["docker"]="kubevirt-plugin=http://host.docker.internal:9001")
```

is not compatible with zsh. Me and others use that

### DIFF
--- a/start-console.sh
+++ b/start-console.sh
@@ -1,50 +1,74 @@
-#!/usr/bin/env bash
-#Please run this script under bash 4+
-set -euo pipefail
+#!/bin/sh
+set -eu
 
-#Cloning, pulling and running other plugins starting from port 9002
-# to add more plugin simply add more properites to dic. [name-of-plugin]={git-repo-url}
-declare -A plugins=(["monitoring-plugin"]="https://github.com/openshift/monitoring-plugin.git", ["networking-console-plugin"]="https://github.com/openshift/networking-console-plugin.git")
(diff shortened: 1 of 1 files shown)


start-console.sh
```diff
@@ -1,49 +1,68 @@
-#!/usr/bin/env bash
-#Please run this script under bash 4+
-set -euo pipefail
+#!/bin/sh
```
metalice, line 1: this is not zsh

metalice:
not working on my mac
image

first run is with #!/bin/zsh
second run is #!/bin/sh

start-console.sh
```diff
@@ -1,50 +1,74 @@
...
-    if [ "$(lsof -t -i:$INITIAL_PORT)" != "" ]; then
-        kill -9 $(lsof -t -i:$INITIAL_PORT)
-    fi
 
-    if [ $arg = "monitoring-plugin" ]; then
+    pid="$(lsof -t -i:$INITIAL_PORT 2>/dev/null || true)"
+    [ -n "$pid" ] && kill -9 "$pid"
+
+
+    [ "$arg" = "monitoring-plugin" ] then
```
pcbailey, line 49: This line needs to be changed back to:

`if [ "$arg" = "monitoring-plugin" ]; then`

PR #2594, opened 2025-04-22
CNV-59218: disable non migratable storageclasses

upalatucci:
## 📝 Description

Disable storageclasses that do not support RWX. RWX is mandatory for the migration to succeed 

## 🎥 Demo

Screenshot 2025-04-22 at 12 01 47

### DIFF
--- a/src/views/virtualmachines/actions/components/VirtualMachineMigration/hooks/useMigratableStorageClasses.ts
+++ b/src/views/virtualmachines/actions/components/VirtualMachineMigration/hooks/useMigratableStorageClasses.ts
@@ -1,0 +1,26 @@
+import { V1beta1StorageSpecAccessModesEnum } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import { StorageProfileModelGroupVersionKind } from '@kubevirt-utils/models';
+import { getName } from '@kubevirt-utils/resources/shared';
+import { K8sResourceCommon, useK8sWatchResource } from '@openshift-console/dynamic-plugin-sdk';
+
+const useMigratableStorageClasses = (): string[] => {
+  const [storageProfiles] = useK8sWatchResource<K8sResourceCommon[]>({
+    groupVersionKind: StorageProfileModelGroupVersionKind,
+    isList: true,
+  });
+
+  const allowedStorageProfilesNames = storageProfiles?.reduce((acc, storageProfile: any) => {
+    if (
+      storageProfile?.status?.claimPropertySets?.some((claimProperty) =>
+        claimProperty?.accessModes?.includes(V1beta1StorageSpecAccessModesEnum.ReadWriteMany),
+      )
+    )
+      acc.push(getName(storageProfile));
+
+    return acc;
+  }, []);
+
+  return allowedStorageProfilesNames;
+};
+
+export default useMigratableStorageClasses;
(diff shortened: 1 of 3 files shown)


src/views/virtualmachines/actions/components/VirtualMachineMigration/hooks/useMigratableStorageClasses.ts
```diff
@@ -0,0 +1,23 @@
...
+    isList: true,
+  });
+
+  const allowedStorageProfilesNames = storageProfiles
+    ?.filter((storageProfile) =>
+      (storageProfile as any)?.status?.claimPropertySets?.some((claimProperty) =>
+        claimProperty?.accessModes?.includes(V1beta1StorageSpecAccessModesEnum.ReadWriteMany),
+      ),
+    )
+    ?.map((storageProfile) => getName(storageProfile));
```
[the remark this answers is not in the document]
upalatucci: no we don't have it . i don't know why but we can find a way later for that

metalice (approved):

src/views/virtualmachines/actions/components/VirtualMachineMigration/tabs/VirtualMachineMigrationDestinationTab.tsx
```diff
@@ -44,10 +50,11 @@ const VirtualMachineMigrationDestinationTab: FC<VirtualMachineMigrationDestinati
                 {defaultStorageClassName === storageClass ? t('(default)') : ''}
               </>
             ),
+            isDisabled: !allowedStorageProfilesNames?.includes(storageClass),
             value: storageClass,
           }))}
           popperProps={{
-            appendTo: () => document.getElementById('virtual-machine-migration-modal'),
+            appendTo: () => document.getElementById(POPPER_CONTAINER_ID),
```
rszwajko, line 57: 1. why we are using a constant ID here instead of the default (`document.body`) ?
2. who creates/owns that element?

src/views/virtualmachines/actions/components/VirtualMachineMigration/hooks/useMigratableStorageClasses.ts
```diff
@@ -0,0 +1,26 @@
...
+import { getName } from '@kubevirt-utils/resources/shared';
+import { K8sResourceCommon, useK8sWatchResource } from '@openshift-console/dynamic-plugin-sdk';
+
+const useMigratableStorageClasses = (): string[] => {
+  const [storageProfiles] = useK8sWatchResource<K8sResourceCommon[]>({
+    groupVersionKind: StorageProfileModelGroupVersionKind,
+    isList: true,
+  });
+
+  const allowedStorageProfilesNames = storageProfiles?.reduce((acc, storageProfile: any) => {
```
rszwajko, line 12: In the #2524 I've assumed that volume/access mode from storage profile are only recommendations (user is allowed to overwrite). Perhaps we should inform the user that he can unblock the storage class by changing configuration in the storage profile or add a way to by-pass  storage profile?

upalatucci: Currently we don't let the user customize anything during the storage migration so it will use the default settings.

PR #2603, opened 2025-04-23
CNV-57378: Show completed QuickStarts in Getting Started card

pcbailey:
## 📝 Description

This PR fixes a bug where the Quick Starts section of the Getting Started card disappears after completing the "Create a virtual machine from a volume" QuickStart. The card disappeared because that's the only QuickStart and QuickStarts were being hidden once completed.

To fix the problem I changed the behavior to show completed QuickStarts at the bottom of the list.

Jira: https://issues.redhat.com/browse/CNV-57378

### Before

[quick-starts-bug--BEFORE--2025-04-23 10-32.webm](https://github.com/user-attachments/assets/9c7f695e-4abd-4166-941d-91fadf6b1aa6)

### After

[quick-starts-bug--AFTER--2025-04-23 10-26.webm](https://github.com/user-attachments/assets/0bceeead-07c0-46e7-9733-1eef3f583876)

### DIFF
--- a/src/views/clusteroverview/OverviewTab/getting-started-card/quick-starts-section/utils.ts
+++ b/src/views/clusteroverview/OverviewTab/getting-started-card/quick-starts-section/utils.ts
@@ -16,10 +16,12 @@ export const orderQuickStarts = (
 ): Merge<QuickStart[], { metadata: ObjectMetadata }>[] => {
   const orderedQuickStarts: Merge<QuickStart, { metadata: ObjectMetadata }>[] = [];
   const filteredQuickStarts = filter ? allQuickStarts.filter(filter) : allQuickStarts;
+
   const isFeatured = (quickStart: Merge<QuickStart, { metadata: ObjectMetadata }>) =>
     featured?.includes(quickStart?.metadata?.name);
   const getStatus = (quickStart: Merge<QuickStart, { metadata: ObjectMetadata }>) =>
     getQuickStartStatus(allQuickStartStates, quickStart?.metadata?.name);
+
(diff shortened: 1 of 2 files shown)


src/views/clusteroverview/OverviewTab/getting-started-card/quick-starts-section/utils.ts
```diff
@@ -44,7 +45,10 @@ export const orderQuickStarts = (
   // Show other not started quick starts (which are not featured)
   orderedQuickStarts.push(
     ...filteredQuickStarts.filter(
-      (q) => !isFeatured(q) && getStatus(q) === QuickStartStatus.NOT_STARTED,
+      (q) =>
+        !isFeatured(q) &&
+        (getStatus(q) === QuickStartStatus.NOT_STARTED ||
+          getStatus(q) === QuickStartStatus.COMPLETE),
```
sjd78, line 51: Probably want to also adjust the comment above to say "not started or complete".

sjd78 (commented):
If no quick starts are returned by the util function, the card is not rendered.  That doesn't exactly make sense.  WDYT about always rendering the card but dropping a "no quickstarts are available" if the list is empty?  But then again, looks like the intention is the list should never be empty and all quick starts, regardless of their quick start status, should be displayed.  If that is the case, is there any visual indication that the user has gone through an available quick start yet (like a filled blue ball for complete vs an empty blue ball for not started)?

It does make sense to render the featured quick starts first, and then the non-featured ones.  Sub-sorting is still a bit of a mystery on first read, but that's ok.

PR #2604, opened 2025-04-24
CNV-60551: show no migratable pvcs in the migration list

upalatucci:
## 📝 Description

When there is no migratable PVC we let the user continue in the migration modal

## 🎥 Demo

Screenshot 2025-04-24 at 17 09 27


src/views/virtualmachines/actions/components/VirtualMachineMigration/tabs/VirtualMachineMigrationDetails.tsx
```diff
@@ -98,6 +99,7 @@ const VirtualMachineMigrationDetails: FC<VirtualMachineMigrationDetailsProps> =
         <Radio
           id="selected-volumes"
           isChecked={!allVolumes}
+          isDisabled={isEmpty(pvcs)}
```
rszwajko, line 102: it seems we should also disable the other radio button - right now it looks as this option was valid  (but the wizard is blocked)

rszwajko (approved):

PR #2605, opened 2025-04-24
CNV-60576: show pvc storageclass

upalatucci:
## 📝 Description

Storageclasse in the VM list was detected from the VM spec. But when VM use the default storage class do not have this info.
Fetch it from the pvcs

IN order to access it faster, create a PVC mapper

## 🎥 Demo

**Before**
Screenshot 2025-04-24 at 17 49 48

**After**

https://github.com/user-attachments/assets/9b6c25c3-a704-4438-8f75-e84e609615cd

Screenshot 2025-04-28 at 09 13 57

### DIFF
--- a/src/views/virtualmachines/list/components/VirtualMachineRow/VirtualMachineRowLayout.tsx
+++ b/src/views/virtualmachines/list/components/VirtualMachineRow/VirtualMachineRowLayout.tsx
@@ -1,20 +1,25 @@
 import React, { FC, ReactNode, useMemo } from 'react';
 
-import { VirtualMachineModelGroupVersionKind } from '@kubevirt-ui/kubevirt-api/console';
+import {
+  modelToGroupVersionKind,
+  StorageClassModel,
+  VirtualMachineModelGroupVersionKind,
+} from '@kubevirt-ui/kubevirt-api/console';
 import {
   V1VirtualMachine,
   V1VirtualMachineInstanceMigration,
 } from '@kubevirt-ui/kubevirt-api/kubevirt';
 import Timestamp from '@kubevirt-utils/components/Timestamp/Timestamp';
 import { getName, getNamespace } from '@kubevirt-utils/resources/shared';
-import { getVirtualMachineStorageClasses } from '@kubevirt-utils/resources/vm';
 import { NO_DATA_DASH } from '@kubevirt-utils/resources/vm/utils/constants';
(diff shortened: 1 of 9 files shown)


src/views/virtualmachines/utils/virtualMachineRowFilter.ts
```diff
@@ -211,10 +211,38 @@ const useIPSearchFilter = (vmiMapper: VMIMapper): RowFilter => ({
...
 });
 
-const useStorageclassFilter = (vms: V1VirtualMachine[]): RowFilter => {
-  const storageClasses = useMemo(
-    () => Array.from(new Set(vms?.map((vm) => getVirtualMachineStorageClasses(vm)).flat())),
-    [vms],
+type StorageClassByVM = { [namespace in string]: { [name in string]: Set<string> } };
+
+const useStorageclassFilter = (vms: V1VirtualMachine[], pvcMapper: PVCMapper): RowFilter => {
+  const { allStorageClasses, storageClassesByVM } = vms.reduce(
```
rszwajko, line 217: ideally we should be able to re-use the next version of `getVirtualMachineStorageClasses()` helper for the first step  i.e.
```
vms
  .map( vm => [name,ns,getVirtualMachineStorageClasses(vm, pvcMapper)]
  .reduce( ...)

File: src/views/virtualmachines/list/components/VirtualMachineRow/VirtualMachineRowLayout.tsx [hunk 2]
```diff
@@ -49,7 +51,19 @@ const VirtualMachineRowLayout: FC<
   const vmName = useMemo(() => getName(obj), [obj]);
   const vmNamespace = useMemo(() => getNamespace(obj), [obj]);
 
-  const storageClasses = useMemo(() => getVirtualMachineStorageClasses(obj), [obj]);
```
rszwajko, line 52: 1. it seems we still need a helper function as this code pattern is repeated in other places
2. `reduce()`  wraps some procedural-style logic - that's OK but I would keep it one style i.e. 
```
getVolumes(vm)
  .map(volume => volume.dataVolume?.name || volume.persistentVolumeClaim?.claimName)
  .map(name => pvcMapper?.[vmNamespace]?.[name])
  .map(volumePVC => volumePVC?.spec?.storageClassName)
  .filter(Boolean)
// + some deduplication
```

rszwajko (approved):
The resource links look great!

PR #2607, opened 2025-04-25
CNV-56585: Change format for Y axis tick based on max value

rszwajko:
## 📝 Description
The change applied to memory and storage charts in VM details.

Calculate the number of digits that should be displayed after decimal point based on a static list of threshold values.
The general rule is to keep the displayed number at 3 ciphers or less. Threshold list:
1. x < 1 GiB :  0
2. 1 GiB <= x < 10 GiB : 2
3. 10 GiB <= x < 100 GiB : 1
4. 100 GiB <= x < 1 TiB : 0
5. 1 TiB <= x < 10 TiB : 2
6. 10 TiB <= x < 100 TiB : 1
7. 100 TiB <= x < 1 PiB :  0
8. x > 1 PiB : 2

### Before - truncated label with corresponding source
truncated_chart_label

### Before - range  100 GiB <= x < 1 TiB 

Screenshot From 2025-04-25 15-20-39

### After
Screenshot From 2025-04-25 14-27-52

### DIFF
--- a/src/utils/components/Charts/MemoryUtil/MemoryThresholdChart.tsx
+++ b/src/utils/components/Charts/MemoryUtil/MemoryThresholdChart.tsx
@@ -26,6 +26,7 @@ import useResponsiveCharts from '../hooks/useResponsiveCharts';
 import { getUtilizationQueries } from '../utils/queries';
 import {
   findMaxYValue,
+  getNumberOfDigitsAfterDecimalPoint,
   MILLISECONDS_MULTIPLIER,
   queriesToLink,
   tickFormat,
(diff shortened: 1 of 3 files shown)


src/utils/components/Charts/MemoryUtil/MemoryThresholdChart.tsx
```diff
@@ -107,8 +108,10 @@ const MemoryThresholdChart: FC<MemoryThresholdChartProps> = ({ vmi }) => {
                 },
                 tickLabels,
               }}
+              tickFormat={(tick: number) =>
```
metalice, line 111: seems like you are missing a run for linters, tickFormat should be next tickValues

rszwajko: on my env it gets pushed to the next line by prettier (the line is too long)

adamviktora: @metalice The linter actually moves the `tickFormat` prop above the `dependentAxis` prop. It seems like it sorts the props which have value over multiple lines first and then sorts the one line props.

adamviktora (approved):

PR #2633, opened 2025-05-06
[release-4.17] Fix rerendering router component

upalatucci:
## 📝 Description

backporting the router component useMemo from here: https://github.com/kubevirt-ui/kubevirt-plugin/pull/2621


metalice:
no need for the params ?

upalatucci:
did that here : https://github.com/kubevirt-ui/kubevirt-plugin/pull/2630

rszwajko (approved):
@upalatucci @metalice 
IMHO we have some other component that sets the params with every render (`useEffect` ?)

PR #2640, opened 2025-05-07
CNV-60417: remove folder filter after VMs are moved out of folder

adamviktora:
## 📝 Description

- Removes folder filter after VMs are moved out of folder
- disables the "Move to folder" option if VMs to move are in different namespaces

## 🎥 Demo

After:

https://github.com/user-attachments/assets/8295c181-e4c6-4b52-a2da-7dab8f478803

https://github.com/user-attachments/assets/7d8689e9-75b6-4486-a505-70bb3780e0bd

### DIFF
--- a/src/utils/components/MoveVMToFolderModal/MoveBulkVMsToFolderModal.tsx
+++ b/src/utils/components/MoveVMToFolderModal/MoveBulkVMsToFolderModal.tsx
@@ -7,6 +7,7 @@ import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTransla
 import { getNamespace } from '@kubevirt-utils/resources/shared';
 import { Popover, PopoverPosition, Stack, StackItem } from '@patternfly/react-core';
 
+import useRemoveFolderQuery from './hooks/useRemoveFolderQuery';
 import BulkVMsPopover from './BulkVMsPopover';
 import SelectedFolderIndicator from './SelectedFolderIndicator';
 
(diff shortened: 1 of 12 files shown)


src/views/virtualmachines/tree/components/TreeViewRightClickActionMenu/DefaultRightClickActionMenu.tsx
```diff
@@ -20,7 +20,7 @@ const DefaultRightClickActionMenu: FC<DefaultRightClickActionMenuProps> = ({
   const { namespace, prefix } = getElementComponentsFromID(triggerElement);
 
   const vms = getVMsTrigger(triggerElement);
-  const actions = useMultipleVirtualMachineActions(vms);
+  const actions = useMultipleVirtualMachineActions(vms, vms);
```
metalice, line 23: If u are sending the same object, make the second one optional in the hook; if it doesn't exist, treat the first as the second

src/utils/components/MoveVMToFolderModal/MoveBulkVMsToFolderModal.tsx
```diff
@@ -7,17 +7,20 @@ import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTransla
 import { getNamespace } from '@kubevirt-utils/resources/shared';
 import { Popover, PopoverPosition, Stack, StackItem } from '@patternfly/react-core';
 
+import useRemoveFolderQuery from './hooks/useRemoveFolderQuery';
 import BulkVMsPopover from './BulkVMsPopover';
 import SelectedFolderIndicator from './SelectedFolderIndicator';
 
 type MoveBulkVMToFolderModalProps = {
+  allVMs: V1VirtualMachine[];
```
upalatucci, line 15: I'm sure we should not let people move vms that have different namespaces. Do not make so much sense as folders are inside namespace.
Maybe we allow that now i don't remember but i don't think does make sense
This will let us remove all those `allVMs`
and fetch the namespace vms with `useVMsInNamespace`  inside `useRemoveFolderQuery` hook

upalatucci (approved):

PR #2644, opened 2025-05-08
CNV-61333: fix storage class migration crash migplan no volumes

upalatucci:
## 📝 Description

List only storage migplans and guard on cases where for some reason there is no volume or no storageclass target (maybe for some error during creation through cli )


src/views/storagemigrations/list/hooks/useStorageMigrationResources.ts
```diff
@@ -28,6 +28,8 @@ const useStorageMigrationResources: UseStorageMigrationResources = () => {
     namespaced: true,
   });
 
+  const storageMigPlan = migPlans.filter((plan) => plan?.spec?.persistentVolumes?.length > 0);
```
rszwajko, line 31: 1. the code needs to handle load-in-progress and error states - I would go with extra guards i.e. `migPlans?.filter(...) ?? []`
2. it's a list but the name suggests single entity

src/views/storagemigrations/list/components/StorageMigrationRow.tsx
```diff
@@ -46,10 +46,10 @@ const StorageMigrationRow: FC<StorageMigrationRowProps> = ({
         />
       </TableData>
       <TableData activeColumnIDs={activeColumnIDs} id="storagemigration">
-        {getSelectedPVFromMigPlan(obj).length} {t('Volumes')}
+        {getSelectedPVFromMigPlan(obj)?.length || '0'} {t('Volumes')}
```
rszwajko, line 49: 1. it feels a param for the translation i.e. `t('{{ count }} Volumes`, {count:  ...)`
2. the existing code has 2 different paths for zero

upalatucci: the `getSelectedPVFromMigPlan` function retrieve all the pvcs that have 'copy' as a method and not 'skip' 
Someone from cli can create a migplan with PVC with all 'skip' and here we should at least show 0 not crash the page.

src/views/storagemigrations/list/hooks/useStorageMigrationResources.ts
upalatucci: solved

rszwajko (approved):

PR #2645, opened 2025-05-08
CNV-61334: Fix rollback request storage migration

upalatucci:
## 📝 Description

patch `/spec/migrateState` only if the `migrateState` is in the object. If it was already removed, do not call the remove again. The cancellation needs some time to be performed, so show `cancelling storage migration` in that time so the user will be informed that the action has been received. 

## 🎥 Demo

Screenshot 2025-05-08 at 23 21 01


src/views/virtualmachines/actions/VirtualMachineActionFactory.tsx
```diff
@@ -92,13 +94,16 @@ export const VirtualMachineActionFactory = {
         k8sPatch({
           data: [
             { op: 'replace', path: '/spec/rollback', value: true },
-            { op: 'remove', path: '/spec/migrateState' },
+            ...(currentStorageMigration?.spec?.migrateState
```
rszwajko, line 97: Just to confirm the use case: if the `migrationState` was already removed and we try to remove it again then the server rejects the request with `422` ?

rszwajko (approved):
It's definitely an improvement - however the success of this approach depends on how up-to-date data is used in the check. It might be possible to hit the problem i.e. if the VM will be modified from CLI and UI won't know about the newest state (yet).

PR #2652, opened 2025-05-12
CNV-60146: edit labels on multiple VMs bug fixes

adamviktora:
## 📝 Description

- editing is enabled even if no label is shared
- editing doesn't overwrite non-shared labels

## 🎥 Demo

Before:

https://github.com/user-attachments/assets/2a223dc7-bc62-4bc2-ab79-ff4d28de9900

After:

https://github.com/user-attachments/assets/66b9d1de-030d-4b2b-ba90-6682402c614c

### DIFF
--- a/src/utils/components/LabelsModal/LabelsModal.tsx
+++ b/src/utils/components/LabelsModal/LabelsModal.tsx
@@ -46,10 +46,17 @@ export const LabelsModal: FC<LabelsModalProps> = memo(
     const [isInputValid, setIsInputValid] = useState(true);
 
     const initLabels = useMemo(() => {
-      if (!isEmpty(initialLabels)) return initialLabels;
-      if (!isEmpty(obj?.metadata?.labels)) return obj?.metadata?.labels;
+      if (initialLabels !== null && initialLabels !== undefined) {
+        return initialLabels;
+      }
+
+      if (!isEmpty(obj?.metadata?.labels)) {
+        return obj?.metadata?.labels;
+      }
+
       return {};
     }, [initialLabels, obj?.metadata?.labels]);
+
     const [labels, setLabels] = useState<string[]>(labelsToArray(initLabels));
 
     const onInputChange = (e: ChangeEvent<HTMLInputElement>) => {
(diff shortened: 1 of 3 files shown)


src/utils/components/LabelsModal/LabelsModal.tsx
```diff
@@ -46,10 +46,17 @@ export const LabelsModal: FC<LabelsModalProps> = memo(
...
+      if (initialLabels !== null && initialLabels !== undefined) {
+        return initialLabels;
+      }
+
+      if (!isEmpty(obj?.metadata?.labels)) {
+        return obj?.metadata?.labels;
+      }
+
       return {};
     }, [initialLabels, obj?.metadata?.labels]);
```
metalice, line 58:
```ts
  const initLabels = useMemo(() => {

      if (!isEmpty(obj?.metadata?.labels)) {
        return obj?.metadata?.labels;
      }

      return initialLabels ?? {}
    }, [initialLabels, obj?.metadata?.labels]);
```

src/views/virtualmachines/actions/utils.ts
```diff
@@ -105,51 +107,53 @@ export const createRollbackPatchData = (vm: V1VirtualMachine): Patch[] => {
   return [...dataVolumeRollback, ...pvcRollback];
 };
 
-export const getCommonLabels = (vms: V1VirtualMachine[]): { [key: string]: string } => {
+export const getCommonLabels = (vms: V1VirtualMachine[]): Labels => {
```
metalice, line 110: I think it will be easier and more readble to use Set.intersection, PTAL https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/intersection

src/utils/components/LabelsModal/LabelsModal.tsx
adamviktora: I don't want to return `obj?.metadata?.labels` if an empty object `{}` is explicitly provided in `initialLabels` prop. We need a way of overwriting the initial labels with empty object.

src/views/virtualmachines/actions/utils.ts
adamviktora: It looks like our eslint version doesn't support this. We have an `intersection` helper, but we would have to convert every label key value pair to string first and then convert it back to object.

I agree that it would be more readable though

PR #2653, opened 2025-05-12
CNV-61181: Surface StorageProfile options

rszwajko:
## 📝 Description

Follow-up to #2524

Make ApplyStorageProfileSettings component independent from
react-hook-form and re-use it for AddBootableVolumeModal and
CreateBootableVolumeModal. Both modals rely on useState hook to keep
the state.
To better encapsulate the functionality, the component will take over
the responsibility of clearing the volume/access modes after storage
class change. The storage class name is stored as ref to avoid using
setters as dependencies in the useEffect hook (component cannot assume
their references will be stable).

### Before
image

image

### After
image

Screenshot From 2025-05-12 20-04-44


src/utils/components/ApplyStorageProfileSettings/utils.ts
```diff
@@ -0,0 +1,89 @@
...
+      return t('The {{priority}}th most suggested option in StorageProfile.', {
+        priority: priority + 1,
+      });
+  }
+};
+
+export const fromPriorityToLabels = (
+  priority: number,
+  recommendationCount: number,
+  t: TFunction,
```
metalice, line 28: no need for t, the console context of trans will trigger all plugins rerender, in this loading, t will be recalculated (what we talk about)

src/utils/components/ApplyStorageProfileSettings/VolumeMode.tsx
```diff
@@ -1,33 +1,39 @@
 import React, { FC, useCallback, useEffect } from 'react';
-import { useFormContext } from 'react-hook-form';
 import { Trans } from 'react-i18next';
 import { Link } from 'react-router-dom-v5-compat';
 import { uniq } from 'lodash';
 
-import { V1beta1StorageSpecVolumeModeEnum } from '@kubevirt-ui/kubevirt-api/kubevirt';
-import HelpTextIcon from '@kubevirt-utils/components/HelpTextIcon/HelpTextIcon';
+import {
```
adamviktora, line 6: (detail not related to this change), but can we move the `{' '}` on line 70 of this file right after the `Learn more about` on line 68? (so the space is not underlined)

src/views/virtualmachines/details/tabs/configuration/storage/components/modal/CreateBootableVolumeModal.tsx
```diff
@@ -131,9 +119,7 @@ const CreateBootableVolumeModal: FC<CreateBootableVolumeModalProps> = ({
           {t('Destination details')}
         </Title>
         <VolumeDestination
-          applyStorageProfileState={applyStorageProfileState}
```
adamviktora, line 134: VolumeDestination is missing Form styles in this modal, the whole Stack inside this TabModal should be wrapped in PatternFly `<Form>` to add the styling

adamviktora (commented):
Looks good!

Only some details regarding styling. 
Another nit is in `ApplyStorageProfileSettings.scss`, the `align-items: anchor-center` doesn't work in Firefox, but `align-items: center` is fine.

src/utils/components/ApplyStorageProfileSettings/utils.ts
rszwajko: @metalice I'll test that behavior and post the results here.

rszwajko: @metalice  it seems we have a bug here. Tested with https://github.com/kubevirt-ui/kubevirt-plugin/pull/2665 and as a result opened https://issues.redhat.com/browse/CNV-61822

adamviktora (approved):

PR #2662, opened 2025-05-14
CNV-61727: Separate reported and desired network link state

rszwajko:
## 📝 Description

Before, the list of available networks and interfaces was created by merging VirtualMachine(desired state) and VirtualMachineInstance (runtime state). Then duplicates were removed. In the process information was lost - interfaces taken from VMI have different/more props then those from VM.
After this PR the networks and interfaces from both sources are linked and aggregated. Item in the list represents a network with all relevant state:
1. config - desired state retrieved from VM that contains: a) network (V1Network) b) iface  (V1Interface)
2. runtime - reported state retrieved from VMI: a) network (V1Network) b) iface (V1Interface) c) state (V1VirtualMachineInstanceNetworkInterface) 
 
The values displayed to the user are a combination of values read from available sources.

Included related fixes:
1. sorting by state - by using the same values in the sort
   as are displayed in the columns
2. default network type changing after delete from masquerade to bridge.
   Fixed by reading the type first from VirtualMachineInstance.
3. deleted default network showed as UP - fix by treating no network
   case differently then no link state prop (inside existing network)
4. all networks marked as pending during startup (for a short period).
   Fixed by checking if the VirtualMachineInstance data is available.
5. un-deleting Pod network by editing

## Before
Screenshot From 2025-05-20 12-53-50

### Network interface added to the configuration with desired state UP but not yet present at runtime

Screenshot From 2025-05-26 14-06-31

### Link state stub for interface types that do not support link state - the one listed explicitly in the docs is SRIOV type
Note the state can be seen only when creating VM.
Screenshot From 2025-05-26 13-57-11

### Interfaces for a stopped VM - configured (desired) state UP is differentiated from runtime (no state)
Screenshot From 2025-05-26 13-58-54

### Ephemeral interfaces during hot-unplug

Screenshot From 2025-05-26 14-23-05

https://github.com/user-attachments/assets/9e29aec5-7a46-41c2-880e-0ab5f5ff6f36

## Before - incorrect sorting

Screenshot From 2025-05-19 13-25-35
Screenshot From 2025-05-19 13-25-26

## After - sorting fixed
Note the screenshot presents first version of the UI.
Screenshot From 2025-05-19 13-36-00

## Before - deleting default network results in incorrect state (UP) and incorrect type (bridge)

https://github.com/user-attachments/assets/d124260a-2988-4f8b-b96d-f32a1fe92f2d

## After - deleting default network
Note the screenshot presents first version of the UI.
https://github.com/user-attachments/assets/2937780f-6e78-448e-a46e-9b71899b6cdf

## Before - adding SRIOV network from a pre-selected NAD allows to change link state and results in incorrect network type (bridge)
https://github.com/user-attachments/assets/b11a2ef8-10a4-4468-aeba-3a80d1d859c7

## Before - all networks are briefly marked as pending hot-plug
https://github.com/user-attachments/assets/330c82f9-8643-4037-bcf3-8de69a8c855d

## After - pending marker no longer visible during start-up
Note the screenshot presents first version of the UI.
https://github.com/user-attachments/assets/0ee0ad72-956e-461b-9373-d7698e84275d

## Before - un-deleting Pod network by editing

https://github.com/user-attachments/assets/3f4f9ab2-61b6-43a2-83cb-480dfb76eadf


locales/en/plugin__kubevirt-plugin.json
yfrimanm, line 1: Hi @rszwajko I think we don't need the 2 columns because they make this too complexed to understand.
I'd suggest to have only one column and add a tooltip that says something like: 
"This process is currently pending as it transitions from [X] to [Y]".
Or "The completion of this process is currently awaiting its transition from [X] to [Y]."
Thanks.

rszwajko: @yfrimanm 
The current representation of the link state (before this PR) suffers from problems related to merging reported and desired link states i.e.
1. After hot-plugging the network interface it's visible as UP (which is the desired state) but the interface is actually not added yet (no state) and is not ready to use. After this PR this will be visible - check screenshot "Network interface added to the configuration with desired state UP but not yet present at runtime"
2. For a stopped VM the interface is visible as UP/DOWN (which corresponds to the desired state of the interface) but actually there are no states reported. After this PR this difference will be easier to understand - check screenshot "Interfaces for a stopped VM - configured (desired) state UP is differentiated from runtime (no state)"
3. During hot-unplug the interfaces are visible as in-progress (again desired state) but the actual state is not described - the interface gets removed immediately and is not available for the guest OS. After this PR this case is more visible - check screenshot and screencast above under "Ephemeral interfaces during hot-unplug"
4. when removing the default/pod network the interface is reported as UP (which corresponds correctly to the runtime state but is completely different behavior then for all other cases). After this PR it's visible that the interface was removed from the configuration but it's still available to the guest OS. Check the screencast "After - deleting default network"

Representing desired and reported link state in one column is possible but
it would require a more complex representation i.e. as tuple reported state/desired state ("UP/UP" for both states UP and  "-/UP" for reported state missing) or as some aggregated state ("BOTH-UP" and "MISSING-BUT-CONFIGURED-UP"). IMHO it wouldn't be simpler to understand but we might try some approaches - any suggestions are welcomed.

yfrimanm:
From a UX standpoint, I suggest we aim for maximum simplicity: representing the current link state in a single column, complemented by tooltips to clarify any nuances.
While the implications of hotplug that require live migration are important, including those details directly might complicate the display for some users. TMO we can provide this information as optional content within the tooltips for users who wish to delve deeper.

rszwajko:
@yfrimanm 
> representing the current link state in a single column, complemented by tooltips to clarify any nuances.
>
What do you mean by `current link state` ? The desired state (configured in VM resource) or the reported state(retrieved from VirtualMachineInstance resource?

src/views/catalog/wizard/tabs/overview/components/WizardOverviewNetworksTable/WizardOverviewNetworksTable.tsx
```diff
@@ -42,9 +43,7 @@ export const WizardOverviewNetworksTable: React.FC<{
         description={
           <Stack>
             {networkData.map((n) => (
-              <StackItem key={n.network.name}>
-                {n.network.pod ? t('Pod networking') : n.network.multus?.networkName || '-'}
-              </StackItem>
+              <StackItem key={n.network.name}>{toNetworkNameLabel(t, n) || '-'}</StackItem>
```
adamviktora, line 46: nit: possibly use `NO_DATA_DASH` here

rszwajko:
@adamviktora 
I've fixed issues pointed in the review in a separate commit (to be squashed later on) for easier second review.

metalice:
@rszwajko what is the status here?

rszwajko:
@metalice 
It was waiting for https://github.com/kubevirt-ui/kubevirt-plugin/pull/2730 - after rebase it's ready for review.

src/utils/resources/vm/utils/network/types.ts
```diff
@@ -1,7 +1,15 @@
...
+import { NetworkPresentation } from './constants';
+
 export enum NetworkInterfaceState {
   'ABSENT' = 'absent',
   'DOWN' = 'down',
   'NONE' = 'none',
   'UNSUPPORTED' = 'unsupported',
   'UP' = 'up',
 }
+export type NicState = {
```
pcbailey, line 12: nit: `NicState` -> `NICState`

src/utils/components/NetworkIcons/LinkStateNoDataIcon.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+  return (
+    <Tooltip
+      content={
+        <StateText
+          {...{
+            configuredState,
+            details: unsupportedText,
+            runtimeState,
+          }}
```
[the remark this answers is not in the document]
rszwajko: @upalatucci 
Technically there is no reason.  If props are only passed to the component (names are the same) then this compact syntax is more readable. Here it's just different :)

src/utils/components/NetworkIcons/LinkStateUnsupportedIcon.tsx
```diff
@@ -0,0 +1,34 @@
...
+const LinkStateUnsupportedIcon: FC<NetworkIconProps> = ({ configuredState, runtimeState }) => {
+  const { t } = useKubevirtTranslation();
+
+  return (
+    <Tooltip
+      content={
+        <StateText
+          {...{
+            configuredState,
+            details: t('Link state is not available for this type of network interface'),
```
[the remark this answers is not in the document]
rszwajko: It could but in our i18n setup that would be an unusual approach. We use a plugin that generates the translation files based on params provided to t-function. Multiple uses are combined into entry. The other claimed benefit is that you can read the message directly in the code (no need to lookup the key).

Note that not all projects follow this approach i.e.  check translations in Konveyor which are based on hierarchical keys:
https://github.com/konveyor/tackle2-ui/blob/33175bb1c5c3d2fcd8634651d749641f01ec001e/client/public/locales/en/translation.json#L6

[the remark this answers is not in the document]
adamviktora: Interesting, thanks for this info!

upalatucci (approved):

PR #2673, opened 2025-05-19
fix po files for memsource

upalatucci:
## 📝 Description

Uploaded PO files are invalid for memsource. The issue is a `;` at the end of the plural definition. Just fix that with a script

### DIFF
--- a/i18n-scripts/export-pos.sh
+++ b/i18n-scripts/export-pos.sh
@@ -1,12 +1,25 @@
-#!/usr/bin/env bash
+#!/bin/sh
 
 set -exuo pipefail
 
 source ./i18n-scripts/languages.sh
 
+
 for f in locales/en/* ; do
   for i in "${LANGUAGES[@]}"
   do
   yarn i18n-to-po -f "$(basename "$f" .json)" -l "$i"
+
+
+  case $i in
+    es) pattern='plural=(n != 1)' ;;
+    fr) pattern='plural=(n >= 2)' ;;
+    *)  pattern='plural=0' ;;
+  esac
+
+  file="./po-files/$i/plugin__kubevirt-plugin.po"
+  sed -i -r "s|${pattern}|${pattern};|g" "$file"
+  sed -i -r 's|;;|;|g' "$file"
+
   done
 done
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -145,7 +145,7 @@
     "cypress-merge": "cd cypress && mochawesome-merge ./gui-test-screenshots/cypress_report*.json > ./gui-test-screenshots/cypress.json",
     "cypress-postreport": "yarn cypress-merge && yarn cypress-generate",
     "dev": "yarn clean && NODE_ENV=development NODE_OPTIONS=--max-old-space-size=8192 yarn ts-node node_modules/.bin/webpack serve --progress",
-    "export-pos": "./i18n-scripts/export-pos.sh",
+    "export-pos": "./i18n-scripts/export-pos.sh && ./i18n-scripts/fix-po-files.sh",
```
metalice, line 148: This cannot be fixed in export-pos.Sh, instead of adding another script file?

upalatucci: Yes

i18n-scripts/fix-po-files.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+sed -i -r "s|plural=(n != 1)|plural=(n != 1);|g" ./po-files/es/plugin__kubevirt-plugin.po
```
rszwajko, line 3: 1. please provide an example of incorrect strings
2. the scrip uses different replace patterns depending on the language - can we make it generic and run all the replacements on all files?  
3. the files are hardcoded - can we use the list of supported locales and calculate them?
4. please rename the script to better reflect it's purpose

rszwajko (commented):
The library supports [the i18next format directly ](https://support.phrase.com/hc/en-us/articles/7278275219612--JSON-i18next-i18nextV4-Strings). Can we drop the translation step?

i18n-scripts/fix-po-files.sh
upalatucci: i made a new script. Please review. 

Wrong:
Plural-Forms: nplurals=2; plural=(n != 1)\n
Correct:
Plural-Forms: nplurals=2; plural=(n != 1);\n

PR #2677, opened 2025-05-20
CNV-46434: Add HTTP DataSource option to the bootable volumes modal

pcbailey:
## 📝 Description

This adds the option to create a new DataSource from a URL to the bootable volumes modal.

Jira: https://issues.redhat.com/browse/CNV-46434

## 🎥 Demo

[create-bootable-volume-http--AFTER--2025-05-20 07-09.webm](https://github.com/user-attachments/assets/0e200c5e-e7ab-4072-ada9-9337d3605bd5)


adamviktora (approved):
Looking good!

One thing noticable from the video and also while I was playing around with it is that sometimes the new DataSource is not shown in the list upon creation at all, sometimes it disappears once cloning is done. But if I remember correctly you said you are working on that issue in a followup, so I am approving this.

PR #2679, opened 2025-05-20
CNV-57330: Use DataVolume size for InstanceType VM creation

pcbailey:
## 📝 Description

This PR updates the Instance Types VM creation flow logic to use the size of the DataVolume over that of the PVC if it exists. The size of PVCs created using a filesystem-based storage class have already been increased to include overhead, so using the already increased size would result in double the overhead.

Jira: https://issues.redhat.com/browse/CNV-57330

### Before

use-dv-size-for-instancetypes-vm-creation--BEFORE--2025-05-20_11-58

### After

use-dv-size-for-instancetypes-vm-creation--AFTER--2025-05-20_11-45

### DIFF
--- a/src/utils/resources/bootableresources/helpers.ts
+++ b/src/utils/resources/bootableresources/helpers.ts
@@ -12,15 +12,21 @@ import {
   V1beta1DataVolume,
 } from '@kubevirt-ui/kubevirt-api/containerized-data-importer/models';
 import { IoK8sApiCoreV1PersistentVolumeClaim } from '@kubevirt-ui/kubevirt-api/kubernetes';
+import {
+  getDataSourcePVCName,
+  getDataSourcePVCNamespace,
+} from '@kubevirt-utils/resources/bootableresources/selectors';
 import { isEmpty, kubevirtConsole } from '@kubevirt-utils/utils/utils';
 import { k8sDelete } from '@openshift-console/dynamic-plugin-sdk';
 
-import { getLabel, getName, getNamespace } from '../shared';
+import { getLabel, getName, getNamespace, NamespacedResourceMap } from '../shared';
 
 import { deprecatedOSNames, KUBEVIRT_ISO_LABEL } from './constants';
 import { BootableVolume } from './types';
 
-export const isBootableVolumePVCKind = (bootableVolume: BootableVolume): boolean =>
+export const isBootableVolumePVCKind = (
+  bootableVolume: BootableVolume,
+): bootableVolume is IoK8sApiCoreV1PersistentVolumeClaim =>
   bootableVolume?.kind === PersistentVolumeClaimModel.kind;
 
 export const getBootableVolumeGroupVersionKind = (bootableVolume: BootableVolume) =>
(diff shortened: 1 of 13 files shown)


src/views/catalog/CreateFromInstanceTypes/components/BootableVolumeList/components/BootableVolumeRow/BootableVolumeRow.tsx
```diff
@@ -69,14 +74,16 @@ const BootableVolumeRow: FC<BootableVolumeRowProps> = ({
   const bootVolumeName = getName(bootableVolume);
   const bootVolumeNamespace = getNamespace(bootableVolume);
   const sizeData = formatBytes(
-    pvcSource?.spec?.resources?.requests?.storage || getVolumeSnapshotSize(volumeSnapshotSource),
+    getDataVolumeSize(dvSource) ||
+      getPVCSize(pvcSource) ||
+      getVolumeSnapshotSize(volumeSnapshotSource),
```
adamviktora, line 79: It looks like your `getDiskSize` helper from utils.ts can be used here

src/views/catalog/CreateFromInstanceTypes/state/utils/types.ts
```diff
@@ -29,11 +33,10 @@ export type UseInstanceTypeAndPreferencesValues = {
 export type UseBootableVolumesValues = {
   bootableVolumes: BootableVolume[];
   dataImportCrons: V1beta1DataImportCron[];
+  dvSources: NamespacedResourceMap<V1beta1DataVolume>;
   error: Error;
   loaded: boolean;
-  pvcSources: {
-    [resourceKeyName: string]: IoK8sApiCoreV1PersistentVolumeClaim;
-  };
+  pvcSources: NamespacedResourceMap<IoK8sApiCoreV1PersistentVolumeClaim>;
```
adamviktora, line 39: Just to be clear, this looks like changing the type from:

`{ [name: string]: IoK8sApiCoreV1PersistentVolumeClaim }`
to:
`{ [namespace: string]: { [name: string]: IoK8sApiCoreV1PersistentVolumeClaim } }`

can it possibly cause issues elsewhere?

But as observing the other code it looks like the type was wrong before and this fixes it.

src/utils/resources/bootableresources/helpers.ts
```diff
@@ -43,6 +43,11 @@ export const getBootableVolumePVCSource = (
       ];
 };
```
adamviktora, line 44: Can the type of `pvcSources` argument of this `getBootableVolumePVCSource` function be changed to `NamespacedResourceMap<IoK8sApiCoreV1PersistentVolumeClaim>`?

Also I am getting a type error for the `bootableVolume` returned. I guess it is safe to return the `IoK8sApiCoreV1PersistentVolumeClaim` type because of the `isBootableVolumePVCKind` check - if so, can we alias it to `(bootableVolume as IoK8sApiCoreV1PersistentVolumeClaim)`?

adamviktora (commented):
Looking great!

A few questions / tips below

src/utils/resources/bootableresources/helpers.ts
adamviktora: Yeah that's what I am concerned about, because it looks like the way `pvcSources` are used here:
```
pvcSources?
.[(bootableVolume as V1beta1DataSource)?.spec?.source?.pvc?.namespace]?
.[(bootableVolume as V1beta1DataSource)?.spec?.source?.pvc?.name]
```

it expects a different type

PR #2680, opened 2025-05-21
CNV-62175: Remove scroll bar from VM yaml & configuration tab

metalice:
## 📝 Description

Unnecessary scroll bar, and unused space for tabs - yam; and configuration under VM details page.

## 🎥 Demo

Before:
image
image

After:
image
image


lkladnit (approved):

src/views/virtualmachines/details/tabs/yaml/virtual-machine-yaml-page.scss
```diff
@@ -1,4 +1,8 @@
 .VirtualMachineYAML--main {
-  height: 100%;
+  //Height of navbar
+  height: calc(100% - 45px);
```
rszwajko, line 3: can we use here some PF variable?
CC: @adamviktora

PR #2681, opened 2025-05-21
CNV-56886: fix dashboard links to VMs with status

adamviktora:
## 📝 Description

Fixes a link from the dashboard to filtered VMs: `rowFilter-vm-status` -> `rowFilter-status`

## 🎥 Demo

Before:

https://github.com/user-attachments/assets/35aca9b6-b3e4-4c5f-ba89-2f8306c4e958

After:

https://github.com/user-attachments/assets/079acd5a-e5ff-44b3-815d-61a895f25567


rszwajko (commented):
We need to link this code with the code that reads the filters - otherwise this will get out of sync again. IMHO the simplest link would be a constant used in both places.

adamviktora:
@rszwajko Good point. I forgot we created some constants for the filters as part of the advanced search updates.

PR #2695, opened 2025-05-26
Fix for table actions column change

lkladnit:
## 📝 Description

> Fix for disappearing actions column, introduced by  [CNV-62226: Add missing attributes to tables and navigation](https://github.com/kubevirt-ui/kubevirt-plugin/pull/2687)

## 🎥 Demo

### DIFF
--- a/src/utils/components/AffinityModal/components/AffinityList/components/AffinityRow.tsx
+++ b/src/utils/components/AffinityModal/components/AffinityList/components/AffinityRow.tsx
@@ -37,7 +37,7 @@ const AffinityRow: React.FC<
       <TableData activeColumnIDs={activeColumnIDs} id="terms">
         <div>{expressionsLabel}</div> <div>{fieldsLabel}</div>
       </TableData>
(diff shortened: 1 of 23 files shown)


src/views/storagemigrations/list/hooks/useStorageMigrationColumns.ts
```diff
@@ -60,8 +60,8 @@ const useStorageMigrationColumns = (
         transforms: [sortable],
       },
       {
-        id: '',
-        props: { className: 'pf-v6-c-table__action' },
+        id: 'actions',
+        props: { className: 'dropdown-kebab-pf pf-v6-c-table__action' },
```
adamviktora, line 64: @lkladnit I think we don't need this `dropdown-kebab-pf` class, it doesn't add any styles anymore.

adamviktora (commented):
One comment below, other than that LGTM

rszwajko (commented):
It seems we need to go back to `id=""` (although it feels wrong). The Console is re-using this attribute for column management (the action column is now visible in the modal and can be hidden).

rszwajko (approved):

PR #2700, opened 2025-05-27
CNV-62314: fix layout reload on project change in Catalog

adamviktora:
## 📝 Description

Fixes a bug where whole card in Catalog -> InstanceTypes was reloading on project change. Now the loading state is displayed just in the volumes list on project change.

## 🎥 Demo

Before:

https://github.com/user-attachments/assets/85e4ce2f-663e-437e-800a-1a5f3bfe5fe0

After:

https://github.com/user-attachments/assets/aca23532-3640-4b22-be14-5773954d0114


adamviktora:
@upalatucci Ok, I had a skeleton there first but then I decided to go with spinner :D because we don't know how many lines will be loaded, but I will do something like 3 lines

src/views/catalog/CreateFromInstanceTypes/components/BootableVolumeList/BootableVolumeList.tsx
```diff
@@ -187,9 +186,17 @@ const BootableVolumeList: FC<BootableVolumeListProps> = ({
           />
           <BootableVolumesPipelinesHint bootableVolumes={bootableVolumes} />
         </>
-      ) : (
-        <BootableVolumeEmptyState />
       )}
+      {!displayVolumes &&
```
rszwajko, line 190: consider splitting `displayVolumes` into simpler  flags  and re-defining this condition . After expanding the condition looks confusing:
```
!isEmpty(bootableVolumes) && loaded && loadedColumns && (loaded ? <Empty> : <Loading>)
```

rszwajko (approved):

PR #2701, opened 2025-05-27
CNV-62203: fixing bottom page display on guided tour step 6

batyana:
## 📝 Description

in the user guided tour on step 6 the bottom page view was missing where the target for the popover appears. the problem occurs through the joyride package- the page has difficulty scrolling down to the target and has known issues with scrolling. i forced a scroll to the target when it appears by id in the tourGuide file in the joyride callback and updated the joyride package. 

## 🎥 Demo
before:
image

after:
image

### DIFF
--- a/src/utils/components/GuidedTour/GuidedTour.tsx
+++ b/src/utils/components/GuidedTour/GuidedTour.tsx
@@ -41,9 +41,6 @@ const GuidedTour: FC = () => {
 
         if (type === EVENTS.STEP_AFTER) {
           if (action === ACTIONS.PREV) {
-            if (step.target === '#tour-step-ssh') {
-              document.getElementById('tour-step-ssh')?.scrollIntoView();
-            }
             prevStep();
 
             return;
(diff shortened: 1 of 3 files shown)


upalatucci:
Did you try to do something like

```
document.queryElement(step.target)?.scrollIntoView()
```

So it will handle also other similar scenarios in the future

src/utils/components/GuidedTour/GuidedTour.tsx
```diff
@@ -37,7 +41,11 @@ const GuidedTour: FC = () => {
 
         if (type === EVENTS.STEP_AFTER) {
           if (action === ACTIONS.PREV) {
+            if (step.target === '#tour-step-ssh') {
+              document.getElementById('tour-step-ssh')?.scrollIntoView();
+            }
```
avivtur, line 46: is this needed? the first if isn't enough?

batyana: thanks for bringing to my attention, it was deleted.

upalatucci (approved):

PR #2712, opened 2025-05-30
CNV-62163: Update Serial Console

rszwajko:
## 📝 Description
  
Changes in Serial Console:
1. update xterm from 4.19 (June 2022) to 5.5 (April 2024)
2. update @xterm/addon-fit from 0.5.0 to 0.10.0
3. Move websocket related logic to custom addon based on
   @xterm/addon-attach. Following addon pattern allows reliable cleanup
   via IDisposable interface.
4. add @xterm/addon-clipboard - it allows direct copy-paste to/from
   terminal - shortcuts depend on the terminal i.e. for pasting to Linux
   terminal Ctr+Shift+V
5. fix xterm styling by importing css provided by the library
6. fix out-of-order text - the root cause was reading from the websocket
   using asynchronous Blob.text() method. The proposed workaround uses
   synchronous FileReader API.

Changes in both Serial and VNC console:
1. strict control of re-rendering via 2 level memoization (more below)
2. common interface for actions (paste, sending keys)
3. avoid collisions when updating share state - a good example is VNC
   disconnect callback triggered asynchronously after the user
   switched to Serial console.
4. avoiding collisions between VNC sessions - disconnect callback
   performing cleanup on the newly created connection
5. switch to copy-paste via navigator.clipboard.writeText/readText
   support for the API was added in FireFox 125 (Apr 2024) which makes
   the workaround not needed anymore. Additionally it allows to treat
   paste from all sources the same - previously only credentials were
   handled.
6. auto-connect only during initial loading - after disconnect the user
   needs to manually trigger reconnect. This prevents race conditions between
   existing console and the newly opened instance (the new instance takes
   the ownership).

Changes in VNC console:
1. remove tabs for displays - main problem was synchronizing state
   between clients. The information which display is used is stored on
   the client only and updated after user chooses an option from the
   dropdown. The other issue is error handling - we have no feedback if
   the display really changed. Technically it's the same component.

2-level memoization:
1. the first level protects from re-renders passed from parents.
   It accepts only simple types(i.e. string,number or boolean) as props.
   The assumption is that they are semi-constant (i.e. VM name).
   Example of first level component: Consoles.tsx.
2. second level is the actual console component which should be rendered
   only once. The received props are stable (i.e. setter from useState
   hook). The component is able to modify the parent state (via setter)
   but is not affected by the resulting state changes. This allows the
   console to operate in a stable environment. Note that memoization
   requires also controlling re-creation - the console component and all
   parent components cannot be re-created when the console is active.

## 🎥 Demo

https://github.com/user-attachments/assets/6019d30e-4973-4694-8a21-2e14c434dc02

### Out-of-order before

Screenshot From 2025-06-25 12-51-37

### Out-of-order after

Screenshot From 2025-06-25 13-12-29


src/utils/components/Consoles/components/AccessConsoles/utils/accessConsoles.tsx
```diff
@@ -1,24 +1,33 @@
...
-  serialSocket?: WSFactoryExtends;
-  setType: Dispatch<SetStateAction<string>>;
-  type: string;
+  setType: Dispatch<SetStateAction<ConsoleType>>;
+  state: ConsoleState;
+  type: ConsoleType;
 };
 
-export const typeMap = (isWindowsVM: boolean) => ({
+export const typeMap = (isWindowsVM: boolean, t: TFunction): { [key in ConsoleType]: string } => ({
```
upalatucci, line 30: here you can use the global t from `useKubevirtTranslation` if you want

src/utils/components/Consoles/utils/utils.ts
```diff
@@ -15,4 +15,7 @@ export const clipboardCopyFunc = (
   );
 };
 
+export const getConsolePath = ({ name, namespace }) =>
```
upalatucci, line 18: i would call it `getConsoleBasePath` as it does not have `/console` or `/vnc`

PR #2717, opened 2025-06-03
CNV-61063: Support user-defined network (UDN) as secondary network

rszwajko:
## 📝 Description

Support user-defined network (UDN) as secondary network

### After

https://github.com/user-attachments/assets/8498cc99-d708-4514-abcd-f69b95c900d6

### DIFF
--- a/src/utils/resources/udn/types.ts
+++ b/src/utils/resources/udn/types.ts
@@ -5,8 +5,8 @@ export type UserDefinedNetworkAnnotations = {
 };
 
 export enum UserDefinedNetworkRole {
-  Primary = 'Primary',
-  Secondary = 'Secondary',
+  primary = 'primary',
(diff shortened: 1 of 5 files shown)


src/utils/resources/udn/types.ts
```diff
@@ -5,8 +5,8 @@ export type UserDefinedNetworkAnnotations = {
 };
 
 export enum UserDefinedNetworkRole {
-  Primary = 'Primary',
-  Secondary = 'Secondary',
+  primary = 'primary',
+  secondary = 'secondary',
 }
```
upalatucci, line 10: in `UserDefinedNetworkLayer2` the `role` have the first letter uppercase like `Primary` and not `primary` . We can create another enum or use the `.toLowerCase`

PR #2718, opened 2025-06-03
CNV-39729: show disk size in TiB and not bytes

adamviktora:
## 📝 Description

Show disk size in TiB and not bytes in Templates / Catalog templates -> Disks

- refactoring of duplicated code, ideally we should use `useDisksTableData` hook everywhere, but there are some differences so for now I only reused `useWizardDisksTableData` in `useTemplateDisksTableData`
- remove of unused class

## 🎥 Demo
Before:

https://github.com/user-attachments/assets/5d86c540-f093-4969-94f0-339503ffdbf5

After:

https://github.com/user-attachments/assets/69a362ec-13d5-41cd-8500-c9350a65888a

### DIFF
--- a/src/views/catalog/wizard/tabs/disks/hooks/useWizardDisksTableData.ts
+++ b/src/views/catalog/wizard/tabs/disks/hooks/useWizardDisksTableData.ts
@@ -1,6 +1,5 @@
 import * as React from 'react';
 
-import { bytesFromQuantity } from '@catalog/utils/quantity';
 import { PersistentVolumeClaimModel } from '@kubevirt-ui/kubevirt-api/console';
 import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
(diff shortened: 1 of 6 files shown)


src/views/catalog/wizard/tabs/disks/hooks/useWizardDisksTableData.ts
```diff
@@ -87,7 +92,7 @@ const useWizardDisksTableData: UseDisksTableDisks = (vm: V1VirtualMachine) => {
         metadata: { name: device?.disk?.name },
         name: device?.disk?.name,
         namespace: device?.pvc?.metadata?.namespace,
-        size: size ? bytesFromQuantity(size, 2).join('') : '-',
+        size: size ? humanizeBinaryBytes(convertToBaseValue(size)).string : NO_DATA_DASH,
```
rszwajko, line 95: It seems we have 2 competing implementations here:) My first choice would be `bytesFromQuantity` as it's unit tested and backed by a library, however it seems that it's not handling `TiB` correctly (after adding a unit test it fails). 
IMHO we should fix  `bytesFromQuantity`  anyway (as it's used in different places). In a follow up we should decide which approach to use and refactor the project accordingly (CC: @upalatucci  @pcbailey )

rszwajko (changes requested):

src/views/catalog/wizard/tabs/disks/hooks/useWizardDisksTableData.ts
adamviktora: I agree we should refactor and only use one. I would prefer `humanizeBinaryBytes` + `convertToBaseValue` as it looks like it is used in more places and it doesn't have the bug. I used it because it was used in `useDisksTableData` quite recently - a hook used in customizing InstanceType VM and in VM details page.

Ideally I think we want to replace `useWizardDisksTableData` and `useTemplateDisksTableData` with the `useDisksTableData` implementation, which seems to be the most recently updated.

upalatucci: Yes but be carefull on this point: `useWizardDisksTableData` and `useTemplateDisksTableData` are created for pages where the vm does not exist yet. So also the disk pvcs and so on. They are slightly different from `useDisksTableData` for this reason.

upalatucci (approved):

rszwajko (approved):
We can fix the library method in a follow-up.

PR #2730, opened 2025-06-06
CNV-61727: Share network column definitions

rszwajko:
## 📝 Description

Changes:
1. strict typing for sort methods - string based selectors tend to get out of sync with code base
2. fix sorting by network - by using the same string representation via toNetworkNameLabel() helper
3. universal string comparator - borrowed from forklift-console-plugin
4. move networking related types/utils to resources/vm/utils/network

Split from #2662.

Reference-Url: https://github.com/kubevirt-ui/kubevirt-plugin/pull/2662
Reference-Url: https://github.com/kubev2v/forklift-console-plugin/blob/9f467af19bc5aea651c07db66fb49251120fc4b2/packages/common/src/components/TableView/sort.ts#L16

## 🎥 Demo

See #2662.


src/utils/resources/vm/utils/network/network-columns.tsx
```diff
@@ -0,0 +1,89 @@
...
+export const toNetworkNameLabel = <
+  T extends { network: { multus?: { networkName: string }; pod?: unknown } },
+>(
+  t: TFunction,
+  nic: T,
+) => (nic?.network?.pod ? t('Pod networking') : nic?.network?.multus?.networkName);
+
+export const isSRIOVNetwork = <T extends { sriov?: object }>(iface: T) => !!iface?.sriov;
+
+export const Name = <T extends { network: { name: string } }>(t: TFunction): TableColumn<T> => ({
```
pcbailey, line 22: Maybe I'm missing something, but is there a reason we're not just using `NetworkPresentation` here?

src/utils/utils/utils.ts
```diff
@@ -169,3 +170,31 @@ export const removeDockerPrefix = (image: string) => image?.replace(DOCKER_PREFI
...
+  localeCompare(String(a ?? ''), String(b ?? ''), locale);
+
+/**
+ * Uses native string localCompare method with numeric option enabled.
+ * @param a -
+ * @param b -
+ * @param locale to be used by string compareFn
+ */
+export const localeCompare = (a: string, b: string, locale: string): number =>
+  a.localeCompare(b, locale, { numeric: true });
```
adamviktora, line 192: Can we possibly reuse this in `comparePathsValues` helper in this file? So we don't end up with two different implementations of the same thing. 

Also I really like the `Nullish entities are converted to empty string` behavior in `universalComparator` helper, which could probably also be used in `comparePathsValues`

rszwajko: Yes. `comparePathsValues` is buggy  - I'll open a separated PR for that.

src/utils/resources/vm/utils/network/network-columns.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+import { NetworkPresentation } from '@kubevirt-utils/resources/vm/utils/network/constants';
+import { getConfigInterfaceState } from '@kubevirt-utils/resources/vm/utils/network/selectors';
+import { sortNICs } from '@kubevirt-utils/resources/vm/utils/network/utils';
+import { sortByDirection, universalComparator } from '@kubevirt-utils/utils/utils';
+import { TableColumn } from '@openshift-console/dynamic-plugin-sdk';
+import { sortable, SortByDirection } from '@patternfly/react-table';
+
+export const compareWithDirection = (direction: SortByDirection, a: any, b: any) =>
+  sortByDirection(universalComparator, direction)(a, b);
```
adamviktora, line 11: Can this be extracted also to the general utils file?

src/utils/resources/vm/utils/network/network-columns.tsx
metalice: why moving t around? get directly

rszwajko: @metalice 
1. regarding changing the location - the main motivation is to re-use column definition - before this PR there were 3 places with (almost) the same code
2. regarding changing the way we build column definition:
      1. the previous way of getting the value for sorting (as string) was prone to bugs (i.e. [sorting by network](https://github.com/kubevirt-ui/kubevirt-plugin/blob/976597487d3653dbbfd6f630f24fb03b8690fc27/src/views/virtualmachines/details/tabs/configuration/network/hooks/useNetworkColumns.ts#L30)). With string typing the props will be in sync.
      2. column builder as function allows to pass translations in a proper way
4. regarding the use of generics - with changes in #2662 the network columns in VM details will have different entity then in catalog

PR #2734, opened 2025-06-09
fixing the rendering on empty servicesList causing react error 130

batyana:
there is an error when trying to access vm details caused by servicesList from console SDK-internal rendering as undefined,  it brings up react error 130.  Since this is an unsupported console SDK I created a workaround to make sure the component only renders when defined. 

> Add a brief description
before:
image

after:
image


adamviktora:
In a followup we must check why the `ServicesList` component is not working

banyartibi:
In which release you will fix this? 1.15.2 still a problem with OKD 4.20

PR #2744, opened 2025-06-11
CNV-62920: Detect single node cluster mode based on number of nodes

rszwajko:
## 📝 Description

The single node cluster node detection is used in the UI to block live migrations between nodes.
Before this PR the check was based on Infrastructure CR  and  `infrastructureTopology` property. If the property was `SingleReplica` then single cluster node was assumed.  
After, this PR the check will be based on actual number of nodes in the cluster.

### DIFF
--- a/src/utils/hooks/useSingleNodeCluster.ts
+++ b/src/utils/hooks/useSingleNodeCluster.ts
@@ -1,36 +1,28 @@
-import { useMemo } from 'react';
-
-import { InfrastructureModel, modelToGroupVersionKind } from '@kubevirt-ui/kubevirt-api/console';
+import { modelToGroupVersionKind, NodeModel } from '@kubevirt-ui/kubevirt-api/console';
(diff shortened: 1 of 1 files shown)


metalice:
@rszwajko  Why is infrastructureTopology not reliable? Something changed in the API?
Isn't the bug about the refresh issue?

rszwajko:
@metalice 
> Why is infrastructureTopology not reliable?
>
It's not a reliable way to detect single node cluster - the main purpose is to distinguish between high-availability mode and non-high-availability. However a non-HA cluster can still be based on multiple nodes - for details see:
https://docs.redhat.com/en/documentation/openshift_dedicated/4/html/operators/developing-operators#osdk-ha-sno-api_osdk-ha-sno

rszwajko:
Replaced with https://github.com/kubevirt-ui/kubevirt-plugin/pull/2751

PR #2748, opened 2025-06-12
CNV-57475: make disk size not editable when using volume snapshot

adamviktora:
## 📝 Description

When adding a bootable volume from "Volume snapshot", changing size of the disk had no effect - the size is always determined from the VolumeSnapshot size.

This PR disables the Disk size input when a "Volume snapshot" option is chosen and adjusts the size in the input + adds an info message for users.

## 🎥 Demo

Before:

 

https://github.com/user-attachments/assets/e65d8716-827d-4761-8cc5-0c1b03896cae

After:

https://github.com/user-attachments/assets/69c1dfbb-03a0-49ca-bdb5-0540abfa973c

### DIFF
--- a/src/utils/components/AddBootableVolumeModal/components/SourceTypeSelection/SourceTypeSelection.tsx
+++ b/src/utils/components/AddBootableVolumeModal/components/SourceTypeSelection/SourceTypeSelection.tsx
@@ -12,12 +12,14 @@ import { DROPDOWN_FORM_SELECTION, optionsValueLabelMapper } from '../../utils/co
 type SourceTypeSelectionProps = {
   formSelection: DROPDOWN_FORM_SELECTION;
   namespace: string;
+  resetDiskSize: () => void;
   setFormSelection: (value: DROPDOWN_FORM_SELECTION) => void;
 };
 
 const SourceTypeSelection: FC<SourceTypeSelectionProps> = ({
   formSelection,
   namespace,
+  resetDiskSize,
   setFormSelection,
 }) => {
   const [isOpen, setIsOpen] = useState(false);
@@ -31,8 +33,15 @@ const SourceTypeSelection: FC<SourceTypeSelectionProps> = ({
       event.preventDefault();
       setFormSelection(value);
       setIsOpen(false);
+
+      if (
+        formSelection === DROPDOWN_FORM_SELECTION.USE_SNAPSHOT &&
+        value !== DROPDOWN_FORM_SELECTION.USE_SNAPSHOT
+      ) {
+        resetDiskSize();
+      }
     },
-    [setFormSelection],
+    [setFormSelection, resetDiskSize],
   );
 
   useEffect(() => {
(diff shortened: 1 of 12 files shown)


src/utils/components/AddBootableVolumeModal/components/VolumeSource/components/SnapshotSource.tsx
```diff
@@ -17,6 +17,7 @@ const SnapshotSource: FC<SnapshotSourceProps> = ({ bootableVolume, setBootableVo
     <SelectSnapshot
       selectSnapshotName={setBootableVolumeField('snapshotName')}
       selectSnapshotNamespace={setBootableVolumeField('snapshotNamespace')}
+      setDiskSize={setBootableVolumeField('size')}
```
pcbailey, line 20: We should eventually refactor this to use an enum or something other than hard-coded strings.

src/utils/components/SelectSnapshot/SelectSnapshot.tsx
```diff
@@ -46,6 +50,9 @@ const SelectSnapshot: FC<SelectSnapshotProps> = ({
     [snapshots],
   );
 
+  const getSnapshotSize = (snapshotName: string) =>
+    snapshots?.find((snapshot) => snapshot.metadata?.name === snapshotName)?.status?.restoreSize;
```
pcbailey, line 54: Please use the `getName` selector.

src/utils/components/AddBootableVolumeModal/components/VolumeSource/components/SnapshotSource.tsx
adamviktora: I wouldn't say this string is hard-coded. The type is `setBootableVolumeField: (key: keyof AddBootableVolumeState, fieldKey?: string) => ...` so the only possible strings are keys in `AddBootableVolumeState`. VSCode also shows a nice hint:

Screenshot 2025-06-23 at 8 11 09

src/utils/components/AddBootableVolumeModal/components/SourceTypeSelection/SourceTypeSelection.tsx
```diff
@@ -31,6 +33,10 @@ const SourceTypeSelection: FC<SourceTypeSelectionProps> = ({
       event.preventDefault();
       setFormSelection(value);
       setIsOpen(false);
+
+      if (value !== DROPDOWN_FORM_SELECTION.USE_SNAPSHOT) {
+        resetDiskSize();
+      }
     },
     [setFormSelection],
```
rszwajko, line 41: `resetDiskSize` is missing in callbacks dependencies

src/utils/components/AddBootableVolumeModal/components/SourceTypeSelection/SourceTypeSelection.tsx
```diff
@@ -31,6 +33,10 @@ const SourceTypeSelection: FC<SourceTypeSelectionProps> = ({
       event.preventDefault();
       setFormSelection(value);
       setIsOpen(false);
+
+      if (value !== DROPDOWN_FORM_SELECTION.USE_SNAPSHOT) {
+        resetDiskSize();
```
rszwajko, line 38: why we need this exception? it leads to confusing behaviour i.e.
1. choose "volume" source and modify the disk size to X
2. switch to "snapshot" source - the disk size is X but cannot be changed

src/utils/components/AddBootableVolumeModal/components/VolumeSource/components/SnapshotSource.tsx
pcbailey: I understand that the type of the value is restricted. By hard-coded I mean that we're typing out string literals. It's not worth holding up the PR over, though.

PR #2753, opened 2025-06-13
Fix storagemigration

upalatucci:
## 📝 Description

The storage team requested us to change a little bit the flow on migrationstorage

Create an empty migplan from the start (and delete that if the user close the modal before starting the migplan)
If there is already a migplan operating in the same namespace (Even if its completed), show an alert and ask the user for some action (delete or move)

And then use the migplan suffix (that in the meanwhile have been populeated by the MTC controller) on the migplan for pvc
 

## 🎥 Demo

ALERT:

Screenshot 2025-06-13 at 16 40 33


src/views/virtualmachines/actions/components/VirtualMachineMigration/VirtualMachineMigrationModal.tsx
```diff
@@ -34,8 +49,13 @@ const VirtualMachineMigrateModal: FC<VirtualMachineMigrateModalProps> = ({
 }) => {
   const { t } = useKubevirtTranslation();
 
+  const migrationNamespace = getNamespace(vms?.[0]);
   const [selectedStorageClass, setSelectedStorageClass] = useState('');
 
+  const currentMigPlanCreation = useCreateEmptyMigPlan(migrationNamespace);
```
rszwajko, line 55: should we wait until the plan gets created?

src/views/virtualmachines/actions/components/VirtualMachineMigration/hooks/useCreateEmptyMigPlan.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  useEffect(() => {
+    const emptyMigPlan = getEmptyMigPlan(namespace);
+
+    k8sCreate({
+      data: emptyMigPlan,
+      model: MigPlanModel,
+    }).then((createdMigPlan) => {
+      setMigPlan(createdMigPlan);
+    });
```
rszwajko, line 19: if we return `loading` then we could also return `error`

rszwajko (approved):

PR #2754, opened 2025-06-16
CNV-63722: fix crashing settings

upalatucci:
## 📝 Description

The error that react show in the crash

```
Objects are not valid as a React child (found: e: hyperconvergeds.hco.kubevirt.io is forbidden: User "test2" cannot list resource "hyperconvergeds" in API group "hco.kubevirt.io" at the cluster scope). If you meant to render a collection of children, use an array instead.
````

The crash is because we are displaing the error object that react cannot handle . We should show the `error.message` that its a string. We have a component that do that . ErrorAlert. And if the error is `null`, it will return `null`


upalatucci:
@rszwajko done it ! Thanks. In other occurrencies the error is represented as a `string` so no need for the fix

PR #2765, opened 2025-06-19
CNV-63729: capacity input fixes

adamviktora:
## 📝 Description

- fixes a bug in `CapacityInput` caused by passing a localized string to `size` parameter and then parsing it. E.g. `1,000 Gi` would result to only a `1 GiB` in the input before

- updates the `CapacityInput` so typing in a floating point number is not possible. And if there initially is a floating point number, pressing +/- will round it to integer

- adds ```type Quantity = {
    unit: QuantityUnit;
    value: number;
}``` for easier work in CapacityInput. And a `toQuantity` helper, which converts a [Quantity string](https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/) to this type

- adds `formatQuantityString` helper which aims to convert a raw bytes string in [Quantity](https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/) format (obtained from YAML) to another string in Quantity format but formatted as integer + unit
   - formatting will occur only if conversion to an integer is possible, e.g. `'1024'` would be converted to `1Ki`, similarly a `'1000'` to `1k`. Unit will be as large as possible so the size is still an integer
   - useful for parsing PVC sizes. When we add a new disk for a VM, a `DataVolume` is created, and the DataVolume's underlying `PVC` will have it's requested storage size written as a bytes string (unfortunately), e.g. `24Gi` in DataVolume will be `'25769803776'` in PVC (looks like a backend feature)
   
- adds `useUnitOptions` helper for the CapacityInput, which for the initial unit returns an appropriate array of options. E.g. if the initial size is `50k`, the returned options will be `Ki, Mi, Gi, Ti, k` - the logic is to prefer binary units so there is always a `Mi, Gi, Ti` as defaults. Then the initial unit is there (`k`) and any other "close" units in the binary base - in this case `Ki`
   - note that other decimal units are not provided on purpose, we want to avoid them. I think it is good though to still show the decimal unit if the storage is specified that way in YAML, rather than either:
      1) converting to closest integer binary (usually bytes) and showing that large number (bad UX)

      2) converting to binary + rounding (`10G` -> `9.31Gi`), this will cause issues because the rounding will not be precise and ultimately when we store something like `9.31Gi`, which is `9996536381.44` bytes (see it is a float), it will be stored in the YAML with `m` unit, which is literally ["milibytes"](https://stackoverflow.com/a/73439953) and it will show a warning
         - floating point numbers are always converted, see [Kubernetes Quantity](https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/) 
         > 1.5 will be serialized as "1500m" - 1.5Gi will be serialized as "1536Mi"

   
### Things to fix in a followup:

- some `CapacityInput`s have problem with generating correct unit options for the dropdown. This is occuring because the size prop may get changed a moment later, but it is already stored in a separate state. 

   - We need the separate state so the dropdown options stay unchanged even after changing the unit. If it wouldn't be in a separate state, then e.g. if `KiB` would be changed to `MiB`, causing the `onChange` changing the underlying `size` prop, which would regenerate new `unitOptions` - and the information about KiB is lost, so this option would disappear (and that is an annoying bug).
   
- remove `CAPACITY_UNITS` enum and refactor accordingly (use `BinaryUnit` + `addByteSuffix` instead)

## 🎥 Demo

Bug before:

https://github.com/user-attachments/assets/36cb5c65-272c-4589-9d53-3743541eefdc

Bug after:

https://github.com/user-attachments/assets/268a60ac-fcb7-4d29-bb23-c7fa1664fe89


src/utils/components/CapacityInput/useUnitOptions.ts
```diff
@@ -0,0 +1,14 @@
...
+
+import { toQuantity } from '@kubevirt-utils/utils/units';
+
+import { getUnitOptions } from './utils';
+
+const useUnitOptions = (size: string) => {
+  const [initialSize] = useState(size);
+  const { unit: initialUnit } = toQuantity(initialSize);
+
+  return getUnitOptions(initialUnit);
```
upalatucci, line 11: why we have `useState` if we consider only the initialSize and don't set anymore the state?

src/utils/utils/unitConstants.ts
```diff
@@ -0,0 +1,62 @@
...
+multipliers.G = multipliers.M * 1000;
+multipliers.T = multipliers.G * 1000;
+multipliers.P = multipliers.T * 1000;
+multipliers.E = multipliers.P * 1000;
+
+/** Matches alphabetic characters at the end of the string */
+export const UNIT_REGEX = /[a-zA-Z]+$/;
+
+/** Matches numeric characters (including decimal) at the start of the string */
+export const NUMBER_REGEX = /^[0-9.]+/;
```
rszwajko, line 62: it seems we do not support decimal exponent type:
https://github.com/kubevirt/kubevirt/blob/205d9455db56d5fcd42fb331122cfef358f19a69/vendor/k8s.io/apimachinery/pkg/api/resource/quantity.go#L131

src/utils/utils/units.ts
```diff
@@ -63,3 +49,90 @@ export const getHumanizedSize = (size: string, bytesOption: 'withB' | 'withoutB'
...
+export const isBinaryUnit = (unit: string): unit is BinaryUnit =>
+  binaryUnitsOrdered.includes(BinaryUnit[unit]);
+
+export const isDecimalUnit = (unit: string): unit is DecimalUnit =>
+  decimalUnitsOrdered.includes(DecimalUnit[unit]);
+
+const isQuantityUnit = (unit: string): unit is QuantityUnit =>
+  isBinaryUnit(unit) || isDecimalUnit(unit);
+
+export const toQuantity = (quantityString: string, keepInitialUnit = true): Quantity => {
```
rszwajko, line 70: 1. IMHO we should unit test the utility functions in this file - they are likely to evolve in the future
2. please provide (perma)link to the format description - original not generated - https://github.com/kubevirt/kubevirt/blob/205d9455db56d5fcd42fb331122cfef358f19a69/vendor/k8s.io/apimachinery/pkg/api/resource/quantity.go#L33

src/utils/utils/units.ts
```diff
@@ -63,3 +49,90 @@ export const getHumanizedSize = (size: string, bytesOption: 'withB' | 'withoutB'
...
+ */
+export const formatQuantityString = (quantityString: string) => {
+  if (!quantityString) {
+    return null;
+  }
+
+  const unit = extractUnitFromQuantityString(quantityString);
+
+  // special case when unit = 'm' can occur, e.g. if we create a PVC with floating point number like 0.2Gi (1/5 Gi) which can't be converted to bytes (power of 2 is not divisible by 5), 'm' then stands for milibytes
+  if (isQuantityUnit(unit) || unit === 'm') {
```
rszwajko, line 120: it seems we have more cases like this - https://github.com/kubevirt/kubevirt/blob/205d9455db56d5fcd42fb331122cfef358f19a69/vendor/k8s.io/apimachinery/pkg/api/resource/suffix.go#L123

rszwajko (changes requested):
It's definitely change in the right direction! The parts that are missing:
1. unit test - if possible aligned with backend tests for parsing quantity format (if exist)
2. aligning our parse algorithm with the backend - if some formats/variants are not supported then we need to handle it gracefully

src/utils/utils/units.ts
```diff
@@ -63,3 +52,138 @@ export const getHumanizedSize = (size: string, bytesOption: 'withB' | 'withoutB'
...
+    return quantityString;
+  }
+
+  if (!isEmpty(unit)) {
+    // ERROR: formatQuantityString helper called with quantityString argument which is not in Quantity format
+    return quantityString;
```
rszwajko, line 177: I'm not sure how we should proceed in case of error (covers also exotic units above). Parsing routines have different strategies is such cases (throwing exception, default values, empty values, returning what's already parsed, etc). 
The components that use this utility have no clear fallback in case of error so throwing exception seems a bad choice.
What do you think about returning some safe but invalid value instead? i.e. zero?
For the code that would need to detect an error we could also provide a `isQuantityString` helper.

adamviktora: Returning a zero seems reasonable. Although ideally we should raise an error before calling `formatQuantityString`, so I like the idea of checking the value with `isQuantityString` helper first.

I briefly looked at the usages of `formatQuantityString`, and it is mostly checking pvc size, which comes from useK8sWatchResource so the PVC's yaml should already be formatted correctly by backend.

Can we do the `isQuantityString` helper in a followup? I don't think we have any bugs raised regarding reading invalid values from pvc's yaml right now, so it seems like a nice-to-have, but not a must-have I would say.

rszwajko (approved):
Thanks for stabilizing this area!

PR #2767, opened 2025-06-20
CNV-56959: Allow snapshot as boot source for Templates

pcbailey:
## 📝 Description

This PR fixes a bug that prevented snapshots that had been added to Templates as boot sources weren't being recognized as boot sources in the Template-based VM creation flow.

Jira: https://issues.redhat.com/browse/CNV-56959

### Before

add-snapshot-as-boot-source--BEFORE--2025-06-20 10-53

### After

add-snapshot-as-boot-source--AFTER--2025-06-20 10-50

### DIFF
--- a/package.json
+++ b/package.json
@@ -136,6 +136,7 @@
     ]
   },
   "name": "kubevirt-plugin",
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e",
   "private": true,
   "resolutions": {
     "@types/react": "^17.0.40",
--- a/src/utils/resources/template/utils/constants.ts
+++ b/src/utils/resources/template/utils/constants.ts
@@ -103,6 +103,7 @@ export enum BOOT_SOURCE {
   NONE = 'NONE',
   PVC = 'PVC',
   REGISTRY = 'REGISTRY',
+  SNAPSHOT = 'SNAPSHOT',
   URL = 'URL',
 }
 
@@ -112,6 +113,7 @@ export enum BOOT_SOURCE {
 // t('Registry')
 // t('Container disk')
 // t('No boot source')
+// t('Snapshot')
 
 export const BOOT_SOURCE_LABELS = {
   [BOOT_SOURCE.CONTAINER_DISK]: 'Container disk',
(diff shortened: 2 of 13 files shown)


package.json
```diff
@@ -136,6 +136,7 @@
     ]
   },
   "name": "kubevirt-plugin",
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e",
```
adamviktora, line 139: Looks like `corepack` or some tool added this line, it should be removed i think

src/utils/resources/template/utils/constants.ts
```diff
@@ -120,6 +122,7 @@ export const BOOT_SOURCE_LABELS = {
   [BOOT_SOURCE.NONE]: 'No boot source',
   [BOOT_SOURCE.PVC]: 'PVC',
   [BOOT_SOURCE.REGISTRY]: 'Registry',
+  [BOOT_SOURCE.SNAPSHOT]: 'Snapshot',
```
adamviktora, line 125: Just curious - why the `t` function is not used directly in the `BOOT_SOURCE_LABELS`? The commented translation is just to generate the translation in the locales I guess. So it will get translated even if not wrapped in `t()` but only because a translation exists in the locales?

start-console.sh
```diff
@@ -63,7 +63,7 @@ for arg in "$@"; do
     INITIAL_PORT=$((INITIAL_PORT + 1))
 done
 
-CONSOLE_IMAGE=${CONSOLE_IMAGE:-"quay.io/openshift/origin-console:latest"}
+CONSOLE_IMAGE=${CONSOLE_IMAGE:-"quay.io/openshift/origin-console:4.19"}
```
adamviktora, line 66: Do we want to commit this too? (I mean it would help the local development for now, we can change it back once the latest version works)

adamviktora (commented):
Only few nits and a question, other than that LGTM!

src/utils/resources/template/utils/constants.ts
pcbailey: This is the way we did it before we started using `t()` in non-component files. These values were being wrapped in `t()` where they were used.

I thought there was a reason this was still using the old way, but `t()` was already imported here, so I updated it to use it.

adamviktora: Cool, thanks

PR #2779, opened 2025-06-26
fixing the CPU unit display in summery and top consumers

batyana:
## 📝 Description

fixing the cpu unit display to display the right amount of units and correct unit measurements in vm summery and in top consumers by using humanizeCpuCores().

## 🎥 Demo
before:
image
image

after:
image
image


adamviktora (commented):
Looks good!

One thing - can we remove the space between "`m`" and the value in the Top consumers tab? Or add a space in the summary. Just so we are consistent.

I would prefer no space so it aligns with the [YAML format](https://medium.com/@betz.mark/understanding-resource-limits-in-kubernetes-cpu-time-9eff74d3161b), on the other hand we add spaces between bytes and unit in Memory...

Curious what others think

upalatucci:
@yfrimanm what do you think about the comment adam wrote?

PR #2792, opened 2025-07-02
CNV-64583:  Add read only access to feature flags

rszwajko:
## 📝 Description
   
Before, the features were accessed via useFeatures hook which provides:
1. values
2. mutators
2. logic for auto-creating resources on the backend
Most of the consuming components use only the value.

After this PR, components that need read only access receive only
the requested functionality (no mutators, no side-effects).
On the Virtual Machine screen the value is provided by context which
allows pre-loading the value for all nested components.

Additionally strict typing was added to capture the list of supported
feature flags.

### Before
https://github.com/user-attachments/assets/f8ad34c5-73f5-404e-ad8f-deb502e3a732
### After

https://github.com/user-attachments/assets/a24f99fb-10e4-48e3-9056-97746669b536


upalatucci:
@rszwajko you can load this context using the extension `console.context-provider` so we'll not need to add the context to all endpoints

PR #2802, opened 2025-07-09
CNV-64839: Moving the copy button out of text element making it always visible

batyana:
## 📝 Description

Moving the copy button out of the hidden text field by creating new copy buttons and hiding only the credential text instead of the whole component, making the copy button always visible while only toggling the text display.

## 🎥 Demo
Before: 
image
image

After:
Screenshot 2025-07-29 at 10 53 14

### DIFF
--- a/src/utils/components/Consoles/components/CloudInitCredentials/CloudInitCredentialsContent.tsx
+++ b/src/utils/components/Consoles/components/CloudInitCredentials/CloudInitCredentialsContent.tsx
@@ -6,10 +6,7 @@ import { getCloudInitCredentials } from '@kubevirt-utils/resources/vmi';
 import { isEmpty } from '@kubevirt-utils/utils/utils';
 import { Flex } from '@patternfly/react-core';
 
-import { LINE_FEED } from '../vnc-console/utils/util';
-
 import CloudInitCredentialsItem from './CloudInitCredentialsItem';
-import InlineCodeClipboardCopy from './InlineCodeClipboardCopy';
 
 type CloudInitCredentialsContentProps = {
   vm: V1VirtualMachine;
@@ -19,7 +16,7 @@ const CloudInitCredentialsContent: FC<CloudInitCredentialsContentProps> = ({ vm 
   const { t } = useKubevirtTranslation();
   const { users } = getCloudInitCredentials(vm);
 
-  if (isEmpty(users)) {
+  if (!Array.isArray(users) || isEmpty(users)) {
     return (
       <div className="pf-v6-u-ml-md">
         {t('No credentials, see operating system documentation for the default username.')}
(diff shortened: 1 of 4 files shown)


src/utils/components/Consoles/components/CloudInitCredentials/CloudInitCredentialsItem.tsx
```diff
@@ -22,7 +24,13 @@ const CloudInitCredentialsItem: FC<CloudInitCredentialsItemProps> = ({
 
   return (
     <FlexItem spacer={{ default: 'spacerSm' }}>
-      {credentialTitle} {isVisible ? credentials : '●●●●●●●●●'}{' '}
+      {credentialTitle}{' '}
```
rszwajko, line 27: we are already using the flex layout here so there is no need to create the gaps with spaces - we can return from this component several [`FlexItem` components](https://www.patternfly.org/layouts/flex#basic).

src/utils/components/Consoles/components/CloudInitCredentials/CloudInitCredentialsContent.tsx
```diff
@@ -27,38 +29,38 @@ const CloudInitCredentialsContent: FC<CloudInitCredentialsContentProps> = ({ vm
...
-  );
+  const usernameCopyButtons = users.map((user, index) => (
+    <InlineCodeClipboardCopy
+      clipboardText={user?.name?.concat(String.fromCharCode(LINE_FEED))}
+      hideText={true}
+      key={`username-copy-${index}`}
+    />
+  ));
+
+  const passwordCopyButtons = users.map((user, index) => (
```
rszwajko, line 40: `map` looks more readble to me but please test the case with no users.

src/utils/components/Consoles/components/CloudInitCredentials/InlineCodeClipboardCopy.tsx
```diff
@@ -3,21 +3,32 @@ import React, { FC } from 'react';
...
+    navigator.clipboard.writeText(clipboardText);
+  };
+
   return (
     <>
       <ClipboardCopy
         clickTip={t('Copied')}
         hoverTip={t('Copy to clipboard')}
         isCode
-        onCopy={clipboardCopyFunc}
```
rszwajko, line 17: remove the utility function - I think it's not used any more

src/utils/components/Consoles/components/CloudInitCredentials/InlineCodeClipboardCopy.tsx
```diff
@@ -3,21 +3,32 @@ import React, { FC } from 'react';
...
 
-const InlineCodeClipboardCopy: FC<{ clipboardText: string }> = ({ clipboardText }) => {
+const InlineCodeClipboardCopy: FC<InlineCodeClipboardCopyProps> = ({
+  clipboardText,
+  hideText = false,
+}) => {
   const { t } = useKubevirtTranslation();
 
+  const handleCopy = () => {
+    navigator.clipboard.writeText(clipboardText);
```
rszwajko, line 18: the write may fail due to browser restrictions and/or missing features - to avoid unhandled promise rejection errors we can add a catch handler and log some meaningful message

src/utils/components/Consoles/components/CloudInitCredentials/InlineCodeClipboardCopy.tsx
```diff
@@ -3,21 +3,32 @@ import React, { FC } from 'react';
...
 import { ClipboardCopy } from '@patternfly/react-core';
 
-import { clipboardCopyFunc } from '../../utils/utils';
+type InlineCodeClipboardCopyProps = {
+  clipboardText: string;
+  hideText?: boolean;
+};
 
-const InlineCodeClipboardCopy: FC<{ clipboardText: string }> = ({ clipboardText }) => {
+const InlineCodeClipboardCopy: FC<InlineCodeClipboardCopyProps> = ({
```
rszwajko, line 11: it seems we have a PatternFly component with similar scope - https://www.patternfly.org/components/clipboard-copy/#inline-compact-with-additional-action
What features are missing there ?

src/utils/components/Consoles/components/CloudInitCredentials/CloudInitCredentialsItem.tsx
batyana: This is outdated- I squashed commits for clarity. The newer fix has no spaces for gaps.

src/utils/components/Consoles/components/CloudInitCredentials/InlineCodeClipboardCopy.tsx
rszwajko: it seems the utility is still there in `src/utils/components/Consoles/utils/utils.ts`

src/utils/components/Consoles/components/CloudInitCredentials/InlineCodeClipboardCopy.tsx
```diff
@@ -3,23 +3,33 @@ import React, { FC } from 'react';
...
+};
+
+const InlineCodeClipboardCopy: FC<InlineCodeClipboardCopyProps> = ({
+  clipboardText,
+  isCredentialsVisible = false,
+}) => {
   const { t } = useKubevirtTranslation();
 
+  const handleCopy = () => {
+    navigator.clipboard.writeText(clipboardText.concat(String.fromCharCode(LINE_FEED)));
```
rszwajko, line 20: 1. please take a look at error handling in the corresponding  read method https://github.com/kubevirt-ui/kubevirt-plugin/blob/15c10986fa04261be46b8e7b0468ee82e16afa8a/src/utils/components/Consoles/utils/utils.ts#L20
2. the method is small but just to keep read and write co-located I would move it to the same utility file

rszwajko (commented):
Looks good - only 2 small comments

rszwajko (approved):

PR #2813, opened 2025-07-14
fix horizontal navbar when vm called with specific names

upalatucci:
## 📝 Description

If you create a vm with a specific name, tabs will not work properly.

For example: vm is called `console`,it will only works on the console tab.  This because the horizontalnavbar is onlylooking at the end of the string and not stripping out the basepath 

## 🎥 Demo

**BEFORE**

https://github.com/user-attachments/assets/82f80228-cd95-451f-a2a6-3ca81c29e894

**AFTER**

https://github.com/user-attachments/assets/453cc4a7-29d5-49e2-9eb0-6c5618f60736

### DIFF
--- a/src/utils/components/HorizontalNavbar/HorizontalNavbar.tsx
+++ b/src/utils/components/HorizontalNavbar/HorizontalNavbar.tsx
@@ -4,12 +4,15 @@ import classNames from 'classnames';
 import { VirtualMachineModel } from 'src/views/dashboard-extensions/utils';
 
 import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import { getName, getNamespace } from '@kubevirt-utils/resources/shared';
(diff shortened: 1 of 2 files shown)


upalatucci:
@metalice the fix do not work with the wizard. It needs a different base path for that

src/utils/components/HorizontalNavbar/HorizontalNavbar.tsx
```diff
@@ -85,11 +89,15 @@ const HorizontalNavbar: FC<HorizontalNavbarProps> = ({
                 key={item.name}
               >
                 <NavLink
+                  to={
+                    basePath +
+                    trimLastHistoryPath(location.pathname.replace(basePath, ''), paths) +
+                    item.href
+                  }
```
adamviktora, line 96: Is the `trimLastHistoryPath` needed here? It feels like it will always return just `/` after these changes.

PR #2818, opened 2025-07-15
CNV-64782: test for cpu hotplug with networkInterfaceMultiqueue

gouyang:
## 📝 Description

> Add a brief description

## 🎥 Demo


cypress/views/catalog-flow.ts
```diff
@@ -91,6 +91,16 @@ export const fillReviewAndCreate = (vmData: VirtualMachineData) => {
     cy.get(cView.winDrivers).uncheck();
   }
   cy.get(cView.vmName).clear().type(name);
+  if (cpu) {
```
upalatucci, line 94: Where did you get `cpu`? seems like its not defined previosuly

PR #2826, opened 2025-07-21
Fix preferences loading with error

upalatucci:
## 📝 Description

When preference fetch is in error, the page keeps loading instead of showing the error.

This as `loaded` from `useK8sWatchResource` is false when the fetch go into the error state but we keep showing the spinner if not loaded 

note: on upstream we should fetch volums fron the upstream namespace


src/views/catalog/CreateFromInstanceTypes/components/BootableVolumeList/BootableVolumeList.tsx
```diff
@@ -109,7 +109,7 @@ const BootableVolumeList: FC<BootableVolumeListProps> = ({
 
   useEffect(() => {
     if (!isAdmin && volumeListNamespace === ALL_PROJECTS) {
-      setVolumeListNamespace(OPENSHIFT_OS_IMAGES_NS);
+      setVolumeListNamespace(isUpstream ? KUBEVIRT_OS_IMAGES_NS : OPENSHIFT_OS_IMAGES_NS);
```
pcbailey, line 112: This seems like something we should wrap in a hook like `useOSImagesNS` instead of doing it manually in every place that we need it. Thoughts?

PR #2832, opened 2025-07-23
CNV-63859: Changing the default project namespace according to user privilege

batyana:
## 📝 Description

when a non-privileged (test) user logs in and navigates to the VirtualMachines section, the "Create VirtualMachine" button is not visible. Changing the default project namespace according to the user privileges and displaying alert warning if no namespace exists. 

## 🎥 Demo
Before:
image

After:
image

### DIFF
--- a/src/views/virtualmachines/tree/VirtualMachineTreeView.tsx
+++ b/src/views/virtualmachines/tree/VirtualMachineTreeView.tsx
@@ -1,10 +1,13 @@
 import React, { CSSProperties, FC, useMemo } from 'react';
 
 import useIsSmallScreen from '@kubevirt-utils/hooks/useIsSmallScreen';
+import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
 import useLocalStorage from '@kubevirt-utils/hooks/useLocalStorage';
 import { getContentScrollableElement } from '@kubevirt-utils/utils/utils';
 import { OnFilterChange } from '@openshift-console/dynamic-plugin-sdk';
 import {
+  Alert,
+  AlertVariant,
   Drawer,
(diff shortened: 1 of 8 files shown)


src/views/virtualmachines/tree/VirtualMachineTreeView.tsx
```diff
@@ -42,11 +51,33 @@ const VirtualMachineTreeView: FC<VirtualMachineTreeViewProps> = ({
...
 
   const [drawerWidth, setDrawerWidth] = useLocalStorage(TREE_VIEW_LAST_WIDTH, OPEN_DRAWER_SIZE);
   const [drawerOpen, setDrawerOpen] = useLocalStorage(SHOW_TREE_VIEW, SHOW);
 
-  const [selected, onSelect] = useTreeViewSelect(onFilterChange);
+  const [searchParams] = useSearchParams();
+  const navigate = useNavigate();
+  const [, setLastNamespace] = useLastNamespace();
+  const isAdmin = useIsAdmin();
+  const [projectNames, projectNamesLoaded] = useProjects();
```
upalatucci, line 63: also this can be in the hook

src/views/virtualmachines/tree/VirtualMachineTreeView.tsx
```diff
@@ -42,11 +51,33 @@ const VirtualMachineTreeView: FC<VirtualMachineTreeViewProps> = ({
...
+  const [alertMessage, setAlertMessage] = React.useState<string>(null);
+
+  const [selected, onSelect, setSelected] = useTreeViewSelect(onFilterChange);
```
upalatucci, line 66: can we have those states in the hook and return it ? Does it make sense?

PR #2839, opened 2025-07-25
ACM-20341: Fix vm actions with acm

upalatucci:
## 📝 Description

Delete protection was not working properly

VM cloning had a bug during the waiting time

snapshot content was not loaded properly

### DIFF
--- a/src/utils/components/CloneVMModal/components/SnapshotContentConfigurationSummary.tsx
+++ b/src/utils/components/CloneVMModal/components/SnapshotContentConfigurationSummary.tsx
@@ -10,7 +10,9 @@ import {
 import Loading from '@kubevirt-utils/components/Loading/Loading';
 import { DEFAULT_NAMESPACE } from '@kubevirt-utils/constants/constants';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
-import { useK8sWatchResource } from '@openshift-console/dynamic-plugin-sdk';
+import { getNamespace } from '@kubevirt-utils/resources/shared';
+import { getCluster } from '@multicluster/helpers/selectors';
+import useK8sWatchData from '@multicluster/hooks/useK8sWatchData';
 import { Alert, AlertVariant } from '@patternfly/react-core';
 
 import ConfigurationSummary from './ConfigurationSummary';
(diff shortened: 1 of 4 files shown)


src/views/virtualmachines/actions/components/DeleteVMModal/DeleteVMModal.tsx
```diff
@@ -84,9 +84,13 @@ const DeleteVMModal: FC<DeleteVMModalProps> = ({ isOpen, onClose, vm }) => {
     }
 
     if (!location.pathname.endsWith('/search')) {
-      navigate(
-        `/k8s/${lastNamespacePath}/${VirtualMachineModelRef}${location.search}${location.hash}`,
-      );
+      const cluster = getCluster(vm);
```
pcbailey, line 87: Is this logic going to be used anywhere else? If yes, then we should move it to a hook that returns a function that can be called wherever. It would get some of the cluster-specific logic out of this component.

src/utils/components/CloneVMModal/components/SnapshotContentConfigurationSummary.tsx
```diff
@@ -24,14 +25,14 @@ const SnapshotContentConfigurationSummary: FC<SnapshotContentConfigurationSummar
...
-        name: snapshot.status.virtualMachineSnapshotContentName,
-        namespace: snapshot.metadata.namespace || DEFAULT_NAMESPACE,
-      },
-    );
+  const [snapshotContent, loaded, error] = useK8sWatchData<V1beta1VirtualMachineSnapshotContent>(
+    snapshot && {
+      cluster: getCluster(snapshot),
+      groupVersionKind: modelToGroupVersionKind(VirtualMachineSnapshotContentModel),
+      name: snapshot.status.virtualMachineSnapshotContentName,
+      namespace: snapshot.metadata.namespace || DEFAULT_NAMESPACE,
```
pcbailey, line 33: Please use the `getNamespace` selector.

src/views/virtualmachines/actions/components/DeleteVMModal/DeleteVMModal.tsx
upalatucci: No it wil not be used anywhere. We redirect only here. I could even redirect to the all-clusters vm list and don't even use this cluster variable.

PR #2865, opened 2025-08-03
CNV-62538: Hetrogeneous cluster multi-arch

avivtur:
## 📝 Description

Adding suuport for hetrogeneous cluster multi arch
- Bootable volume and Template list pages introduce a new column and filters "Architecture" 
- Bootable volume and Template catalogs introduce a new column and filters "Architecture" 
- Adding multi select checkbox in the add bootable volume modal. If multiple selected, a different DS will be created for each architecture.
- Adding nodeSelector when creating a VM if the template/volume has the architecture label to match the node with the correct architecture

## 🎥 Demo

https://github.com/user-attachments/assets/8fb86f2c-feb6-400e-b529-125867a1e62c


src/views/catalog/CreateFromInstanceTypes/components/BootableVolumeList/hooks/useBootVolumeFilters.ts
```diff
@@ -34,6 +52,20 @@ const useBootVolumeFilters = (isModal: boolean): RowFilter<BootableVolume>[] =>
       reducer: (obj) => isDeprecated(getName(obj)) && SHOW_DEPRECATED_BOOTABLE_VOLUMES_LABEL,
       type: SHOW_DEPRECATED_BOOTABLE_VOLUMES,
     },
+    {
+      filter: (availableArchitectures, obj) => {
+        return (
+          availableArchitectures?.selected?.length === 0 ||
+          availableArchitectures?.selected?.includes(
+            getItemNameWithOther(getArchitecture(obj), OS_NAMES),
```
adamviktora, line 60: These should be `workloadsArchitecturesItems` instead of `OS_NAMES`.

Also it looks like an `includeFilter` helper can be used here (like in `useBootableVolumesFilters.ts`)

adamviktora (commented):
Looking great! 

One comment below + one bug with column management modal:

https://github.com/user-attachments/assets/89b975f0-352e-477b-9b1b-0f345b76f815

src/views/catalog/CreateFromInstanceTypes/components/BootableVolumeList/hooks/useBootVolumeFilters.ts
avivtur: that's a funny miss I made, great catch

avivtur:
/hold
Waiting for backend status

PR #2871, opened 2025-08-06
ACM-20341: make ssh secret modal multicluster

upalatucci:
## 📝 Description

Make the secret modal fetch the cluster

### DIFF
--- a/src/utils/resources/udn/hooks/useNamespaceUDN.ts
+++ b/src/utils/resources/udn/hooks/useNamespaceUDN.ts
@@ -3,6 +3,7 @@ import { useMemo } from 'react';
 import { NetworkAttachmentDefinitionModelGroupVersionKind } from '@kubevirt-utils/models';
 import { UserDefinedNetworkRole } from '@kubevirt-utils/resources/udn/types';
 import { isEmpty } from '@kubevirt-utils/utils/utils';
+import useClusterParam from '@multicluster/hooks/useClusterParam';
 import useK8sWatchData from '@multicluster/hooks/useK8sWatchData';
 import { NetworkAttachmentDefinitionKind } from '@overview/OverviewTab/inventory-card/utils/types';
 
@@ -10,12 +11,13 @@ import { LAYER3_TOPOLOGY, PrimaryTopologies } from '../constants';
 
 const useNamespaceUDN = (
   namespace: string,
-  cluster?: string,
 ): [
   isNamespaceManagedByUDN: boolean,
   vmsNotSupported: boolean,
   nad?: NetworkAttachmentDefinitionKind,
 ] => {
+  const cluster = useClusterParam();
+
   const [nads] = useK8sWatchData<NetworkAttachmentDefinitionKind[]>({
     cluster,
     groupVersionKind: NetworkAttachmentDefinitionModelGroupVersionKind,
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "@patternfly/react-tokens": "^6.2.2",
     "@patternfly/react-topology": "^6.2.0",
     "@preact/signals-react": "^2.0.1",
-    "@stolostron/multicluster-sdk": "0.5.0",
+    "@stolostron/multicluster-sdk": "0.5.1",
     "@xterm/addon-clipboard": "^0.1.0",
     "@xterm/addon-fit": "^0.10.0",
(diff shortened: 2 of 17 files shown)


avivtur (approved):

src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/TemplatesCatalogDrawerCreateForm.tsx
```diff
@@ -46,9 +47,10 @@ type TemplatesCatalogDrawerCreateFormProps = {
 export const TemplatesCatalogDrawerCreateForm: FC<TemplatesCatalogDrawerCreateFormProps> = memo(
   ({ authorizedSSHKey, namespace, onCancel, subscriptionData }) => {
     const { t } = useKubevirtTranslation();
+    const cluster = useClusterParam();
 
     const { isBootSourceAvailable, templateLoadingError } = useDrawerContext();
-    const [__, vmsNotSupported] = useNamespaceUDN(namespace);
+    const [__, vmsNotSupported] = useNamespaceUDN(namespace, cluster);
```
pcbailey, line 53: Why are we passing the cluster into hooks instead of calling `useClusterParam` inside the hooks directly? We should be minimizing instances of passing the cluster around as much as possible, IMO.

src/views/catalog/CreateFromInstanceTypes/components/VMDetailsSection/components/DetailsRightGrid.tsx
```diff
@@ -24,6 +25,7 @@ import './details-right-grid.scss';
 const DetailsRightGrid: FC = () => {
   const { t } = useKubevirtTranslation();
   const { createModal } = useModal();
+  const cluster = useClusterParam();
```
pcbailey, line 28: Why not call this directly in `SSHSecretModal`? Is it possible for the VM cluster to be different from the cluster in the params?

pcbailey: 😆

src/views/catalog/CreateFromInstanceTypes/components/VMDetailsSection/components/DetailsRightGrid.tsx
```diff
@@ -24,8 +25,9 @@ import './details-right-grid.scss';
...
   const [{ clusterDefaultStorageClass, sortedStorageClasses, virtDefaultStorageClass }, loaded] =
-    useDefaultStorageClass();
+    useDefaultStorageClass(cluster);
```
pcbailey, line 30: Can you update this hook to get the cluster itself as well?

src/utils/resources/udn/hooks/useNamespaceUDN.ts
```diff
@@ -3,19 +3,21 @@ import { useMemo } from 'react';
...
 
 const useNamespaceUDN = (
   namespace: string,
-  cluster?: string,
 ): [
   isNamespaceManagedByUDN: boolean,
   vmsNotSupported: boolean,
   nad?: NetworkAttachmentDefinitionKind,
 ] => {
+  const cluster = useClusterParam();
```
pcbailey, line 19: It doesn't look like this is being used anywhere, unless I'm missing it.

pcbailey: Ah, I missed that. Sorry about that.

src/views/catalog/CreateFromInstanceTypes/components/VMDetailsSection/components/DetailsRightGrid.tsx
pcbailey: I looked at the places where `useDefaultStorageClass` is used. It looks like we pass the VM cluster in one place. Maybe the approach we should take with hooks is to pull the cluster inside each hook that needs it, but take an optional `cluster` parameter that overrides that value if a cluster is passed in. That way, we minimize the places where we're passing the cluster value around while making hooks friendly for lists to use. This way will further decouple the cluster logic from the rest of the code base.

upalatucci: done it

PR #2906, opened 2025-08-19
CNV-63859: Fixing displayed errors and namespace selection in virtualization views for non privileged users

batyana:
## 📝 Description

making sure in all virtualization views if the user is non privileged "all-projects" is hidden and system namespaces are displayed

## 🎥 Demo
Before:
https://github.com/user-attachments/assets/110c5d59-dcb6-48e7-83c9-5b1cbc8da4ae

After:
Uploading Screen Recording 2025-08-19 at 17.31.32.mov…

### DIFF
--- a/src/utils/hooks/useForceProjectSelection.ts
+++ b/src/utils/hooks/useForceProjectSelection.ts
@@ -1,0 +1,49 @@
+import { useEffect } from 'react';
+import { useLocation, useNavigate } from 'react-router-dom-v5-compat';
+
+import { DEFAULT_NAMESPACE } from '@kubevirt-utils/constants/constants';
+import { useIsAdmin } from '@kubevirt-utils/hooks/useIsAdmin';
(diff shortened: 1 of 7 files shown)


src/utils/hooks/useForceProjectSelection.ts
```diff
@@ -0,0 +1,49 @@
...
+import { VIRTUALIZATION_PATHS } from '@virtualmachines/tree/utils/constants';
+
+import useProjects from './useProjects';
+
+type UseForceProjectSelectionOptions = {
+  enabled?: boolean;
+  pathConditions?: string[];
+};
+
+export const useForceProjectSelection = (options: UseForceProjectSelectionOptions = {}) => {
```
upalatucci, line 17: with only two parameters, instead of using the object i would rather use two distinct parameters pathConditions and enable that can also have default values `[VIRTUALIZATION_PATHS.BASE]` and `true`

lkladnit (approved):

PR #2930, opened 2025-08-27
CNV-52340: better descriptive help text for configuring IP adress in cloudinit

batyana:
## 📝 Description

Changing the help text to a pop over to allow better detail and description for configuring IP addresses in cloudinit

## 🎥 Demo
Before:
image

After:
image


src/utils/components/CloudinitModal/CloudInitNetworkForm.tsx
```diff
@@ -54,6 +53,20 @@ export const CloudinitNetworkForm: React.FC<CloudinitNetworkFormProps> = ({
             />
           </FormGroup>
           <FormGroup
+            labelHelp={
+              <HelpTextIcon
+                bodyContent={
+                  <div>
```
pcbailey, line 59: This would look cleaner as a `<Trans>` block. You can use a `<br />` tag in between the lines. Let me know if you need help.

src/utils/components/CloudinitModal/CloudInitNetworkForm.tsx
```diff
@@ -54,6 +54,19 @@ export const CloudinitNetworkForm: React.FC<CloudinitNetworkFormProps> = ({
             />
           </FormGroup>
           <FormGroup
+            labelHelp={
+              <HelpTextIcon
+                bodyContent={
+                  <Trans>
```
pcbailey, line 60: The only thing you're missing is telling the `Trans` component which namespace to use. This can be done by either setting `ns="plugin__kubevirt-plugin"` or `t={t}` on the `Trans` component. 

We're very inconsistent on which to use. You'll see both used in some places. From what I can tell the namespace is enough, so just set `ns="plugin__kubevirt-plugin"`.

PR #2936, opened 2025-08-28
Fix web console standalone url

upalatucci:
## 📝 Description

Make sure we follow the new url also for the standalone console. Now its broken from when we changed the url

### DIFF
--- a/src/multicluster/urls.ts
+++ b/src/multicluster/urls.ts
@@ -4,22 +4,24 @@ import { ResourceRouteHandler } from '@stolostron/multicluster-sdk';
 
 import { VirtualMachineModel } from '../views/dashboard-extensions/utils';
 
+import { KUBEVIRT_VM_PATH } from './constants';
+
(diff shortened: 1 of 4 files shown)


src/multicluster/urls.ts
```diff
@@ -29,6 +29,11 @@ export const getCatalogURL = (cluster: string, namespace?: string): string => {
     : `/k8s/${namespacePath}/catalog`;
 };
```
galkremer1, line 31: If cluster can be undefined maybe we can make it optional and also reuse the common path:
`
export const getConsoleStandaloneURL = (namespace: string, name: string, cluster?: string): string => {
  const commonPath = `/ns/${namespace}/kubevirt.io~v1~VirtualMachine/${name}/console/standalone`;

  const clusterPath = cluster ? `/cluster/${cluster}` : '';
  return `/k8s${clusterPath}${commonPath}`;
};
`

PR #2943, opened 2025-08-29
CNV-68347: Force re-creating NetworkSelect on network pre-selection

rszwajko:
## 📝 Description

Depends on: #2933 
Follow-up to: #2919

### Before

https://github.com/user-attachments/assets/e25e323b-a8e7-4c39-b2d7-22407d265eba

### After
https://github.com/user-attachments/assets/3d7f9d8b-2051-4bad-ae98-0c9b1022c288

### DIFF
--- a/src/utils/components/NetworkInterfaceModal/components/NetworkInterfaceNetworkSelect/NetworkInterfaceNetworkSelect.tsx
+++ b/src/utils/components/NetworkInterfaceModal/components/NetworkInterfaceNetworkSelect/NetworkInterfaceNetworkSelect.tsx
@@ -54,6 +54,7 @@ const NetworkInterfaceNetworkSelect: FC<NetworkInterfaceNetworkSelectProps> = ({
   const { t } = useKubevirtTranslation();
   const vmiNamespace = vm?.metadata?.namespace || namespace;
   const { loaded, loadError, nads } = useNADsData(vmiNamespace);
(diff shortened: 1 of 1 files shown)


upalatucci (approved):

src/utils/components/NetworkInterfaceModal/components/NetworkInterfaceNetworkSelect/NetworkInterfaceNetworkSelect.tsx
```diff
@@ -54,6 +54,7 @@ const NetworkInterfaceNetworkSelect: FC<NetworkInterfaceNetworkSelectProps> = ({
   const { t } = useKubevirtTranslation();
   const vmiNamespace = vm?.metadata?.namespace || namespace;
   const { loaded, loadError, nads } = useNADsData(vmiNamespace);
+  const [selectedFirstOnLoad, setSelectedFirstOnLoad] = useState(false);
```
upalatucci, line 57: Can this be a ref? So we'll not trigger a re-rendering? and would not go into the useEffect dependencies

rszwajko: IMHO we need a re-render to make sure  we re-create the network select component - but I can check if `handleChange` guarantees that.

PR #2944, opened 2025-09-02
CNV-48181: fix CPU and memory metrics

upalatucci:
## 📝 Description

For CPU requested metric we don't have to fetch it form prometheus using `kube_pod_resource_request` 
We need to take it from the VMI. `kube_pod_resource_request` is not directly related to the VMI CPU requested, but it's scaled using HCO configuration. 

For memory we ofter use `kubevirt_vmi_memory_available_bytes - kubevirt_vmi_memory_usable_bytes` but the right metric is `kubevirt_vmi_memory_used_bytes` 

Screenshot 2025-09-04 at 10 05 52

### DIFF
--- a/src/utils/components/Charts/CPUUtil/CPUThresholdChart.tsx
+++ b/src/utils/components/Charts/CPUUtil/CPUThresholdChart.tsx
@@ -1,14 +1,13 @@
-import React, { FC, useMemo } from 'react';
+import React, { FC } from 'react';
 import { Link } from 'react-router-dom-v5-compat';
 
 import { V1VirtualMachineInstance } from '@kubevirt-ui/kubevirt-api/kubevirt';
 import useVMQueries from '@kubevirt-utils/hooks/useVMQueries';
 import { getNamespace } from '@kubevirt-utils/resources/shared';
-import { getVMIPod } from '@kubevirt-utils/resources/vmi';
-import { humanizeCpuCores } from '@kubevirt-utils/utils/humanize.js';
+import { getCPU, getVCPUCount } from '@kubevirt-utils/resources/vm';
 import { isEmpty } from '@kubevirt-utils/utils/utils';
 import { getCluster } from '@multicluster/helpers/selectors';
-import { K8sResourceCommon, PrometheusEndpoint } from '@openshift-console/dynamic-plugin-sdk';
+import { PrometheusEndpoint } from '@openshift-console/dynamic-plugin-sdk';
 import {
   Chart,
   ChartArea,
@@ -28,7 +27,6 @@ import ComponentReady from '../ComponentReady/ComponentReady';
 import useResponsiveCharts from '../hooks/useResponsiveCharts';
 import {
   addTimestampToTooltip,
-  findMaxYValue,
   formatCPUUtilTooltipData,
   MILLISECONDS_MULTIPLIER,
   queriesToLink,
@@ -37,16 +35,14 @@ import {
 } from '../utils/utils';
 
 type CPUThresholdChartProps = {
-  pods: K8sResourceCommon[];
   vmi: V1VirtualMachineInstance;
 };
 
-const CPUThresholdChart: FC<CPUThresholdChartProps> = ({ pods, vmi }) => {
-  const vmiPod = useMemo(() => getVMIPod(vmi, pods), [pods, vmi]);
+const CPUThresholdChart: FC<CPUThresholdChartProps> = ({ vmi }) => {
   const { currentTime, duration, timespan } = useDuration();
   const { height, ref, width } = useResponsiveCharts();
 
-  const queries = useVMQueries(vmi, vmiPod?.metadata?.name);
+  const queries = useVMQueries(vmi);
 
   const prometheusProps = {
     cluster: getCluster(vmi),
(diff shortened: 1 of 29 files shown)


src/utils/components/Charts/CPUUtil/CPUThresholdChart.tsx
```diff
@@ -79,17 +71,17 @@ const CPUThresholdChart: FC<CPUThresholdChartProps> = ({ pods, vmi }) => {
...
-    return {
+  const thresholdData = [
+    {
       name: 'CPU requested',
-      x: new Date(x * MILLISECONDS_MULTIPLIER),
-      y: humanizeCpuCores(Number(y)).value,
-    };
-  });
+      x: new Date(currentTime),
+      y: cpuRequested * 1000,
```
rszwajko, line 78: Do we need this multiplication? `humanizeCpuCores` did not multiply by 1000 if the value was above 1.

src/utils/resources/vm/utils/utils.ts
```diff
@@ -18,3 +19,6 @@ export const getVMListPathWithRowFilters = (
   namespace: string,
   rowFilters: Record<string, string>,
 ) => getVMListPath(namespace, getRowFiltersString(rowFilters));
+
+export const getVCPUCount = (cpu: V1CPU) =>
+  (cpu.sockets || 1) * (cpu.cores || 1) * (cpu.threads || 1);
```
rszwajko, line 24: the method receives (most of the time) the output from `getCpu(vmi)` which may be undefined. Some cases are guarded with `isEmpty` or equivalent but some are not. We need either make this method return reasonable default data for undefined or verify each usage.

rszwajko (changes requested):
There are some cases where the conversion of CPU values may fail.

src/utils/components/Charts/CPUUtil/CPUThresholdChart.tsx
upalatucci: If CPU usage is in `m`, we should also write in the chart in `m`

upalatucci: I think here we should always show in m the chart . After 999 m, the usage would decome 1 and not 1000. This will lead to some issues in the chart

src/views/virtualmachines/details/tabs/overview/components/VirtualMachinesOverviewTabUtilization/components/CPUUtil/CPUUtil.tsx
```diff
@@ -70,7 +72,7 @@ const CPUUtil: FC<CPUUtilProps> = ({ pods, vmi }) => {
             }}
             animate
             constrainToVisibleArea
-            labels={({ datum }) => (datum.x ? `${datum.x}: ${(cpuUsage || 0)?.toFixed(2)}m` : null)}
+            labels={({ datum }) => (datum.x ? `${datum.x}: ${cpuUsageHumanized?.string}` : null)}
```
rszwajko, line 75: if we mix humanized and raw values then we need to handle milli-cores

src/views/virtualmachines/details/tabs/overview/components/VirtualMachinesOverviewTabUtilization/components/CPUUtil/CPUUtil.tsx
```diff
@@ -34,18 +32,19 @@ const CPUUtil: FC<CPUUtilProps> = ({ pods, vmi }) => {
...
   const [dataCPUUsage] = useFleetPrometheusPoll({
     ...prometheusProps,
     query: queries?.CPU_USAGE,
   });
 
-  const cpuUsage = humanizeCpuCores(+dataCPUUsage?.data?.result?.[0]?.value?.[1]).value;
-  const cpuRequested = humanizeCpuCores(+dataCPURequested?.data?.result?.[0]?.value?.[1]).value;
+  const vmCPU = getCPU(vmi);
+
+  const cpuUsage = +dataCPUUsage?.data?.result?.[0]?.value?.[1];
```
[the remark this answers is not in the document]
upalatucci: okay but not for `getCPU`. Just for  `+dataCPUUsage?.data?.result?.[0]?.value?.[1]` right ?

src/utils/components/Charts/CPUUtil/CPUThresholdChart.tsx
```diff
@@ -118,7 +113,7 @@ const CPUThresholdChart: FC<CPUThresholdChartProps> = ({ pods, vmi }) => {
               }}
               dependentAxis
               tickFormat={(tick: number) => tick?.toFixed(2)}
-              tickValues={[0, yMax]}
+              tickValues={[0, cpuRequested * 1000]}
```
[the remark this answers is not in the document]
upalatucci: cool thank you

rszwajko (approved):
Looks good!

src/views/virtualmachines/details/tabs/overview/components/VirtualMachinesOverviewTabUtilization/components/CPUUtil/CPUUtil.tsx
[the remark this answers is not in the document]
rszwajko: yes. Probably we shouldn't add fallback to `getCPU` - I see now we do have fallback in the selectors for properties inside V1CPU i.e. `getCPUCores`

PR #2955, opened 2025-09-03
CNV-68329: Enable focus after pasting values to console

galkremer1:
## 📝 Description

After pasting values to console we automatically focus on the console, improving the user experience. 
Previously before this enhancement the user had to click on the console in order to gain focus and be able to type inside.

## 🎥 Demo

https://github.com/user-attachments/assets/98d7b597-f033-417d-8fc7-6dec5432a1a1

### DIFF
--- a/src/utils/components/Consoles/components/AccessConsoles/AccessConsoles.tsx
+++ b/src/utils/components/Consoles/components/AccessConsoles/AccessConsoles.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/no-shadow */
-import React, { FC, useState } from 'react';
+import React, { FC, MouseEvent, useState } from 'react';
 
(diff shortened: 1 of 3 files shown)


src/utils/components/Consoles/components/AccessConsoles/AccessConsoles.tsx
```diff
@@ -59,8 +58,11 @@ export const AccessConsoles: FC<AccessConsolesProps> = ({
             <PasteIcon /> {t('Paste to console')}
           </>
         }
+        onClick={(e: MouseEvent<HTMLButtonElement>) => {
+          actions.sendPaste(true);
+          e.currentTarget.blur();
```
galkremer1, line 63: Focus was previously set in the paste button, so if the user pressed enter it would send the paste command again to the console. While we are setting the focus to the console I am still explicitly removing the focus from the button itself.

upalatucci (approved):

PR #2967, opened 2025-09-09
CNV-69002: make sure catalog works with all namespaces selected

upalatucci:
## 📝 Description

`useActiveNamespace` have been replaced by `useNamespaceParam` in a previous pr to support the namespace when the catalog page is in a multicluster mode. In the multicluster, we can't use `useActiveNamespace`.  [HERE](https://github.com/kubevirt-ui/kubevirt-plugin/pull/2935)

The multicluster catalog do not follow the console active namespace so active namespace can be different from what the user select.

To have a valid workaround for the future, i created in this pr a hook that returns the namespace if it's selected and  `ALL_NAMESPACES_SESSION_KEY` when the all-namespace is selected. Using the same name so it will be easier to replace in other pages

## 🎥 Demo

**BUG**

https://github.com/user-attachments/assets/2904d561-85c3-41bc-af13-ae6dc623cee6

### DIFF
--- a/src/utils/components/SSHSecretModal/components/SSHKeyUpload/SSHKeyUpload.tsx
+++ b/src/utils/components/SSHSecretModal/components/SSHKeyUpload/SSHKeyUpload.tsx
@@ -2,8 +2,8 @@ import React, { Dispatch, FC, SetStateAction, useState } from 'react';
 
 import { IoK8sApiCoreV1Secret } from '@kubevirt-ui/kubevirt-api/kubernetes';
 import FormGroupHelperText from '@kubevirt-utils/components/FormGroupHelperText/FormGroupHelperText';
+import useActiveNamespace from '@kubevirt-utils/hooks/useActiveNamespace';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
-import useNamespaceParam from '@kubevirt-utils/hooks/useNamespaceParam';
 import { getValidNamespace, validateSSHPublicKey } from '@kubevirt-utils/utils/utils';
 import {
   FileUpload,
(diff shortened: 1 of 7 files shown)


adamviktora (approved):

src/utils/resources/vm/utils/utils.ts
```diff
@@ -1,10 +1,14 @@
 import { V1CPU } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import { ALL_NAMESPACES, ALL_NAMESPACES_SESSION_KEY } from '@kubevirt-utils/hooks/constants';
 import { isEmpty } from '@kubevirt-utils/utils/utils';
 import { getVMListNamespacesURL, getVMListURL } from '@multicluster/urls';
 import { getRowFilterQueryKey } from '@search/utils/query';
 
 export const getVMListPath = (namespace: string, params: string, cluster?: string) => {
-  const vmListURL = namespace ? getVMListNamespacesURL(cluster, namespace) : getVMListURL(cluster);
+  const vmListURL =
+    namespace && namespace !== ALL_NAMESPACES_SESSION_KEY && namespace !== ALL_NAMESPACES
```
pcbailey, line 9: We should be able to replace this with the `isAllNamespaces` utility function.

src/utils/components/SSHSecretModal/components/SSHKeyUpload/SSHKeyUpload.tsx
```diff
@@ -43,7 +43,7 @@ const SSHKeyUpload: FC<SSHKeyUploadProps> = ({ secrets, setSSHDetails, sshDetail
       ...sshDetails,
       secretOption: SecretSelectionOption.addNew,
       sshPubKey: sshPublicKey?.trim(),
-      sshSecretNamespace: activeNamespace,
+      sshSecretNamespace: vmNamespaceTarget,
```
galkremer1, line 46: Why is activeNamespace is replaced with vmNamespaceTarget?

PR #2974, opened 2025-09-10
CNV-67097: Fix bootable volume creation from a registry with credentials

pcbailey:
## 📝 Description

This PR fixes a bug that occurs when a bootable volume is created from a registry with username and password via the "Add bootable volume" modal that is accessed by clicking the "Add Volume" button on the "Bootable volumes" list page. The import of some volume types fails with a "Secret not found" error.

This occurs because the Secret is needed in both the `openshift-cnv` namespace and the target namespace in which the bootable volume is being created if there is no ImageStream for the container type. When there is no ImageStream a pod is created in the `openshift-cnv` namespace to determine the digest of the container and that pod requires access to the Secret.

This PR creates the Secret in both namespaces to ensure images are able to import successfully. It also strips `http://` and `https://` from the beginning of registry URLs to prevent errors.

Jira: https://issues.redhat.com/browse/CNV-67097

### DIFF
--- a/src/utils/components/AddBootableVolumeModal/utils/utils.ts
+++ b/src/utils/components/AddBootableVolumeModal/utils/utils.ts
@@ -9,6 +9,7 @@ import {
   V1beta1DataSource,
 } from '@kubevirt-ui/kubevirt-api/containerized-data-importer/models';
 import { V1beta1DataVolumeSource } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import { OPENSHIFT_CNV } from '@kubevirt-utils/constants/constants';
 import { UploadDataProps } from '@kubevirt-utils/hooks/useCDIUpload/useCDIUpload';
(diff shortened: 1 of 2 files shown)


src/utils/components/AddBootableVolumeModal/utils/utils.ts
```diff
@@ -333,7 +338,14 @@ export const createDataSourceWithImportCron: CreateDataSourceWithImportCronType
 
   if (addRegistrySecret) {
```
galkremer1, line 339: I see that we have a similar functionality in `src/views/catalog/templatescatalog/components/TemplatesCatalogDrawer/hooks/useCreateDrawerForm.tsx`
Do we need to update it there as well?

galkremer1: Also in `src/views/catalog/utils/useWizardVMCreate.ts`

PR #2975, opened 2025-09-10
CNV-68705: Add deletion protection option when customizing a VM from catalog

galkremer1:
## 📝 Description
[CNV-68705](https://issues.redhat.com/browse/CNV-68705)
Added the option to configure delete protection to a VM when creating and customizing it from catalog. 

## 🎥 Demo

https://github.com/user-attachments/assets/eccef917-c541-42b9-8e57-a1ce230aab54


src/views/catalog/wizard/tabs/overview/components/WizardOverviewGrid.tsx
```diff
@@ -362,6 +368,45 @@ const WizardOverviewGrid: FC<WizardOverviewGridProps> = ({ tabsData, updateVM, v
             testId="guest-system-log-access"
             title={t('Guest system log access')}
           />
```
galkremer1, line 370: This component is getting a bit too big - we probably want to break it down to smaller components later.

PR #2979, opened 2025-09-12
CNV-69370: Disable hot-unplug for Pod Networks

rszwajko:
## 📝 Description

Fix disabling hot-unplug for Pod Network  - drop assumption that the Pod Network has the first interface and detect only based on the `pod` property.

### Before
https://github.com/user-attachments/assets/8aa1bad3-b2d8-48eb-9377-c7283cb3386c

### After
https://github.com/user-attachments/assets/53b91c32-26da-4bd5-876c-00ac7c586cf3

### DIFF
--- a/src/utils/resources/vm/utils/network/selectors.ts
+++ b/src/utils/resources/vm/utils/network/selectors.ts
@@ -1,4 +1,4 @@
-import { V1Interface, V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import { V1Interface, V1Network, V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
 import { getAutoAttachPodInterface, getInterfaces } from '@kubevirt-utils/resources/vm';
 import { isNetworkInterfaceState } from '@kubevirt-utils/utils/typeGuards';
 
@@ -68,3 +68,6 @@ export const getConfigInterfaceState = (
 
   return isNetworkInterfaceState(ifaceState) ? ifaceState : NetworkInterfaceState.UP;
 };
+
+export const isPodNetwork = (network: Partial<V1Network>): boolean =>
+  Boolean(network?.pod) && typeof network.pod === 'object';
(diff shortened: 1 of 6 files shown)


src/utils/resources/vm/utils/network/selectors.ts
```diff
@@ -68,3 +68,6 @@ export const getConfigInterfaceState = (
 
   return isNetworkInterfaceState(ifaceState) ? ifaceState : NetworkInterfaceState.UP;
 };
+
+export const isPodNetwork = (network: Partial<V1Network>): boolean =>
+  Boolean(network?.pod) && typeof network.pod === 'object';
```
galkremer1, line 73: Can't we just use `!isEmpty(network?.pod)`?

galkremer1:
```suggestion
  export const isPodNetwork = (network: Partial<V1Network>): boolean =>  !isEmpty(network?.pod);
```

rszwajko: @galkremer1 
We cannot if we want to have a strict check - `isEmpty` returns true for various data types other then object.
Note also that it doesn't cover all the cases - compare with https://github.com/lodash/lodash/blob/ddfd9b11a0126db2302cb70ec9973b66baec0975/lodash.js#L11479

src/views/catalog/wizard/tabs/network/components/list/NetworkInterfaceActions.tsx
```diff
@@ -63,8 +66,11 @@ const NetworkInterfaceActions: FC<NetworkInterfaceActionsProps> = ({
       const vmInterfaces = getInterfaces(draftVM);
 
       const isExistingNetwork = getNetworks(draftVM)?.find(({ name }) => name === nicName);
-      const isPodNetwork = nicPresentation?.network?.pod;
-      if (!isExistingNetwork && hasAutoAttachedPodNetwork(draftVM) && isPodNetwork) {
+      if (
+        !isExistingNetwork &&
+        hasAutoAttachedPodNetwork(draftVM) &&
+        isPodNetwork(nicPresentation?.network)
+      ) {
```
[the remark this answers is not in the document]
rszwajko: It seems the only screen where we add fake NICs to the table to visualize an auto-attached Pod Network.
I would wait with adding a helper until there will be at least one other place that needs it.

src/utils/resources/vm/utils/network/selectors.ts
pcbailey: We already have lodash installed as a dependency, so if you think lodash.isEmpty works better here you can use it.

orelmisan (commented):
Thanks for the PR, @rszwajko!

Could you please clarify if the UI prevents a user from hot-unplugging an interface connected to the pod network? The KubeVirt backend explicitly blocks this action [1], so we should ensure the UI aligns with that constraint.
[1] https://github.com/kubevirt/kubevirt/blob/e36bb0bd799764901e5dade8e4b2a5e906230d15/pkg/network/admitter/admit.go#L64

PR #2981, opened 2025-09-12
CNV-69370: Hot-unplug NIC in stopped state

rszwajko:
## 📝 Description
Depends on: #2979

Before, the UI used 2 different approaches for deleting a hot-unplug
capable NIC:
1. for running VM - hot-unplug via setting state to "absent"
2. for stopped VM - standard way via deleting sections in config

On a heavily used cluster the VM state was sometimes detected as stopped
too early which resulted in not using hot-unplug.
    
After this PR, only the hot-unplug approach will be used.

### DIFF
--- a/src/utils/components/NetworkInterfaceModal/utils/helpers.tsx
+++ b/src/utils/components/NetworkInterfaceModal/utils/helpers.tsx
@@ -27,7 +27,6 @@ import {
 import { NetworkInterfaceState } from '@kubevirt-utils/resources/vm/utils/network/types';
 import { kubevirtConsole } from '@kubevirt-utils/utils/utils';
 import { ABSENT } from '@virtualmachines/details/tabs/configuration/network/utils/constants';
-import { isStopped } from '@virtualmachines/utils';
 
 import { NetworkAttachmentDefinition } from '../components/hooks/types';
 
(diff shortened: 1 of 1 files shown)


orelmisan (commented):
Thank you for the PR, @rszwajko!

For stopped VMs, I suggest we directly remove the interface and its matching network from the spec. This approach seems better for two key reasons:
1. It's more reliable. It avoids the constraints of hot-unplug, which is only supported for `virtio` interfaces using `bridge` binding and connected to secondary networks.

~~2. It's cleaner. IIRC, setting the interface state to `absent` would cause the interface and network to linger until the VM restarts. A direct removal is more immediate.~~
Edit: We have fixed that.

Please let me know what you think.

rszwajko:
@orelmisan 
>It avoids the constraints of hot-unplug, which is only supported for virtio interfaces using bridge binding and connected to secondary networks.
>
We already test if an interface supports hot-unplug and it won't change after the PR. The improvement is less ways of deleting an interface and dropping time-dependent condition that sometimes may deliver false results (i.e. VM is already running but the UI sees stale state). 

Currently we have 3 ways to delete an interface/network:
1. hot-unplug capable interface
     1. running VM - set `absent`
     2. other  state  - delete sections from config
2. other interface - delete sections from config

After this PR we would have 2 paths:
1. hot-unplug capable interface - set `absent`
2. other interface - delete sections from config

If you still think it's safer to keep the current flow (3 paths) then we can close this PR - the cost of false VM state detection is not very high (user will see RestartRequired condition).

orelmisan:
@rszwajko, I would suggest framing the problem from the VM state perspective:
When the VM is off, we can remove any interface and its matching network.
When the VM is running, we can only hot-unplug certain interfaces (and their matching networks).

PR #2984, opened 2025-09-15
CNV-68471: Fixing installation CD-ROM with registry image to allow eject action

batyana:
## 📝 Description

fixing logic to correctly identify empty container disk image,
allowing the eject action on an installation cd-rom with registry image. 

## 🎥 Demo
Before:
image

After:
image

### DIFF
--- a/src/utils/resources/vm/utils/disk/utils.ts
+++ b/src/utils/resources/vm/utils/disk/utils.ts
@@ -8,5 +8,7 @@ export const isEmptyContainerDiskImage = (volume: V1Volume): boolean => {
   const image = getContainerDiskImage(volume);
   if (isEmpty(image)) return true;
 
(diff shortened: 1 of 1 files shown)


src/utils/resources/vm/utils/disk/utils.ts
```diff
@@ -8,5 +8,7 @@ export const isEmptyContainerDiskImage = (volume: V1Volume): boolean => {
   const image = getContainerDiskImage(volume);
   if (isEmpty(image)) return true;
 
-  return Object.values(EMPTY_DISK_IMAGE_PATTERNS).some((pattern) => image.includes(pattern));
```
galkremer1, line 11: We should be able to iterate over the array only once using:
`return Object.values(EMPTY_DISK_IMAGE_PATTERNS).some(
  (pattern) => pattern !== '' && image.includes(pattern)
);`

PR #2987, opened 2025-09-16
CNV-69441: infinite loading on vm with preference but no instancetype

upalatucci:
## 📝 Description

VMs with preference but no instancetypes are having an infinite loading issue.

The fix is to make sure the code waits for the expanded spec VM instead of searching for the instancetype vm (that its `undefined`)

In case we don't have an expanded spec VM because of an error or something, the Horizontal NavBar will present an error
instead of the details tab 

Note: would like in the future to rename all the `instancetypeVM` to `expandedSpecVM` as the vm can also be expanded but not connected to an instancetype. I would not do that now as i can disrupt a release 
 

## 🎥 Demo
**BEFORE**
https://github.com/user-attachments/assets/82424837-0b32-4dd4-8b45-fd9e3e35eefb

**AFTER**
Screenshot 2025-09-16 at 11 04 22

### DIFF
--- a/src/utils/components/PendingChanges/hooks/usePendingChanges.tsx
+++ b/src/utils/components/PendingChanges/hooks/usePendingChanges.tsx
@@ -31,7 +31,7 @@ import {
 import useHyperConvergeConfiguration from '@kubevirt-utils/hooks/useHyperConvergeConfiguration';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
 import useKubevirtUserSettings from '@kubevirt-utils/hooks/useKubevirtUserSettings/useKubevirtUserSettings';
-import { isInstanceTypeVM } from '@kubevirt-utils/resources/instancetype/helper';
+import { isExpandableSpecVM } from '@kubevirt-utils/resources/instancetype/helper';
 import { getCPU, getGPUDevices, getHostDevices } from '@kubevirt-utils/resources/vm';
 import { DESCHEDULER_EVICT_LABEL } from '@kubevirt-utils/resources/vmi';
 import { isEmpty } from '@kubevirt-utils/utils/utils';
@@ -152,7 +152,7 @@ export const usePendingChanges = (
           <CPUMemoryModal isOpen={isOpen} onClose={onClose} onSubmit={onSubmit} vm={vm} />
         )),
       ),
-      hasPendingChange: !isInstanceTypeVM(vm) && cpuMemoryChanged && restartRequired(vm),
+      hasPendingChange: !isExpandableSpecVM(vm) && cpuMemoryChanged && restartRequired(vm),
       label: t('CPU | Memory'),
     },
     {
@@ -166,7 +166,7 @@ export const usePendingChanges = (
           />
         )),
       ),
-      hasPendingChange: isInstanceTypeVM(vm) && instanceTypeChanged && restartRequired(vm),
+      hasPendingChange: isExpandableSpecVM(vm) && instanceTypeChanged && restartRequired(vm),
       label: t('InstanceType'),
     },
     {
--- a/src/views/virtualmachines/details/VirtualMachineNavPage.tsx
+++ b/src/views/virtualmachines/details/VirtualMachineNavPage.tsx
@@ -46,7 +46,8 @@ const VirtualMachineNavPage: FC = () => {
 
   const vmToShow = useMemo(() => (runningTourSignal.value ? tourGuideVM : vm), [vm]);
 
-  const [instanceTypeExpandedSpec, expandedSpecLoading] = useInstanceTypeExpandSpec(vmToShow);
+  const [instanceTypeExpandedSpec, expandedSpecLoading, expandedSpecError] =
+    useInstanceTypeExpandSpec(vmToShow);
 
   const pages = useVirtualMachineTabs();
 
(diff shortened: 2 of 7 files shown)


src/views/virtualmachines/details/tabs/configuration/details/DetailsSection.tsx
```diff
@@ -96,7 +96,7 @@ const DetailsSection: FC<DetailsSectionProps> = ({ allInstanceTypes, instanceTyp
   const isInstanceType = isInstanceTypeVM(vm);
   const deletionProtectionEnabled = isDeletionProtectionEnabled(vm);
 
-  const loadingInstanceType = isInstanceType && isEmpty(instanceType);
+  const loadingInstanceType = isInstanceType && isEmpty(instanceTypeVM);
```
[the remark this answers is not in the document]
upalatucci: `isEmpty(instanceType)` will be empty in this case, and we'll have infinite loading again. 
`isPreferenceType` would return true, but we should wait for the `expand-spec` api to respond. 
So here we are loading and waiting for the `expand-spec` to return an expanded spec vm that is represented by `instanceTypeVM`

src/views/virtualmachines/details/VirtualMachineNavPage.tsx
```diff
@@ -65,7 +66,7 @@ const VirtualMachineNavPage: FC = () => {
         <div className="VirtualMachineNavPage--tabs__main">
           <HorizontalNavbar
             basePath={getVMURL(cluster, namespace, name)}
-            error={loadError}
+            error={loadError || expandedSpecError}
```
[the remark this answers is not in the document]
upalatucci: Yes in fact it's a prerequisite . The majority of the details tab informations comes from the expanded spec vm like cpu and memory

src/views/virtualmachines/details/tabs/configuration/details/DetailsSection.tsx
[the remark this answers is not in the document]
upalatucci: i would do a `isInstanceType  function that only see for instancetype and rename the current one `isExpandableSpecVM`

src/utils/components/PendingChanges/hooks/usePendingChanges.tsx
```diff
@@ -152,7 +152,7 @@ export const usePendingChanges = (
           <CPUMemoryModal isOpen={isOpen} onClose={onClose} onSubmit={onSubmit} vm={vm} />
         )),
       ),
-      hasPendingChange: !isInstanceTypeVM(vm) && cpuMemoryChanged && restartRequired(vm),
+      hasPendingChange: !isExpandableSpecVM(vm) && cpuMemoryChanged && restartRequired(vm),
```
rszwajko, line 155: user can override preferences - I believe we need a strict `isInstanceTypeVM` check here

src/views/virtualmachines/details/tabs/configuration/details/DetailsSection.tsx
```diff
@@ -157,12 +157,12 @@ const DetailsSection: FC<DetailsSectionProps> = ({ allInstanceTypes, instanceTyp
             <VirtualMachineDescriptionItem
               descriptionHeader={
                 <SearchItem id="cpu-memory">
-                  {isInstanceType ? t('InstanceType') : t('CPU | Memory')}
+                  {isExpandableSpec ? t('InstanceType') : t('CPU | Memory')}
```
rszwajko, line 160: as with pending changes - it seems we need `isInstanceTypeVM` here

src/views/virtualmachines/details/tabs/configuration/details/DetailsSection.tsx
```diff
@@ -157,12 +157,12 @@ const DetailsSection: FC<DetailsSectionProps> = ({ allInstanceTypes, instanceTyp
...
                 </SearchItem>
               }
               onEditClick={() =>
                 createModal(({ isOpen, onClose }) => {
-                  return isInstanceType ? (
+                  return isExpandableSpec ? (
```
rszwajko, line 165: `isInstanceTypeVM` ?

src/views/virtualmachines/details/tabs/configuration/details/DetailsSection.tsx
```diff
@@ -185,7 +185,7 @@ const DetailsSection: FC<DetailsSectionProps> = ({ allInstanceTypes, instanceTyp
                 instanceType && getAnnotation(instanceType, INSTANCETYPE_CLASS_DISPLAY_NAME)
               }
               additionalContent={hasNUMAConfiguration(cpuMemoryVM) && <NUMABadge />}
-              bodyContent={isInstanceType ? null : <CPUDescription cpu={getCPU(vm)} />}
+              bodyContent={isExpandableSpec ? null : <CPUDescription cpu={getCPU(vm)} />}
```
rszwajko, line 188: `isInstanceTypeVM`?

src/views/virtualmachines/details/tabs/configuration/scheduling/components/SchedulingSectionRightGrid.tsx
```diff
@@ -70,8 +70,8 @@ const SchedulingSectionRightGrid: FC<SchedulingSectionRightGridProps> = ({
             ))
           }
           data-test-id="dedicated-resources"
-          descriptionData={<DedicatedResources vm={isInstanceTypeVM(vm) ? instanceTypeVM : vm} />}
-          isDisabled={isInstanceTypeVM(vm)}
+          descriptionData={<DedicatedResources vm={isExpandableSpecVM(vm) ? instanceTypeVM : vm} />}
```
rszwajko, line 73: not sure about dedicated resources - are they allowed or not for preferences-only VM?

rszwajko (commented):
@upalatucci 
looks good but let's double check the few places below(see my comments)

PR #2993, opened 2025-09-16
CNV-69485: prevent user for accessing Dedicated resources dialog when VirtualMachine is created from Instance Type

galkremer1:
## 📝 Description

[CNV-69485](https://issues.redhat.com/browse/CNV-69485): After user is manually editing the VM YAML, they are able to access the Dedicated resources Modal through the pending changes, even though we are blocking it in the configuration page based on the rule: 
- Can not configure dedicated resources if the VirtualMachine is created from Instance Type

## 🎥 Demo

Before: 

https://github.com/user-attachments/assets/c5a3fb56-411e-4173-b0a1-7af493946d06

After: 

https://github.com/user-attachments/assets/c6e663a9-7d4c-4a29-84bb-020430124614

### DIFF
--- a/src/utils/components/PendingChanges/utils/helpers.ts
+++ b/src/utils/components/PendingChanges/utils/helpers.ts
@@ -21,7 +21,10 @@ import {
   VirtualMachineConfigurationTabInner,
   VirtualMachineDetailsTab,
 } from '@kubevirt-utils/constants/tabs-constants';
-import { getInstanceTypeNameFromAnnotation } from '@kubevirt-utils/resources/instancetype/helper';
+import {
+  getInstanceTypeNameFromAnnotation,
+  isInstanceTypeVM,
+} from '@kubevirt-utils/resources/instancetype/helper';
 import { getName, getNamespace } from '@kubevirt-utils/resources/shared';
 import {
   getAffinity,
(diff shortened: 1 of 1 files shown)


src/utils/components/PendingChanges/utils/helpers.ts
```diff
@@ -290,7 +293,7 @@ export const getChangedDedicatedResources = (
   vmi: V1VirtualMachineInstance,
   currentSelection: boolean,
 ): boolean => {
-  if (isEmpty(vm) || isEmpty(vmi)) {
+  if (isEmpty(vm) || isEmpty(vmi) || isInstanceTypeVM(vm)) {
```
upalatucci, line 296: Should we check if the instancetype has `dedicatedCPUPlacement` customization? Does make sense to let the user customize this if the instancetype do not have the option?

galkremer1: I think it should be done on the backend side - they should reject the change. Right?

upalatucci, line 296: `isInstanceTypeVM` needs to change as this function test if vm its connected to instancetype or preference. ANd we experiment that [some customers use vms with preference but no instancetype](https://issues.redhat.com//browse/CNV-69441). So here we should test only if vm is connected to an instance type. I'll update my pr to change this behaviour. We assumed that those two things comes together every time . but its not always the case

pcbailey: @upalatucci Are you good with this PR now? I don't want to lgtm it if you have any remaining concerns.

upalatucci (approved):

PR #2997, opened 2025-09-17
CNV-69444: Fixing routing to correctly display operators in the software catalog

batyana:
## 📝 Description

changing the routing in virtualization feature links and overview related operators from operatorhub to catalog after console change to display the operator items in the catalog instead of 404 error page.

## 🎥 Demo
Before:
https://github.com/user-attachments/assets/9909ee51-b5da-444d-8a33-23dcaa60ae05

After:
https://github.com/user-attachments/assets/df1bcda4-f5ee-4512-91d6-8976f75fbcd7


src/views/clusteroverview/OverviewTab/getting-started-card/related-operators-section/RelatedOperatorsSection.tsx
```diff
@@ -14,6 +15,7 @@ const RelatedOperatorsSection: FC = () => {
   const { t } = useKubevirtTranslation();
   const { mtvLink, mtvLoaded } = useMTVResources();
   const isAdmin = useIsAdmin();
+  const [activeNamespace] = useActiveNamespace();
```
upalatucci, line 18: `activeNamespace` can also have a special value that we have under the constant `ALL_NAMESPACES_SESSION_KEY` when the user have `all-namespaces` selected. You can use  `getValidNamespace` to make sure the string do not have that value.

PR #2999, opened 2025-09-17
CNV-66097: Fixing the ability to change cpu when creating a vm from windows 11 template

batyana:
## 📝 Description

the vcpu logic was set so when updating the number it updated the socket amount while hardcoding core number and thread number to 1 each. in windows11 template there is a minimal requierment for 2 cores causing a failure to update. 
This fix allows to keep updating the sockets alone when updating vcpu but displaying in the CPU|Memory actual vcpu number and adds validation checks to define minimum core number in windows11 template as 2. 

## 🎥 Demo
before:
image

After:
image

### DIFF
--- a/src/utils/components/CPUMemoryModal/components/CPUInput/components/CPUTopologyInput/components/CPUComponentInput.tsx
+++ b/src/utils/components/CPUMemoryModal/components/CPUInput/components/CPUTopologyInput/components/CPUComponentInput.tsx
@@ -11,6 +11,7 @@ import { Content, GridItem, NumberInput } from '@patternfly/react-core';
 type CPUComponentInputProps = {
   cpu: V1CPU;
   cpuComponent: CPUComponent;
+  cpuLimits?: Record<string, number>;
   isDisabled: boolean;
   setCPU: Dispatch<SetStateAction<V1CPU>>;
 };
@@ -18,9 +19,19 @@ type CPUComponentInputProps = {
 const CPUComponentInput: FC<CPUComponentInputProps> = ({
   cpu,
   cpuComponent,
+  cpuLimits,
   isDisabled,
   setCPU,
 }) => {
+  // Get minimum value from validation rules
+  const minValue = cpuLimits?.[cpuComponent] || 1;
+
+  const updateCPU = (newValue: number) => {
+    if (newValue >= minValue) {
+      setCPU(getUpdatedCPU(cpu, newValue, cpuComponent));
+    }
+  };
+
   return (
     <>
       <GridItem span={3}>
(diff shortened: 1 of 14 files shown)


src/utils/components/CPUMemoryModal/components/CPUInput/utils/utils.ts
```diff
@@ -20,6 +20,5 @@ export const convertTopologyToVCPUs = (cpu: V1CPU): number =>
   cpu?.cores * cpu?.sockets * (cpu?.threads || 1);
 
 export const formatVCPUsAsSockets = (cpu: V1CPU): V1CPU => {
```
galkremer1, line 22: This function is now redundant and can be removed.

galkremer1, line 22: This function is not doing anything as it only returns its input value. 
It's also being used in 
```
          <VCPUInput
            cpu={formatVCPUsAsSockets(userEnteredCPU)}
            isDisabled={selectedRadioOption !== CPUInputType.editVCPU}
            setCPU={setUserEnteredCPU}
          />
```
Does it have any affect on it?

Do you think you can move the following logic into this function?
(from `src/utils/components/CPUMemoryModal/components/CPUInput/components/vCPUInput/components/CPUHelperText/CPUHelperText.tsx`)
```
const cores = formattedCpu?.cores || 1;
const threads = formattedCpu?.threads || 1;
const sockets = formattedCpu?.sockets || 1;
```

src/utils/components/CPUMemoryModal/components/CPUInput/components/CPUTopologyInput/components/CPUComponentInput.tsx
```diff
@@ -32,12 +38,19 @@ const CPUComponentInput: FC<CPUComponentInputProps> = ({
         <NumberInput
           onChange={(e: ChangeEvent<HTMLInputElement>) => {
             const newNumber = +e?.target?.value;
-            setCPU(getUpdatedCPU(cpu, newNumber, cpuComponent));
+            if (newNumber >= minValue) {
+              setCPU(getUpdatedCPU(cpu, newNumber, cpuComponent));
+            }
+          }}
+          onMinus={() => {
+            const newValue = +cpu?.[cpuComponent] - 1;
```
galkremer1, line 46: We can move
```
if (newNumber >= minValue) {
   setCPU(getUpdatedCPU(cpu, newNumber, cpuComponent));
}
```
And use it for `onChange`, `onMinus` and `onPlus`

rszwajko (changes requested):
The behaviour should be driven by the information in the template not by the template name. Templates may have different requirements which may vary in time. Please check if such functionality is possible with templates. If not perhaps we can suggest different solution i.e. instance types?
Ideally we should retrieve some kind of spec and pass it down to the widget.

src/utils/components/CPUMemoryModal/components/CPUInput/utils/utils.ts
batyana: after consulting backend and UX when it is a template or vm with minimum requirements for more than 1 core or thread the manual topology settings mode will display to avoid confusion.

src/utils/components/CPUMemoryModal/components/CPUInput/utils/utils.ts
```diff
@@ -20,6 +21,38 @@ export const convertTopologyToVCPUs = (cpu: V1CPU): number =>
...
+  return { ...cpu };
+};
+
+const parseValidationAnnotations = (
+  annotations?: Record<string, string>,
+): Record<string, number> => {
+  const validations = parseJSONAnnotation<any[]>(annotations, 'vm.kubevirt.io/validations', {});
+
+  if (!validations) {
+    return { cores: 1, sockets: 1, threads: 1 };
```
galkremer1, line 33: We can define `const defaultCPU = { cores: 1, sockets: 1, threads: 1 }`

PR #3028, opened 2025-09-24
CNV-69438: Update Add CD-ROM / Mount ISO modals

galkremer1:
## 📝 Description

**Tech improvements**
- User is now able to select an existing ISO when creating a new CD-ROM drive
- Fixed the structure of the disk drive when uploading a DV ISO (the PVC source was populated which is incorrect). 

**UX improvements:** 
- When mounting an ISO to a CD-ROM drive, the drive name is indicated in the modal. 
- When ejecting an ISO from a CD-ROM drive, the ISO name is indicated in the modal.

**Notes -**
Once confirms to be working as expected, we should re-visit the logic of mounting an ISO and adding a new CD-ROM drive, as some functions can be reused. 

## 🎥 Demo

https://github.com/user-attachments/assets/bdc7f28c-c244-40d1-b69d-7200c29a126d


src/utils/components/DiskModal/utils/helpers.ts
```diff
@@ -328,3 +328,17 @@ export const isHotPluggableEnabled = (featureGates: string[]) => {
   }
   return false;
 };
+
+export const convertDataVolumeToTemplate = (
```
galkremer1, line 332: Moved this helper here from another file.

src/utils/components/DiskModal/utils/helpers.ts
```diff
@@ -328,3 +328,17 @@ export const isHotPluggableEnabled = (featureGates: string[]) => {
...
+  dataVolume: V1beta1DataVolume,
+): V1DataVolumeTemplateSpec => ({
+  metadata: dataVolume.metadata,
+  spec: {
+    source: dataVolume.spec.source,
+    sourceRef: dataVolume.spec.sourceRef,
```
pcbailey, line 338: `metadata` is marked as optional in `V1beta1DataVolume` and `source` and `sourceRef` are marked as optional in `V1beta1DataVolumeSpec`. I know you just moved this pre-existing function, but could you please add the optional chaining operators to those fields?

src/utils/components/DiskModal/AddCDROMModal.tsx
```diff
@@ -65,16 +88,27 @@ const AddCDROMModal: FC<V1SubDiskModalProps> = ({
...
+    if (selectedISO) {
+      data.volume = {
+        name: data.volume.name,
+        persistentVolumeClaim: {
+          claimName: selectedISO,
+          ...(isHotPluggable && { hotpluggable: true }),
+        },
+      };
+      delete data.dataVolumeTemplate;
+    } else if (uploadEnabled && data.uploadFile?.file) {
```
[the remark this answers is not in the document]
galkremer1: Yes, since otherwise the form is not going to be valid and the user won't be able to submit it. I will add the optional though just to be on the safe side.

upalatucci (approved):

PR #3029, opened 2025-09-25
CNV-69912: Remove any links to the obervability plugin for multicluster views

upalatucci:
## 📝 Description

Remove metrics links, as those do not work in ACM vm details view. Observability is not capable of showing multicluster metrics right now


src/utils/components/Charts/MigrationUtil/MigrationThresholdChartDiskRate.tsx
```diff
@@ -64,7 +67,7 @@ const MigrationThresholdChartDiskRate: React.FC<MigrationThresholdChartDiskRateP
 
   const isReady = !isEmpty(chartDataProcessed);
   const yMax = findMaxYValue(chartDataProcessed);
-  const linkToMetrics = queriesToLink(queries.MIGRATION_DISK_TRANSFER_RATE);
+  const linkToMetrics = !isACMPage && queriesToLink(queries.MIGRATION_DISK_TRANSFER_RATE);
```
galkremer1, line 70: What do you think about integrating `queriesToLink` into `useVMQueries`? Then inside we can check `isACMPage` so we can consolidate the use cases.

pcbailey:
/lgtm
That looks much cleaner!

PR #3044, opened 2025-09-29
CNV-69165: Use preserveSession flag for VNC connections

rszwajko:
## 📝 Description

Key points:
1. connect by default with preserve-session flag
2. sync the preserve-session flag with VNC shared flag (RFC 6143)
3. allow users to force exclusive mode (after initial connection)

Depends on backend feature: Add preserve-session option to help avoid disconnect VNC sessions (#15267)

Reference-Url: https://datatracker.ietf.org/doc/html/rfc6143#section-7.3.1
Reference-Url: https://github.com/kubevirt/kubevirt/pull/15267

## 🎥 Demo

Screenshot From 2025-09-29 18-27-44


rszwajko:
Replaced by #3085

PR #3051, opened 2025-09-30
CNV-68721: Adding a second disk changes the boot order

galkremer1:
## 📝 Description

Jira Ticket: [CNV-68721](https://issues.redhat.com/browse/CNV-68721)

This PR fixes a fundamental issue where a `vm` is created without a bootOrder value. 
When it's created without one, the `rootDisk` does not show up in the VirtualMachine details under boot order, so in case the user changes the boot order the VM cannot be restored to a working state without manually editing the YAML. 

`getBootableSortedDevices` is using `instanceTypeVM` (`instanceTypeExpandedSpec` - the expanded YAML object that we are getting from `/expand-spec`) - if `instanceTypeVM` is not available we default to `vm`.
Getting the data from `instanceTypeExpandedSpec` ensures that the `root disk` will be present in the disk list. 

## 🎥 Demo

Before:

https://github.com/user-attachments/assets/89069dd2-fa12-486f-9a95-30c5c05a7458

After:

https://github.com/user-attachments/assets/784f14fb-4d9f-4ec2-b3a4-92b4af9c0340

### DIFF
--- a/src/views/topology/components/vm/VMDetailsPanel/TopologyVMDetailsPanel.tsx
+++ b/src/views/topology/components/vm/VMDetailsPanel/TopologyVMDetailsPanel.tsx
@@ -1,6 +1,7 @@
 import React, { FC } from 'react';
 
 import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import useInstanceTypeExpandSpec from '@kubevirt-utils/resources/vm/hooks/useInstanceTypeExpandSpec';
 import { Grid, GridItem } from '@patternfly/react-core';
 import { observer } from '@patternfly/react-topology';
 import { VMNode } from '@topology/utils/types/types';
@@ -16,6 +17,7 @@ const TopologyVMDetailsPanel: FC<TopologyVMDetailsPanelProps> = observer(({ vmNo
   const vmData = vmNode.getData();
   const vm = vmData.resource as V1VirtualMachine;
   const { vmi } = vmData?.data;
+  const [instanceTypeExpandedSpec] = useInstanceTypeExpandSpec(vm);
   return (
     <div className="overview__sidebar-pane-body resource-overview__body">
       <Grid hasGutter>
@@ -23,7 +25,7 @@ const TopologyVMDetailsPanel: FC<TopologyVMDetailsPanelProps> = observer(({ vmNo
           <VMDetailsPanelLeftColumn vm={vm} vmi={vmi} />
         </GridItem>
         <GridItem span={6}>
-          <VMDetailsPanelRightColumn vm={vm} vmi={vmi} />
+          <VMDetailsPanelRightColumn instanceTypeVM={instanceTypeExpandedSpec} vm={vm} vmi={vmi} />
         </GridItem>
       </Grid>
     </div>
--- a/src/views/topology/components/vm/VMDetailsPanel/components/VMDetailsPanelRightColumn/components/VMBootOrderDetailsItem.tsx
+++ b/src/views/topology/components/vm/VMDetailsPanel/components/VMDetailsPanelRightColumn/components/VMBootOrderDetailsItem.tsx
@@ -7,20 +7,19 @@ import { useModal } from '@kubevirt-utils/components/ModalProvider/ModalProvider
 import VirtualMachineDescriptionItem from '@kubevirt-utils/components/VirtualMachineDescriptionItem/VirtualMachineDescriptionItem';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
 import { getName } from '@kubevirt-utils/resources/shared';
-import useInstanceTypeExpandSpec from '@kubevirt-utils/resources/vm/hooks/useInstanceTypeExpandSpec';
 import { updateBootOrder } from '@virtualmachines/details/tabs/configuration/details/utils/utils';
 
 import '../../../TopologyVMDetailsPanel.scss';
 
 type VMBootOrderDetailsItemProps = {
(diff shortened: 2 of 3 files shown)


src/globals.d.ts
```diff
@@ -13,6 +13,11 @@ declare module '@kubevirt-ui/kubevirt-api/kubevirt' {
     cluster?: string;
   }
```
galkremer1, line 15: This fix is not related to CNV-68721 - it just prevents this type of linter error: 
Screenshot 2025-09-30 at 4 42 50 PM

src/views/catalog/CreateFromInstanceTypes/utils/utils.ts
```diff
@@ -216,7 +216,15 @@ export const generateVM: GenerateVMCallback = ({
           domain: {
             devices: {
               autoattachPodInterface: false,
```
galkremer1, line 218: Adding a default bootORder to `ROOTDISK` and adding it to the disk list when creating a vm from an Instance type.

src/views/catalog/utils/quick-create-vm.ts
```diff
@@ -101,6 +101,10 @@ export const quickCreateVM: QuickCreateVMType = async ({
     if (!isEmpty(templateArchitecture) && enableMultiArchBootImageImport) {
       draftVM.spec.template.spec.architecture = templateArchitecture;
     }
+    const rootDisk = getDisks(draftVM)?.find((disk) => disk.name === ROOTDISK);
+    if (rootDisk && !rootDisk.bootOrder) {
+      rootDisk.bootOrder = 1;
```
galkremer1, line 106: Adding a bootOrder=1 when creating a vm from template.

src/views/topology/components/vm/VMDetailsPanel/TopologyVMDetailsPanel.tsx
```diff
@@ -16,15 +17,15 @@ const TopologyVMDetailsPanel: FC<TopologyVMDetailsPanelProps> = observer(({ vmNo
   const vmData = vmNode.getData();
   const vm = vmData.resource as V1VirtualMachine;
   const { vmi } = vmData?.data;
-
+  const [instanceTypeExpandedSpec] = useInstanceTypeExpandSpec(vm);
```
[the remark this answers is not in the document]
galkremer1: You are not missing anything - that's a great suggestion - I can just move the hook into the BootOrder Modal.

src/utils/resources/vm/utils/boot-order/bootOrder.ts
```diff
@@ -49,3 +50,33 @@ export const transformDevices = (
...
+  if (b.value.bootOrder) {
+    return 1;
+  }
+
+  return 0;
+};
+export const hasBootOrderOnVM = (vm: V1VirtualMachine) =>
+  getDisks(vm).some((disk) => disk.bootOrder);
+
+export const getBootableSortedDevices = ({
```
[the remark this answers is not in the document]
galkremer1: Done, thank you

src/utils/resources/vm/utils/boot-order/bootOrder.ts
```diff
@@ -49,3 +50,33 @@ export const transformDevices = (
 
   return [...transformedDisks, ...transformedNics];
 };
+
+export const sortBootOrder = (a: BootableDeviceType, b: BootableDeviceType) => {
+  if (a.value.bootOrder && b.value.bootOrder) {
```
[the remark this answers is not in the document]
galkremer1: I agree

src/utils/resources/vm/utils/boot-order/bootOrder.ts
```diff
@@ -49,3 +53,34 @@ export const transformDevices = (
...
+};
+export const getSortedBootableDevices = ({
+  instanceTypeVM,
+  vm,
+}: {
+  instanceTypeVM: V1VirtualMachine;
+  vm: V1VirtualMachine;
+}): BootableDeviceType[] | undefined => {
+  const filteredInstanceTypeVMDisks = getDisks(instanceTypeVM || vm)?.filter(
+    (disk) => disk?.name !== CLOUDINITDISK,
```
galkremer1, line 80: We are removing the 'cloudinit' disk from the extended VM and then merging it with the disks from the VM. This way we ensure that the root disk will be in the list and if the user named a disk "cloudinit" we are not going to filter it out (as it's going to be included in the VM object).

src/views/topology/components/vm/VMDetailsPanel/components/VMDetailsPanelRightColumn/components/VMBootOrderDetailsItem.tsx
```diff
@@ -19,13 +20,15 @@ type VMBootOrderDetailsItemProps = {
 const VMBootOrderDetailsItem: FC<VMBootOrderDetailsItemProps> = ({ vm, vmi }) => {
   const { t } = useKubevirtTranslation();
   const { createModal } = useModal();
+  const [instanceTypeVM] = useInstanceTypeExpandSpec(vm);
```
upalatucci, line 23: Because `useInstanceTypeExpandSpec` is not in a centralized state and do not cache anything, this will fire another call to the `expand-spec` endpoint

galkremer1: I moved back `useInstanceTypeExpandSpec` into the parent component. I agree - we should improve the hook but I think that we need to implement it in a separate PR.

upalatucci (approved):

PR #3060, opened 2025-10-03
CNV-70171:  Prevent YAML corruption when changing boot order while customizing a new vm

galkremer1:
## 📝 Description
Jira Ticket: [CNV-70233](https://issues.redhat.com/browse/CNV-70233)
When customizing a new VM - changing the boot order corrupts the YAML due to an improper handling of `updateCustomizeInstanceType`. 
The test case of the exact scenario that caused the YAML corruption is tested in the test file under `object corruption prevention`

## 🎥 Demo

Before:

https://github.com/user-attachments/assets/d36445ec-05e3-4d06-9004-1b16d4613aea

After:

https://github.com/user-attachments/assets/cfe40033-7a50-40e8-a2fb-fab18896fce1

### DIFF
--- a/src/utils/store/customizeInstanceType.ts
+++ b/src/utils/store/customizeInstanceType.ts
@@ -1,6 +1,7 @@
 import produce from 'immer';
 
 import { V1VirtualMachine } from '@kubevirt-ui/kubevirt-api/kubevirt';
+import { ensurePath } from '@kubevirt-utils/utils/utils';
 import { effect, signal } from '@preact/signals-react';
 
 import { isEmpty } from '../utils/utils';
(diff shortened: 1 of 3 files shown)


src/utils/store/customizeInstanceType.ts
```diff
@@ -39,22 +40,44 @@ export const clearCustomizeInstanceType = () => {
...
+      if (validPathParts.length === 0) {
+        return;
+      }
+
       let obj = vmDraft;
 
-      pathParts.forEach((part: string, index: number) => {
-        if (index < pathParts.length - 1) {
-          obj = obj?.[part] ? obj[part] : Object.assign(obj, { [part]: {} })[part];
+      validPathParts.forEach((part: string, index: number) => {
```
upalatucci, line 74: we have a function that does that. Its called `ensurePath` or something like that

PR #3061, opened 2025-10-06
CNV-69204: Instancetype page multicluster

upalatucci:
## 📝 Description

Search api does not have some parameters so the columns for cpu and memory are empty. But they'll be populated when search api will index those fields

Make sure to have the dropdowns for cluster and projects. ClusterInstaceTypes do not have namespace so no project dropdown in that case.

## 🎥 Demo

https://github.com/user-attachments/assets/d545d746-31c4-4c3e-a8e0-fe20030275b3


galkremer1 (approved):
/lgtm

PR #3062, opened 2025-10-06
CNV-70150: Fix stale vm data in in useInstanceTypeExpandSpec hook

galkremer1:
## 📝 Description

Jira: [CNV-70150](https://issues.redhat.com/browse/CNV-70150)
This PR fixes two issues; 
1. The original issue in the Jira ticket - the VM gets stale data from the `useInstanceTypeExpandSpec` hook, leading to incorrect data for the boot mode (and potentially other parts which show incorrect data as well). 
2. Right clicking a VM in the list after reloading the page while selecting the configuration tab crashes the app. 

## 🎥 Demo

Before:

https://github.com/user-attachments/assets/536f1df3-78e5-400a-b50b-408ea1f20467

After:

https://github.com/user-attachments/assets/b7851a36-4086-4434-ac01-75e7882a8e26

### DIFF
--- a/src/utils/resources/vm/hooks/useInstanceTypeExpandSpec.ts
+++ b/src/utils/resources/vm/hooks/useInstanceTypeExpandSpec.ts
@@ -43,7 +43,13 @@ const useInstanceTypeExpandSpec: UseInstanceTypeExpandSpec = (vm) => {
         setLoadingExpandedSpec(false);
       }
     };
-    !isEmpty(innerVM) && isExpandableSpec && fetch();
+
+    if (!isEmpty(innerVM) && isExpandableSpec) {
+      fetch();
+    } else {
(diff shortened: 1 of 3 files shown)


src/utils/resources/vm/hooks/useInstanceTypeExpandSpec.ts
```diff
@@ -43,7 +43,13 @@ const useInstanceTypeExpandSpec: UseInstanceTypeExpandSpec = (vm) => {
         setLoadingExpandedSpec(false);
       }
     };
-    !isEmpty(innerVM) && isExpandableSpec && fetch();
+
```
galkremer1, line 46: Clear stale data when VM doesn't have instance type or is empty

upalatucci (approved):

PR #3097, opened 2025-10-21
CNV-70940: Adding the option to open the vm console through tree view

batyana:
## 📝 Description

Adding the option to open a vm console from the virtual machine tree view and the virtual machine list through the action menu. the console opens in a new tab allowing users flexibility in opening the console for multiple vm's without having to go to each vm's details page.

## 🎥 Demo

https://github.com/user-attachments/assets/61c9dfcd-ad15-4379-a75b-e89b751474e1


src/views/virtualmachines/actions/VirtualMachineActionFactory.tsx
```diff
@@ -262,6 +263,22 @@ export const VirtualMachineActionFactory = {
       label: t('Move to folder'),
     };
   },
+  openConsole: (vm: V1VirtualMachine): ActionDropdownItemType => {
+    const vmCluster = getCluster(vm);
+    const vmNamespace = getNamespace(vm);
+    const vmName = getName(vm);
+
+    return {
+      accessReview: asAccessReview(VirtualMachineModel, vm, 'get'),
```
upalatucci: This will fire an unecessary call for a 'get' permission. But if u get the vm, u have get access

PR #3103, opened 2025-10-22
CNV-71112: add ServicesList component

adamviktora:
## 📝 Description

- adds `ServicesList` component to kubevirt, because the one from `@openshift-console/dynamic-plugin-sdk-internal` was removed in 4.20

- some code is copied from [networking plugin](https://github.com/openshift/networking-console-plugin) - LabelList, PodSelectorModal, useServiceActions, Selector (which is all mostly a copy of console code)
- other code is taken from [ServicesList from console before it was removed](https://github.com/openshift/console/commit/4730440c7223dc9edd05b3d3d67f9fca9b3834c5#diff-4c15e217d79b0f4047ce481fb02ec07d8068c7b05445f177f3d6e560a37c2183) and customized for usage in kubevirt

- TODO: I think we share a lot of common code and components between our 3 plugins and also from console. We should think about a common library or trying to use console extensions more where possible

## 🎥 Demo

Before:
Screenshot 2025-10-22 at 14 23 56

After:
Screenshot 2025-10-22 at 13 27 39


adamviktora:
@upalatucci I updated the PR to use the `LazyActionMenu`. I had to add a `console.action/resource-provider` extension so it works.

>  This will allow us to not even copying the PodSelectorModal

But I was not able to do it without copying this. I still had to keep all the actions there

PR #3132, opened 2025-10-29
CNV-69204: migration policy list page multicluster

upalatucci:
## 📝 Description

Followup for creation form and yaml

## 🎥 Demo

https://github.com/user-attachments/assets/b7842577-c6e7-48aa-b2e1-61d231e66629


src/utils/hooks/useProjectFilter.ts
```diff
@@ -0,0 +1,53 @@
...
+import { RowFilter } from '@openshift-console/dynamic-plugin-sdk';
+
+import useSelectedRowFilterClusters from './useSelectedRowFilterClusters';
+
+export const useProjectFilter = <R extends K8sResourceCommon>(): RowFilter<R> => {
+  const allClustersSelected = useSelectedRowFilterClusters();
+  const isACMPage = useIsACMPage();
+
+  const [projects] = useProjects();
+  const { allNamespaces: multiclusterProjects } = useMulticlusterNamespaces(allClustersSelected);
```
galkremer1: We should use the same convention: `namespaces` instead of `projects`

src/utils/hooks/useSelectedRowFilterClusters.ts
```diff
@@ -0,0 +1,11 @@
+import { useMemo } from 'react';
+import { useSearchParams } from 'react-router-dom-v5-compat';
+
+import { CLUSTER_LIST_FILTER_PARAM } from '@kubevirt-utils/utils/constants';
+
+const useSelectedRowFilterClusters = (): string[] => {
+  const [searchParams] = useSearchParams();
+  return useMemo(() => searchParams.get(CLUSTER_LIST_FILTER_PARAM)?.split(','), [searchParams]);
```
galkremer1: if `searchParams.get(CLUSTER_LIST_FILTER_PARAM)` is undefined or null the return type is going to be incorrect. 
I suggest:
```
const clusterParam = searchParams.get(CLUSTER_LIST_FILTER_PARAM);
return useMemo(() => clusterParam?.split(',') ?? [], [clusterParam]);
```

upalatucci: cool thanks

src/views/virtualmachines/tree/utils/utils.tsx
```diff
@@ -279,7 +279,7 @@ export const createMultiClusterTreeViewData = (
   vms: V1VirtualMachine[],
   pathname: string,
   foldersEnabled: boolean,
-  projectsByClusters: UseMulticlusterNamespacesReturnType[0],
+  projectsByClusters: UseMulticlusterNamespacesReturn['namespacesByCluster'],
```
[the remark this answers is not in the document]
upalatucci: typescript gives me errors with ur annotation and suggest to switch to my annotation

PR #3136, opened 2025-10-30
CNV-69204: change cluster select for instancetype to a rowFilter

upalatucci:
## 📝 Description

Change the top cluster and namespace filter to a regular rowFilter in the table 

## 🎥 Demo

https://github.com/user-attachments/assets/678cfde5-53e8-4455-bd15-3e86ee5d577b

Screenshot 2025-10-30 at 16 43 34

## Summary by CodeRabbit

* **New Features**
  * Expanded multi-cluster and multi-namespace support for instance type listing and retrieval.
  * Cluster and project selection integrated directly into list filters for clearer selection.

* **Bug Fixes**
  * Improved filter stability and performance by memoizing filter configurations.
  * More accurate empty-state and loading behavior when filters are applied.

* **Refactor**
  * Simplified UI by removing the separate cluster/project header dropdown and consolidating into list filters.

### DIFF
--- a/src/views/instancetypes/list/ClusterInstancetypeList.tsx
+++ b/src/views/instancetypes/list/ClusterInstancetypeList.tsx
@@ -1,9 +1,10 @@
-import React, { FC } from 'react';
+import React, { FC, useMemo } from 'react';
 
 import useClusterInstanceTypes from '@catalog/CreateFromInstanceTypes/state/hooks/useClusterInstanceTypes';
 import { VirtualMachineClusterInstancetypeModelRef } from '@kubevirt-ui/kubevirt-api/console';
 import { V1beta1VirtualMachineClusterInstancetype } from '@kubevirt-ui/kubevirt-api/kubevirt';
 import ListPageFilter from '@kubevirt-utils/components/ListPageFilter/ListPageFilter';
+import { useClusterFilter } from '@kubevirt-utils/hooks/useClusterFilter';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
 import usePagination from '@kubevirt-utils/hooks/usePagination/usePagination';
 import { paginationDefaultValues } from '@kubevirt-utils/hooks/usePagination/utils/constants';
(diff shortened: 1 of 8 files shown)


src/views/instancetypes/list/UserInstancetypeList.tsx
```diff
@@ -38,19 +42,26 @@ const UserInstancetypeList: FC<UserInstancetypeListProps> = ({
...
     V1beta1VirtualMachineInstancetype
-  >(instanceTypes, null, {
+  >(instanceTypes, filtersWithSelect, {
     name: { selected: [nameFilter] },
   });
 
   const [columns, activeColumns, loadedColumns] = useUserInstancetypeListColumns(pagination, data);
 
-  if (loaded && isEmpty(unfilteredData)) {
+  if (loaded && isEmpty(unfilteredData) && isEmpty(selectedClusters) && isEmpty(selectedNamespaces)) {
```
adamviktora: I think that the EmptyState will not be shown in case there are no filters and `selectedNamespaces` or `selectedClusters` comes from the URL path. (`useListNamespaces` -> `const namespace = useNamespaceParam()`).

src/views/instancetypes/list/ClusterInstancetypeList.tsx
```diff
@@ -64,10 +68,11 @@ const ClusterInstancetypeList: FC<ListPageProps> = ({
             });
           }}
           data={unfilteredData}
+          filtersWithSelect={filtersWithSelect}
           hideColumnManagement={hideColumnManagement}
           hideLabelFilter={hideTextFilter}
           hideNameLabelFilters={hideNameLabelFilters}
-          loaded={loaded && loadedColumns}
+          loaded={loadedColumns}
```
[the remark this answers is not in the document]
upalatucci: When changing stuff in the filters at every click that i do in the project or cluster selection, the filter section goes into loading state. But in reality we don't need to load anything for the filters. The loading is just for the stuff  in the table

src/views/instancetypes/list/ClusterInstancetypeList.tsx
```diff
@@ -32,7 +33,10 @@ const ClusterInstancetypeList: FC<ListPageProps> = ({
   const [instanceTypes, loaded, loadError] = useClusterInstanceTypes(fieldSelector, selector);
 
   const { onPaginationChange, pagination } = usePagination();
-  const [unfilteredData, data, onFilterChange] = useListPageFilter(instanceTypes, null, {
+
+  const clusterFilter = useClusterFilter();
+  const filtersWithSelect = useMemo(() => [clusterFilter], [clusterFilter]);
```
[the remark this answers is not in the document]
upalatucci: I should memoize the cluster filter to have a true memoization. I'll do that in the next prs

src/views/instancetypes/list/UserInstancetypeList.tsx
adamviktora: Yes, my point is that when `selectedNamespaces` will have one namespace (from url path), then `isEmpty(selectedNamespaces)` will be false, so we will not show EmptyState, although we should show it.

upalatucci: oooh yes right so i'll switch by getting only the filtered clusters from the row filter

adamviktora (approved):

PR #3123, opened 2025-10-27
CNV-71428: Lint staged only

rszwajko:
## 📝 Description
Depends on: #3122 

1. ensure that prettier and eslint do not overlap.
2. run i18n as pre-commit hook after linting

Reference-Url: https://github.com/lint-staged/lint-staged?tab=readme-ov-file#task-concurrency

PR #3137, opened 2025-10-30
CNV-71417: Fix running prettier in IDE and via yarn lint

rszwajko:
## 📝 Description

The previous improvements #3122  and #3123  were incorrectly tested (different repo or newer packages).
Problems:
1. Restrictive .prettierignore files disables the formatting in the IDE (if prettier plugin is used)
2. prettier formatting for code files is effectively disabled - older prettier (<3.x) is missing features required for this use case

Fix: use different, more restrictive prettierignore only for `yarn lint` and `yarn lint:fix`. Lint-staged is not affected as it uses it's own file patterns.

## 🎥 Demo

https://github.com/user-attachments/assets/3cfa80f0-1c3f-40e9-87a0-709d90db217e


galkremer1 (approved):
/lgtm

PR #3150, opened 2025-11-03
CNV-71151: Update service label selectors to use vmi.kubevirt.io/id for KubeVirt v1.7+ compatibility

galkremer1:
## 📝 Description

Adds VMI labels as a fallback in `getServicesForVmi` to improve service matching when pods are unavailable.

- Updated getServicesForVmi to accept optional vmi parameter
- Added VMI labels check in fallback chain: pod labels → VMI labels → VM template labels → VM name
- Updated call sites to pass VMI when available

KubeVirt v1.7+ introduces `vmi.kubevirt.io/id` on pods for service matching. Labels like `vm.kubevirt.io/name` can also exist on the VMI resource (e.g., manually set or inherited from templates). 
Checking VMI labels ensures matching when the pod is not available (stopped VMs, etc.).

Jira ticket: [CNV-71151](https://issues.redhat.com/browse/CNV-71151)
[KubeVirt PR #15783 introducing the change](https://github.com/kubevirt/kubevirt/pull/15783)

## 🎥 Demo

In addition to adding the support for the new selector (`vmi.kubevirt.io/id`), this PR also fixes a bug where the services are not listed under the VMI:

Before:

https://github.com/user-attachments/assets/d5392fea-1969-489a-847b-9ad82a3d790e

After:

https://github.com/user-attachments/assets/5518ccd6-f17a-481f-b258-76d01f189d1a

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved RDP and SSH service discovery by making label selection pod-aware and strengthening fallbacks when pod/VMI context is missing.

* **Refactor**
  * Centralized and optimized service lookup and label handling for VM consoles and access paths to ensure consistent selector generation.

* **Tests**
  * Updated unit tests to cover pod-aware selector logic and new fallback scenarios.

### DIFF
--- a/src/utils/components/Consoles/components/DesktopViewer/utils/utils.ts
+++ b/src/utils/components/Consoles/components/DesktopViewer/utils/utils.ts
@@ -9,10 +9,14 @@ import {
   IoK8sApiCoreV1ServicePort,
 } from '@kubevirt-ui-ext/kubevirt-api/kubernetes';
 import { V1VirtualMachine, V1VirtualMachineInstance } from '@kubevirt-ui-ext/kubevirt-api/kubevirt';
+import { buildOwnerReference } from '@kubevirt-utils/resources/shared';
+import {
+  getServicesForVmi,
+  getVMILabelForServiceSelector,
+} from '@kubevirt-utils/resources/vmi/utils/services';
 import { escapeJsonPointerToken, kubevirtConsole } from '@kubevirt-utils/utils/utils';
 import { k8sCreate, k8sPatch, K8sResourceCommon } from '@openshift-console/dynamic-plugin-sdk';
 
-import { buildOwnerReference } from '../../../../../resources/shared';
 import { RDP_CONSOLE_TYPE, SPICE_CONSOLE_TYPE, VNC_CONSOLE_TYPE } from '../../utils/ConsoleConsts';
 
 import {
@@ -21,8 +25,6 @@ import {
   DEFAULT_VV_MIMETYPE,
   MULTUS,
   POD,
-  TEMPLATE_VM_NAME_LABEL,
-  VMI_LABEL_AS_RDP_SERVICE_SELECTOR,
 } from './constants';
 import { ConsoleDetailPropType, Network } from './types';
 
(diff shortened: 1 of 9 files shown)


src/utils/components/Consoles/components/DesktopViewer/utils/utils.ts
```diff
@@ -228,19 +233,25 @@ export const getDefaultNetwork = (networks: Network[]) => {
...
   const { name, namespace } = vm?.metadata || {};
-  const vmiLabels = vm?.spec?.template?.metadata?.labels;
-  const labelSelector = vmiLabels?.[VMI_LABEL_AS_RDP_SERVICE_SELECTOR] || name;
+
+  const labelSelector = pod
+    ? getVMILabelForServiceSelector(pod, vm)
+    : {
+        labelKey: VM_NAME_LABEL,
+        labelValue: vm?.spec?.template?.metadata?.labels?.[VM_NAME_LABEL] || name,
+      };
```
adamviktora: Question / suggestion - every time we call the `getVMILabelForServiceSelector` there is this fallback for when the `pod` is undefined.

Here the `labelValue` is different from the other occasions, where the `getName(vm)` is used directly. Is there a reason for not using the `vm?.spec?.template?.metadata?.labels?.[VM_NAME_LABEL]` in the other occasions?

If the same fallback could be used everywhere, we could refactor to put it inside `getVMILabelForServiceSelector`

src/utils/resources/vmi/utils/services.ts
```diff
@@ -1,18 +1,85 @@
...
-    [key: string]: string;
-  },
+  pod?: IoK8sApiCoreV1Pod,
+  vm?: V1VirtualMachine,
+  vmi?: V1VirtualMachineInstance,
 ): IoK8sApiCoreV1Service[] => {
-  if (!vmiLabels) return [];
+  if (!services || services.length === 0) return [];
+
+  let labelsToMatch: { [key: string]: string } = {};
```
adamviktora: nit: we could extract the part to get `labelsToMatch` to a separate function

src/utils/components/SSHAccess/utils.ts
```diff
@@ -3,41 +3,58 @@ import produce from 'immer';
...
+      };
+
+  return {
+    apiVersion: ServiceModel.apiVersion,
+    kind: ServiceModel.kind,
+    metadata: {
+      // max name length is 63 characters
+      name: `${vm?.metadata?.name}-${type.toLowerCase()}-ssh-service`
+        .substring(0, 56)
+        .concat(`-${getRandomChars(4)}`),
```
adamviktora: we can probably use `truncateToK8sName` helper for this

src/utils/components/Consoles/components/DesktopViewer/utils/utils.ts
galkremer1: Good catch! It wasn't intentional - I will refactor it to make `pod` in `getVMILabelForServiceSelector` optional, so the fallback lives in one place and all three call sites (SSHAccess/utils.ts, SSHCommand.tsx, DesktopViewer/utils.ts) simply call `getVMILabelForServiceSelector(pod, vm)` directly.

adamviktora (approved):

PR #3160, opened 2025-11-05
CNV-71545: Filtering top consumers by namespace for vm and node scopes

batyana:
## 📝 Description

previously, when in top consumers tab the charts would show metrics for all vm's in all projects regardless of what name space the user is on. this fix creates filtering according to the active name space showing only vm's and nodes in the current project. We also hide the "projects" chart scope for all namespaces except "all-projects"

## 🎥 Demo
Before:
image
image

After:
image
image

## Summary by CodeRabbit

* **New Features**
  * Charts and queries are namespace-aware: show namespace-specific data when a namespace is selected and cluster-wide data when viewing all namespaces.
  * Scope selector adapts to namespace context, offering only relevant scopes for the current namespace.
  * Metric queries now respect the selected namespace for more accurate results.

* **Bug Fixes**
  * Improved default scope selection so charts fall back sensibly when a chosen scope is unavailable.

### DIFF
--- a/src/views/clusteroverview/TopConsumersTab/utils/queries.ts
+++ b/src/views/clusteroverview/TopConsumersTab/utils/queries.ts
@@ -1,62 +1,76 @@
+import { ALL_NAMESPACES_SESSION_KEY } from '@kubevirt-utils/hooks/constants';
+
 import { TopConsumerMetric as Metric } from './topConsumerMetric';
 import { TopConsumerScope as Scope } from './topConsumerScope';
 
+const getNamespaceFilter = (namespace: string) =>
+  namespace && namespace !== ALL_NAMESPACES_SESSION_KEY ? `{namespace="${namespace}"}` : '';
+
 const topConsumerQueries = {
   [Scope.NODE.getValue()]: {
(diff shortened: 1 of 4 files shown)


adamviktora (commented):
@batyana This change looks great. Can you also fix the `by (node)` -> `by (namespace)` as coderabbit suggests?

src/views/clusteroverview/TopConsumersTab/utils/queries.ts
```diff
@@ -1,62 +1,124 @@
...
 };
 
-export const getTopConsumerQuery = (metric, scope, numItemsToShow = 5, duration = '5m') =>
-  topConsumerQueries?.[scope]?.[metric]?.(numItemsToShow, duration);
+export const getTopConsumerQuery = (
+  metric,
+  scope,
+  numItemsToShow = 5,
+  duration = '5m',
+  namespace?: string,
```
galkremer1: Instead of calling `getNamespaceFilter` inside every query - can't we just call it here when passing it to `topConsumerQueries`?

adamviktora (approved):

upalatucci (approved):

PR #3164, opened 2025-11-06
[release-4.18] CNV-66469: CNV-70478: fix filtering VMs by IP when proxy pod is used

adamviktora:
## 📝 Description

- bug fix is on line 141:
```
const matchedVMS = filteredData?.filter(
```
  - `matchedVMS` variable is used when proxy is used
  - if the proxy filtering fails, it now uses `filteredData` (VMs with applied frontend filter), same as in 4.19 and newer
- the other changes are just refactoring to align with 4.19 branch

### DIFF
--- a/src/views/virtualmachines/list/VirtualMachinesList.tsx
+++ b/src/views/virtualmachines/list/VirtualMachinesList.tsx
@@ -119,10 +119,10 @@ const VirtualMachinesList: FC<VirtualMachinesListProps> = forwardRef(({ kind, na
 
   const [pagination, setPagination] = useState(paginationInitialState);
 
-  const [unfilterData, dataFilters, onFilterChange] = useListPageFilter<
-    V1VirtualMachine,
-    V1VirtualMachine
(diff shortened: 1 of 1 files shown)


src/views/virtualmachines/list/VirtualMachinesList.tsx
```diff
@@ -119,7 +119,7 @@ const VirtualMachinesList: FC<VirtualMachinesListProps> = forwardRef(({ kind, na
 
   const [pagination, setPagination] = useState(paginationInitialState);
 
-  const [unfilterData, dataFilters, onFilterChange] = useListPageFilter<
+  const [_unfilteredData, filteredData, onFilterChange] = useListPageFilter<
```
upalatucci:
```suggestion
  const [_, filteredData, onFilterChange] = useListPageFilter<
```

upalatucci: Okay so u can put double underscore. __

PR #3171, opened 2025-11-10
CNV-70560: fix guided tour

upalatucci:
## 📝 Description

Non priv users were having trouble as we were firing instancetype expand spec request anyway. Even for the guided tour vm. It's unnecessary. 

Then we were having also other issues for all users . Some ids changed and 

## 🎥 Demo

https://github.com/user-attachments/assets/9a8c3fca-ccac-4aec-aa64-13ac495671c2

## Summary by CodeRabbit

* **New Features**
  * Start and stop guided tours from the UI; tours now run consistently across more views.

* **Bug Fixes**
  * Fixed configuration search target used by the guided tour.

* **Chores**
  * Guided tour state refined to avoid interrupting background navigation and data fetching.
  * Certain controls remain enabled during a tour so walkthroughs aren’t blocked.

### DIFF
--- a/src/utils/hooks/useForceProjectSelection.ts
+++ b/src/utils/hooks/useForceProjectSelection.ts
@@ -1,7 +1,7 @@
 import { useEffect } from 'react';
 import { useLocation, useNavigate } from 'react-router-dom-v5-compat';
 
-import { runningTourSignal } from '@kubevirt-utils/components/GuidedTour/utils/constants';
+import { runningTourSignal } from '@kubevirt-utils/components/GuidedTour/utils/guidedTourSignals';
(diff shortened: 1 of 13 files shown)


src/utils/hooks/useForceProjectSelection.ts
```diff
@@ -1,6 +1,7 @@
 import { useEffect } from 'react';
 import { useLocation, useNavigate } from 'react-router-dom-v5-compat';
 
+import { runningTourSignal } from '@kubevirt-utils/components/GuidedTour/utils/constants';
```
pcbailey: This is outside the scope of this PR, but I think we should keep signals in their own files and the file name should indicate that it holds signals. They aren't normal constants.

PR #3176, opened 2025-11-11
CNV-72270: remove resource limit in the resource fetch for multicluster

upalatucci:
## 📝 Description

multicluster sdk and search api have a different concept of LIMIT in the request. 

For console sdk `useK8sWatchResource` hook, the limit is PER REQUEST. This means that if we have more than 10k vms we'll continue to fetch the other 10 001 vms using the `continue` token in the first response.

Multicluster sdk and search api do not have this mechanism so the user will just see 10k vms in their acm view even tho can be more vms 

## Release Notes

* **Refactor**
  * Improved fleet cluster data queries by removing limit restrictions to ensure complete data retrieval.
  * Introduced optimized search polling mechanism for enhanced data fetching efficiency.
  * Updated internal hooks to use consolidated search functionality.


galkremer1 (approved):
/lgtm

PR #3180, opened 2025-11-11
CNV-71976: fixing the navigation to overview tab when changing projects in top consumers

batyana:
## 📝 Description

Changing projects while on top consumers would navigate back to the overview tab. Created a hook to store the active tab and automatically restores it after changing project.

## 🎥 Demo
Before:

https://github.com/user-attachments/assets/98419b2a-4485-4be4-a212-91765264cabe

After:

https://github.com/user-attachments/assets/7f9c93e3-ea85-4868-b213-d9ac0288688c

## Summary by CodeRabbit

* **New Features**
  * Persistent tab state: the cluster overview now remembers and restores the last selected tab when you navigate away and return.
  * Namespace-aware restore: the previously selected tab is preserved when switching namespaces or returning to the overview, keeping your context intact.
  * Smoother return navigation: returning to the overview restores your prior tab so you can continue work without reselecting tabs.

### DIFF
--- a/src/utils/hooks/usePreserveTabDisplay.ts
+++ b/src/utils/hooks/usePreserveTabDisplay.ts
@@ -1,0 +1,49 @@
+import { useEffect, useRef } from 'react';
+import { useLocation, useNavigate, useParams } from 'react-router-dom-v5-compat';
+
+type UsePreserveTabDisplayProps = {
+  basePath: string;
+  storageKey: string;
+};
+
+export const usePreserveTabDisplay = ({ basePath, storageKey }: UsePreserveTabDisplayProps) => {
+  const { ns: namespace } = useParams<{ ns: string }>();
+  const location = useLocation();
+  const navigate = useNavigate();
+  const currentNsRef = useRef(namespace);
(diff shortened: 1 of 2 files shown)


src/utils/hooks/usePreserveTabDisplay.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export const usePreserveTabDisplay = ({ basePath, storageKey }: UsePreserveTabDisplayProps) => {
+  const { ns: namespace } = useParams<{ ns: string }>();
+  const location = useLocation();
+  const navigate = useNavigate();
+  const prevNsRef = useRef(namespace);
+  const prevTabRef = useRef<string>('');
+
+  useEffect(() => {
+    const regex = new RegExp(`${basePath}/([^/]+)`);
```
adamviktora: Can we update this regex to just `${basePath}/(.+)` so it matches everything after `virtualization-overview/` including nested `/`?

In the settings tab, this would keep us on the nested path `settings/user` or `settings/features` without resetting to `settings`

PR #3197, opened 2025-11-18
CNV-72262:Control RBAC for adding volume

aviavissar:
Thanks for checking my  a Pull Request 💖!
## 📝 Description

**after checking with Ronen And Ugo: the demand is Only for cd rom we need the user to have permissions to create data 
For others If the user do not have permission to edit the vm, the user cannot do anything in this menu it should be disabld**

## readUserOnly

image  

## so i disable the Add 

image

## edit with no create user 

image

## Summary by CodeRabbit

* **Improvements**
  * Implemented permission-based access controls for disk operations in virtual machine configuration. Disk creation and modification capabilities are now restricted based on user authorization levels, allowing administrators to enforce fine-grained security policies and control which users can manage storage configurations for individual virtual machines.

### DIFF
--- a/src/views/virtualmachines/details/tabs/configuration/storage/components/tables/disk/DiskList.tsx
+++ b/src/views/virtualmachines/details/tabs/configuration/storage/components/tables/disk/DiskList.tsx
@@ -1,5 +1,7 @@
 import React, { FC } from 'react';
 
+import { useInstanceTypeVMStore } from '@catalog/CreateFromInstanceTypes/state/useInstanceTypeVMStore';
+import VirtualMachineModel from '@kubevirt-ui/kubevirt-api/console/models/VirtualMachineModel';
 import { V1VirtualMachine, V1VirtualMachineInstance } from '@kubevirt-ui/kubevirt-api/kubevirt';
 import DiskListTitle from '@kubevirt-utils/components/DiskListTitle/DiskListTitle';
 import DiskSourceSelect from '@kubevirt-utils/components/DiskModal/components/DiskSourceSelect/DiskSourceSelect';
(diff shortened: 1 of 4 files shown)


src/views/virtualmachines/details/tabs/configuration/storage/components/tables/disk/DiskList.tsx
```diff
@@ -35,9 +39,30 @@ const DiskList: FC<DiskListProps> = ({ customize = false, onDiskUpdate, vm, vmi
...
+        group: VirtualMachineModel.apiGroup,
+        namespace: getNamespace(vm),
+        resource: VirtualMachineModel.plural,
+        verb: 'update' as K8sVerb,
+      },
+    ],
+    [vm],
+  );
+
+  const [results] = useMultipleAccessReviews(accessReviewAttributes);
```
upalatucci: please don't use this function as its not multicluster. use `useFleetAccessReview`  for `create` DataVolumeModel.plural and `update` VirtualMachineModel.plural.

Name the two variables `canCreateDataVolume` and `canUpdate` and pass both into the DiskSourceSelect.

src/views/virtualmachines/details/tabs/configuration/storage/components/tables/disk/DiskList.tsx
```diff
@@ -57,6 +73,8 @@ const DiskList: FC<DiskListProps> = ({ customize = false, onDiskUpdate, vm, vmi
             />
```
galkremer1: Why do you pass 
```
canCreateDataVolume={canCreateDataVolume}
canUpdate={canUpdate}
```
to the child component instead of moving the logic to src/utils/components/DiskModal/components/DiskSourceSelect/DiskSourceSelect.tsx?

galkremer1: Yes it does make sense, nm.

galkremer1 (commented):
/lgtm

PR #3240, opened 2025-12-02
CNV-72907: fix showing 2 default NICs on customizing VM from template

adamviktora:
## 📝 Description

- fixes showing 2 default NICs in `Catalog -> Template -> Customize VirtualMachine -> Disks`
  - the fix is in `getNetworkInterfaceRowData` - removing the part where we add an interface just for viewing purposes in case an `autoattachPodInterface` is true. I think it is misleading to show an interface which is not present in the VM YAML.
  - showing an interface which is not in YAML actually leads to more bugs, e.g. try editing that interface

- the `autoattachPodInterface` (true by default) is now represented by an empty state with an appropriate message (same as on VM details network page)

Possible improvement:
- add a checkbox to empty state to control `autoattachPodInterface: true / false` (right now configurable only from yaml)

## 🎥 Demo

Before:

https://github.com/user-attachments/assets/1a1effeb-4e97-4356-ab7d-2d1d42301bc0

After:

https://github.com/user-attachments/assets/e144221e-fc70-4500-8b64-130020cc2302

Screenshot 2025-12-15 at 12 41 46

## Summary by CodeRabbit

* **Refactor**
  * Simplified network interface data handling and deletion logic for more predictable behavior; removed fallback injection of a default network entry.

* **UI**
  * Updated empty-state for auto-attached networks.
  * Added a contextual warning in the network selector explaining how to enable Pod Networking when applicable.

* **Utilities**
  * Added helper to detect explicitly defined Pod networks and updated pod-network existence checks.

* **Localization**
  * Added guidance strings for Pod Networking (en, es, fr, ja, ko, zh).

### DIFF
--- a/locales/en/plugin__kubevirt-plugin.json
+++ b/locales/en/plugin__kubevirt-plugin.json
@@ -1589,6 +1589,7 @@
   "This user is not allowed to edit this boot source": "This user is not allowed to edit this boot source",
   "This VirtualMachine has": "This VirtualMachine has",
   "This VirtualMachine is down. Please start it to access its console.": "This VirtualMachine is down. Please start it to access its console.",
+  "This VM is using auto attached Pod Network. To disable it, set spec.template.spec.domain.devices.autoattachPodInterface to false.": "This VM is using auto attached Pod Network. To disable it, set spec.template.spec.domain.devices.autoattachPodInterface to false.",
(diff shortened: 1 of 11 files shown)


rszwajko (commented):
We might need to backport the template fixing so I would make it visible in the commit message (maybe even a small dedicated PR for easy automated backporting?).

PR #3249, opened 2025-12-03
CNV-74433: update @openshift-console/dynamic-plugin-sdk dependencies to 4.20

adamviktora:
## 📝 Description

- updates `@openshift-console/dynamic-plugin-sdk` dependencies to 4.20
- changes in `ResourceYAMLEditor` - `initialResource` now needs to be a K8sResourceKind object

## Summary by CodeRabbit

* **Chores**
  * Updated OpenShift dynamic plugin SDK to version 4.20.0.

* **Refactor**
  * Switched several YAML-related flows to use structured resource objects instead of raw YAML strings, improving editor compatibility and type safety.
  * Improved YAML handling across VM templates and resource editors for more reliable parsing and editing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,9 +68,9 @@
   "devDependencies": {
     "@cypress/webpack-preprocessor": "^6.0.2",
     "@kubevirt-ui/kubevirt-api": "^1.4.0",
-    "@openshift-console/dynamic-plugin-sdk": "4.19.0-prerelease.2",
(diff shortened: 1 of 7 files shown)


adamviktora:
I think we can merge Batya's PR first and then I'll rebase this PR and do the change from `defaultVMYamlTemplate` string to an object (just so we don't do conversion if not necessary)

PR #3256, opened 2025-12-08
CNV-74367: backporting fix to cpu and memory metrics from pull #2944

batyana:
## 📝 Description

backporting https://github.com/kubevirt-ui/kubevirt-plugin/pull/2944 changes to cpu and memory metrics. 
For CPU requested metric we don't have to fetch it from prometheus using `kube_pod_resource_request`
We need to take it from the VMI. `kube_pod_resource_request` is not directly related to the VMI CPU requested, but it's scaled using HCO configuration.

For memory we often use `kubevirt_vmi_memory_available_bytes `- `kubevirt_vmi_memory_usable_bytes` but the right metric is `kubevirt_vmi_memory_used_bytes`

### DIFF
--- a/src/utils/components/Charts/utils/queries.ts
+++ b/src/utils/components/Charts/utils/queries.ts
@@ -1,7 +1,6 @@
 import { K8sResourceCommon } from '@openshift-console/dynamic-plugin-sdk';
 
 enum VMQueries {
-  CPU_REQUESTED = 'CPU_REQUESTED',
   CPU_USAGE = 'CPU_USAGE',
   FILESYSTEM_READ_USAGE = 'FILESYSTEM_READ_USAGE',
   FILESYSTEM_USAGE_TOTAL = 'FILESYSTEM_TOTAL_USAGE',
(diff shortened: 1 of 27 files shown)


src/utils/components/Charts/utils/queries.ts
```diff
@@ -24,23 +23,17 @@ enum VMQueries {
 
 type UtilizationQueriesArgs = {
   duration?: string;
-  launcherPodName?: string;
   nic?: string;
   obj: K8sResourceCommon;
 };
 
-type GetUtilizationQueries = ({ duration, launcherPodName, nic, obj }: UtilizationQueriesArgs) => {
+type GetUtilizationQueries = ({ duration, nic, obj }: UtilizationQueriesArgs) => {
```
batyana: removed hubClusterName param- multi-cluster not supported in 4.19

src/views/virtualmachines/list/utils/totalsQueries.ts
```diff
@@ -1,26 +1,21 @@
...
 export const VMTotalsQueries = {
-  TOTAL_CPU_REQUESTED: 'TOTAL_CPU_REQUESTED',
   TOTAL_CPU_USAGE: 'TOTAL_CPU_USAGE',
   TOTAL_MEMORY_USAGE: 'TOTAL_MEMORY_USAGE',
   TOTAL_STORAGE_CAPACITY: 'TOTAL_STORAGE_CAPACITY',
   TOTAL_STORAGE_USAGE: 'TOTAL_STORAGE_USAGE',
 };
 
-export const getVMTotalsQueries = (namespace: string, namespacesList: string[]) => {
+export const getVMTotalsQueries = (namespace: string) => {
```
batyana: removed cluster and allClusters params as 4.19 doesn't support multi-cluster

PR #3277, opened 2025-12-15
CNV-69216: fix checkup not found after creation

upalatucci:
## 📝 Description

After creation the checkup was not found. Getting the checkup by name instead of searching through the namespace checkups do not have this issue and its faster. 

Note: fixing the on delete redirect by trimming the latest path part so

`k8s/ns/namespace/checkups/self-validation/checkupname/yaml` (when deleting from the yaml tab) becomes `k8s/ns/namespace/checkups/self-validation` and `k8s/ns/namespace/checkups/self-validation/checkupname/`  becomes `k8s/ns/namespace/checkups/self-validation` 

Redirecting to the list


galkremer1 (approved):
/lgtm

PR #3325, opened 2026-01-06
CNV-75956: VNC console element has 0px height

avivtur:
## 📝 Description

vnc console was having a 0px inline height which basically made it hidden, adding min-height to fix it.

#### Before

vnc-no-height

#### After

vnc-min-height

## Summary by CodeRabbit

* **Style**
  * Updated the VNC console layout to maintain a minimum height, ensuring consistent visual appearance across different screen configurations.

### DIFF
--- a/src/utils/components/Consoles/components/vnc-console/vnc-console.scss
+++ b/src/utils/components/Consoles/components/vnc-console/vnc-console.scss
@@ -11,6 +11,7 @@
   }
   canvas {
     margin: 0 !important;
+    min-height: 400px;
   }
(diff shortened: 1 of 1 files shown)


adamviktora (approved):
We don't need this change once we merge this PR: https://github.com/kubevirt-ui/kubevirt-plugin/pull/3324

The issue is in `HorizontalNavbar` that had `<Routes>` wrapped in another div recently, without height 100% on it.

But it is good to have this min-height anyways so we don't break it easily next time.

PR #3330, opened 2026-01-07
CNV-76078: Allow non-priv user to create new projects by exposing projects root

batyana:
## 📝 Description

exposed projects root for non-priv user displaying only accessible projects and allowed creating new project from right click menu. Changed the root name in tree view from "All projects" to "Local cluster" for all users to create better consistency with ACM.

## 🎥 Demo

After:
Screenshot at Jan 07 10-16-36
Screenshot at Jan 07 10-16-55

## Summary by CodeRabbit

* **New Features**
  * Added a "Local cluster" label with multilingual support.

* **Refactor**
  * Tree view simplified to always include the Local cluster / All-namespaces item and removed admin-based branching for selection behavior.

* **Localization**
  * Removed an obsolete "no accessible namespaces" translation entry across multiple languages.


src/views/virtualmachines/tree/hooks/useClusterActions.tsx
```diff
@@ -17,8 +16,8 @@ const useClusterActions = (cluster?: string) => {
     () => [
       {
         accessReview: {
-          group: NamespaceModel.apiGroup,
-          resource: NamespaceModel.plural,
+          group: 'project.openshift.io',
+          resource: 'projectrequests',
```
upalatucci: ProjectRequestModel.plural

PR #3345, opened 2026-01-09
CNV-75945: Overview page does not load properly when MCO is not installed

galkremer1:
## 📝 Description

Jira Ticket: [CNV-75945](https://issues.redhat.com/browse/CNV-75945)

When Multi Cluster Observability plugin is not installed on the hub cluster we cannot show any data for the spoke clusters in the overview screen. 

## 🎥 Demo

Before:

https://github.com/user-attachments/assets/ab8e7477-a901-4b9a-b819-9c4e14689b1f

After:

https://github.com/user-attachments/assets/4ff673de-e092-41a7-bf1d-b2b0ad8188da

After (MCO Installed):

https://github.com/user-attachments/assets/d8ffc6c2-428c-4676-9954-36324b0eb66d

## Summary by CodeRabbit

* **New Features**
  * Detects whether Multicluster Observability (MCO) is installed on the hub cluster and shows a loading state while checking.
  * Adds localized MCO-unavailable message entries (en, es, fr, ja, ko, zh) and a tooltip explaining MCO absence.

* **Improvements**
  * Automatically disables the "All Clusters" option when MCO is not installed and surfaces contextual tooltips.
  * Refined cluster redirect behavior to avoid selecting disabled clusters and skip unnecessary redirects.

### DIFF
--- a/src/utils/components/ClusterProjectDropdown/hooks/useDisabledClusterRedirect.ts
+++ b/src/utils/components/ClusterProjectDropdown/hooks/useDisabledClusterRedirect.ts
@@ -33,14 +33,7 @@ export const useDisabledClusterRedirect = ({
 }: UseDisabledClusterRedirectProps): void => {
   useEffect(() => {
     if (!isACMPage) return;
-    if (
-      !clustersLoaded ||
-      !clusterNames ||
-      !cluster ||
-      cluster === ALL_CLUSTERS_KEY ||
-      !disabledClusters ||
-      disabledClusters.length === 0
-    ) {
+    if (!clustersLoaded || !clusterNames || !cluster || !disabledClusters) {
       return;
     }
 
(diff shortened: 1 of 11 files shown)


src/utils/components/ClusterProjectDropdown/hooks/useDisabledClusterRedirect.ts
```diff
@@ -53,8 +46,17 @@ export const useDisabledClusterRedirect = ({
...
-    // If includeAllClusters is true, prefer "all clusters"
-    if (includeAllClusters) {
+    const redirectTo = (next: string | undefined): void => {
+      if (next && next !== cluster) {
+        onClusterChange(next);
+      }
+    };
+
+    // If "All clusters" is disabled (e.g. MCO not installed), don't redirect to it
+    const isAllClustersDisabled = disabledSet.has(ALL_CLUSTERS_KEY);
```
pcbailey: Would naming the variable `mcoNotInstalled` provide a clearer intention of what the variable is meant to convey or are there other possible reasons for all clusters being disabled?

src/utils/hooks/useAlerts/utils/useClusterObservabilityDisabled.ts
```diff
@@ -80,15 +91,46 @@ export const useClusterObservabilityDisabled = (
       }
     });
 
-    return { disabledClusters: disabled, enabledClusters: enabled };
+    return { allClusterNames: all, disabledClusters: disabled, enabledClusters: enabled };
   }, [isACMPage, managedClusterData]);
 
+  // When MCO is not installed, only the hub cluster should be enabled
+  // All spoke clusters should be disabled
+  const enabledClustersWithMCO = useMemo(() => {
```
pcbailey: Do we need to use `useMemo` for all of these values? It doesn't look like we're doing any heavy work within them.

galkremer1: Since we are using the useClusterObservabilityDisabled hook on a base level - it prevents unnecessary renders to the component which is using it. If we don't memoize it, a new array is going to be created each time the hook is called.

src/utils/components/ClusterProjectDropdown/hooks/useDisabledClusterRedirect.ts
galkremer1: I think we should leave it as it is - When the MCO is not installed we are explicitly disabling it as it's being added to `disabledSet`, however we might want to disable the "All clusters" key in other scenarios (and then the suggested variable name is not going describe it accurately).

PR #3432, opened 2026-02-04
CNV-77599: don't connect VMs on IPv6 single-stack cluster to Pod network

adamviktora:
## 📝 Description

This PR adds IPv6 single-stack cluster detection and ensures VMs are not connected to Pod networking on such clusters.

- Adds a `useIPStackType` hook that reads the cluster's IP stack type from the `kubevirt-ui-features` ConfigMap

For an IPv6 single-stack cluster, the default interface and Pod network are omitted from the VM spec during creation across all flows:

- Create from InstanceType (both old catalog and new creation wizard)
- Quick Create from Template (old catalog)
- Customize from Template (old catalog)
- Create from YAML
- New creation wizard

**UI changes**
- Template drawer (both old catalog and new wizard) -- shows a warning alert when the template has a Pod network on an IPv6 single-stack cluster

- Customization step (new wizard) -- removes Pod network from the list

- Add / Edit network interface modal -- "Pod networking" option is hidden on IPv6 single-stack clusters

## 🎥 Demo

After:

https://github.com/user-attachments/assets/7b2de278-cca7-4f95-964a-f561852797a9

After (in new wizard):

https://github.com/user-attachments/assets/1b21704e-6f12-436c-8fe8-ca7b8f7a3f82

## Release Notes

* **New Features**
  * Added support for IPv6 single-stack clusters with automatic Pod networking restrictions in VM creation and customization flows

* **Bug Fixes**
  * Corrected network interface selection label typo

* **Documentation**
  * Added warning messages explaining Pod networking unavailability in IPv6 single-stack cluster environments
  * Updated localization strings across multiple languages


src/utils/hooks/useIPStackType/useIPStackType.ts
```diff
@@ -0,0 +1,19 @@
+import useFeaturesConfigMap from '@kubevirt-utils/hooks/useFeatures/useFeaturesConfigMap';
+
+import { IP_STACK_TYPE_KEY } from './constants';
+import { IPStackType } from './types';
+
+const useIPStackType = (cluster?: string): IPStackType | undefined => {
+  const { featuresConfigMapData } = useFeaturesConfigMap(cluster);
+  const [configMap] = featuresConfigMapData;
+
+  const rawValue = configMap?.data?.[IP_STACK_TYPE_KEY];
```
galkremer1: You can use the type here so you won't need to repeat it twice in lines 12 / 13
```
const rawValue = configMap?.data?.[IP_STACK_TYPE_KEY] as IPStackType;
```

src/utils/resources/vm/utils/network/utils.ts
```diff
@@ -57,3 +62,11 @@ export const getInterfacesAndNetworks = (
 
   return [...withNetwork, ...runtimeInterfacesOnly];
 };
+
+export const removePodNetworkFromVM = (vm: V1VirtualMachine): void => {
```
galkremer1, line 70: We can use current selectors:
```
export const removePodNetworkFromVM = (vm: V1VirtualMachine): void => {
  const devices = getDevices(vm);
  devices.interfaces = getInterfaces(vm)?.filter(
    (iface) => iface.name !== DEFAULT_NETWORK_INTERFACE.name,
  );
  devices.autoattachPodInterface = false;
  vm.spec.template.spec.networks = getNetworks(vm)?.filter((network) => !isPodNetwork(network));
};
```

src/views/virtualmachines/creation-wizard/steps/TemplateStep/components/TemplatesCatalogDrawer/components/TemplateInfoSection.tsx
```diff
@@ -35,6 +40,7 @@ const TemplateInfoSection: FC = memo(() => {
   const interfaces = getTemplateInterfaces(template);
   const [disks] = useWizardDisksTableData(vm);
   const isDefaultTemplate = isDefaultVariantTemplate(template);
+  const hasPodNetwork = useMemo(() => networks?.some((n) => isPodNetwork(n)), [networks]);
```
galkremer1: This is repeated twice so it might be useful to create a helper function for that purpose.

galkremer1 (commented):
A couple of nits

src/utils/resources/vm/utils/network/utils.ts
adamviktora: I'd rather avoid using selectors when mutating the `vm` object. I think we usually provide the full path in these cases.

upalatucci (approved):

PR #3784, opened 2026-04-15
CNV-84748: Add VirtualMachineTemplates to new VM wizard

pcbailey:
## 📝 Description

This PR adds VirtualMachineTemplates support to the new VM wizard.

Note: There is a known issue where VirtualMachineTemplate parameters aren't being generated.

Jira: https://redhat.atlassian.net/browse/CNV-84748

## 🎥 Demo

https://github.com/user-attachments/assets/615a8c4a-b2c6-460f-ae12-c0370e59d328

## Summary by CodeRabbit

* **New Features**
  * Improved template catalog: unified handling of OpenShift and VM templates, generated parameters, and available boot-source detection.
  * Added dynamic storage availability detection for template boot sources.

* **Refactor**
  * Streamlined resource fetching API for more flexible, cluster-aware searches.
  * Consolidated template utilities and selectors for consistent name/label/parameter resolution.

* **Style/Docs**
  * UI text and tab rendering tweaks in template info and drawer views.

### DIFF
--- a/src/utils/resources/template/utils/helpers.ts
+++ b/src/utils/resources/template/utils/helpers.ts
@@ -1,17 +1,27 @@
+import { SetStateAction } from 'react';
 import produce from 'immer';
 
 import {
+  ProcessedTemplatesModel,
   TemplateModel,
   TemplateParameter,
   V1Template,
+  VirtualMachineTemplateRequestModel,
 } from '@kubevirt-ui-ext/kubevirt-api/console';
 import { VirtualMachineModel } from '@kubevirt-ui-ext/kubevirt-api/console';
 import { V1VirtualMachine } from '@kubevirt-ui-ext/kubevirt-api/kubevirt';
-import { getAnnotation, getName, getNamespace } from '@kubevirt-utils/resources/shared';
+import { V1alpha1VirtualMachineTemplateSpecParameters } from '@kubevirt-ui-ext/kubevirt-api/virt-template';
+import { getAnnotation, getLabels, getName, getNamespace } from '@kubevirt-utils/resources/shared';
+import {
+  getParameters,
+  isOpenShiftTemplate,
+  isVirtualMachineTemplate,
+  Template,
+} from '@kubevirt-utils/resources/template';
 import { vmBootDiskSourceIsRegistry } from '@kubevirt-utils/resources/vm/utils/source';
 import { generatePrettyName } from '@kubevirt-utils/utils/utils';
 import { getCluster } from '@multicluster/helpers/selectors';
-import { kubevirtK8sUpdate } from '@multicluster/k8sRequests';
+import { kubevirtK8sCreate, kubevirtK8sUpdate } from '@multicluster/k8sRequests';
 
 import { ANNOTATIONS } from './annotations';
 import {
@@ -37,8 +47,8 @@ export const poorManProcess = (template: V1Template): V1Template => {
   return JSON.parse(templateString);
 };
 
-export const isCommonTemplate = (template: V1Template): boolean =>
-  template?.metadata?.labels?.[TEMPLATE_TYPE_LABEL] === TEMPLATE_TYPE_BASE;
+export const isCommonTemplate = (template: Template): boolean =>
+  getLabels(template)?.[TEMPLATE_TYPE_LABEL] === TEMPLATE_TYPE_BASE;
 
 export const isDeprecatedTemplate = (template: K8sResourceCommon): boolean =>
   getAnnotation(template, ANNOTATIONS.deprecated) === 'true';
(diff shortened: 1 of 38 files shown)


src/utils/resources/template/utils/helpers.ts
```diff
@@ -56,19 +69,20 @@ export const replaceTemplateVM = (template: V1Template, vm: V1VirtualMachine) =>
  * @param {V1Template} template - template
```
adamviktora: change the JSDoc here to `Template`

src/utils/resources/template/utils/helpers.ts
```diff
@@ -126,3 +163,35 @@ export const updateTemplate = (template: V1Template) => {
     ns: getNamespace(template),
   });
 };
+
+export const createProcessedTemplate = <T extends V1alpha1VirtualMachineTemplate | V1Template>(
```
adamviktora: can do `<T extends Template>`

src/utils/resources/template/utils/helpers.ts
```diff
@@ -78,6 +92,29 @@ export const generateParamsWithPrettyName = (template: V1Template) => {
...
+  produce(template, (draftTemplate) => {
+    if (isOpenShiftTemplate(draftTemplate)) draftTemplate.parameters = parameters;
+    if (isVirtualMachineTemplate(draftTemplate))
+      draftTemplate.spec.parameters = parameters as V1alpha1VirtualMachineTemplateSpecParameters[];
+  });
+
+export const createTemplateDraft = (
+  template: Template,
+  namespace: string,
+  parameters: TemplateParameter[] | V1alpha1VirtualMachineTemplateSpecParameters[],
```
adamviktora, line 105: we could possibly unify the type, the only difference of the two types is in the `generate` prop, which is more strict `"expression"` in the latter vs `string` in `TemplateParameter`

src/views/virtualmachines/search/hooks/useAccessibleResources.ts
```diff
@@ -38,24 +55,28 @@ export const useAccessibleResources: UseAccessibleResources = <T>(
     shouldFetchClusterWide
       ? {
           cluster: clusters ? undefined : cluster,
+          fieldSelector,
           groupVersionKind,
           isList: true,
           limit: OBJECTS_FETCHING_LIMIT,
+          namespace,
+          namespaced: true,
```
adamviktora: We should probably use the `namespaced: true` only if `namespace` is provided. It falls back to active namespace if no namespace is provided - this could maybe cause issues

src/views/templates/list/hooks/useVirtualMachineTemplates.ts
```diff
@@ -17,18 +16,20 @@ const useVirtualMachineTemplates: UseVirtualMachineTemplates = (namespace, enabl
...
+    loaded,
+    loadError,
+    resources: vmTemplates,
+  } = useAccessibleResources<V1alpha1VirtualMachineTemplate>(
     enabled
       ? {
-          cluster,
-          groupVersionKind: getGroupVersionKindForModel(VirtualMachineTemplateModel),
-          isList: true,
+          clusters: [cluster],
```
adamviktora: this is problematic to pass the clusters array here. It would change the way the VMTemplates are fetched. In `useAccessibleResources` it ultimately results to `multiSearchData` being returned from `useRedirectWatchHooks`.

This actually might be good and it can solve the bug with line 17:
`const cluster = clusters?.length === 1 ? clusters[0] : undefined;`
Instead of this we can now pass the entire `clusters` from `useListClusters()`.

Not sure if the Fleet search then returns the full resources as expected (e.g. VMs from fleet search have shorter spec). But if so, we can use it.

src/utils/resources/template/utils/helpers.ts
pcbailey: I think that's a good idea, but should be handled in a follow-up.

src/views/templates/list/hooks/useVirtualMachineTemplates.ts
pcbailey: I reverted the changes.

src/utils/resources/template/utils/helpers.ts
```diff
@@ -37,8 +47,8 @@ export const poorManProcess = (template: V1Template): V1Template => {
   return JSON.parse(templateString);
 };
 
-export const isCommonTemplate = (template: V1Template): boolean =>
-  template?.metadata?.labels?.[TEMPLATE_TYPE_LABEL] === TEMPLATE_TYPE_BASE;
+export const isCommonTemplate = (template: Template): boolean =>
+  getLabels(template)?.[TEMPLATE_TYPE_LABEL] === TEMPLATE_TYPE_BASE;
```
galkremer1, line 51: NIT: `getLabel` is probably better for this purpose.

src/utils/resources/template/utils/selectors.ts
```diff
@@ -39,7 +39,7 @@ export const getTemplateVirtualMachineObject = (template: Template): V1VirtualMa
  * @param {Template} template - template
  */
 export const isDefaultVariantTemplate = (template: Template): boolean =>
-  template?.metadata?.labels?.[TEMPLATE_DEFAULT_VARIANT_LABEL] === 'true';
+  getLabels(template)?.[TEMPLATE_DEFAULT_VARIANT_LABEL] === 'true';
```
galkremer1, line 42: `getLabel`

src/views/virtualmachines/search/utils.ts
```diff
@@ -15,3 +17,29 @@ export const getTooltipContent = (filterType: VirtualMachineRowFilterType, t: TF
...
+
+export const getSearchQueries = (
+  searchQueries: AdvancedSearchFilter,
+  clusters: string[],
+): AdvancedSearchFilter | null => {
+  const searchQueriesExist = !isEmpty(searchQueries);
+  const clustersExist = !isEmpty(clusters);
+  const clusterPropertyExists = searchQueries?.find(
+    (query) =>
+      query.property === 'cluster' &&
```
galkremer1, line 29: I think we should use a const for `'cluster'`

galkremer1 (approved):
/lgtm

PR #3813, opened 2026-04-19
CNV-84432: prevent both popovers and guided tour from both appearing and render them conditionally

batyana:
## 📝 Description

Implements a guided tour, welcome modal, and sequential onboarding popover chain for first-time users.

- **Guided tour**: 6-step tour navigates within the user's actual namespace; records seen steps to suppress redundant onboarding hints
- **Welcome modal**: tracks dismissal via signal so onboarding popovers don't appear while it's on screen; hidden on Fleet/ACM pages
- **Onboarding popover chain** (VMsTab → Catalog → CreateProject): each popover waits for its predecessor; popovers whose content was covered by the tour are auto-skipped
- **Non-privileged users**: `useForceProjectSelection` redirects users away from all-namespaces to their first available namespace to fix a non-related existing bug

## 🎥 Demo

After:

https://github.com/user-attachments/assets/528c0fa4-c2c6-408e-a123-4be7b522a333

https://github.com/user-attachments/assets/821776c6-99e7-4fe2-b811-5057dc1b9dfb

https://github.com/user-attachments/assets/0b91c015-b31e-4f63-9d2e-966c7953eec3

https://github.com/user-attachments/assets/a38e6e31-f9bc-44d2-b7da-758d8d2dd613

## Summary by CodeRabbit

* **New Features**
  * Guided tour now records which steps you’ve seen and respects that across sessions; welcome modal dismissal is tracked.
  * Onboarding popovers coordinate with tours (auto-hide when a tour covers them), persist dismissal, and improve VM-tab trigger behavior.

* **Behavior Changes**
  * VM list/tree and overview use namespace-aware sources and conditionally enable watches for more accurate loading and error states.

### DIFF
--- a/src/utils/components/GuidedTour/GuidedTour.tsx
+++ b/src/utils/components/GuidedTour/GuidedTour.tsx
@@ -3,13 +3,14 @@ import Joyride, { ACTIONS, CallBackProps, EVENTS, TooltipRenderProps } from 'rea
 import { useLocation, useNavigate } from 'react-router-dom-v5-compat';
 
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
+import useKubevirtUserSettings from '@kubevirt-utils/hooks/useKubevirtUserSettings/useKubevirtUserSettings';
 import { isEmpty } from '@kubevirt-utils/utils/utils';
 import { useSignals } from '@preact/signals-react/runtime';
 
 import TourPopover from './components/TourPopover/TourPopover';
 import useTour from './hooks/useTour';
 import { getTourSteps } from './utils/constants';
-import { runningTourSignal, stepIndexSignal } from './utils/guidedTourSignals';
+import { runningTourSignal, stepIndexSignal, tourStepsSeenSignal } from './utils/guidedTourSignals';
 import { handleClose, handleNext, handlePrev } from './utils/utils';
 
 const GuidedTour: FC = () => {
@@ -20,6 +21,7 @@ const GuidedTour: FC = () => {
   const navigate = useNavigate();
 
   const { resetTour } = useTour();
+  const [quickStarts, setQuickStarts] = useKubevirtUserSettings('quickStart');
 
   const steps = useMemo(() => getTourSteps(t), [t]);
 
@@ -33,16 +35,35 @@ const GuidedTour: FC = () => {
           document.querySelector(step.target)?.scrollIntoView({ block: 'nearest' });
         }
 
-        if (!isEmpty(route) && location.pathname !== route && runningTourSignal.value) {
-          navigate(route);
-        }
+        const markStepSeen = (stepIndex: number) => {
+          const mergedSeen = Array.from(
+            new Set([
+              ...(quickStarts?.tourStepsSeen || []),
+              ...tourStepsSeenSignal.value,
+              stepIndex,
+            ]),
+          );
+          if (mergedSeen.length !== (quickStarts?.tourStepsSeen || []).length) {
+            setQuickStarts?.({ ...quickStarts, tourStepsSeen: mergedSeen });
+          }
+          if (mergedSeen.length !== tourStepsSeenSignal.value.length) {
+            tourStepsSeenSignal.value = mergedSeen;
+          }
+        };
 
         if (action === ACTIONS.CLOSE) {
+          markStepSeen(index);
           handleClose(resetTour);
           return;
         }
 
+        if (!isEmpty(route) && location.pathname !== route && runningTourSignal.value) {
+          navigate(route);
+        }
+
         if (type === EVENTS.STEP_AFTER) {
+          markStepSeen(index);
+
           if (index !== stepIndexSignal.value) return;
 
           const currentIndex = stepIndexSignal.value;
--- a/src/utils/components/OnboardingPopover/OnboardingPopover.tsx
+++ b/src/utils/components/OnboardingPopover/OnboardingPopover.tsx
@@ -1,81 +1,103 @@
-import React, { FC, ReactNode, useRef, useState } from 'react';
+import React, { FC, ReactNode, useCallback, useEffect } from 'react';
 
+import {
+  runningTourSignal,
+  tourStepsSeenSignal,
+  welcomeModalDismissedSignal,
+} from '@kubevirt-utils/components/GuidedTour/utils/guidedTourSignals';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
 import useKubevirtUserSettings from '@kubevirt-utils/hooks/useKubevirtUserSettings/useKubevirtUserSettings';
 import { Button, ButtonVariant, Popover, PopoverPosition } from '@patternfly/react-core';
+import { useSignals } from '@preact/signals-react/runtime';
 
 import { ONBOARDING_POPOVER_CHAIN } from './constants';
+import { dismissedPopoverKeysSignal } from './onboardingSignals';
 import { OnboardingPopoverKey } from './types';
 
 type OnboardingPopoverProps = {
   bodyContent: ReactNode;
+  coveredByTourSteps?: number[];
   headerContent: ReactNode;
   hideOnTriggerClick?: boolean;
   popoverKey: OnboardingPopoverKey;
-  triggerElement?: HTMLElement;
+  triggerElement: HTMLElement | null;
 };
 
 const OnboardingPopover: FC<OnboardingPopoverProps> = ({
   bodyContent,
-  children,
+  coveredByTourSteps,
   headerContent,
   hideOnTriggerClick = false,
   popoverKey,
   triggerElement,
 }) => {
+  useSignals();
   const { t } = useKubevirtTranslation();
-  const [onboardingPopoversHidden, setOnboardingPopoversHidden, userSettingsLoaded] =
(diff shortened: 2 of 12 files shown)


src/views/virtualmachines/navigator/VirtualMachineNavigator.tsx
```diff
@@ -81,6 +85,9 @@ const VirtualMachineNavigator: FC = () => {
             {!isFleetPage && <GuidedTour />}
             {!isFleetPage && <WelcomeModal />}
             <CatalogOnboardingPopover />
+            <VMsTabOnboardingPopover
+              triggerElement={document.querySelector<HTMLElement>('[data-test="vm-list-tab"]')}
```
adamviktora: This prop can be an inner variable of `VMsTabOnboardingPopover`, like in the `CatalogOnboardingPopover` component.

src/utils/components/OnboardingPopover/components/VMsTabOnboardingPopover.tsx
```diff
@@ -19,9 +26,11 @@ const VMsTabOnboardingPopover: FC = ({ children }) => {
           high-level summary of your current view, including health and alerts.
         </Trans>
       }
+      coveredByTourSteps={[2, 3]}
       headerContent={t('Looking for your VMs?')}
       hideOnTriggerClick
       popoverKey={OnboardingPopoverKey.VMsTab}
+      triggerElement={triggerElement ?? undefined}
     >
       {children}
```
adamviktora: we don't need the `children` prop anymore

src/utils/components/OnboardingPopover/OnboardingPopover.tsx
```diff
@@ -18,34 +26,60 @@ type OnboardingPopoverProps = {
...
   const isVisible =
     userSettingsLoaded &&
     !isHidden &&
     !onboardingPopoversHidden?.[popoverKey] &&
     predecessorsDismissed &&
-    triggerElement !== null;
+    !coveredByTour &&
+    !runningTourSignal.value &&
+    welcomeExperienceDone &&
+    (!!triggerElement || (!triggerElement && React.Children.count(children) > 0));
```
adamviktora: We always use the `triggerElement` now, so the `React.Children.count` check is probably not needed

src/utils/components/OnboardingPopover/OnboardingPopover.tsx
```diff
@@ -59,7 +93,9 @@ const OnboardingPopover: FC<OnboardingPopoverProps> = ({
 
   return (
     <>
-      {!triggerElement && <span ref={spanRef}>{children}</span>}
+      {!triggerElement && React.Children.count(children) > 0 && (
+        <span ref={spanRef}>{children}</span>
```
adamviktora: we can make `triggerElement` a required prop and remove this span, since it's not used anymore

src/utils/components/OnboardingPopover/OnboardingPopover.tsx
```diff
@@ -18,34 +26,60 @@ type OnboardingPopoverProps = {
...
+    chainIndex === -1 ||
+    ONBOARDING_POPOVER_CHAIN.slice(0, chainIndex).every(
+      (key) => onboardingPopoversHidden?.[key] || dismissedPopoverKeysSignal.value.includes(key),
+    );
+
+  const tourStepsSeen = [...(quickStarts?.tourStepsSeen || []), ...tourStepsSeenSignal.value];
+  const coveredByTour = coveredByTourSteps?.some((step) => tourStepsSeen.includes(step)) ?? false;
+
+  useEffect(() => {
+    if (coveredByTour && !dismissedPopoverKeysSignal.value.includes(popoverKey)) {
```
adamviktora: I think we don't need the `dismissedPopoverKeysSignal` to keep state. We already have this information in `onboardingPopoversHidden` + `isHidden` state. 

The reason for using `isHidden` state is because updating `useKubevirtUserSettings` is slow.

We could call `hide()` inside this if statement

src/views/virtualmachines/navigator/VirtualMachineNavigator.tsx
```diff
@@ -30,8 +31,11 @@ import useNavigatorTabs from './useNavigatorTabs';
 
 import './VirtualMachineNavigator.scss';
 
+const VM_PATH_CONDITIONS = [VirtualMachineModelRef];
+
 const VirtualMachineNavigator: FC = () => {
   useSignals();
+  useForceProjectSelection(VM_PATH_CONDITIONS);
```
adamviktora: What is the bug that is being fixed with this? I thought that non-priv users can now view the `all-namespaces` path, when the `useAccessibleResources` is used

src/utils/components/GuidedTour/GuidedTour.tsx
```diff
@@ -33,16 +41,39 @@ const GuidedTour: FC = () => {
           document.querySelector(step.target)?.scrollIntoView({ block: 'nearest' });
         }
 
-        if (!isEmpty(route) && location.pathname !== route && runningTourSignal.value) {
-          navigate(route);
-        }
+        const tourNamespace = namespaceSignal.value;
+        const isRealNamespace = tourNamespace && tourNamespace !== ALL_NAMESPACES_SESSION_KEY;
+        const resolvedRoute = isRealNamespace
+          ? route?.replace(ALL_NAMESPACES, `ns/${tourNamespace}`)
```
adamviktora: why do we need to navigate away from the `ALL_NAMESPACES` path? Also it seems like `namespaceSignal.value` will never have the `ALL_NAMESPACES_SESSION_KEY` value, because its value comes from `useNamespaceParam`

adamviktora (commented):
Nice work! Couple of comments below

src/utils/components/GuidedTour/GuidedTour.tsx
```diff
@@ -33,16 +41,39 @@ const GuidedTour: FC = () => {
...
+          : route;
+
+        const markStepSeen = (stepIndex: number) => {
+          const alreadySeen = quickStarts?.tourStepsSeen || [];
+          if (!alreadySeen.includes(stepIndex)) {
+            setQuickStarts?.({ ...quickStarts, tourStepsSeen: [...alreadySeen, stepIndex] });
+          }
+          if (!tourStepsSeenSignal.value.includes(stepIndex)) {
```
adamviktora: Do we need the `tourStepsSeenSignal`? Isn't `setQuickStarts` enough?

src/utils/components/OnboardingPopover/OnboardingPopover.tsx
```diff
@@ -1,81 +1,109 @@
...
+  const tourDismissedRef = useRef(false);
 
   const chainIndex = ONBOARDING_POPOVER_CHAIN.indexOf(popoverKey);
-  const predecessorsDismissed = ONBOARDING_POPOVER_CHAIN.slice(0, chainIndex).every(
-    (key) => onboardingPopoversHidden?.[key],
-  );
+  const predecessorsDismissed =
+    chainIndex === -1 ||
+    ONBOARDING_POPOVER_CHAIN.slice(0, chainIndex).every(
+      (key) => onboardingPopoversHidden?.[key] || dismissedPopoverKeysSignal.value.includes(key),
```
adamviktora: I see that `dismissedPopoverKeysSignal` is used in order to load the `predecessorsDismissed` variable faster. So then we can remove the `isHidden` variable and only use the signal. We could maybe use a `Set` for the signal and just do `.has` and `.add`

src/utils/components/OnboardingPopover/OnboardingPopover.tsx
```diff
@@ -1,81 +1,109 @@
...
-  const [onboardingPopoversHidden, setOnboardingPopoversHidden, userSettingsLoaded] =
-    useKubevirtUserSettings('onboardingPopoversHidden');
+  const [userSettings, setUserSettings, userSettingsLoaded] = useKubevirtUserSettings();
+  const onboardingPopoversHidden = userSettings?.onboardingPopoversHidden;
+  const quickStarts = userSettings?.quickStart;
   const [isHidden, setIsHidden] = useState(false);
-  const spanRef = useRef<HTMLSpanElement>(null);
-
-  const triggerRef = triggerElement ? () => triggerElement : spanRef;
+  const tourDismissedRef = useRef(false);
```
adamviktora: nit: maybe rename to `popoverDismissedRef`

src/views/virtualmachines/list/components/OverviewTab/OverviewTab.tsx
```diff
@@ -42,20 +44,33 @@ const OverviewTab: FC<OverviewTabProps> = ({ cluster, namespace }) => {
...
   );
+  const {
+    loaded: accessibleVmsLoaded,
+    loadError: accessibleVmsError,
+    resources: accessibleVms,
+  } = useAccessibleResources<V1VirtualMachine>({
+    groupVersionKind: VirtualMachineModelGroupVersionKind,
+  });
+
+  const vms = namespace ? watchedVms : accessibleVms;
```
adamviktora, line 72: Something for a followup - I noticed this pattern of fetching namespaced resources and then falling back to `useAccessibleResources`, we could probably refactor so that the namespace fetching is also done inside `useAccessibleResources` hook.

adamviktora (approved):

PR #3825, opened 2026-04-21
Update dependencies for upcoming Console 4.22 release

vojtechszocs:
## Plugin SDK and shared modules

- Console plugin SDK packages - bump to 4.22 latest prerelease
- React - bump to Console compatible 18.x version
- React i18next - bump to Console compatible 16.x version
- React Router - bump to Console compatible 7.x version
- Remove `react-router-dom` and `react-router-dom-v5-compat`
- Bump `@testing-library/react`, `@testing-library/user-event` and `@testing-library/jest-dom`
- Remove `@testing-library/react-hooks` - replaced by `renderHook` in `@testing-library/react`
- TypeScript - bump to 5.9.x to satisfy peer dependency in `@openshift-console/dynamic-plugin-sdk-webpack`
- Bump `webpack`, `webpack-cli`, `webpack-dev-server` and `fork-ts-checker-webpack-plugin`
- Bump `use-immer` to latest version to avoid unmet peer dependency on React 17.x

## PatternFly

- Update PatternFly 6.x dependencies based on current [Console 4.22 version resolutions](https://raw.githubusercontent.com/openshift/console/refs/heads/release-4.22/frontend/yarn.lock)

## Code changes

- React `FCC` type to work around a type breaking change where `FC` type no longer supports optional `children`
  - Keep using regular `FC` type in React components that don't work with `children`
  - Note - we can't use [TypeScript Declaration Merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) to extend `FC` type since type alias declarations are final and cannot be merged; redefining a type with the same name triggers a naming collision
- Replace `ExtensionDeclaration` type with `Extension` type due to changes in Console [#15778](https://github.com/openshift/console/pull/15778)
- React and i18next - `TFunction` imported directly from `i18next` (no longer re-exported by `react-i18next`)
- Remove unused `window.SERVER_FLAGS.telemetry` object type declaration
- Import all React Router code from `react-router`

## Follow-up improvements

- Remove `FCC` type usage
  - React components that work with `children` should use e.g. `children?: ReactNode` in their prop types
  - Similar change was done in Console as part of React 18 upgrade - https://github.com/openshift/console/pull/15893/commits/695ad334e1251ababc4aa57fee813508e63ecdab
- Remove `FirehoseResource` and `FirehoseResult` type usage
  - Console [#16096](https://github.com/openshift/console/pull/16096) has removed `Firehose` component and related code


@types/kubevirt-plugin/index.d.ts
```diff
@@ -0,0 +1,12 @@
+import './i18next';
+import './react';
+
+declare global {
+  interface Window {
+    SERVER_FLAGS: {
+      authDisabled: boolean;
+      branding: string;
+      nodeArchitectures?: string[];
+    };
```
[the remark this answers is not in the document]
vojtechszocs: `window.SERVER_FLAGS.telemetry` is not used in the code base so I've removed the associated type declaration in `SERVER_FLAGS` object type.

`console-telemetry.md` is meant to document Console Bridge server supported env. variables which are not directly related to `window.SERVER_FLAGS.telemetry` type used when building the plugin.

src/utils/components/DescriptionItem/components/DescriptionItemAnnotations.tsx
```diff
@@ -65,14 +64,8 @@ const DescriptionItemAnnotations: FC<DescriptionItemAnnotationsProps> = ({
...
-      bodyContent={(hide) => (
-        <PopoverContentWithLightspeedButton
-          content={t(
-            'Annotations is an unstructured key value map stored with a resource that may be set by external tools to store and retrieve arbitrary metadata. They are not queryable and should be preserved when modifying objects.',
-          )}
-          hide={hide}
-          promptType={OLSPromptType.ANNOTATIONS}
-        />
+      bodyContent={t(
+        'Annotations is an unstructured key value map stored with a resource that may be set by external tools to store and retrieve arbitrary metadata. They are not queryable and should be preserved when modifying objects.',
```
galkremer1, line 68: Why is `PopoverContentWithLightspeedButton` removed? @pcbailey  can you please confirm this change?

src/utils/components/HelpTextIcon/HelpTextIcon.tsx
```diff
@@ -1,13 +1,20 @@
...
+  PopoverPosition,
+  PopoverProps,
+} from '@patternfly/react-core';
 import { HelpIcon } from '@patternfly/react-icons';
 
 import './HelpTextIcon.scss';
 
 type HelpTextIconProps = {
-  bodyContent: ((hide: () => void) => ReactNode) | ReactNode;
+  bodyContent: PopoverProps['bodyContent'];
```
galkremer1, line 17: Also here - is it going to be an issue @pcbailey with the Lightspeed component?

galkremer1 (commented):
@vojtechszocs I didn't find any issues except these two changes which I would like @pcbailey to confirm that won't cause any regressions.

src/utils/components/HelpTextIcon/HelpTextIcon.tsx
vojtechszocs: @galkremer1
```ts
bodyContent: PopoverProps['bodyContent']
```
infers to
```ts
bodyContent: React.ReactNode | ((hide: () => void) => React.ReactNode)
```
so this change aligns us with underlying PatternFly types.

src/utils/components/DescriptionItem/components/DescriptionItemAnnotations.tsx
vojtechszocs: @galkremer1 

`DescriptionItem` prop `bodyContent` inferred type is
```ts
bodyContent?: React.ReactNode
```
which means that `DescriptionItem` does not appear to support the callback variant
```ts
(hide: () => void) => React.ReactNode
```

vojtechszocs: But ultimately the `bodyContent` prop is passed down to PatternFly `Popover` so it could be possible to retain existing callback above but it would yield some code refactoring.

Check out `DescriptionItemPopoverContent` where `bodyContent` and `hide` are passed to `renderBodyContent` function.

adamviktora: This is the correct change I think. We shouldn't pass the entire `PopoverContentWithLightspeedButton` into the `bodyContent` - because it is already wrapped in `PopoverContentWithLightspeedButton` later on in `DescriptionItemHeader` -> `DescriptionItemPopoverContent`.

It follows the same principle as the other components that use `DescriptionItem`

adamviktora (approved):

galkremer1 (approved):
/lgtm

upalatucci:
@coderabbitai why `configure public ssh key` test is failing every time?

PR #3840, opened 2026-04-26
CNV-85419: Fix create vm wizard to reset upon closing

batyana:
## 📝 Description
The `useInstanceTypeVMStore` (holding `operatingSystemType` and preference) was never cleared when the VM creation wizard was initialized or closed, causing the Guest OS selection to persist across wizard sessions.

Added a `resetInstanceTypeVMState` action to `useInstanceTypeVMStore` and call it both on wizard mount and on wizard close/cancel, ensuring the Guest OS step always starts fresh.

## 🎥 Demo

Before:

https://github.com/user-attachments/assets/f336f331-213c-439d-ab57-b5f765e054a0

After:

https://github.com/user-attachments/assets/72ac1e9d-7ba0-4045-8a38-3cd3b3df1b7d

## Summary by CodeRabbit

* **Bug Fixes**
  * VM Creation Wizard now reliably resets to default settings when starting or closing a session, ensuring previous customization choices do not persist and preventing stale data from appearing between uses.

### DIFF
--- a/src/views/virtualmachines/creation-wizard/hooks/useCloseWizard.ts
+++ b/src/views/virtualmachines/creation-wizard/hooks/useCloseWizard.ts
@@ -1,6 +1,5 @@
 import { useNavigate, useParams } from 'react-router';
 
-import { clearCustomizeInstanceType } from '@kubevirt-utils/store/customizeInstanceType';
 import useClusterParam from '@multicluster/hooks/useClusterParam';
 import { getVMListURL } from '@multicluster/urls';
 import useVMWizardStore from '@virtualmachines/creation-wizard/state/vm-wizard-store/useVMWizardStore';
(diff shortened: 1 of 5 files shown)


src/views/virtualmachines/creation-wizard/hooks/useCloseWizard.ts
```diff
@@ -3,19 +3,22 @@ import { useNavigate, useParams } from 'react-router';
...
 const useCloseWizard: UseCloseWizard = () => {
   const navigate = useNavigate();
   const { resetWizardState } = useVMWizardStore();
+  const { resetInstanceTypeVMState } = useInstanceTypeVMStore();
   const clusterParam = useClusterParam();
   const { ns } = useParams<{ ns: string }>();
   const vmListURL = getVMListURL(clusterParam, ns);
 
   const closeWizard = () => {
     resetWizardState();
```
galkremer1, line 17: Is `resetWizardState` called somewhere else without `resetInstanceTypeVMState` and `clearCustomizeInstanceType`? If not - I think that `resetInstanceTypeVMState` and `clearCustomizeInstanceType` should be called within `resetWizardState`.

galkremer1 (approved):
/lgtm

PR #3841, opened 2026-04-26
CNV-85192: Fix vm name display to load immediatly with badge

batyana:
## 📝 Description

The VM name was missing from the title while the resource loaded, even though it's already in the URL. Pass `name/namespace` from URL params as a fallback so the heading renders immediately. Status and actions still wait for the full resource.

## 🎥 Demo

Before:

https://github.com/user-attachments/assets/1140b171-dcad-47d2-ae23-1debb0c758a7

After:

https://github.com/user-attachments/assets/b37c9741-5e97-41ba-a7c5-b6659361a003

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Virtual Machine details page title stability: when VM data is missing or incomplete, the page now falls back to URL parameters (name and namespace) so a meaningful title is shown consistently during load or error states.

### DIFF
--- a/src/views/virtualmachines/details/VirtualMachineNavPageTitle.tsx
+++ b/src/views/virtualmachines/details/VirtualMachineNavPageTitle.tsx
@@ -1,5 +1,5 @@
 import React, { FC } from 'react';
-import { useLocation } from 'react-router';
+import { useLocation, useParams } from 'react-router';
 
 import { V1VirtualMachine } from '@kubevirt-ui-ext/kubevirt-api/kubevirt';
 import DetailsPageTitle from '@kubevirt-utils/components/DetailsPageTitle/DetailsPageTitle';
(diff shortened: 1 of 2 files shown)


adamviktora (approved):

src/views/virtualmachines/details/VirtualMachineNavPageTitle.tsx
```diff
@@ -27,19 +27,23 @@ import { vmTabsWithYAML } from './utils/constants';
...
+  namespaceFromUrl,
   vm,
 }) => {
   const { t } = useKubevirtTranslation();
   const location = useLocation();
 
-  const name = getName(vm);
-  const namespace = getNamespace(vm);
+  const name = getName(vm) || nameFromUrl;
+  const namespace = getNamespace(vm) || namespaceFromUrl;
```
upalatucci: Can we directly use the name and namespace from `useParams` hook ? Wihtout addining additional props to this component?

adamviktora (approved):

PR #3870, opened 2026-04-29
CNV-85902: fix creating vm for non-priv users

batyana:
## 📝 Description

This PR fixes several regressions that prevented non-privileged users from creating VMs through the new creation wizard.

Template catalog (non-admin users)

Skip cluster-wide Template watch for non-admin users; use per-namespace watches instead to avoid 403 floods
Fix useVMTemplateFeatureFlag stuck loading when HyperConverged CR is inaccessible
Fix processedtemplates API calls using the correct namespace (wizard's selected project)
Restrict OpenShift template watches to openshift + user namespace via watchNamespaces
VM provisioning (CDI smart clone)

Add vm.kubevirt.io/template and vm.kubevirt.io/template.namespace labels to VMs created from templates. Without these labels, virt-api cannot generate a CDI clone token, forcing host-assisted clone which fails for non-admin users due to cross-namespace RBAC and causes ResourceQuota exhaustion from orphaned PVCs
Boot source (instance type flow)

Non-admin users now default to openshift-virtualization-os-images (or kubevirt-os-images upstream) instead of a blank cluster-wide watch that returns 403. Users can still switch to another project via the dropdown
Wizard namespace

Wizard initializes with the correct project: non-admin users always get their active namespace; admin users keep their previously selected project (falling back to active namespace)

## 🎥 Demo

Before:

https://github.com/user-attachments/assets/a2d1e1d7-17ea-4ef2-8c7d-668111d87ae7

After:

https://github.com/user-attachments/assets/ac7aad3c-a15f-462f-b055-ba75ed2fb791

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable role-based template visibility and access for admins vs non-admins.
  * VM creation wizard preserves/initializes project selection correctly for non-admins and admins.
  * Loading state reporting improved for template-related actions.

* **Refactor**
  * Boot source and bootable volume lists scoped correctly by effective namespace based on role/selection.
  * Template watching and namespace handling optimized to reduce unnecessary watches and improve load signaling.
  * Generated VMs always include stable template identity labels.

* **New Features**
  * Template generation and resource-watching accept explicit namespace overrides and watch-list options.

### DIFF
--- a/src/views/virtualmachines/creation-wizard/steps/InstanceTypesSteps/BootSourceStep/BootSourceStep.tsx
+++ b/src/views/virtualmachines/creation-wizard/steps/InstanceTypesSteps/BootSourceStep/BootSourceStep.tsx
@@ -1,9 +1,11 @@
 import React, { FC } from 'react';
 
+import { ALL_PROJECTS } from '@kubevirt-utils/hooks/constants';
 import useInstanceTypesAndPreferences from '@kubevirt-utils/hooks/useInstanceTypesAndPreferences';
+import { useIsAdmin } from '@kubevirt-utils/hooks/useIsAdmin';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
 import useBootableVolumes from '@kubevirt-utils/resources/bootableresources/hooks/useBootableVolumes';
-import { getValidNamespace } from '@kubevirt-utils/utils/utils';
+import { getValidNamespace, OS_IMAGES_NS } from '@kubevirt-utils/utils/utils';
 import {
   Radio,
   Split,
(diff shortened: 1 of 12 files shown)


src/views/virtualmachines/creation-wizard/steps/InstanceTypesSteps/BootSourceStep/BootSourceStep.tsx
```diff
@@ -21,13 +23,20 @@ import AddBootableVolumeButton from './components/AddBootableVolumeButton';
...
   );
-  const { volumeListNamespace } = useInstanceTypeVMStore();
-  const bootableVolumesData = useBootableVolumes(volumeListNamespace);
+
+  // For non-admin users with no explicit selection, always scope to the OS images
+  // namespace — they cannot do cluster-wide watches. Respect any explicit selection.
+  const effectiveNamespace =
+    isAdmin || (volumeListNamespace && volumeListNamespace !== ALL_PROJECTS)
+      ? volumeListNamespace
+      : OS_IMAGES_NS;
```
upalatucci, line 38: i think OS_IMAGES_NS can be configured using the HCO configuration. We should have a way to fetch this namespace. HCO is not available for all users. Only `useKubevirtHyperconvergeConfiguration` will be available to all users

src/views/virtualmachines/search/hooks/useAccessibleResources.ts
```diff
@@ -40,14 +44,28 @@ export const useAccessibleResources: UseAccessibleResources = <T>({
...
 
+  const namespacesToWatch = useMemo(() => {
+    if (watchNamespaces) return watchNamespaces;
+    if (!onlyUserProjects || !projectNames) return projectNames;
+    const filtered = projectNames.filter(
+      (ns) => !isSystemNamespace(ns) || ns === OPENSHIFT_NAMESPACE,
+    );
+    if (!filtered.includes(OPENSHIFT_NAMESPACE)) {
+      filtered.push(OPENSHIFT_NAMESPACE);
+    }
```
upalatucci: `OPENSHIFT_NAMESPACE` is specific to just template. Not all resources

src/views/virtualmachines/creation-wizard/steps/InstanceTypesSteps/BootSourceStep/BootSourceStep.tsx
batyana: as far as i can tell this is an existing issue as there isn't a way to access `commonBootImageNamespace` not through HCO and non-admin doesn't have permissions to HCO. previously there was no default for non-priv users now we're defaulting to it but I don't see how a user can fetch the ns.
maybe we can bring it up as a followup but  `OS_IMAGES_NS` is defined to determine the image many other places as well.

upalatucci (approved):

upalatucci:
/override ci/prow/kubevirt-e2e-aws

unrelated test failing

upalatucci (approved):

PR #3872, opened 2026-04-29
CNV-85594: fix operation insecure error when creating new vm through firefox browser

batyana:
## 📝 Description

in Firefox caused by exceeding the 50 `history.pushState() `calls/10s limit in the VM template catalog. All filter URL updates now use `replaceState`, `clearAll` calls `navigate()` once, stale debounce timers are cancelled on clear, and `useHideDeprecatedTemplateTiles` runs only on mount. Also adds a contextual empty state for namespaces with no templates, and consolidates duplicate catalog code shared between the standalone catalog and VM creation wizard.

## 🎥 Demo
Before:

https://github.com/user-attachments/assets/d357b6fa-a91a-43b1-8e35-03f297a8d1e4

https://github.com/user-attachments/assets/0225e385-97a4-46c8-b69a-ef7c85b4653d

After:

https://github.com/user-attachments/assets/179f28a1-483e-40b9-895f-99ab77130b94

## Release Notes

* **New Features**
  * Improved empty-state messaging to distinguish between "no templates in your project" and "no templates matching current filters."
  * Added "View all projects" action in empty-state scenarios.
  * Filter dropdown now auto-closes when an option is selected.

* **Bug Fixes**
  * Search input now properly clears when filters are reset.
  * Deprecated templates are now hidden by default during template selection.

* **Localization**
  * Added new UI strings in English, Spanish, French, Japanese, Korean, and Chinese.


upalatucci (approved):

adamviktora (dismissed):
about this part:

> and consolidates duplicate catalog code shared between the standalone catalog and VM creation wizard

Can you please change it so we keep it duplicated?

Because we will eventually remove all the code from `views/catalog` page.

src/views/catalog/templatescatalog/components/TemplatesCatalogEmptyState.tsx
adamviktora, line 1: I think we don't need to do changes to any files in `views/catalog` as the page is unused

upalatucci (approved):

PR #3878, opened 2026-04-30
CNV-85836: VM creation wizard: keep creation method tiles visible when resizing

kuklas:
https: //issues.redhat.com/browse/CNV-85836
Made-with: Cursor

## 📝 Description
Use a wrapping Flex layout and a fluid tile group width instead of a fixed-width Split row so all three options stay reachable in narrower wizard panels and browser widths.

## 🎥 Demo
Before
Before resizing

After
Resizing create option

## Summary by CodeRabbit

* **Style**
  * Enhanced responsive layout for the virtual machine creation wizard's creation method selection, with flexible tile sizing and improved spacing that adapts across various screen resolutions and devices.


upalatucci (approved):
thank u @kuklas

PR #3879, opened 2026-04-30
CNV-86227: Use radios for specific node in compute migration

kuklas:
https://redhat.atlassian.net/browse/CNV-86227
Made-with: Cursor

## 📝 Description
Replace row checkboxes with radio buttons in the migrate-compute nodes table so manual node selection reads as single-select. Selection no longer toggles off when clicking the same row again.

## 🎥 Demo
Before
image

After
image


upalatucci (approved):
thanks @kuklas

PR #3918, opened 2026-05-08
CNV-85829: auto-select default size when switching Compute Resource series

upalatucci:
## Summary
- Fixes a bug where switching to a different instance type series (e.g., from General Purpose to Compute Exclusive) in the VM Wizard Compute Resources step left the size dropdown blank and the Next button disabled
- Default size is now selected at the point of user interaction in `RedHatSeriesMenuCard`: the first standard (non-hugepages) size for the chosen series is set immediately when the series card is clicked

## Root cause
`setSelectedSeries` resets `selectedSize` to `''` but nothing re-populated it when the user picked a new series. The initial "General Purpose" selection had a size pre-filled from the boot volume's instance type annotation — switching series lost that.

## Changes
- `RedHatSeriesMenuCard.tsx`: `handleSeriesClick` now computes the first standard size from `rhSeriesItem.sizes` and calls `setSelectedSize` right after `setSelectedSeries`

### Demo

**BEFORE**

https://github.com/user-attachments/assets/f82ea9df-145f-4d24-93fd-9b8e75cd6ec8

**AFTER**

https://github.com/user-attachments/assets/3b63e5e4-b4ed-41e6-ad4f-5354b1295028

Jira: https://issues.redhat.com/browse/CNV-85829

## Release Notes

* **Bug Fixes**
  * When selecting an instance type series in the virtual machine creation wizard, a default size is now automatically selected, streamlining the configuration process.
  * Selecting an already-selected series is now handled more efficiently.


adamviktora (commented):
The [coderabbit comment](https://github.com/kubevirt-ui/kubevirt-plugin/pull/3918#discussion_r3210410442) is worth addressing, otherwise LGTM

adamviktora (approved):

PR #3920, opened 2026-05-11
CNV-86103: remove the project filter from the create from clone in create wizard

batyana:
## 📝 Description

VMs listed in the Clone source step were only fetched from the namespace selected in step 1, but the project filter dropdown allowed users to select a different project — showing no results. Removed the interactive project filter and now displays the project as a read-only label instead.

## 🎥 Demo

Before:

https://github.com/user-attachments/assets/e4a84c36-d23a-4853-ba8e-795dcfa051f6

After:

https://github.com/user-attachments/assets/19aec2e5-85fe-4918-a158-6b7e59960120

## Summary by CodeRabbit

* **Refactor**
  * Streamlined clone-source filtering in the VM creation wizard: project filter handling simplified and moved to an explicit "Project:" label when a target is set; dropdown filters consolidated to focus on row-based filtering for status, OS, storage, hardware, scheduling, nodes, guest agent, and architecture.

### DIFF
--- a/locales/en/plugin__kubevirt-plugin.json
+++ b/locales/en/plugin__kubevirt-plugin.json
@@ -1566,6 +1566,7 @@
   "Project name to clone the template to": "Project name to clone the template to",
   "Project root": "Project root",
   "Project-scoped": "Project-scoped",
+  "Project: {{project}}": "Project: {{project}}",
   "Projects": "Projects",
   "Provider": "Provider",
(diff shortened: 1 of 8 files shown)


upalatucci:
@batyana i would add a little bit of margin between the label and the list

batyana:
@upalatucci @adamviktora done :)
image

adamviktora (approved):

PR #3923, opened 2026-05-11
CNV-86642: tooltip doesn't appear in treeview right click menu

batyana:
## 📝 Description

Disabled "Migrate Compute" and "Migrate Storage" actions in the tree view right-click menu were not showing tooltips explaining why they are disabled, unlike the kebab menu and Actions dropdown which do. The root cause was PatternFly's `MenuItem` setting `pointer-events: none` on disabled items, blocking all hover events. Fixed by using `isAriaDisabled` instead of `isDisabled` and wrapping disabled items with an explicit Tooltip component. Tooltip position is set to right in the tree view context to avoid overlapping the menu.

## 🎥 Demo

Before:

https://github.com/user-attachments/assets/89cb0301-e033-47e1-aaef-8d1551ccace9

After:

https://github.com/user-attachments/assets/98c4951c-d813-4c65-a72a-29c40cd0a468

## Summary by CodeRabbit

* **Style**
  * Tooltips for dropdown and right-click menus now consistently appear on the right and respect z-index for proper layering.
  * Submenu items inherit tooltip positioning for consistent behavior.

* **Bug Fixes**
  * Disabled menu items now display explanatory tooltips via conditional wrapping, improving discoverability while preserving accessibility.

### DIFF
--- a/src/utils/components/ActionDropdownItem/ActionDropdownItem.tsx
+++ b/src/utils/components/ActionDropdownItem/ActionDropdownItem.tsx
@@ -1,6 +1,13 @@
 import React, { Dispatch, FC, SetStateAction } from 'react';
 
-import { Menu, MenuContent, MenuItem, MenuList, TooltipPosition } from '@patternfly/react-core';
+import {
+  Menu,
+  MenuContent,
+  MenuItem,
+  MenuList,
+  Tooltip,
+  TooltipPosition,
+} from '@patternfly/react-core';
 import { useFleetAccessReview } from '@stolostron/multicluster-sdk';
 
 import { ActionDropdownItemType } from '../ActionsDropdown/constants';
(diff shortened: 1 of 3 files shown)


src/utils/components/ActionDropdownItem/ActionDropdownItem.tsx
```diff
@@ -64,6 +73,20 @@ const ActionDropdownItem: FC<ActionDropdownItemProps> = ({ action, setIsOpen })
       )}
     </MenuItem>
   );
+
+  if (showTooltip) {
+    return (
+      <Tooltip
+        content={action.disabledTooltip}
+        position={tooltipPosition ?? TooltipPosition.left}
+        zIndex={10002}
```
adamviktora: I'd add a comment explaining the specific zIndex number, also maybe there is a semantic variable for this in `@patternfly/react-tokens`

adamviktora (commented):
Looking good! Just one nit

adamviktora (approved):

PR #3951, opened 2026-05-15
CNV-86754: filter migration node list by VM architecture

upalatucci:
## 📝 Description

In heterogeneous clusters (mixed amd64/arm64), the "Migrate to Specific Node" dialog showed all worker nodes regardless of CPU architecture compatibility. This caused users to select incompatible nodes, leading to migration failures.

**Changes:**
- Filter candidate nodes by CPU architecture
- Architecture is resolved via a fallback chain:
  1. `spec.template.spec.architecture` — explicit KubeVirt field (set by virt-api defaulting on new VMs)
  2. Current node's `status.nodeInfo.architecture` — inferred from where the VMI is already running. Safe fallback because migration requires a running VM, so the current node is always available. Handles old VMs created before the `MultiArchitecture` feature gate, which have no architecture field set
- Node architecture is read from `node.status.nodeInfo.architecture` (kubelet-reported, authoritative) 
- Add `getNodeArchitecture()` utility in node selectors
- Show a helper text banner when architecture filtering is active (e.g. "Showing only nodes with arm64 architecture, matching this VirtualMachine.")
- Show an arch-specific empty state when no compatible nodes exist (e.g. "No nodes with arm64 architecture found")

## 🔗 Links

- https://issues.redhat.com/browse/CNV-86754

## 🎥 Demo

Screenshot 2026-05-15 at 17 29 45

NO NODES FOUND WITH VM ARCH

Screenshot 2026-05-15 at 17 37 34

## Summary by CodeRabbit

* **Improvements**
  * Migration now detects the VM's architecture and only shows compatible target nodes.
  * UI updates: search bar and architecture helper text are better laid out; table message changes to indicate when no nodes match the VM architecture.
* **Bug Fixes**
  * Default architecture handling added so nodes without reported architecture fall back to a sensible default.


locales/en/plugin__kubevirt-plugin.json
```diff
@@ -1848,6 +1849,7 @@
   "Show virtualization health alerts": "Show virtualization health alerts",
   "Show VirtualMachine per InstanceTypes": "Show VirtualMachine per InstanceTypes",
   "Show VirtualMachine per Templates": "Show VirtualMachine per Templates",
+  "Showing only nodes with {{arch}} architecture, matching this VirtualMachine.": "Showing only nodes with {{arch}} architecture, matching this VirtualMachine.",
```
galkremer1, line 1852: I think it can be more user friendly to highlight `{{arch}}`

galkremer1: Also maybe we can move it to the "Manual node selection" info box. In my opinion it makes sense as we are also filtering out nodes that do not meet the required criteria such as unscheduled nodes. 
@yfrimanm  WDYT?

src/utils/resources/node/utils/selectors.ts
```diff
@@ -8,3 +9,12 @@ import { IoK8sApiCoreV1Node } from '@kubevirt-ui-ext/kubevirt-api/kubernetes';
...
   node?.status?.conditions;
+
+/**
+ * Returns the architecture reported by the node via status.nodeInfo.architecture,
+ * falling back to DEFAULT_ARCHITECTURE if not available.
+ * @param {IoK8sApiCoreV1Node} node
+ * @returns {string} the node's architecture
+ */
+export const getNodeArchitecture = (node: IoK8sApiCoreV1Node): string =>
+  node?.status?.nodeInfo?.architecture || ARCHITECTURES.AMD64;
```
galkremer1: I suggest adding a `getDefaultArchitecture` function and then return `ARCHITECTURES.AMD64`

locales/en/plugin__kubevirt-plugin.json
upalatucci: talked with yifat. We'll stick with this solution. Better to separate the two different things that we explain. but thanks for the feedback

galkremer1 (approved):
/lgtm

PR #3953, opened 2026-05-15
CNV-75009: Use a shared component for ack confirmation modals

galkremer1:
## 📝 Description

We currently have two modals which are very similar:

HeavyLoadCheckupConfirmationModal.tsx

ConfirmVMActionBaseModal.tsx

Create a shared component and reuse it instead of having two separate components. 

## 🔗 Links

Jira ticket: [CNV-75009](https://redhat.atlassian.net/browse/CNV-75009)

## 🎥 Demo

Working as expected after the code changes:

https://github.com/user-attachments/assets/0005aedd-8c1b-4ec4-983a-25df48afca49

## Summary by CodeRabbit

* **Refactor**
  * Consolidated confirmation modal UI across multiple dialogs for a consistent experience.

* **New Features**
  * Optional "check to confirm" checkbox for risky actions.
  * Confirm buttons show loading/disabled state while actions run.
  * Improved warning/danger severity visuals and messaging in confirmation dialogs.

* **Bug Fixes**
  * More robust handling of bulk VM actions to reduce partial-failure surprises.

### DIFF
--- a/src/utils/components/AckConfirmationModal/AckConfirmationModal.tsx
+++ b/src/utils/components/AckConfirmationModal/AckConfirmationModal.tsx
@@ -1,4 +1,4 @@
-import React, { FC, ReactNode, useState } from 'react';
+import React, { FC, ReactNode, useEffect, useState } from 'react';
 
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
 import {
@@ -13,24 +13,26 @@ import {
   StackItem,
 } from '@patternfly/react-core';
 
-type ConfirmVMActionBaseModalProps = {
-  action: () => Promise<string | void>;
+export type AckConfirmationModalProps = {
+  action: () => Promise<string | void> | void;
   actionLabel: string;
-  actionType: string;
+  actionType?: string;
   checkToConfirmMessage?: string;
   children?: ReactNode;
+  className?: string;
   isOpen: boolean;
   onClose: () => void;
   severityVariant?: 'danger' | 'warning' | undefined;
   title: string;
 };
 
-const ConfirmVMActionBaseModal: FC<ConfirmVMActionBaseModalProps> = ({
+const AckConfirmationModal: FC<AckConfirmationModalProps> = ({
   action,
   actionLabel,
   actionType,
   checkToConfirmMessage,
   children,
+  className,
   isOpen,
   onClose,
   severityVariant,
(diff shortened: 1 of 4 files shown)


upalatucci (approved):

src/utils/components/AckConfirmationModal/AckConfirmationModal.tsx
```diff
@@ -0,0 +1,105 @@
...
+          )}
+        </Stack>
+      </ModalBody>
+      <ModalFooter>
+        <Button
+          isDisabled={!isChecked || isSubmitting}
+          isLoading={isSubmitting}
+          key="confirm"
+          onClick={submitHandler}
+          variant={severityVariant ?? ButtonVariant.primary}
```
adamviktora: I checked [PF guidelines for Modals](https://www.patternfly.org/components/modal/design-guidelines#icon-use-in-modal-dialogs) and we were previously doing some things wrong regarding the severity. We shouldn't show warning button and danger icon.

The variants should be:

| severity | Button variant | titleIconVariant |
|----------|----------------|------------------|
| danger   | danger         | warning          |
| warning  | primary        | warning          |
| undefined  | primary        | -          |

Can you please fix this as part of this PR?

Maybe we could also then remove the `titleIconVariant` prop. The only time when the `titleIconVariant` should be "danger" is when indicating errors.

Btw maybe we could use a `WarningModal` component: https://www.patternfly.org/component-groups/error-communication/warning-modal#warning-modal-with-a-checkbox - not sure if it's suitable though, and if so, can be a followup

galkremer1: Thanks Adam - I just pushed the relevant changes. 
As for the `WarningModal` suggestion - I tested it and it has some limitations from what I can see regarding the spacing of the body text and the checkbox. It might be OK to do the conversion however we will need a confirmation from Yifat.

src/views/virtualmachines/actions/components/ConfirmMultipleVMActionsModal/components/ConfirmVMActionBaseModal.tsx
```diff
@@ -35,54 +24,20 @@ const ConfirmVMActionBaseModal: FC<ConfirmVMActionBaseModalProps> = ({
...
-    </Modal>
-  );
-};
+}) => (
+  <AckConfirmationModal
+    action={action}
+    actionLabel={actionLabel}
+    actionType={actionType}
+    checkToConfirmMessage={checkToConfirmMessage}
+    className="confirm-multiple-vm-actions-modal"
```
adamviktora: Can we remove this `ConfirmVMActionBaseModal` component and use `AckConfirmationModal` directly? We are just passing the same props. 

The `confirm-multiple-vm-actions-modal` class is unused anyway (the only class used applied from that scss file is `confirm-multiple-vm-actions-modal__popover`)

galkremer1: Done

adamviktora (approved):

PR #3955, opened 2026-05-17
CNV-84105: adjust the create vm wizard footer padding

batyana:
## 📝 Description

The cancel button in the create vm wizard was hidden by lightspeed icon when enabled. The Console's QuickStartDrawer.scss applies `justify-content: space-between` to all `.pf-v6-c-action-list` elements, pushing the Cancel button away from Back/Next in the VM creation wizard footer. This overrides it with `flex-start `to keep the buttons grouped together according to [PF guidelines](https://www.patternfly.org/components/wizard/react-deprecated/#wizard-in-modal). 

## 🔗 Links

Jira ticket: https://redhat.atlassian.net/browse/CNV-84105

## 🎥 Demo

Before:
image

After:
image

## Summary by CodeRabbit

* **Style**
  * Unified and improved alignment of actions in the VM creation wizard footer for consistent button spacing and layout across all steps.
* **Chores**
  * Removed conditional spacing behavior so cancel/next buttons render consistently throughout the wizard.

### DIFF
--- a/src/views/catalog/wizard/Wizard.scss
+++ b/src/views/catalog/wizard/Wizard.scss
@@ -19,6 +19,12 @@
   z-index: 1000;
 }
 
+.vm-creation-wizard {
+  .pf-v6-c-wizard__footer .pf-v6-c-action-list {
+    justify-content: flex-start;
+  }
+}
+
 .wizard-header {
   .pf-l-split {
     justify-content: space-between;
--- a/src/views/virtualmachines/creation-wizard/VMCreationWizard.tsx
+++ b/src/views/virtualmachines/creation-wizard/VMCreationWizard.tsx
@@ -1,12 +1,10 @@
 import React, { FC, useEffect, useRef } from 'react';
-import classnames from 'classnames';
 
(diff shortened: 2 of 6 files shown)


adamviktora (commented):
Hi @batyana, this is fine, but there are places in code [like this](https://github.com/kubevirt-ui/kubevirt-plugin/blob/main/src/views/virtualmachines/creation-wizard/components/DefaultWizardFooter.tsx#L21) which were meant to address this issue

`cancelButtonProps={{ className: classnames({ 'pf-v6-u-mr-4xl': hasOLSConsole }) }}`

It should apply the class if lightspeed is enabled, but apparently it doesn't work. Can you please check what is wrong there?

If we can apply the padding conditionally, it would be nice

upalatucci (approved):

adamviktora:
we should remove the parts of code that do the `cancelButtonProps={{ className: classnames({ 'pf-v6-u-mr-4xl': hasOLSConsole }) }}` if that class is not being applied, otherwise looking good

adamviktora (approved):

PR #4017, opened 2026-06-03
CNV-88903: Lock-in the bootable volume in the custom configuration fl…

benbassatido:
## 📝 Description
When opening the "Add volume" modal from the VM creation wizard, the preference field is now pre-filled and locked to match the Guest OS selected in the previous step.

## 🔗 Links

https://redhat.atlassian.net/browse/CNV-88903

## 🎥 Demo

Before: 
image

After: 
image

## Summary by CodeRabbit

* **New Features**
  * Modal can be opened with a locked preference so users cannot change the preference during bootable volume creation; UI shows an explanatory message and disables the selector.

* **Bug Fixes**
  * Rejects invalid/unmapped preference selections and improves preference-based filtering and empty-state behavior.

* **Refactor**
  * Preference storage changed from plain strings to typed objects and default preference cleared to null for more consistent handling.

### DIFF
--- a/src/utils/components/AddBootableVolumeModal/AddBootableVolumeModal.tsx
+++ b/src/utils/components/AddBootableVolumeModal/AddBootableVolumeModal.tsx
@@ -23,23 +23,25 @@ import {
   emptyDataSource,
   SOURCE_DETAILS_SECTION_ID,
 } from './utils/constants';
-import { createBootableVolume, extractCreatedDataSources } from './utils/utils';
+import { createBootableVolume, extractCreatedDataSources, PreferenceOption } from './utils/utils';
 
 type AddBootableVolumeModalProps = {
   isOpen: boolean;
+  lockedPreference?: PreferenceOption;
   onClose: () => void;
   onCreateVolume?: (createdVolume: BootableVolume) => void;
 };
 
 const AddBootableVolumeModal: FC<AddBootableVolumeModalProps> = ({
   isOpen,
+  lockedPreference,
   onClose,
   onCreateVolume,
 }) => {
   const { t } = useKubevirtTranslation();
   const { createModal } = useModal();
   const { bootableVolume, setBootableVolume, setSourceType, sourceType, upload, uploadData } =
-    useAddBootableVolumeModalData();
+    useAddBootableVolumeModalData(lockedPreference);
 
   const isFormValid = useAddBootableVolumeFormValidation({
     bootableVolume,
--- a/src/utils/components/AddBootableVolumeModal/components/VolumeMetadata/components/PreferenceSelect/PreferenceSelect.tsx
+++ b/src/utils/components/AddBootableVolumeModal/components/VolumeMetadata/components/PreferenceSelect/PreferenceSelect.tsx
@@ -15,7 +15,7 @@ import {
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
 import PopoverContentWithLightspeedButton from '@lightspeed/components/PopoverContentWithLightspeedButton/PopoverContentWithLightspeedButton';
 import { OLSPromptType } from '@lightspeed/utils/prompts';
-import { FormGroup, PopoverPosition } from '@patternfly/react-core';
+import { FormGroup, HelperText, PopoverPosition } from '@patternfly/react-core';
 
 import { getSelectedKeyByLabel } from './utils/utils';
 import PreferencePopoverContent from './PreferencePopoverContent';
@@ -44,6 +44,7 @@ const PreferenceSelect: FC<PreferenceSelectProps> = ({
 
   const handleSelect = (value: string) => {
     const selectedOption = preferenceSelectOptions.find((option) => option.value === value);
+    if (!selectedOption) return;
     setBootableVolumeField('labels', DEFAULT_PREFERENCE_LABEL)(selectedOption.label);
   };
 
@@ -60,11 +61,12 @@ const PreferenceSelect: FC<PreferenceSelectProps> = ({
   );
 
   useEffect(() => {
+    if (!preferencesLoaded || bootableVolume.lockedPreference) return;
     if (!isExistingOption) {
       deleteLabel(DEFAULT_PREFERENCE_LABEL);
       deleteLabel(DEFAULT_PREFERENCE_KIND_LABEL);
     }
-  }, [deleteLabel, isExistingOption]);
+  }, [deleteLabel, isExistingOption, bootableVolume.lockedPreference, preferencesLoaded]);
 
   if (!preferencesLoaded) return <Loading />;
 
@@ -86,12 +88,18 @@ const PreferenceSelect: FC<PreferenceSelectProps> = ({
       label={t('Preference')}
     >
       <InlineFilterSelect
+        toggleProps={{
+          isDisabled: isDisabled || !!bootableVolume.lockedPreference,
+          isFullWidth: true,
+        }}
         options={preferenceSelectOptions}
         placeholder={t('Select preference')}
         selected={selectedPreferenceKey}
         setSelected={handleSelect}
-        toggleProps={{ isDisabled, isFullWidth: true }}
       />
+      {bootableVolume.lockedPreference && (
+        <HelperText>{t('Automatically set by the VM Guest OS selection.')}</HelperText>
+      )}
     </FormGroup>
   );
 };
--- a/src/utils/components/AddBootableVolumeModal/hooks/useAddBootableVolumeModalData.ts
+++ b/src/utils/components/AddBootableVolumeModal/hooks/useAddBootableVolumeModalData.ts
@@ -5,6 +5,11 @@ import {
   DROPDOWN_FORM_SELECTION,
   initialBootableVolumeState,
 } from '@kubevirt-utils/components/AddBootableVolumeModal/utils/constants';
+import { PreferenceOption } from '@kubevirt-utils/components/AddBootableVolumeModal/utils/utils';
+import {
+  DEFAULT_PREFERENCE_KIND_LABEL,
+  DEFAULT_PREFERENCE_LABEL,
+} from '@kubevirt-utils/constants/instancetypes-and-preferences';
 import {
   DataUpload,
   UploadDataProps,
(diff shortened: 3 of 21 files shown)


src/utils/components/AddBootableVolumeModal/hooks/useAddBootableVolumeModalData.ts
```diff
@@ -33,6 +38,13 @@ const useAddBootableVolumeModalData: UseAddBootableVolumeModalData = () => {
     ...initialBootableVolumeState,
     bootableVolumeCluster: selectedCluster,
     bootableVolumeNamespace: namespace,
+    labels: initialPreference
+      ? {
+          [DEFAULT_PREFERENCE_KIND_LABEL]: VirtualMachineClusterPreferenceModel.kind,
```
upalatucci: the preference can also be a `VirtualMachinePreference` so this means we should not just pass the preference name but also the kind. We can pass an object with both `kind` and `name`.  In the `useInstanceTypeVMStore` we only store the preference name. To fix that we can add the kind in the store. We have `usePreferenceSelectOptions` hook in the addbootablevolumemodal that create the selectoption with the kind. the `usePreferenceSelectOptions` in the `creation-wizard` do not do this.

src/utils/components/AddBootableVolumeModal/hooks/useAddBootableVolumeModalData.ts
```diff
@@ -23,7 +28,7 @@ type UseAddBootableVolumeModalData = () => {
   uploadData: ({ dataVolume, file }: UploadDataProps) => Promise<void>;
 };
 
-const useAddBootableVolumeModalData: UseAddBootableVolumeModalData = () => {
+const useAddBootableVolumeModalData: UseAddBootableVolumeModalData = (initialPreference) => {
```
upalatucci: U know what? I would call this prop `lockedPreference` and add a flag in the bootableVolume state. So we don't have to drill down this prop in the components, but they will get it automatically as we already pass the whole state

benbassatido: fixed!

src/utils/components/AddBootableVolumeModal/hooks/useAddBootableVolumeModalData.ts
benbassatido: Fixed!

src/views/virtualmachines/creation-wizard/steps/InstanceTypesSteps/GuestOSStep/components/PreferenceSelectMenu/PreferenceSelectMenu.tsx
```diff
@@ -23,14 +23,17 @@ const PreferenceSelectMenu: FC = () => {
...
-        selectedLabel={preference || t('Select guest operating system type')}
+        selected={preference?.name || ''}
+        selectedLabel={preference?.name || t('Select guest operating system type')}
         toggleProps={{ isFullWidth: true }}
       >
-        {preferenceNames.map((name) => (
+        {preferences.map(({ name }) => (
           <SelectOption key={name} value={name}>
             {name}
           </SelectOption>
```
upalatucci, line 39: We should not use `name` as `value` but a custom value . Would be great to use the whole preference. So that we don't have to do strange stuff in the `onSelect`

src/views/virtualmachines/creation-wizard/steps/InstanceTypesSteps/BootSourceStep/components/BootableVolumeList/components/AddBootableVolumeLink/AddBootableVolumeLink.tsx
```diff
@@ -33,6 +33,7 @@ const AddBootableVolumeLink: FC<AddBootableVolumeLinkProps> = ({
         hidePopover?.();
         createModal((props) => (
           <AddBootableVolumeModal
+            lockedPreference={preference?.name}
```
upalatucci: Here we should pass the whole preference as we need the name and also kind

src/utils/components/AddBootableVolumeModal/hooks/useAddBootableVolumeModalData.ts
```diff
@@ -33,6 +38,14 @@ const useAddBootableVolumeModalData: UseAddBootableVolumeModalData = () => {
     ...initialBootableVolumeState,
     bootableVolumeCluster: selectedCluster,
     bootableVolumeNamespace: namespace,
+    labels: lockedPreference
+      ? {
+          [DEFAULT_PREFERENCE_KIND_LABEL]: VirtualMachineClusterPreferenceModel.kind,
```
upalatucci: here we need to use the preference kind that we should have as a prop in this hook. Not just the name but also the kind

src/utils/components/AddBootableVolumeModal/components/VolumeMetadata/components/PreferenceSelect/PreferenceSelect.tsx
```diff
@@ -60,11 +61,13 @@ const PreferenceSelect: FC<PreferenceSelectProps> = ({
   );
 
   useEffect(() => {
+    if (!preferencesLoaded) return;
```
galkremer1: nit: We can use a single if
`if (!preferencesLoaded || bootableVolume.lockedPreference) return;`

src/utils/components/AddBootableVolumeModal/utils/constants.ts
```diff
@@ -51,6 +51,7 @@ export type AddBootableVolumeState = {
   cronExpression?: string;
   isIso?: boolean;
   labels?: { [key: string]: string };
+  lockedPreference?: boolean;
```
galkremer1: You use the same var name with different types (`boolean` and `string`)
I thing it would be better if you keep it as string and just disable the dropdown when this value is provided. 
Also as I mentioned earlier in the meeting - check with @yfrimanm if we should provide the user with a tooltip in order to indicate why the dropdown is disabled.

src/views/virtualmachines/creation-wizard/steps/InstanceTypesSteps/GuestOSStep/components/PreferenceSelectMenu/PreferenceSelectMenu.tsx
benbassatido: Fixed !

src/utils/components/AddBootableVolumeModal/hooks/useAddBootableVolumeModalData.ts
benbassatido: Fixed !

src/views/virtualmachines/creation-wizard/steps/InstanceTypesSteps/BootSourceStep/components/BootableVolumeList/components/AddBootableVolumeLink/AddBootableVolumeLink.tsx
benbassatido: Fixed !

src/utils/components/AddBootableVolumeModal/components/VolumeMetadata/components/PreferenceSelect/PreferenceSelect.tsx
benbassatido: Fixed !

src/utils/components/AddBootableVolumeModal/utils/constants.ts
benbassatido: Fixed, and I'll check with Yifat

yfrimanm:
TMO we should add helper text to explain why the dropdown is disabled. Something like:
"Locked to match the selected VM Guest OS." or "This preference is locked to match the selected Guest OS"
@awolfe@redhat.com please review the suggested text.

@benbassatido https://github.com/benbassatido now I'm thinking - If it can never be changed from this modal, why do we need it as a dropdown?

src/utils/components/AddBootableVolumeModal/AddBootableVolumeModal.tsx
```diff
@@ -13,6 +13,7 @@ import { BootableVolume } from '@kubevirt-utils/resources/bootableresources/type
 import { isEmpty, kubevirtConsole } from '@kubevirt-utils/utils/utils';
 import { kubevirtK8sDelete } from '@multicluster/k8sRequests';
 import { ButtonVariant, Content } from '@patternfly/react-core';
+import { PreferenceOption } from '@virtualmachines/creation-wizard/state/instance-type-vm-store/utils/types';
```
upalatucci: Lets have the preferenceOPtion in the ADdBootableVOlumeModal folder . The utils folder should not import things from the views folder

galkremer1:
> TMO we should add helper text to explain why the dropdown is disabled. Something like: "Locked to match the selected VM Guest OS." or "This preference is locked to match the selected Guest OS" @[awolfe@redhat.com](mailto:awolfe@redhat.com) please review the suggested text.
> 
> @benbassatido https://github.com/benbassatido now I'm thinking - If it can never be changed from this modal, why do we need it as a dropdown?

@yfrimanm In my opinion it should stay in a disabled dropdown to maintain the same structure as in a modal where it's not disabled.

yfrimanm:
Thinking it over, "Locked" implies an error so how about "Automatically set by the VM Guest OS selection."

galkremer1:
> Thinking it over, "Locked" implies an error so how about "Automatically set by the VM Guest OS selection."

@yfrimanm WDYT about that?
Screenshot 2026-06-03 at 2 50 23 PM
Using a `FormGroupHelperText` component is a bit more explicit than a tooltip.

yfrimanm:
@Gal Kremer ***@***.***> I was suggesting to use helper text (not a
tooltip).
I think this text is clearer "Automatically set by the VM Guest OS
selection."

src/utils/components/AddBootableVolumeModal/AddBootableVolumeModal.tsx
benbassatido: Fixed

galkremer1 (approved):
/lgtm

PR #4032, opened 2026-06-05
CNV-87886: replace useListPageFilter on VirtualMachines page

adamviktora:
### PART 2 of replacing the old way of filtering data

Replace `useListPageFilter` on `VirtualMachines` page

- use `useKubevirtDataViewFilters` instead
- use `KubevirtFilter` instead of `RowFilter`
- adapt URL params to not use "rowFilter-" prefix
  - there is a `useMigratedSearchParams` hook to convert old URLs stored in bookmarks
- adapt URL params so that multiple values for single key are chained as individual `key1=valueA&key1=valueB` instead of `key1=valueA,valueB` (default PatternFly DataView behavior)

## 🔗 Links

Jira: https://redhat.atlassian.net/browse/CNV-87886

## 🎥 Demo

Functional changes:
- URL params on VirtualMachines page now don't use `rowFilter-` prefix and key=value pairs are stored individually

BEFORE:
`rowFilter-status=Running,Stopped`

AFTER:
`status=Running&status=Stopped`

## Summary by CodeRabbit

* **New Features**
  * Hidden filter chips shown; many new VM filters added (Guest Agent, Scheduling, HW devices, Date range, Description, NADs, Architecture, StorageClass, OS, IP, CPU/Memory). URL-backed label syncing and legacy-params migration; search/navigation now drives filter state.

* **Bug Fixes**
  * Safer handling of missing filter options and improved chip/label fallbacks.

* **Refactor**
  * Simplified filter state, clear workflows and toolbar behavior; removed legacy/ref-based filter plumbing.


src/utils/hooks/useKubevirtDataViewFilters/hooks/useMigratedSearchParams.ts
```diff
@@ -0,0 +1,28 @@
...
+  );
+  const searchParams = migratedParams ?? rawSearchParams;
+
+  const migrationDone = useRef(false);
+  useEffect(() => {
+    if (migratedParams && !migrationDone.current) {
+      migrationDone.current = true;
+      setSearchParams(migratedParams, { replace: true });
+    }
+  }, [migratedParams, setSearchParams]);
```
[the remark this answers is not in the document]
adamviktora: How can it happen that a user would navigate to another `rowFilter-*` URL without unmounting the page? 

I am removing `rowFilter-` for VMs page in this PR. The purpose of the hook is to make current bookmarks that users might have saved somewhere, to still continue working.

src/utils/components/KubevirtFilterToolbar/KubevirtFilterToolbar.tsx
```diff
@@ -40,13 +43,22 @@ const KubevirtFilterToolbar: FC<KubevirtFilterToolbarProps> = ({
   const { t } = useKubevirtTranslation();
 
   const groupedFilters = useMemo(
-    () => filterDefinitions.filter((f) => f.filterLayout !== KubevirtFilterLayout.SELECT),
+    () =>
+      filterDefinitions.filter(
+        (f) =>
+          f.filterLayout !== KubevirtFilterLayout.SELECT &&
```
galkremer1: NIT: 
```
![KubevirtFilterLayout.SELECT, KubevirtFilterLayout.HIDDEN].includes(f.filterLayout)
```

src/views/virtualmachines/list/VirtualMachinesList.tsx
```diff
@@ -262,29 +251,26 @@ const VirtualMachinesList: FC<VirtualMachinesListProps> = forwardRef((props, ref
...
-                  onFilterChange(...args);
-                  setPagination((prevPagination) => ({
-                    ...prevPagination,
-                    endIndex: prevPagination?.perPage,
-                    page: 1,
-                    startIndex: 0,
-                  }));
+                  resetPagination();
+                  clearAllFilters();
                 }}
```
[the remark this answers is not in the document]
adamviktora: This should be fine - `SearchBar` uses clearAllFilters later in `navigateToSearchResults`, where a new component is mounted, so there is no need to call `resetPagination` or `deselectAllVMs`. Please check that and let me know if there is a potential issue.

src/utils/hooks/useKubevirtDataViewFilters/hooks/useMigratedSearchParams.ts
```diff
@@ -0,0 +1,28 @@
...
+
+import { migrateLegacyFilterParams } from '../utils';
+
+type UseMigratedSearchParams = () => [URLSearchParams, SetURLSearchParams];
+
+const useMigratedSearchParams: UseMigratedSearchParams = () => {
+  const [rawSearchParams, setSearchParams] = useSearchParams();
+
+  const migratedParams = useMemo(
+    () => migrateLegacyFilterParams(rawSearchParams),
```
galkremer1, line 12: Does it make sense for `migrateLegacyFilterParams` to return `rawSearchParams` as a fallback?

src/views/virtualmachines/list/filters/getDateFilter.ts
```diff
@@ -0,0 +1,35 @@
...
+import { V1VirtualMachine } from '@kubevirt-ui-ext/kubevirt-api/kubevirt';
+import {
+  KubevirtFilter,
+  KubevirtFilterLayout,
+} from '@kubevirt-utils/hooks/useKubevirtDataViewFilters/types';
+import { getCreationTimestamp } from '@kubevirt-utils/resources/shared';
+import { VirtualMachineRowFilterType } from '@virtualmachines/utils';
+
+const getMatchDateFunction =
+  (variant: 'from' | 'to') => (obj: V1VirtualMachine, selected: string[]) => {
```
galkremer1: We have this const `const FROM_QUERY_PARAM = 'from';` - it can be extracted to a shared const object along with `to` and we can create a type based on it.

src/views/virtualmachines/list/filters/getDescriptionFilter.ts
```diff
@@ -0,0 +1,20 @@
...
+import { VirtualMachineRowFilterType } from '@virtualmachines/utils';
+
+export const getDescriptionFilter = (t: TFunction): KubevirtFilter<V1VirtualMachine> => ({
+  categoryLabel: t('Description'),
+  filterLayout: KubevirtFilterLayout.HIDDEN,
+  id: VirtualMachineRowFilterType.Description,
+  match: (obj, selected) => {
+    const search = selected[0];
+    if (!search) return true;
+    return getDescription(obj)?.toLowerCase().includes(search.toLowerCase());
```
galkremer1: Maybe we can create a helper function for that `getDescription(obj)?.toLowerCase().includes(search.toLowerCase());`

src/utils/hooks/useKubevirtDataViewFilters/hooks/useMigratedSearchParams.ts
adamviktora: It would be cleaner, but on line 19 we check `if (migratedParams ...` to check if there were any legacy params. If we had returned `params` from `migrateLegacyFilterParams` instead of null, we wouldn't be able to check it that easily.

src/views/virtualmachines/list/filters/getDateFilter.ts
adamviktora: I wouldn't mix the two, here it is used in a very different context, it could be a boolean flag.

src/views/virtualmachines/list/filters/getDescriptionFilter.ts
adamviktora: I used `fuzzyCaseInsensitive` - it changes the behavior of that filter to be a fuzzy search, but I think it makes sense for description field, rather than being an exact search.

galkremer1 (commented):
@adamviktora please address / resolve remaining Coderabbit comments, otherwise lgtm.

galkremer1 (approved):
/lgtm

PR #4038, opened 2026-06-07
CNV-88997: Style VM creation location display in create VM Wizard

mayarub:
## 📝 Description

Styles the VM creation location field in the Deployment Details step to match the TextInput appearance (border, background, padding, and width). 
* When the user clicks Edit, the input styling is removed so the location text appears as plain text above the edit form. 
* Changed generate VM name icon to blue icon

## 🔗 Links

[Enhance the UI/UX for the location path in the “Create VM” wizard
](https://docs.google.com/document/d/1UTM5IE1zEQ70zwniT_MmLtG2w4T1k2Q_9Iqf7h1hVo8/edit?tab=t.0)

[Jira ticket: [CNV-88997](https://redhat.atlassian.net/browse/CNV-88997)](https://redhat.atlassian.net/browse/CNV-88997)
## 🎥 Demo

before:
image

after:
image

## Summary by CodeRabbit

* **New Features**
  * Added an editing mode for VM creation location in the wizard.

* **Improvements**
  * Replaced the VM name generation button icon with a unified BlueSync icon.
  * Shortened and clarified the Location section message.
  * Improved layout and styling of the VM creation location display for better alignment and clearer editing visuals.


adamviktora (commented):
Great work @mayarub!

You should also run `npm run i18n` - this will regenerate the translation files (it is needed because of the change in `t('Your VirtualMachine will be created in the following location')`)

Also make sure to set the Fix Version field in Jira to `4.23.0`

mayarub:
> Great work @mayarub!
> 
> You should also run `npm run i18n` - this will regenerate the translation files (it is needed because of the change in `t('Your VirtualMachine will be created in the following location')`)
> 
> Also make sure to set the Fix Version field in Jira to `4.23.0`

Thank you!
I ran i18n and updated the Fix version field in Jira

src/views/virtualmachines/creation-wizard/components/GenerateVMNameButton.tsx
```diff
@@ -2,8 +2,8 @@ import React, { FC } from 'react';
 
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
 import { generatePrettyName } from '@kubevirt-utils/utils/utils';
+import BlueSyncIcon from '@overview/OverviewTab/status-card/utils/health-state-icons/BlueSyncIcon';
```
upalatucci: can we move this icon in the utils folder? 
A general rule that we have is to not use components from different views (src/views/template, src/views/virtualmachines) but in this case, we move things in the utils folder

mayarub: fixed

upalatucci (approved):

PR #4055, opened 2026-06-10
CNV-89379 closing the storage migration with X cancels the migration instead of closing the dialog

benbassatido:
## 📝 Description

Fix X button on storage migration progress modal to dismiss the dialog without canceling the migration.

## 🔗 Links

https://redhat.atlassian.net/browse/CNV-89379

## 🎥 Demo

Before: 

https://github.com/user-attachments/assets/80f84ed4-5ca9-4f9f-8a2e-f137798228fb

After: 

https://github.com/user-attachments/assets/0c31c193-98e6-43e1-acda-2a9d0ef762af

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized the close button label in virtual machine migration dialogs to consistently display "Close" for improved user clarity and consistency.

### DIFF
--- a/src/views/virtualmachines/actions/components/VirtualMachineMigration/components/StorageMigrationProgress.tsx
+++ b/src/views/virtualmachines/actions/components/VirtualMachineMigration/components/StorageMigrationProgress.tsx
@@ -46,13 +46,8 @@ const StorageMigrationProgress: FC<StorageMigrationProgressProps> = ({
     storageMigrationPlan,
   });
 
-  const onHeaderClose = migrationCompleted || hasFailed ? onClose : onCancelMigration;
-
   return (
-    <MigrationStatusShell
-      isTerminal={migrationCompleted || hasFailed}
-      onHeaderClose={onHeaderClose}
-    >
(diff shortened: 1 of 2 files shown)


src/views/virtualmachines/actions/components/VirtualMachineMigration/components/StorageMigrationProgress.tsx
```diff
@@ -46,13 +46,8 @@ const StorageMigrationProgress: FC<StorageMigrationProgressProps> = ({
     storageMigrationPlan,
   });
 
-  const onHeaderClose = migrationCompleted || hasFailed ? onClose : onCancelMigration;
```
metalice, line 49: Please double-check the logic you are removing here is correct. @upalatucci

upalatucci: yes its okay. The logic is: When the user just close the modal, we don't cancel the migration. When the user clicks on "Cancel" we cancel the migration. 

Image

upalatucci (approved):

PR #4061, opened 2026-06-12
CNV-85847: [release-4.18] DOMPurify: Cross-Site Scripting (XSS) via inconsistent tag sanitization

UdayYendva:
Fix raised to bump the vulnerable dompurify package to patched version in resolutions section in package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -126,7 +126,7 @@
     "@types/d3-dispatch": "3.0.6",
     "@types/react": "^17.0.40",
     "@types/react-dom": "^17.0.0",
-    "dompurify": "3.4.0"
+    "dompurify": "^3.4.0"
   },
   "scripts": {
     "build": "yarn clean && NODE_ENV=production NODE_OPTIONS=--max-old-space-size=8192 yarn ts-node node_modules/.bin/webpack",
--- a/package.json
+++ b/package.json
@@ -125,7 +125,8 @@
     "@openshift-console/dynamic-plugin-sdk/immutable": "^3.8.3",
     "@types/d3-dispatch": "3.0.6",
     "@types/react": "^17.0.40",
-    "@types/react-dom": "^17.0.0"
+    "@types/react-dom": "^17.0.0",
+    "dompurify": "3.4.0"
   },
   "scripts": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -125,7 +125,8 @@
     "@openshift-console/dynamic-plugin-sdk/immutable": "^3.8.3",
     "@types/d3-dispatch": "3.0.6",
     "@types/react": "^17.0.40",
-    "@types/react-dom": "^17.0.0"
+    "@types/react-dom": "^17.0.0",
+    "dompurify": "^3.4.0"
```
upalatucci, line 129: we should be ablet o upgrade it wihtout adding it to the dependencies

UdayYendva: @upalatucci, the parent dependency is not OpenShift-related, which is why I added only the vulnerable package. However, for versions 4.12 through 4.14, the parent dependencies are OpenShift-related packages.
Given that this dependency is not being hoisted by any OpenShift package, could you please confirm whether we can close this PR?

UdayYendva:
```
yarn why dompurify
yarn why v1.22.22
[1/4] Why do we have the module "dompurify"...?
[2/4] Initialising dependency graph...
[3/4] Finding dependency...
[4/4] Calculating file sizes...
=> Found "dompurify@3.4.9"
info Reasons this module exists
   - "@patternfly#quickstarts" depends on it
   - Hoisted from "@patternfly#quickstarts#dompurify"
info Disk size without dependencies: "832KB"
info Disk size with unique dependencies: "832KB"
info Disk size with transitive dependencies: "832KB"
info Number of shared dependencies: 0
Done in 0.47s.


```

upalatucci: We can use the `override` package.json term at this point

UdayYendva: So @upalatucci do we need to include "@patternfly/quickstarts/dompurify": "3.4.0" or else good with current changes

upalatucci: sorry not the `override`. Here we are using `yarn` so in the `resolution` object in the `pakcage.json`. But yes this is the fix

upalatucci (approved):

PR #4070, opened 2026-06-15
CNV-90113: fix useVMTemplateFeatureFlag to allow non-priv to create vm from template

batyana:
## 📝 Description

`useK8sWatchResource` never sets loaded: true on error, so `useVMTemplateFeatureFlag` blocks the template catalog indefinitely. Handles failed watches (403) by treating errors as a terminal loaded state.

## 🔗 Links

Jira ticket: https://redhat.atlassian.net/browse/CNV-90113

## 🎥 Demo

Before:
image

After:
Screenshot at Jun 15 17-48-52

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of the “VM templates” enablement check, including more accurate loading behavior and better surfaced error handling when template configuration can’t be retrieved.
* **Refactor**
  * Updated template-related feature gating to use a dedicated enablement hook across filters, template resource loading, and virtual machine actions/wizard flows.

### DIFF
--- a/src/utils/hooks/useVMTemplateFeatureFlag/useVMTemplateFeatureFlag.ts
+++ b/src/utils/hooks/useVMTemplateFeatureFlag/useVMTemplateFeatureFlag.ts
@@ -1,39 +1,33 @@
-import { useMemo } from 'react';
-
 import { HyperConvergedModel } from '@kubevirt-ui-ext/kubevirt-api/console';
 import { K8S_OPS } from '@kubevirt-utils/constants/constants';
 import useHyperConvergeConfiguration from '@kubevirt-utils/hooks/useHyperConvergeConfiguration';
 import { useIsAdmin } from '@kubevirt-utils/hooks/useIsAdmin';
-import useKubevirtHyperconvergeConfiguration from '@kubevirt-utils/hooks/useKubevirtHyperconvergeConfiguration';
 import { getAnnotations } from '@kubevirt-utils/resources/shared';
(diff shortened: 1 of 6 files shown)


src/utils/hooks/useVMTemplateFeatureFlag/useVMTemplateFeatureFlag.ts
```diff
@@ -33,7 +34,7 @@ const useVMTemplateFeatureFlag = (clusterOverride?: string) => {
   return {
     canEdit: isAdmin,
     featureEnabled,
-    loading: !hcLoaded || !hcConfigLoaded,
+    loading: (!hcLoaded && !hcError) || (!hcConfigLoaded && !hcConfigError),
```
galkremer1: I think it would be better to separate it to two states: `loading` and `error`
```
loading: !hcLoaded || !hcConfigLoaded ,
error: hcError || hcConfigError 
```
Then in the consumer we can handle this error. 
Can you check for a specific type of error that the UI can ignore and still show the catalog?
BTW we have a similar pattern in the `usePasstFeatureFlag` hook - please check if we need to handle it as well.

upalatucci: Regular users do not have permissions to fetch HCO so it will allways fail and `hcConfigError` will always have some kind of error

upalatucci (approved):

PR #4072, opened 2026-06-15
CNV-89814: Warn user when navigating away while uploads are in progress

galkremer1:
## 📝 Description

When one or more file uploads are running (CD-ROM ISO, disk, or bootable volume), the user should be warned before navigating away from the current page.

This prevents accidental loss of in-progress uploads that would leave orphaned DataVolumes in the cluster.

## 🔗 Links

Jira ticket: [CNV-89814](https://redhat.atlassian.net/browse/CNV-89814)

## 🎥 Demo

https://github.com/user-attachments/assets/4ecdc6d6-a47d-4e76-a3ed-1d3596654cd2

## New Features
* Added page navigation protection to prevent accidental leaving when file uploads are in progress, reducing the risk of losing unfinished uploads.

## Tests
* Added coverage for upload navigation protection, verifying that the `beforeunload` listener is correctly registered/removed based on upload state and that the handler blocks navigation during active uploads.

### DIFF
--- a/src/utils/hooks/useUploadProgressToast/UploadNavigationGuard/useUploadBeforeUnload.ts
+++ b/src/utils/hooks/useUploadProgressToast/UploadNavigationGuard/useUploadBeforeUnload.ts
@@ -1,0 +1,26 @@
+import { useEffect } from 'react';
+
+import { useUploadProgressStore } from '../uploadProgressStore';
+import { UPLOAD_PROGRESS_STATUS } from '../utils/constants';
+
+const useUploadBeforeUnload = (): void => {
+  const hasActiveUploads = useUploadProgressStore((state) =>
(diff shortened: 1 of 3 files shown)


src/utils/hooks/useUploadProgressToast/UploadNavigationGuard/useUploadBeforeUnload.ts
```diff
@@ -0,0 +1,25 @@
...
+    Object.values(state.uploads).some((entry) => entry.status === UPLOAD_PROGRESS_STATUS.UPLOADING),
+  );
+
+  useEffect(() => {
+    if (!hasActiveUploads) {
+      return;
+    }
+
+    const handler = (event: BeforeUnloadEvent) => {
+      event.preventDefault();
```
galkremer1, line 17: Unfortunately we cannot customize the message so the default one is used (`Changes you made may not be saved.`)

adamviktora (approved):

PR #4084, opened 2026-06-17
CNV-90088: show permission tooltip on disabled action menu items

kuklas:
## Summary
- The kebab menu items were greyed out for unprivileged users but showed no tooltip because the tooltip logic lived in a component that didn't know about the permission check result.
- Moved the tooltip into the component that knows the full disabled state, so "You don't have permission to perform this action" now appears on hover.

## Test plan
- [ ] Log in as an unprivileged user (without `create` permission on `virtualmachineinstancemigrations`)
- [ ] Navigate to a running VM's kebab menu → Migration → Compute
- [ ] Verify the item is greyed out and hovering shows "You don't have permission to perform this action"
- [ ] Repeat for Migration → Storage
- [ ] Verify that tooltips still show for other disabled reasons (VM not running, VM not live migratable)
- [ ] Verify that enabled actions still work normally

New
nopermission5
nopermissions4
nopremissiosn3
Nopermissions-2
Permissions-No1
nopermission6

### DIFF
--- a/src/utils/components/LazyActionMenu/ActionMenuItem.tsx
+++ b/src/utils/components/LazyActionMenu/ActionMenuItem.tsx
@@ -54,7 +54,17 @@ const ActionItem: FC<ActionMenuItemProps & { isAllowed: boolean }> = ({
     ...(external ? { isExternalLink: external, to: href } : {}),
   };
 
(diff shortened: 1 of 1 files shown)


galkremer1 (approved):
/lgtm

PR #4105, opened 2026-06-18
Ignore missing Project API for cluster-wide VM listing

chrisayoub:
## 📝 Description

Fixes regression introduced in version `v4.21.0`. This was not present in `v4.20.0`. Tested against a cluster with vanilla K8s (no OpenShift CRDs installed)

URL tested: <http://localhost:9000/k8s/all-namespaces/kubevirt.io~v1~VirtualMachine/search>

## 🎥 Demo

With change, renders correctly rather than error:
Screenshot 2026-06-18 at 5 51 32 AM

## Release Notes

* **Bug Fixes**
  * Improved virtual machine resource discovery when performing cluster-wide searches.
  * The app now treats certain cluster-wide readiness/error states as non-fatal, avoiding premature failures and allowing searches to proceed more reliably.
  * Namespace-scoped behavior remains unchanged for error handling.

### DIFF
--- a/src/views/virtualmachines/search/hooks/useAccessibleResources.ts
+++ b/src/views/virtualmachines/search/hooks/useAccessibleResources.ts
@@ -106,12 +106,14 @@ export const useAccessibleResources = <T>({
     return vms || [];
   }, [allResources, allowedResources, loadPerNamespace]);
 
-  const loaded =
-    projectNamesLoaded &&
-    (loadPerNamespace
-      ? isEmpty(allowedResources) ||
-        Object.values(allowedResources).some((resource) => resource.loaded || resource.loadError)
-      : allResourcesLoaded);
+  const loaded = shouldFetchClusterWide
+    ? allResourcesLoaded
+    : projectNamesLoaded &&
+      (isEmpty(allowedResources) ||
+        Object.values(allowedResources).some((resource) => resource.loaded || resource.loadError));
 
-  return { loaded, loadError: projectNamesError, resources };
+  // Cluster-wide watches do not need OpenShift Project discovery.
+  const loadError = shouldFetchClusterWide ? undefined : projectNamesError;
+
+  return { loaded, loadError, resources };
 };


chrisayoub:
cc @batyana

chjohnst (approved):

chrisayoub:
cc @upalatucci

src/views/virtualmachines/search/hooks/useAccessibleResources.ts
```diff
@@ -113,5 +116,12 @@ export const useAccessibleResources = <T>({
         Object.values(allowedResources).some((resource) => resource.loaded || resource.loadError)
       : allResourcesLoaded);
 
-  return { loaded, loadError: projectNamesError, resources };
+  // Cluster-wide watches do not need OpenShift Project discovery. Plain Kubernetes clusters can
+  // return either an HTTP 404 or the SDK's missing model error before the request is made.
+  const loadError =
+    shouldFetchClusterWide && isMissingProjectAPIError(projectNamesError)
```
metalice: Why only if shouldFetchClusterWide?

chrisayoub: The namespace path doesn’t trigger the error case in the first place

src/views/virtualmachines/search/hooks/useAccessibleResources.ts
```diff
@@ -113,5 +116,12 @@ export const useAccessibleResources = <T>({
         Object.values(allowedResources).some((resource) => resource.loaded || resource.loadError)
       : allResourcesLoaded);
 
-  return { loaded, loadError: projectNamesError, resources };
+  // Cluster-wide watches do not need OpenShift Project discovery. Plain Kubernetes clusters can
+  // return either an HTTP 404 or the SDK's missing model error before the request is made.
+  const loadError =
```
upalatucci: We can completely remove the need to return `projectNamesError` if `shouldFetchClusterWide` and also for the loaded we do that. So we can do 


```suggestion
  const loadError = shouldFetchClusterWide ? undefined : projectNamesError

  // move the projectNamesLoaded && inside the ternary and make it valid only when loadPerNamespace condition is true
  const loaded = loadPerNamespace
      ? projectNamesLoaded && isEmpty(allowedResources) ||
        Object.values(allowedResources).some((resource) => resource.loaded || resource.loadError)
      : allResourcesLoaded);
```

upalatucci (approved):

upalatucci:
thanks @chrisayoub for the contribution

PR #4114, opened 2026-06-22
CNV-71181: fix VM advanced search for Memory and CPU

adamviktora:
## 📝 Description

Fixes VM advanced search for Memory and CPU for stopped VMs created from `InstanceType`

**BUG:** If a VM doesn't have a running VMI and doesn't have `cpu` or `memory` specified in its YAML, then memory and cpu filters didn't work

**FIX:** read memory and cpu fields from corresponding `VirtualMachineClusterInstanceType` or `VirtualMachineInstanceType`

_____

Also adds `getClusterKey` helper:

```
export const getClusterKey = (resource: K8sResourceCommon): string =>
  getCluster(resource) || SINGLE_CLUSTER_KEY;
```

## 🔗 Links

Jira: https://redhat.atlassian.net/browse/CNV-71181

## 🎥 Demo

Before:

https://github.com/user-attachments/assets/87f58ccb-171c-4662-ae64-42123f18b10e

After:

https://github.com/user-attachments/assets/218ec0d0-a142-42ab-9824-d962b3ed0665

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Virtual Machine list CPU and memory filters with a more reliable fallback using instance-type data when direct VM details are missing.
  * Filters now update correctly once instance-type information finishes loading.
  * Fixed multi-cluster consistency for Virtual Machine metrics caching, self-validation “current job” detection, and bootable-volume source mapping.

### DIFF
--- a/src/views/virtualmachines/list/hooks/useInstanceTypeMapper.ts
+++ b/src/views/virtualmachines/list/hooks/useInstanceTypeMapper.ts
@@ -1,0 +1,51 @@
+import { useMemo } from 'react';
+
+import {
+  VirtualMachineClusterInstancetypeModelGroupVersionKind,
+  VirtualMachineInstancetypeModelGroupVersionKind,
+} from '@kubevirt-ui-ext/kubevirt-api/console';
+import {
+  V1beta1VirtualMachineClusterInstancetype,
+  V1beta1VirtualMachineInstancetype,
+} from '@kubevirt-ui-ext/kubevirt-api/kubevirt';
+import { getClusterKey, getName, getNamespace } from '@kubevirt-utils/resources/shared';
+import { useAccessibleResources } from '@virtualmachines/search/hooks/useAccessibleResources';
+import { InstanceTypeMapper } from '@virtualmachines/utils/mappers';
+
+export const useInstanceTypeMapper = () => {
+  const { loaded: clusterITsLoaded, resources: clusterITs } =
+    useAccessibleResources<V1beta1VirtualMachineClusterInstancetype>({
+      groupVersionKind: VirtualMachineClusterInstancetypeModelGroupVersionKind,
+    });
+
+  const { loaded: namespacedITsLoaded, resources: namespacedITs } =
+    useAccessibleResources<V1beta1VirtualMachineInstancetype>({
+      groupVersionKind: VirtualMachineInstancetypeModelGroupVersionKind,
+    });
+
+  const instanceTypeMapper: InstanceTypeMapper = useMemo(() => {
+    const clusterInstanceTypes = (clusterITs ?? []).reduce<
+      InstanceTypeMapper['clusterInstanceTypes']
+    >((acc, it) => {
+      const cluster = getClusterKey(it);
+      if (!acc[cluster]) acc[cluster] = {};
+      acc[cluster][getName(it)] = it;
+      return acc;
+    }, {});
+
+    const namespacedInstanceTypes = (namespacedITs ?? []).reduce<
+      InstanceTypeMapper['namespacedInstanceTypes']
+    >((acc, it) => {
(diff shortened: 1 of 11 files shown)


src/views/virtualmachines/list/hooks/useInstanceTypeMapper.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  const { loaded: namespacedITsLoaded, resources: namespacedITs } =
+    useAccessibleResources<V1beta1VirtualMachineInstancetype>({
+      groupVersionKind: VirtualMachineInstancetypeModelGroupVersionKind,
+    });
+
+  const instanceTypeMapper: InstanceTypeMapper = useMemo(() => {
+    const clusterInstanceTypes = (Array.isArray(clusterITs) ? clusterITs : []).reduce(
+      (acc, it) => {
+        const cluster = getCluster(it) || SINGLE_CLUSTER_KEY;
```
galkremer1: The logic here and in lines `39-49` is  repeated. 
Can you extract the logic and move it to `mappers.ts` or another util file?
Also we can use consts for `clusterInstanceTypes` and `namespacedInstanceTypes`

src/views/virtualmachines/utils/mappers.ts
```diff
@@ -33,6 +35,30 @@ export const getVMIMFromMapper = (
...
+    [cluster: string]: { [namespace: string]: { [name: string]: InstanceTypeUnion } };
+  };
+};
+
+export const getInstanceTypeFromMapper = (
+  mapper: InstanceTypeMapper,
+  vm: V1VirtualMachine,
+): InstanceTypeUnion | undefined => {
+  const matcher = getInstanceTypeMatcher(vm);
+  if (!matcher?.name) return undefined;
```
galkremer1: We can early return here is `mapper` is also undefined.

src/views/virtualmachines/utils/mappers.ts
```diff
@@ -33,6 +35,30 @@ export const getVMIMFromMapper = (
...
+
+  const cluster = getCluster(vm) || SINGLE_CLUSTER_KEY;
```
galkremer1: `getCluster(vm) || SINGLE_CLUSTER_KEY` is used 3 (or 2 times if the other code will be consolidated). 
We can create a helper function for getting the cluster key.

src/views/virtualmachines/list/hooks/useInstanceTypeMapper.ts
```diff
@@ -0,0 +1,55 @@
...
+        if (!acc[cluster]) acc[cluster] = {};
+        acc[cluster][getName(it)] = it;
+        return acc;
+      },
+      {} as InstanceTypeMapper['clusterInstanceTypes'],
+    );
+
+    const namespacedInstanceTypes = (Array.isArray(namespacedITs) ? namespacedITs : []).reduce(
```
galkremer1: Can we use the `toArray` helper here? 
I know that there is a difference since if `namespacedITs` is not an array you are returning an empty array in this case instead of [namespacedITs].

src/views/virtualmachines/list/hooks/useInstanceTypeMapper.ts
adamviktora: The logic is a bit different in the other case because of the namespace. I think it is not worth extracting

src/views/virtualmachines/list/hooks/useInstanceTypeMapper.ts
adamviktora: It is probably way too defensive anyway. Replacing with `namespacedITs ?? []`

galkremer1 (approved):
/lgtm

PR #4128, opened 2026-06-24
CNV-90382 added eslinter file for new code

benbassatido:
## 📝 Description

Introduces an incremental linting upgrade that applies stricter rules only to new/changed files without breaking the existing codebase. A separate eslint-v10/ directory isolates ESLint v10 with its own dependencies.

## 🔗 Links

## Summary by CodeRabbit

* **New Features**
  * Added stricter commit-message validation with updated commit hooks.
  * Introduced targeted “lint only changed files” workflow, plus style linting for SCSS.

* **Bug Fixes**
  * Updated inline lint suppressions to the newer rule identifiers.
  * Refreshed script output formatting for more consistent validation comments/status reporting.

* **Chores**
  * Overhauled repository lint/format tooling (ESLint, Stylelint, Prettier) and added supporting configs (commitlint, knip, strict lint TypeScript config).


package.json
```diff
@@ -151,19 +161,20 @@
...
-    "lint": "eslint . --color && prettier . --ignore-unknown --check --ignore-path .prettierignorecode",
-    "lint:fix": "eslint . --color --fix && prettier . --ignore-unknown --write --ignore-path .prettierignorecode",
+    "i18n": "i18next --config i18next-parser.config.cjs --silent && node ./i18n-scripts/set-english-defaults.cjs && ./i18n-scripts/replace-br.sh",
+    "i18n:verbose": "i18next --config i18next-parser.config.cjs && node ./i18n-scripts/set-english-defaults.cjs && ./i18n-scripts/replace-br.sh",
+    "i18n-dummy-locale": "node ./i18n-scripts/dummy-locale.cjs",
+    "i18n-to-po": "node ./i18n-scripts/i18n-to-po.cjs",
+    "lint": "eslint src --color && prettier . --ignore-unknown --check --ignore-path .prettierignorecode",
+    "lint:fix": "eslint src --color --fix && prettier . --ignore-unknown --write --ignore-path .prettierignorecode",
     "lint-staged": "lint-staged",
+    "lint:pr": "FILES=$(git diff --name-only --diff-filter=d origin/main...HEAD -- '*.ts' '*.tsx' '*.js' '*.jsx'); [ -z \"$FILES\" ] && echo 'No changed files to lint.' || echo \"$FILES\" | xargs eslint --color",
```
metalice: Why?

package.json
```diff
@@ -151,19 +161,20 @@
...
-    "i18n-dummy-locale": "node ./i18n-scripts/dummy-locale.js",
-    "i18n-to-po": "node ./i18n-scripts/i18n-to-po.js",
-    "lint": "eslint . --color && prettier . --ignore-unknown --check --ignore-path .prettierignorecode",
-    "lint:fix": "eslint . --color --fix && prettier . --ignore-unknown --write --ignore-path .prettierignorecode",
+    "i18n": "i18next --config i18next-parser.config.cjs --silent && node ./i18n-scripts/set-english-defaults.cjs && ./i18n-scripts/replace-br.sh",
+    "i18n:verbose": "i18next --config i18next-parser.config.cjs && node ./i18n-scripts/set-english-defaults.cjs && ./i18n-scripts/replace-br.sh",
+    "i18n-dummy-locale": "node ./i18n-scripts/dummy-locale.cjs",
+    "i18n-to-po": "node ./i18n-scripts/i18n-to-po.cjs",
+    "lint": "eslint src --color && prettier . --ignore-unknown --check --ignore-path .prettierignorecode",
+    "lint:fix": "eslint src --color --fix && prettier . --ignore-unknown --write --ignore-path .prettierignorecode",
```
metalice: Why?

benbassatido: fixed

package.json
benbassatido: fixed

src/utils/components/DurationOption/DurationOption.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable perfectionist/sort-classes */
```
metalice, line 1: ?

src/views/clusteroverview/TopConsumersTab/utils/topConsumerMetric.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable perfectionist/sort-classes */
```
metalice, line 1: ?

eslint.config.js
```diff
@@ -0,0 +1,86 @@
...
+
+const tsConfigs = [{ ...tseslint.configs.base, files: ['**/*.{ts,tsx}'] }];
+
+const prettierOverrides = {
+  rules: {
+    'prettier/prettier': ['error', { endOfLine: 'auto' }],
+  },
+};
+
+export default [ignoresConfig, baseConfig, ...tsConfigs, prettier, prettierOverrides];
```
metalice: u tried running this lint locally and checked that we are good? This should run on the entire code base, check how many new fixes this config is presenting, if any

package.json
```diff
@@ -159,12 +169,13 @@
     "lint": "eslint . --color && prettier . --ignore-unknown --check --ignore-path .prettierignorecode",
     "lint:fix": "eslint . --color --fix && prettier . --ignore-unknown --write --ignore-path .prettierignorecode",
     "lint-staged": "lint-staged",
+    "lint:pr": "FILES=$(git diff --name-only --diff-filter=d origin/main...HEAD -- '*.ts' '*.tsx' '*.js' '*.jsx'); [ -z \"$FILES\" ] && echo 'No changed files to lint.' || echo \"$FILES\" | xargs eslint --color",
```
metalice: where it used??

tsconfig.eslint.json
```diff
@@ -0,0 +1,15 @@
+{
```
metalice, line 1: why another tsconfig is needed?

.github/scripts/src/validate-pr.ts
upalatucci, line 1: can we ignore `no-console` rule just for the .github folder ? I think we can do that by adding a .eslintrc file in the folder right ?

knip.json
```diff
@@ -0,0 +1,7 @@
+{
```
upalatucci, line 1: cool

eslint.config.js
benbassatido: Yes, this lint config file is running on our entire code base, all the fixes that it created are fixed in this PR. So no more fixes are needed. This config is intentionally matches the old eslintrc.json behavior (with only updated versions that supports Eslint V10)

tsconfig.eslint.json
benbassatido: It is used by eslintv10.config.js for type aware lint rules. it extends the tsconfig.json but enables stricter rules options. We can't enable it these rules in the main tsconfig.json, we will get hundreds or thousand of TS errors. I will delete this config file after the entire code base will compile under the strict rules, do you want that I'll add a comment in the file explaining that this file will be deleted?

src/views/clusteroverview/TopConsumersTab/utils/topConsumerMetric.ts
benbassatido: The perfectionist/sort-classes rule wants to alphabetically sort all class members. But TopConsumerMetric uses a static initialization where order matters. so CPU, Memory etc... must be defined before all, and all must exist before the dropdownLabelMapper / stringMapper. In this case alphabetically sorting would break this dependency chain and can cause runtime crashes. So I disabled the sorting alphabetically to preserve the required initialization order.
When I'll start fixing the entire code base this will be fixed as well (can add comment that explains that it will be fixed)

src/utils/components/DurationOption/DurationOption.ts
benbassatido: Same reason as topConsumerMetric.ts, static initialization order. sorting this file alphabetically breaks this dependency chain and will cause runtime crashes.

.github/scripts/src/validate-pr.ts
benbassatido: Yes. It is not by adding .eslintrc file because eslint version 10 doesn't support eslintrc files.
It will be added to eslint.config.js and eslintv10.config.js files

package.json
benbassatido: Removed

src/utils/components/DurationOption/DurationOption.ts
metalice: it dont make sense

PR #4153, opened 2026-06-30
CNV-90382: Added testLibrary and i18next, changed husky from pre-commit to pre-push  

benbassatido:
## 📝 Description

Adds i18next/no-literal-string, testing-library ESLint rules to the root config. replaces the commit-msg hook with a pre-push hook for commit message validation. And splits few files exceeding the 150 line limit. 

## 🔗 Links

## Summary by CodeRabbit

* **New Features**
  * Improved network interface selection with smarter auto-selection, better option handling, and clearer validation messages.
  * Added localized labels in editor and network-related UI text.

* **Bug Fixes**
  * Preserved valid empty values in several places by using safer fallback logic.
  * Adjusted commit message checks to better handle `TYPE-123:` style headers.

* **Chores**
  * Updated linting, formatting, and editor settings for a smoother development experience.

### DIFF
--- a/locales/es/plugin__kubevirt-plugin.json
+++ b/locales/es/plugin__kubevirt-plugin.json
@@ -736,7 +736,7 @@
   "Domain": "Dominio",
   "Down": "Abajo",
   "Download results": "Descargar resultados",
-  "Download the MSI from ": "Descargar el MSI desde ",
+  "Download the MSI from <2>virt-manager.org</2>": "Descargar el MSI desde <2>virt-manager.org</2>",
   "Download URL is not needed when an existing Windows DataSource is selected.": "Download URL is not needed when an existing Windows DataSource is selected.",
   "Drag and drop a certificate file or paste PEM content": "Arrastre y suelte un archivo de certificado o pegue el contenido PEM",
   "Drag and drop an image or upload one": "Arrastrar y soltar una imagen o cargar una",
--- a/src/perspective/virtualization-icon.tsx
+++ b/src/perspective/virtualization-icon.tsx
@@ -1,6 +1,6 @@
-import React from 'react';
+import React, { type JSX } from 'react';
 
-export default () => (
+export default (): JSX.Element => (
   <>
     <svg
       aria-hidden="true"
@@ -11,6 +11,7 @@ export default () => (
       width="1em"
       xmlns="http://www.w3.org/2000/svg"
     >
+      {/* eslint-disable-next-line i18next/no-literal-string */}
       <title>Red Hat OpenShift Virtualization icon</title>
       <desc>RHOCPV</desc>
       <path
--- a/src/utils/components/Consoles/components/DesktopViewer/Components/MoreInformationDefault.tsx
+++ b/src/utils/components/Consoles/components/DesktopViewer/Components/MoreInformationDefault.tsx
@@ -1,4 +1,4 @@
-import React, { FC } from 'react';
+import React, { type FC } from 'react';
 import { Trans } from 'react-i18next';
 
 import DescriptionItem from '@kubevirt-utils/components/DescriptionItem/DescriptionItem';
@@ -6,15 +6,14 @@ import { documentationURL } from '@kubevirt-utils/constants/documentation';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
 import { Content, DescriptionList } from '@patternfly/react-core';
 
-import { MoreInformationDefaultProps } from '../utils/types';
-
+import { type MoreInformationDefaultProps } from '../utils/types';
 import Detail from './Detail';
 
 const MoreInformationDefault: FC<MoreInformationDefaultProps> = ({ textMoreInfoContent }) => {
   const { t } = useKubevirtTranslation();
   return (
     <>
-      {textMoreInfoContent || (
+      {textMoreInfoContent ?? (
         <Content>
           <Trans ns="plugin__kubevirt-plugin" t={t}>
             <p>
@@ -34,8 +33,8 @@ const MoreInformationDefault: FC<MoreInformationDefaultProps> = ({ textMoreInfoC
         <Detail title={'Ubuntu, Debian'} value={'sudo apt-get install virt-viewer'} />
         <DescriptionItem
           descriptionData={
-            <div>
-              {t('Download the MSI from ')}
+            <Trans ns="plugin__kubevirt-plugin" t={t}>
+              Download the MSI from{' '}
               <a
                 href={documentationURL.VIRT_MANAGER_DOWNLOAD}
                 rel="noopener noreferrer"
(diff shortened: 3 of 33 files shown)


src/perspective/virtualization-icon.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 
-export default () => (
+export default (): React.JSX.Element => (
```
galkremer1: Import `JSX` from `React` - BTW can you add it to the rules?

src/utils/components/Consoles/components/DesktopViewer/Components/MoreInformationDefault.tsx
```diff
@@ -36,13 +35,15 @@ const MoreInformationDefault: FC<MoreInformationDefaultProps> = ({ textMoreInfoC
           descriptionData={
             <div>
               {t('Download the MSI from ')}
+              {/* eslint-disable i18next/no-literal-string */}
               <a
                 href={documentationURL.VIRT_MANAGER_DOWNLOAD}
                 rel="noopener noreferrer"
                 target="_blank"
               >
```
galkremer1, line 42: We can use a single `{/* eslint-disable-next-line i18next/no-literal-string */}` instead of repeating `{/* eslint-disable i18next/no-literal-string */}` twice in the beginning and the end of the block. 
It saves a line and makes the exception more explicit.

src/utils/components/SidebarEditor/SidebarEditor.tsx
```diff
@@ -40,48 +41,53 @@ const SidebarEditor = <Resource extends K8sResourceCommon>({
...
 
-  const onUpdate = (newResource: Resource) => {
+  const onUpdate = (newResource: Resource): void => {
     if (!onResourceUpdate) return;
 
     setSuccess(false);
     setError(null);
     setLoading(true);
-    return onResourceUpdate(newResource)
+    void onResourceUpdate(newResource)
```
galkremer1, line 78: Check if we can just call `onResourceUpdate` without using `void`

galkremer1: Return type is `void` so it should be fine.

src/utils/components/SidebarEditor/SidebarEditor.tsx
```diff
@@ -40,48 +41,53 @@ const SidebarEditor = <Resource extends K8sResourceCommon>({
...
 
-  const onReload = () => {
+  const onReload = (): void => {
     setSuccess(false);
     setError(null);
     setEditableYAML(resourceYAML);
   };
 
-  const childrenComponent = useMemo(() => {
+  // eslint-disable-next-line
```
galkremer1: We can use `// eslint-disable-next-line react-hooks/exhaustive-deps` rather than just disabling all eslint rules using `// eslint-disable-next-line`
This comment applies to everywhere else `// eslint-disable-next-line` is used

galkremer1 (commented):
Please see my comments

src/perspective/virtualization-icon.tsx
benbassatido: I'll check if there's a rule that we can use to force this

src/utils/components/SidebarEditor/SidebarEditor.tsx
benbassatido: The linter requires void. I tried calling without it and it throws an error

src/utils/components/SidebarEditor/SidebarEditor.tsx
benbassatido: We use eslint-disable-next-line and not specific one is because this is rule is being triggered from sonarjs/function-return-type and in eslint.config.js file we don't have this rule enabled yet. so if I'll use this specific and not generic it will throw another error that I'm trying to disable a rule that is not enabled. 
This line will be removed once sonarjs will be added to the root rules.

src/utils/components/Consoles/components/DesktopViewer/Components/MoreInformationDefault.tsx
benbassatido: The eslint disable comment was removed because I changed it from literal string to Trans like coderabbit suggested

galkremer1:
/override "Hot Cluster E2E / Fork PR"

src/utils/components/SidebarEditor/SidebarEditor.tsx
```diff
@@ -40,48 +41,55 @@ const SidebarEditor = <Resource extends K8sResourceCommon>({
   onResourceUpdate,
   pathsToHighlight = PATHS_TO_HIGHLIGHT.DEFAULT,
   resource,
-}: SidebarEditorProps<Resource>): JSX.Element => {
-  const [editableYAML, setEditableYAML] = useState('');
+}: SidebarEditorProps<Resource>): React.JSX.Element => {
```
galkremer1: import `JSX` from React

locales/es/plugin__kubevirt-plugin.json
```diff
@@ -736,7 +729,7 @@
   "Domain": "Dominio",
   "Down": "Abajo",
   "Download results": "Descargar resultados",
-  "Download the MSI from ": "Descargar el MSI desde ",
+  "Download the MSI from <2>virt-manager.org</2>": "Download the MSI from <2>virt-manager.org</2>",
```
galkremer1: I think it's safe to re-use the existing translations - for example in this case it would be:
`"Download the MSI from <2>virt-manager.org</2>": "Descargar el MSI desde <2>virt-manager.org</2>",`

galkremer1: Same goes for the rest of the translation files.

src/views/virtualmachines/actions/hooks/storageMigrationApi/useClusterStorageMigrationApiProbe.fallback.test.ts
```diff
@@ -0,0 +1,156 @@
+import {
```
galkremer1, line 1: I think it would be better to move tests to a dedicated sub-folder when there is more than a single test file. 
@upalatucci  WDYT?

galkremer1 (changes requested):

src/views/virtualmachines/actions/hooks/storageMigrationApi/useClusterStorageMigrationApiProbe.fallback.test.ts
upalatucci: Yes it's great

benbassatido: Moved all storage migration API tests to a subfolder named storageMigrationAPITests

locales/es/plugin__kubevirt-plugin.json
benbassatido: Fixed

src/utils/components/SidebarEditor/SidebarEditor.tsx
benbassatido: Fixed

galkremer1 (approved):
/lgtm

PR #4166, opened 2026-07-01
CNV-89616: Create VM Wizard refactor - Part 9

mayarub:
## 📝 Description

Replace remaining instance type store usage with react-hook-form fields across wizard steps, and refactor the clone source VirtualMachines list component
## 🔗 Links

Jira ticket: https://redhat.atlassian.net/browse/CNV-89616
## 🎥 Demo

no visual changes

https://github.com/user-attachments/assets/d4d4fa3b-7df3-4154-bcdc-90390a5a7b83

### DIFF
--- a/src/utils/components/ConfigurationSearch/utils/configurationSearch.ts
+++ b/src/utils/components/ConfigurationSearch/utils/configurationSearch.ts
@@ -1,0 +1,33 @@
+import { isEmpty } from '@kubevirt-utils/utils/utils';
+import { SearchItemWithTab } from '@virtualmachines/details/tabs/configuration/utils/search';
+
+const MAX_RESULTS_TO_DISPLAY = 9;
+
+export const getOptions = (
+  searchItems: SearchItemWithTab[],
+  query: string,
+): SearchItemWithTab[] => {
+  if (isEmpty(query)) {
+    return searchItems;
+  }
+  const lowerCaseQueryValue = query.toLowerCase();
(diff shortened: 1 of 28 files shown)


src/utils/components/ConfigurationSearch/utils/configurationSearch.ts
```diff
@@ -0,0 +1,33 @@
...
+
+const MAX_RESULTS_TO_DISPLAY = 9;
+
+export const getOptions = (
+  searchItems: SearchItemWithTab[],
+  query: string,
+): SearchItemWithTab[] => {
+  if (isEmpty(query)) {
+    return searchItems;
+  }
```
upalatucci, line 12: in the original code here we do something different. We don't show all the searchItems but only a part of it : https://github.com/kubevirt-ui/kubevirt-plugin/pull/4166/changes#diff-f7cb74e5728a1b8878356de4bb5a9dfc6a1aecca75c538d63260f81060ef250eL84-L86 


Can u explain the difference here?

mayarub: "query" corresponds to "newValue" in the previous code. I noticed that on line 86, we check if newValue is empty — if so, we set searchItems as-is; otherwise, we use the filtered result
I extracted it to a function in order to make it easier to read

Image

upalatucci (approved):

PR #4169, opened 2026-07-01
CNV-91884: Console stand alone - doc title VM name.

metalice:
## 📝 Description

Fix return types 

image

Add exceptions for id-length rule 

## 🔗 Links

## 🎥 Demo

## Summary by CodeRabbit

* **New Features**
  * Console pages now set a clearer browser/document title for standalone virtual machine views.

* **Bug Fixes**
  * Improved watch-hook error handling so failures are surfaced more consistently across cluster and fleet data updates.
  * SSH access hook now returns an explicit error and separates loading state from failure state.
  * Descheduler details now expose a proper `Error` value and use safer cluster selection.
  * CSV details now reliably produce an `Error[]` when multiple load issues occur.

* **Chores**
  * Updated lint rules to allow additional short identifier names.


src/views/clusteroverview/utils/hooks/useKubevirtCSVDetails.ts
```diff
@@ -40,7 +44,7 @@ export const useKubevirtCSVDetails = (cluster?: string): UseKubevirtCSVDetails =
     },
   );
 
-  const loadErrors = loadSourceError || loadCSVError;
+  const loadErrors = [loadSourceError, loadCSVError].filter(Boolean);
```
upalatucci, line 47: i think was better before. I find this more difficult to read  here

src/utils/components/SSHAccess/useSSHService.tsx
```diff
@@ -39,7 +43,7 @@ const useSSHService = (vm: V1VirtualMachine): UseSSHServiceReturnType => {
     service?.spec?.ports?.find((port) => parseInt(port.targetPort, 10) === SSH_PORT),
   );
 
-  return [sshVMIService, servicesLoaded || servicesError];
+  return [sshVMIService, servicesLoaded, servicesError];
```
upalatucci, line 46: I think the `||` condition was there because when we have an error on fetching the service, we get an infinite loading skeleton. We should pass the error to the `SSHAccess` then and handle this case there

src/views/clusteroverview/utils/hooks/useKubevirtCSVDetails.ts
metalice: Actually, because it was any, it wasnt in the correct type, loadErrors type is Error[] . so actually now its right. before that it was error anod not error[] as type expected.

src/utils/components/SSHAccess/useSSHService.tsx
metalice: Oh, I see, in this case, the error handling is wrong in the SSH component, it should also check for error, as the type definition is boolean, and because it was any, it was ok, but it is actually wrong and should be amended in the SSH component, - @upalatucci could u open a bug for it please?

upalatucci: Sure

upalatucci (approved):
jira bug to solve https://redhat.atlassian.net/browse/CNV-91998

upalatucci:
/override "Run E2E Tests / Build Kubevirt Plugin Image"

PR #4099, opened 2026-06-17
CNV-74265: Hot cluster CI update

galkremer1:
## 📝 Description

Continues [CNV-74265](https://redhat.atlassian.net/browse/CNV-74265) / PR #3713: move KubeVirt plugin E2E from Prow (new AWS cluster per run) to a **persistent IBM Cloud ROKS hot cluster** with GitHub Actions + ARC.

This PR rebases Scott's hot-cluster POC onto current `main` and hardens it for upstream use:

- **Playwright** gating tests (replacing Cypress)
- **Workflow cleanup:** `hot-cluster-e2e.yml` (PR + manual entry) and `hot-cluster-e2e-run.yml` (build plugin, provision test stack, run tests); removed duplicate `poc-e2e-ci-test*.yml`
- **Secrets:** use org `IC_KEY` for IBM Cloud (replacing `IBM_CLOUD_API_KEY`)
- **Hardening:** secure `oc` install via `install-oc-client.sh`, HCO InstallPlan polling, fail-fast runner checks, per-run plugin image tags

Cluster setup/teardown workflows (`ibmc-cluster-*.yml`), ARC, ci-env-controller, and ci-test-stack Helm charts are included from the POC. See [ci-scripts/README.md](./ci-scripts/README.md) for architecture and testing steps.

## 🔗 Links

Jira ticket: https://redhat.atlassian.net/browse/CNV-74265

Related: https://github.com/kubevirt-ui/kubevirt-plugin/pull/3713

## 🎥 Demo

WIP — first green run pending IBM Cloud IAM completion.

**Latest test runs (2026-06-24):**
- [IBM Cloud Hot Cluster Setup #28109623659](https://github.com/kubevirt-ui/kubevirt-plugin/actions/runs/28109623659) — failed at cluster create (**E73e6**: classic infrastructure permissions; A0010 appears resolved)
- [Hot Cluster E2E #28109713900](https://github.com/kubevirt-ui/kubevirt-plugin/actions/runs/28109713900) — failed at health check (**G0004**: cluster not found)

See [testing update comment](https://github.com/kubevirt-ui/kubevirt-plugin/pull/4099#issuecomment-4790904630) for IAM details and next steps.

## Test plan

- [x] Push branch and open PR against `main`
- [ ] Run **IBM Cloud Hot Cluster Setup** if no hot cluster / ARC exists — blocked on Classic infrastructure Super User for `IC_KEY`
[...]

PR #4175, opened 2026-07-01
CNV-91940 resolve-console-image.sh overrides CONSOLE_IMAGE default in start-console.sh

galkremer1:
## 📝 Description

`resolve-console-image.sh` was added in PR #4099 ([CNV-74265](https://redhat.atlassian.net/browse/CNV-74265)) for CI hot-cluster E2E. It auto-detects the cluster's OpenShift version and sets `CONSOLE_IMAGE` accordingly. When wired into start-console.sh, it made the existing fallback line unreachable.

In order for the `CONSOLE_IMAGE` version to be overridden `CONSOLE_IMAGE_SKIP_DETECTION` needs to be set to `false`.

## 🔗 Links

Jira ticket: [CNV-91940](https://redhat.atlassian.net/browse/CNV-91940)

## 🎥 Demo

https://github.com/user-attachments/assets/b3b5b27f-0eca-4b67-a3b3-3ea974dea56c

### DIFF
--- a/ci-scripts/resolve-console-image.sh
+++ b/ci-scripts/resolve-console-image.sh
@@ -4,6 +4,7 @@
 set -euo pipefail
 
 CONSOLE_IMAGE_REGISTRY="${CONSOLE_IMAGE_REGISTRY:-quay.io/openshift/origin-console}"
+CONSOLE_IMAGE_DEFAULT_TAG="${CONSOLE_IMAGE_DEFAULT_TAG:-latest}"
 
 if [[ -n "${CONSOLE_IMAGE:-}" ]]; then
   echo "CONSOLE_IMAGE already set (${CONSOLE_IMAGE}); skipping cluster version lookup." >&2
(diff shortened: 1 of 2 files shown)


galkremer1:
/hold
Waiting for @metalice approval

ci-scripts/resolve-console-image.sh
```diff
@@ -4,23 +4,44 @@
 set -euo pipefail
 
 CONSOLE_IMAGE_REGISTRY="${CONSOLE_IMAGE_REGISTRY:-quay.io/openshift/origin-console}"
+CONSOLE_IMAGE_DEFAULT_TAG="${CONSOLE_IMAGE_DEFAULT_TAG:-latest}"
```
upalatucci, line 7: can we just say something like: if we have CONSOLE_IMAGE env variable defined, use this and do not detect ?

galkremer1: When we use the explicit env var in the `cli` it's not an issue - the problem is when someone is explicitly changing it using the file. Since we have the auto discovery it is getting overridden by this script.

upalatucci (approved):

PR #4178, opened 2026-07-01
CNV-91510 Auto-provision hot cluster when PR E2E finds no active cluster

galkremer1:
## 📝 Description

Make the hot-cluster CI self-managing so no manual cluster provisioning is needed when a PR triggers E2E.

## 🔗 Links

Jira ticket: [CNV-91510](https://redhat.atlassian.net/browse/CNV-91510)

## 🎥 Demo

N/A

### DIFF
--- a/.github/workflows/ok-to-test-reset.yml
+++ b/.github/workflows/ok-to-test-reset.yml
@@ -1,0 +1,26 @@
+name: Reset ok-to-test on push
+
+on:
+  pull_request_target:
+    branches: [main]
+    types: [synchronize]
+
+permissions:
+  pull-requests: write
+
+jobs:
+  remove-label:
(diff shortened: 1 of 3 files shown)


.github/workflows/ok-to-test-reset.yml
upalatucci, line 1: We have to try this one as i'm sure that the event trigger the other actions anyway but maybe this remove it before the other actions test for the existing of this label

upalatucci (approved):

upalatucci:
@galkremer1  if we add more workflows i would create a directory just for this hot cluster

galkremer1:
> @galkremer1 if we add more workflows i would create a directory just for this hot cluster

Yes once everything is working I will do a cleanup.

PR #4189, opened 2026-07-02
CNV-92017 Upload file via bootable volume triggers React error #185 (infinite loop) on Console 4.22

galkremer1:
## 📝 Description

Fixes a React error that crashes the app when `useToast` is not available from `console`.
This should not be an issue when `useToast` is available however `noopResult` should be a reliable backup option and should not crash the app. 

## 🔗 Links

Jira ticket: [CNV-92017](https://redhat.atlassian.net/browse/CNV-92017)

## 🎥 Demo

Before:

https://github.com/user-attachments/assets/e325313c-fe1d-4768-b3fe-976bbfcfed79

After (note that the toast notification is not shown as expected due to a mismatch on console version): 

https://github.com/user-attachments/assets/b568361d-0f36-4b1a-8c3c-d549685e8acc

### DIFF
--- a/src/utils/hooks/useKubevirtToast.ts
+++ b/src/utils/hooks/useKubevirtToast.ts
@@ -24,7 +24,7 @@ type UseKubevirtToastResult = ToastActions & {
   addToast: AddToast;
 };
 
-const noopAddToast = () => '';
+const noopAddToast = () => 'noop';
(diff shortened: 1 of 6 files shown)


upalatucci (approved):

upalatucci:
/override "Run E2E Tests / Build Kubevirt Plugin Image"

PR #4208, opened 2026-07-05
CNV-90382: remove jsdoc rule and add promise rule

benbassatido:
## 📝 Description

Removes jsdoc rule and adds promise rule to root lint config. 

## 🔗 Links

## Summary by CodeRabbit

* **New Features**
  * Added a streamlined PVC-backed disk upload workflow with template selection, access checks, validation, certificate checks, progress tracking, and upload status handling.
  * Improved synchronization between form-based and YAML-based editors, including conversion warnings and clearer switching behavior.
  * Added reusable template processing and update operations for creating drafts and applying generated parameters.

* **Bug Fixes**
  * Improved upload, template, settings, and migration error handling so loading states consistently reset.
  * Preserved valid empty values in modal actions and configuration settings.
  * Improved detection of bootable volumes that are ready, cloning, uploading, or still importing.

### DIFF
--- a/src/views/virtualmachines/actions/hooks/storageMigrationApi/probeEffects.ts
+++ b/src/views/virtualmachines/actions/hooks/storageMigrationApi/probeEffects.ts
@@ -4,18 +4,18 @@ import {
   VirtualMachineStorageMigrationPlanModel,
 } from '@kubevirt-utils/models';
 import {
-  type StorageMigrationAPI,
   MTC_MIGRATION_NAMESPACE,
   STORAGE_MIGRATION_API,
+  type StorageMigrationAPI,
 } from '@kubevirt-utils/resources/migrations/constants';
 import { kubevirtConsole } from '@kubevirt-utils/utils/utils';
 import { kubevirtK8sListItems } from '@multicluster/k8sRequests';
 
 import {
-  type StorageMigrationProbeFallbackPhase,
   csvLoadedIndicatesMultiNsStorageMigrationApi,
   STORAGE_MIGRATION_PROBE_PHASE_IDLE,
   STORAGE_MIGRATION_PROBE_PHASE_WAITING_CSV_AFTER_MULTI_NS_404,
+  type StorageMigrationProbeFallbackPhase,
 } from './constants';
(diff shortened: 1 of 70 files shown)


.husky/pre-push
```diff
@@ -3,17 +3,4 @@ if [ "$branch" = "main" ]; then
...
-done
-
-if [ -z "$merge_base" ]; then
-  echo "Warning: could not find upstream/main or origin/main — skipping commitlint"
-  exit 0
-fi
-
-npx commitlint --from "$merge_base" --to HEAD
\ No newline at end of file
+npx commitlint --from "$(git merge-base HEAD upstream/main)" --to HEAD
```
upalatucci: is this necessary? Gal changed that as he does not have upstream/main locally. He called it origin/main. If possible, we would like to support either upstream/main or origin/main

src/views/virtualmachines/actions/hooks/storageMigrationApi/probeEffects.ts
```diff
@@ -35,6 +35,7 @@ export const probeSingleNsApiOrNone = (
   })
     .then(() => {
       finish(STORAGE_MIGRATION_API.SINGLE_NS);
+      return undefined;
```
upalatucci: i don't know if i like this rule

benbassatido: Fixed, added { ignoreLastCallback: true } to the rule.

.husky/pre-push
benbassatido: Fixed, changed to Gal's pre-push

galkremer1 (approved):
/lgtm

upalatucci:
I agree with coderabbit. If we force to have a catch in all promises we need to make sure that we have a re-throw function or a noop function that at least `kubevirtConsole.error` something and use them depending on what the developer wants. 
So on one hand we'll force developers to explicitly tell what needs to be done in each promise without having empty functions that will be added automatically without thinking .

I don't like we add `.catch(() => {})` to all promises . seems like this eslint rule will not have much sense

PR #4212, opened 2026-07-07
CNV-87133: allign operators installation pr3

batyana:
## 📝 Description
Rebuilds the data layer for the Recommended Capabilities tab with proper
install tracking and bundle installation support.
- **Context refactor**: replaces `RecommendedCapabilitiesContext` with a
  split `CapabilitiesDataProvider` / `useCapabilitiesData` / `useCapabilitiesActions`
  pattern for cleaner separation of data and actions
- **`buildRecommendedDetailsMap`** (`detailsMap.ts`): computes install state
  for each operator by looking up its Subscription and CSV. Handles operators
  whose PackageManifest isn't visible in the catalog (e.g. installed from a
  non-standard source) via the `missingOperators` fallback path
- **`useInstallBundle`**: orchestrates parallel bundle installation with
  deduplication, error collection, and a post-install delay (`isAwaitingOLM`)
  that keeps the button disabled until OLM picks up the subscriptions
- **`useInstallResources`**: extracted hook that watches namespaces and
  console operator config needed to perform installations
- **`capabilityFeatures`**: defines the full capability→operator mapping for
  the recommended bundle

## 🔗 Links

Jira ticket: https://redhat.atlassian.net/browse/CNV-87133

## Summary by CodeRabbit

* **New Features**
  * Recommended capabilities can now be installed using predefined bundles in a single action, with installation progress updating once required operator resources are ready.
  * Improved handling and display of alternative operator choices during installation.
  * Refreshed capability installation status labels for clearer, consistent UI presentation.
* **Bug Fixes**
  * Fixed capability matching for Node Feature Discovery by updating the detected operator identifier.
* **UI Changes**
  * Adjusted the Recommended Capabilities gallery sizing for improved layout.


upalatucci:
lgtm for me the second commit

src/views/settings/tabs/RecommendedCapabilitiesTab/hooks/useInstallBundle.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  useEffect(() => {
+    if (isAwaitingOLM && isBundleInstallInProgress) {
+      setIsAwaitingOLM(false);
+    }
+  }, [isAwaitingOLM, isBundleInstallInProgress]);
+
+  useEffect(() => {
+    if (!isAwaitingOLM) return undefined;
+    const timer = setTimeout(() => setIsAwaitingOLM(false), 30_000);
```
galkremer1: Please use a const for 30_000

src/views/settings/tabs/RecommendedCapabilitiesTab/hooks/useInstallBundle.ts
```diff
@@ -0,0 +1,127 @@
...
+    if (isInstallingRef.current) return;
+
+    const manifestsToInstall = getNonInstalledBundleManifests(
+      features,
+      detailsMap,
+      filteredPackageManifests,
+    );
+
+    if (isEmpty(manifestsToInstall)) {
+      addDangerToast({
```
galkremer1, line 69: Do you want this toast to be ephemeral? 
If not - after my [PR](https://github.com/kubevirt-ui/kubevirt-plugin/pull/4241) is merged you can make it persistent in the drawer.

src/views/settings/tabs/RecommendedCapabilitiesTab/hooks/useInstallBundle.ts
```diff
@@ -0,0 +1,127 @@
...
+    isInstallingRef.current = false;
+
+    if (isEmpty(errors)) {
+      addSuccessToast({ title: t('Bundle installation started successfully') });
+      setIsAwaitingOLM(true);
+      return;
+    }
+
+    addDangerToast({
+      title: t('Some operators failed to install ({{count}} error(s))', {
```
galkremer1: If you are using `error(s)`, it's better to use `errorCount` instead of `count` so the `_other` translation would not be auto generated.

src/views/settings/tabs/RecommendedCapabilitiesTab/hooks/useInstallBundle.ts
galkremer1: Also - it it possible to avoid using a `setTimeout`?

src/views/settings/tabs/RecommendedCapabilitiesTab/hooks/useInstallBundle.ts
```diff
@@ -0,0 +1,127 @@
...
+    );
+
+    if (isEmpty(manifestsToInstall)) {
+      addDangerToast({
+        title: t('No operators available to install. OperatorHub catalog may not be configured.'),
+      });
+      return;
+    }
+
+    setIsInstalling(true);
```
galkremer1: This is going to be a bit better stuctured and will make sure that `updateInstallingState` is executed in finally
```
const updateInstallingState = (installing: boolean) => {
  setIsInstalling(installing);
  isInstallingRef.current = installing;
};
```

```
updateInstallingState(true); // Using the helper from above
const errors: Error[] = [];

try {
  await Promise.allSettled(
    manifestsToInstall.map((pkg) =>
      // ... your operator logic
    )
  );
} finally {
  // This will ALWAYS run, even if something throws above
  updateInstallingState(false);
}
```

src/views/settings/tabs/RecommendedCapabilitiesTab/RecommendedCapabilitiesTab.tsx
```diff
@@ -4,21 +4,21 @@ import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTransla
...
+import { getRecommendedCapabilityFeatures } from './utils/capabilityFeatures';
 
 const RecommendedCapabilitiesTab: FC = () => {
   const { t } = useKubevirtTranslation();
   const features = useMemo(() => getRecommendedCapabilityFeatures(t), [t]);
 
   return (
-    <RecommendedCapabilitiesContextProvider>
+    <CapabilitiesDataProvider>
       <Gallery hasGutter minWidths={{ default: '280px' }}>
```
galkremer1: Not a big deal but we should use `rem` instead of `px` (`17.5rem` instead of `280px`)

src/views/settings/tabs/RecommendedCapabilitiesTab/utils/utils.ts
```diff
@@ -89,21 +23,55 @@ export const computeCapabilityInstallState = (
...
+      .filter(isBundleFeature)
+      .flatMap(({ operators }) => operators.map(({ packageName }) => packageName)),
+  );
+
+  const seen = new Set<string>();
+
+  return filteredPackageManifests.filter((pkg) => {
+    const name = getName(pkg);
+    if (!defaultBundlePackageNames.has(name)) return false;
+    if (detailsMap[name]?.installState !== InstallState.NOT_INSTALLED) return false;
```
galkremer1, line 64: We can make it a bit shorter by replacing lines 65-67 with 
```
return seen.has(name) ? false : (seen.add(name), true);
```

src/views/settings/tabs/RecommendedCapabilitiesTab/hooks/useInstallBundle.ts
batyana: currently it should be ephemeral, maybe as a follow up i will incorporate update status of the amount of operators installed out of total. I will talk to Yifat and Stefan about it

src/views/settings/tabs/RecommendedCapabilitiesTab/hooks/useInstallBundle.ts
batyana: the timeout is just a safety fallback, it only fires if OLM never picks up the subscriptions so the button doesn't stay disabled forever.

galkremer1 (approved):
/lgtm

PR #4215, opened 2026-07-07
CNV-91998: Surface SSH access fetch errors instead of infinite loading

upalatucci:
## 📝 Description

When loading SSH service data fails (e.g. API or permission error), the SSH access section previously showed an infinite loading spinner because only the `loaded` boolean was checked.

This change passes the service fetch error from `useSSHService` through `SSHAccess` into `SSHCommand`, combines it with VMI/pod load errors, and renders the result via `StateHandler` so users see a clear error state (including 403 access denied) instead of perpetual loading.

## 🔗 Links

- [CNV-91998](https://redhat.atlassian.net/browse/CNV-91998)

## 🎥 Demo

N/A — error state replaces infinite loading when the services API call fails.

## Test plan

- [ ] Open a VM details page → Configuration → SSH access with normal permissions; verify SSH controls and command still load correctly
- [ ] Revoke the user's permission to list services in the namespace; verify the SSH access section shows an error instead of a loading spinner
- [ ] Change SSH service type on a VM with permissions; verify mutation errors still display below the controls
- [ ] Verify SSH access on the VMI details page behaves the same way

### DIFF
--- a/src/utils/components/SSHAccess/SSHAccess.tsx
+++ b/src/utils/components/SSHAccess/SSHAccess.tsx
@@ -10,6 +10,7 @@ import SSHCommand from './components/SSHCommand';
 type SSHAccessProps = {
   isCustomizeInstanceType?: boolean;
   sshService: IoK8sApiCoreV1Service;
+  sshServiceError?: Error;
   sshServiceLoaded?: boolean;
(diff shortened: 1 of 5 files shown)


galkremer1 (approved):
/lgtm

PR #4173, opened 2026-07-01
CNV-91888: Add Konflux hermetic build CI check

upalatucci:
## Summary

Adds a `hermetic-build` CI job that simulates Konflux offline installs:

1. Prefetch npm dependencies with [Hermeto](https://hermetoproject.github.io/hermeto/npm/) (`fetch-deps` + `inject-files`)
2. Run `npm ci --offline --ignore-scripts` with no npm registry access

This mirrors the cachi2/hermeto prefetch flow used in Konflux container builds.

## Expected behavior

This PR is **expected to fail on `main`** because `package-lock.json` hoists `cosmiconfig@8.3.6` while `cosmiconfig-typescript-loader` requires `cosmiconfig >= 9`. Unresolved peer dependencies are not prefetched by Hermeto, causing offline `npm ci` to fail.

## Fix

Merge https://github.com/kubevirt-ui/kubevirt-plugin/pull/4167 first — it pins `cosmiconfig@^9.0.2` as a devDependency so the lockfile satisfies the peer and hermetic installs succeed.

- Jira: https://redhat.atlassian.net/browse/CNV-91888

## Test plan

- [ ] CI `hermetic-build` job fails on this PR (demonstrates the problem on main)
- [ ] Rebase on #4167 or merge #4167 first, then re-run — `hermetic-build` should pass

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Chores**
  * Enhanced CI build checks to more reliably validate dependency installs when the lockfile changes.
  * Added an offline installation verification step using pre-fetched dependencies to detect environment or caching issues earlier.
  * Tightened workflow permissions and disabled persistent checkout credentials to improve security.

PR #4279, opened 2026-07-10
[release-4.18] CNV-92658: Add Konflux hermetic build CI check

upalatucci:
## Summary

Backport of #4173 to `release-4.18`.

Adds a conditional Hermeto prefetch/offline `npm ci` verification step to the `build` job when `package-lock.json` changes, mirroring Konflux container builds.

Also aligns Victory chart dependencies so the hermetic/offline build succeeds without webpack warnings:

- `victory-core`: `~36.8.1` → `^36.9.2`
- `@patternfly/react-charts`: `7.1.1` → `7.3.1`

### Why the Victory dependency changes are in this PR

The PatternFly pin commit (`9f7b0bb`) locked `victory-core` to `~36.8.1` and `@patternfly/react-charts` to `7.1.1`. That combination is internally inconsistent once npm resolves the full lockfile:

1. **Victory subpackages expect 36.9.x APIs** — packages like `victory-brush-container` import `Helpers.isFunction`, `mapValues`, and `isNil`, which exist only in `victory-core@36.9.x`. Webpack resolves all `victory-core` imports to the root copy (see `webpack.config.ts`), so the pinned `36.8.6` is used everywhere, producing ~50 build warnings.
2. **`react-charts@7.1.1` imports `CommonProps` removed in 36.9.0** — upgrading `victory-core` alone fixes (1) but introduces a `CommonProps` warning from `@patternfly/react-charts`. PatternFly fixed this in `7.2.1+` ([patternfly-react#10082](https://github.com/patternfly/patternfly-react/pull/10082)); `7.3.1` targets Victory `36.9.x` and is the minimal compatible bump.
3. **Hermetic build must compile cleanly** — the new Konflux/Hermeto CI step runs `npm ci --offline` and a full webpack build. These Victory export mismatches surface as build warnings/errors during that verification, so the dependency alignment is required for the CI check added by this PR to pass reliably.

A small type fix in `NetworkThresholdChartSingleSource.tsx` adapts chart tooltip formatting to stricter `react-charts@7.3.1` types (`datum.x` is typed as `number`, not `Date`). Runtime behavior is unchanged.

## Jira

- [CNV-92658](https://redhat.atlassian.net/browse/CNV-92658) - Fix version: CNV v4.18.z

## Test plan

- [ ] CI `build` job runs Hermeto prefetch + offline install when `package-lock.json` changes
- [ ] `npm run build` — no `Helpers` / `CommonProps` / `victory-core` warnings
- [ ] CI passes on this PR
- [ ] Smoke-test chart tooltips (network utilization chart shows `HH:MM` on hover)


upalatucci:
Pushed an additional commit to fix Victory chart dependency mismatches uncovered during local `npm run build` while validating the lockfile changes.

**Commit:** `c272cd7` — *[CNV-92658](https://redhat.atlassian.net/browse/CNV-92658): Align Victory chart dependencies for hermetic build*

**Changes:**
- `victory-core` `~36.8.1` → `^36.9.2`
- `@patternfly/react-charts` `7.1.1` → `7.3.1`
- Small tooltip type fix in `NetworkThresholdChartSingleSource.tsx`

**Why this belongs in this PR:** The hermetic CI step added here runs a full offline `npm ci` + webpack build. The existing PatternFly/victory pins produced ~50 `victory-core` export warnings (`Helpers.isFunction`, `CommonProps`, etc.) that would fail or noise that verification. Both dependency bumps are needed together — upgrading only `victory-core` swaps one set of warnings for another.

PR #4322, opened 2026-07-14
CNV-91407: fix default ssh key not applied in create vm wizard

batyana:
## 📝 Description

The new VM creation wizard was missing SSH key handling — default SSH keys configured in user settings were never applied to VMs created via either the instance-type or template flows. This fix integrates the default SSH key directly into the VM definition during generation, with proper detection of dynamic credentials support based on the boot volume label. It also fixes the template flow which was using the wrong hook to resolve the cluster value, causing the user settings lookup to silently fail and adds testing. 

## 🔗 Links

Jira ticket: https://redhat.atlassian.net/browse/CNV-91407

## 🎥 Demo

Before:

https://github.com/user-attachments/assets/4f4efdd7-1875-4cc6-a54d-c700a4734996

After:

https://github.com/user-attachments/assets/db562fab-f4cf-4653-b1e4-75cc4753c945

## Summary by CodeRabbit

* **New Features**
  * Virtual machines created through the wizard or from templates can now automatically use the authorized SSH key secret for the selected namespace.
  * SSH credentials are applied appropriately for workloads supporting dynamic credentials.
* **Bug Fixes**
  * Improved cluster selection handling when creating a virtual machine from a template.
  * Ensured driver information is available consistently during VM generation.

### DIFF
--- a/src/views/virtualmachines/wizard/steps/InstanceTypesSteps/hooks/useGenerateVM/utils/generateVM.ts
+++ b/src/views/virtualmachines/wizard/steps/InstanceTypesSteps/hooks/useGenerateVM/utils/generateVM.ts
@@ -1,5 +1,7 @@
 import { VirtualMachineModel } from '@kubevirt-ui-ext/kubevirt-api/console';
 import { V1VirtualMachine } from '@kubevirt-ui-ext/kubevirt-api/kubevirt';
+import { DYNAMIC_CREDENTIALS_SUPPORT } from '@kubevirt-utils/components/DynamicSSHKeyInjection/constants/constants';
+import { addSecretToVM } from '@kubevirt-utils/components/SSHSecretModal/utils/utils';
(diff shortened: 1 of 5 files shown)


upalatucci (approved):

src/views/virtualmachines/wizard/steps/InstanceTypesSteps/hooks/useGenerateVM/utils/generateVM.ts
```diff
@@ -59,6 +62,11 @@ export const generateVM: GenerateVMCallback = ({
     }),
   };
 
+  if (sshSecretName) {
+    const isDynamic = getLabel(selectedBootableVolume, DYNAMIC_CREDENTIALS_SUPPORT) === 'true';
```
metalice, line 66: we have about 60 places in our code base that check === true , please create another subtask in the bug, converting all those places to a utility function

galkremer1 (approved):
/lgtm

PR #4358, opened 2026-07-16
[release-4.22] CNV-92871: Fix Fleet Virtualization Customization step hang

mayarub:
## 📝 Description

- On Fleet, Templates were loaded via multicluster search, which often omitted nested `objects`, so VM create could not resolve a VirtualMachine.
- When a single managed cluster is selected, load Templates with fleet k8s watch instead of search, and enrich watch results with `kind`/`apiVersion`

## 🔗 Links

Jira ticket: https://redhat.atlassian.net/browse/CNV-92871
## 🎥 Demo

before:

https://github.com/user-attachments/assets/ed22c5db-4be2-4a70-b89c-a139f66f3c9d

after:

https://github.com/user-attachments/assets/c813f58a-63b1-4bdf-9554-f0b263923922


galkremer1 (approved):
/lgtm

PR #4382, opened 2026-07-19
CNV-74205: labels and annotations tab redesign

batyana:
## 📝 Description

Redesigns the VM Configuration > Metadata tab into a structured "Labels and Annotations" view with:

- Key-value tables for labels and annotations
- System-created entries are displayed but non-deletable
- Clickable label values navigate to CNV advanced search (VM list with filter)
- "Advanced view" toggle preserves access to the classic metadata editor

## 🔗 Links

Jira ticket: https://redhat.atlassian.net/browse/CNV-74205

## 🎥 Demo

https://github.com/user-attachments/assets/c33f5e36-467e-4342-91e4-eee233989bbf

## Summary by CodeRabbit

* **New Features**
  * Updated the VM metadata experience to emphasize **“Labels and annotations”**, with an improved basic view and an advanced YAML sidebar editor.
  * Added dedicated label/annotation management components, including search integration and support for adding and deleting entries.
  * Expanded configuration search to include **metadata**, **labels**, and **annotations** sections.

* **Improvements**
  * System-managed labels/annotations (by known prefixes) are now protected from deletion.
  * VM details and the customization wizard now share the same labels/annotations editor experience.

PR #4383, opened 2026-07-19
CNV-74205: labels edit modal redesign

batyana:
## 📝 Description

Replaces the labels editor with a row-based key-value modal featuring:

- Individual key/value input fields per row with add/delete controls
- Key input with suggested label keys
- Inline validation for new labels 
- Save disabled when keys are empty or validation errors exist

## 🔗 Links

Jira ticket: https://redhat.atlassian.net/browse/CNV-74205

## 🎥 Demo

https://github.com/user-attachments/assets/b843ed7d-2f40-4460-b2a0-236f208bd7cd

## Summary by CodeRabbit

* **New Features**
  * Added an “Edit labels” modal with per-row editing, inline validation, duplicate-key protection, and suggested label key suggestions.
  * Updated labels editing to use the new modal flow and keep label add/edit/delete interactions consistent.
* **Bug Fixes**
  * Improved system label protection by aligning deletion checks with system key detection for both labels and annotations.
* **Changes**
  * Simplified label submission behavior by removing previous system-label preservation/merging during updates.


batyana:
/hold 
needs https://github.com/kubevirt-ui/kubevirt-plugin/pull/4382 to merge first

src/utils/components/LabelsModal/utils.ts
```diff
@@ -14,3 +27,64 @@ export const labelsArrayToObject = (labels: string[]): { [key: string]: string }
...
+
+export const isSystemKey = (key: string): boolean =>
+  SYSTEM_LABEL_PREFIXES.some((prefix) => key.startsWith(prefix));
+
+export const hasDuplicateKeys = (entries: LabelEntry[]): boolean => {
+  const keys = entries.map(({ key }) => key);
+  return new Set(keys).size !== keys.length;
+};
+
+export const validateLabelEntry = (
```
upalatucci: In this function we do some validations that backend do for us. I don't want to do multiple validations of the same thing like: `K8S_LABEL_REGEX` or `MAX_KEY_PREFIX_LENGTH`. 

I think we should alert user for the duplicate key (as backend just take the last value and ignore the previous one) and `Cannot use system-managed key prefix`. Maybe we can consider something else if it has a reason to be in the frontend

src/utils/components/LabelsModal/hooks/useLabelsModalState.ts
```diff
@@ -0,0 +1,101 @@
...
+const useLabelsModalState = ({
+  initialLabels,
+  obj,
+  onLabelsSubmit,
+}: UseLabelsModalStateParams): UseLabelsModalStateReturn => {
+  const { t } = useKubevirtTranslation();
+
+  const initLabels = useMemo(() => {
+    if (initialLabels !== undefined) return initialLabels;
+    if (!isEmpty(obj?.metadata?.labels)) return obj?.metadata?.labels;
```
upalatucci: Use getLabels

upalatucci:
@coderabbitai why tests are failing ?

PR #4385, opened 2026-07-20
fix: View storage migrations 404 — route to MultiNamespaceVirtualMachineStorageMigrationPlan

sarpatil-07:
## 📝 Description

**Context:** From OpenShift Virtualization 4.20, live storage class migration uses the **native** API (`MultiNamespaceVirtualMachineStorageMigrationPlan`). MTC is decommissioned for this flow — the bug reproduces with **MTC uninstalled**.

**Bug (4.20.15):** After initiating Migration → Storage on a running VM, the status popup shows **View storage migrations**. Clicking it navigates to `/k8s/.../storagemigrations`. That plural is not a CRD, so the console shows:

> 404: Page Not Found  
> The server doesn't have a resource type "storagemigrations".

**Expected (works on 4.22):** Open the native `MultiNamespaceVirtualMachineStorageMigrationPlan` list / workflow.

**Fix:** Build the list URL from the native plan model GVK via `getResourceUrl` (same approach 4.22 uses when the custom `/storagemigrations` route is not available — spokes / CNV < 4.23). Register `console.page/resource/list` for that model so the existing Storage Migration list page still renders. Point Migration nav (admin + virt perspective) at the native model instead of the synthetic `storagemigrations` href / legacy MigPlan.

### Changes
- `getStorageMigrationsListUrl()` → native `MultiNamespaceVirtualMachineStorageMigrationPlan` GVK path
- **View storage migrations** (post-initiate popup) uses that URL
- Nav + `console.page/resource/list` for the native plan model
- Legacy `/storagemigrations` `console.page/route` kept for old bookmarks only

### Duplicate check
No other open PR covers this. Related historical work (#3977 / CNV-88330) only ungated the custom route and is **not** in `v4.20.15-rc.0`; this PR fixes the link target itself for the native path.

## 🔗 Links

Jira: _please add CNV ticket_ (PR currently blocked by `do-not-merge/jira-invalid`)

## Test plan

- [ ] Cluster: OCP/CNV **4.20.x**, native live storage migration enabled, **MTC not installed**
- [ ] VM Actions → Migration → Storage → complete initiate flow until status popup
- [ ] Click **View storage migrations** → opens `MultiNamespaceVirtualMachineStorageMigrationPlan` list (no 404)
- [ ] Migration → Storage migrations nav opens the same native list
- [ ] Confirm behavior matches 4.22 for this link

### DIFF
--- a/src/perspective/extensions.ts
+++ b/src/perspective/extensions.ts
@@ -194,6 +194,9 @@ const migrationSection = [
     type: 'console.navigation/section',
   } as EncodedExtension<NavSection>,
   {
+    flags: {
+      required: ['SHOW_MIGRATION_SECTION'],
+    },
     properties: {
       dataAttributes: {
         'data-quickstart-id': 'qs-nav-storagemigrations-virt-perspective',
@@ -201,8 +204,8 @@ const migrationSection = [
       },
       id: 'storagemigrations-virt-perspective',
       model: {
-        group: 'migration.openshift.io',
-        kind: 'MigPlan',
+        group: 'migrations.kubevirt.io',
+        kind: 'MultiNamespaceVirtualMachineStorageMigrationPlan',
         version: 'v1alpha1',
       },
       name: '%plugin__kubevirt-plugin~Storage migrations%',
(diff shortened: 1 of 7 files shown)


sarpatil-07:
Closing as not needed / overlapping with already-merged #3977.

#3977 ungates the `/storagemigrations` `console.page/route` so **View storage migrations** keeps working with the same URL pattern as 4.22. This PR took a different approach (GVK `MultiNamespaceVirtualMachineStorageMigrationPlan` links) that is redundant if #3977 is present in the build.

PR #4393, opened 2026-07-21
CNV-93099: [release-4.17] DOMPurify: Cross-site scripting vulnerability allows code execution

UdayYendva:
Fix raised to bump vulnerable dompurify package to patched version

### DIFF
--- a/package.json
+++ b/package.json
@@ -124,6 +124,7 @@
     },
     "@types/react": "^17.0.40",
     "@types/react-dom": "^17.0.0",
+    "dompurify": "^3.4.6",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -124,6 +124,7 @@
     },
     "@types/react": "^17.0.40",
     "@types/react-dom": "^17.0.0",
+    "dompurify": "^3.4.6",
```
upalatucci, line 127: we don't have to override. we just have to bump dompurify to `2.5.9` version in this case were we only use version 2

PR #4397, opened 2026-07-21
CNV-93554: Add optional category when saving a VM as a template

upalatucci:
## 📝 Description

Adds an optional **Category** field to the **Save as template** flow, and fixes ACM fleet listing so VirtualMachineTemplates are discovered across both `v1alpha1` and `v1beta1`.

### Category on save
- Persist category as label `vm.kubevirt.io/category` via VMTR `spec.templateLabels` (controller propagates labels to the created VMT)
- Defaults use label-safe IDs (e.g. `operating-systems`) with display labels in the typeahead (e.g. "Operating systems")
- Custom categories normalize spaces to `-` and validate against Kubernetes label-value rules
- Always create **v1beta1** VirtualMachineTemplateRequest

### ACM / multi-version listing
- `useKubevirtSearchPoll` clears `groupVersionKind.version` when polling via fleet search so results include objects from **all served API versions**
- Fixes missing VMTs on clusters that only serve `template.kubevirt.io/v1beta1` when the client previously pinned `v1alpha1`

### Tests / i18n
- Playwright: selects "Operating systems" and asserts label `operating-systems` on the created VMT
- Locale strings for the new Category field

https://github.com/user-attachments/assets/eb6a5351-bba5-42ca-ac0d-40c837db498b

Screenshot 2026-07-23 at 14 38 28

## 🔗 Links

- https://issues.redhat.com/browse/CNV-93554
- Epic: https://issues.redhat.com/browse/CNV-92485
- Follow-up: list/filter/catalog PR (#4398)

## Test plan
- [ ] Save a VM as template with category "Operating systems" → VMTR `spec.templateLabels` and VMT label `vm.kubevirt.io/category=operating-systems`
- [ ] Save without a category → no category label / `templateLabels`
- [ ] Create a new category name from the typeahead (spaces become `-`)
- [ ] ACM: templates list shows VMTs from clusters serving only `v1beta1` (and mixed alpha/beta fleets)


src/views/templates/components/VirtualMachineTemplateRequest/useIsVMTemplateRequestBeta.ts
```diff
@@ -0,0 +1,45 @@
...
+    !isEmpty(namespace)
+      ? {
+          cluster,
+          model: VirtualMachineTemplateRequestBetaModel,
+          namespace,
+        }
+      : null,
+  );
+  const loaded = listResult[1];
+  const loadError = listResult[2] as Error | undefined;
```
metalice: ?

metalice: why same approche as HCO?

src/utils/components/CategorySelect/utils/options.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import { type TFunction } from 'i18next';
+
+import { isEmpty } from '@kubevirt-utils/utils/utils';
+import { HelperText, HelperTextItem, type SelectOptionProps } from '@patternfly/react-core';
+import { InfoIcon } from '@patternfly/react-icons';
+
+export const getCreateNewCategoryOption = (
```
metalice: a util that returns JSX?

src/views/templates/components/VirtualMachineTemplateRequest/useIsVMTemplateRequestBeta.ts
upalatucci: oh yes i have to fix also here. its for the same eslint issue

upalatucci: here we need to understand which request we need to fire as for BetaModel we have a different way to add the label. There is a new api with `templateLabel`. Alpha is deprecated and firing a alpha request when we support beta it will generate a toaster alert to the top left.

src/utils/components/SaveAsTemplateModal/useSaveAsTemplateModal.ts
```diff
@@ -23,20 +32,29 @@ const useSaveAsTemplateModal = (vm: V1VirtualMachine) => {
...
 
-  const onSubmit = async () => {
-    await createVMTemplateRequest(vm, templateName, selectedProject);
     logTemplateCreated({
       osType: getOperatingSystem(vm),
       sourceVmId: getUID(vm),
       workloadProfile: getWorkload(vm),
     });
-    navigate(isACMPage ? getACMTemplateListURL() : getTemplateListURL(selectedProject));
+    void navigate(isACMPage ? getACMTemplateListURL() : getTemplateListURL(selectedProject));
```
adamviktora: do we have some new eslint rules enforcing this? I find it less readable like this, so I'd remove the rule

src/utils/resources/template/utils/vmTemplateRequest.ts
```diff
@@ -0,0 +1,63 @@
...
+import { getCluster } from '@multicluster/helpers/selectors';
+import { kubevirtK8sCreate } from '@multicluster/k8sRequests';
+import { type K8sModel } from '@openshift-console/dynamic-plugin-sdk';
+
+import { TEMPLATE_CATEGORY_LABEL } from './constants';
+
+/** Pinned to v1beta1 so category can be set via spec.templateLabels. */
+export const VirtualMachineTemplateRequestBetaModel: K8sModel = {
+  ...VirtualMachineTemplateRequestModel,
+  apiVersion: 'v1beta1',
```
adamviktora: do we now have the beta K8sModel exported in the new version of `kubevirt-api` package?

src/utils/resources/template/utils/vmTemplateRequest.ts
```diff
@@ -0,0 +1,63 @@
...
+  templateNamespace: string;
+  vm: V1VirtualMachine;
+};
+
+export const createVMTemplateRequest = async ({
+  category,
+  templateName,
+  templateNamespace,
+  vm,
+}: CreateVMTemplateRequestOptions): Promise<V1alpha1VirtualMachineTemplateRequest> => {
```
adamviktora: same here, should we use the `V1beta1` types now?

src/utils/resources/template/utils/constants.ts
```diff
@@ -14,9 +14,20 @@ export const TEMPLATE_DATA_SOURCE_NAMESPACE_PARAMETER = 'DATA_SOURCE_NAMESPACE';
...
 
+export const DEFAULT_TEMPLATE_CATEGORIES = [
+  'Databases',
+  'Operating systems',
+  'Monitoring',
+  'Networking',
+  'Observability',
+  'Security',
+  'Storage',
+] as const;
```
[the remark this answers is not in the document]
adamviktora: We should address this. When choosing "Operating systems", the VMTR will end up with this fail:

```
	VirtualMachineTemplate.template.kubevirt.io "aquamarine-whippet-96-template-new" is invalid: metadata.labels: Invalid value: "Operating systems": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue', or 'my_value', or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')
```

Also this VMTR is marked as "In progress", it should be "Failed" - this is a pre-existing bug in `getVMTemplateRequestStatus`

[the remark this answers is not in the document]
adamviktora: @upalatucci Maybe we can process the text which user types, to replace " " with "-", and then reconstruct the visible label?

Or use annotation instead of label?

src/utils/components/CategorySelect/CategorySelect.tsx
```diff
@@ -0,0 +1,53 @@
...
+    [availableCategories, createdCategories, selectedCategory],
+  );
+
+  return (
+    <SelectTypeahead
+      addOption={(input) => {
+        setCreatedCategories((prev) => (prev.includes(input) ? prev : [...prev, input]));
+        return true;
+      }}
+      canCreate
```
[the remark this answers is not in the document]
adamviktora: we should address this, similar as https://github.com/kubevirt-ui/kubevirt-plugin/pull/4397/changes#r3621142889

playwright/tests/tier1/templates/template-creation-flows.spec.ts
```diff
@@ -32,6 +32,7 @@ test.describe.serial('Template creation flows', { tag: [T1_TAG, '@tier1-template
 
     const vmName = utils.generateRandomVmName('vm-save-tpl');
     const templateName = utils.generateRandomTemplateName('saved-tpl');
+    const category = 'Databases';
```
adamviktora: can we add test for the value with space like "Operating systems"?

playwright/src/components/vm/vm-detail-component.ts
```diff
@@ -544,7 +544,11 @@ export default class VmDetailComponent extends PageCommons {
     await this.performVmAction('restart');
   }
 
-  async saveAsTemplate(templateName: string, project: string): Promise<void> {
+  async saveAsTemplate(
```
adamviktora, line 532: Isn't this just a duplication of what is in [virtual-machine-detail-actions-component.ts](https://github.com/kubevirt-ui/kubevirt-plugin/pull/4397/changes#diff-c9f1646fef7ea7d0734735c9bd851ab39ee405f05a5939beee5792a7e3f606f9)?

upalatucci: bruno will clean this up in a furhter pr

src/utils/resources/template/utils/constants.ts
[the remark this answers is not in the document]
upalatucci: Moved category to annotation `vm.kubevirt.io/category` so values with spaces (like "Operating systems") are valid. VMTR only exposes `templateLabels`, so we create the VMTR then patch the annotation on the created VMT.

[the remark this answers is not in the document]
upalatucci: Went with the annotation approach you suggested — no need to slugify/display-map spaces.

src/utils/components/CategorySelect/CategorySelect.tsx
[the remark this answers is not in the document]
upalatucci: Addressed by storing category as an annotation instead of a label value.

playwright/tests/tier1/templates/template-creation-flows.spec.ts
upalatucci: Done — the save-as-template test now uses "Operating systems" and asserts the category annotation on the created VMT.

src/utils/components/SaveAsTemplateModal/useSaveAsTemplateModal.ts
upalatucci: Kept `void navigate(...)` — `@typescript-eslint/no-floating-promises` requires it (or `await`). Happy to drop the `void` if we relax that rule separately.

upalatucci: Updated — \`NavigateFunction\` is now in \`allowForKnownSafeCalls\` in \`eslintv10.config.js\`, so plain \`navigate(...)\` is allowed without \`void\`.

playwright/tests/tier1/templates/template-creation-flows.spec.ts
upalatucci: i'll stick with labels as VMTR do have a way to create labels in the VMT using `templateLabel` but do not for the annotations

src/utils/components/CategorySelect/utils/getCategorySelectOptions.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export const getCategorySelectOptions = (
+  categories: string[],
+  t: TFunction,
+  selectedCategory?: string,
+): SelectTypeaheadOptionProps[] => {
+  const uniqueCategories = new Set(categories);
+
+  if (selectedCategory) {
+    uniqueCategories.add(selectedCategory);
```
adamviktora, line 10: why do we need to add the `selectedCategory` to the Set?

src/utils/resources/template/utils/vmTemplateRequest.ts
```diff
@@ -0,0 +1,52 @@
...
+};
+
+export const createVMTemplateRequest = async ({
+  category,
+  templateName,
+  templateNamespace,
+  vm,
+}: CreateVMTemplateRequestOptions): Promise<V1beta1VirtualMachineTemplateRequest> => {
+  const cluster = getCluster(vm);
+  const categoryLabelValue = category ? toTemplateCategoryLabelValue(category) : undefined;
```
adamviktora: isn't `category` the already processed value? So the `toTemplateCategoryLabelValue` is not needed

src/utils/resources/template/utils/getTemplateCategoryLabel.ts
```diff
@@ -0,0 +1,20 @@
...
+  const knownLabel = TEMPLATE_CATEGORY_LABELS[category as keyof typeof TEMPLATE_CATEGORY_LABELS];
+  return knownLabel ? t(knownLabel) : category;
+};
+
+/**
+ * Normalize user input into a Kubernetes-safe label value (spaces → '-').
+ * @param category - raw category text from the typeahead
+ */
+export const toTemplateCategoryLabelValue = (category: string): string =>
+  category.trim().replace(/\s+/g, '-');
```
adamviktora, line 6: shouldn't here be also `.toLowerCase()` ? So e.g. 'operating-systems' key is matched correctly?

But then users can't create new labels starting with capital letter, so maybe change the `TEMPLATE_CATEGORY_LABELS` keys to start with capital?

adamviktora: Or maybe it is OK like this, because this helper is not used for the `TEMPLATE_CATEGORY_LABELS`

playwright/tests/tier1/templates/template-creation-flows.spec.ts
```diff
@@ -32,6 +32,9 @@ test.describe.serial('Template creation flows', { tag: [T1_TAG, '@tier1-template
 
     const vmName = utils.generateRandomVmName('vm-save-tpl');
     const templateName = utils.generateRandomTemplateName('saved-tpl');
+    // Display label with a space; persisted label value is label-safe.
+    const categoryDisplay = 'Operating systems';
```
adamviktora, line 36: maybe this test will fail due to `.toLowerCase` missing? and I think we pass the `'operating-systems'` value as a category in the actual saveAsTemplate helper code

src/utils/components/CategorySelect/utils/getCategorySelectOptions.ts
upalatucci: because if the user select something different we want this to become an option in the select

src/utils/resources/template/utils/getTemplateCategoryLabel.ts
upalatucci: we don't ned lowercase. labels can have uppercase. I just fixed it by having default categories with uppercase. and `Operating-system`

src/utils/components/CategorySelect/utils/getCategorySelectOptions.ts
adamviktora: I thought this is already done by calling  `setCreatedCategories` in the `addOption` prop in the SelectTypeahead.

Oh but if user reloads the page, we need the initally selected value there (which doesn't need to be in `DEFAULT_TEMPLATE_CATEGORIES`), so we need this

adamviktora (approved):
/lgtm

PR #4404, opened 2026-07-21
CNV-92254: Make HCO v1 availability check ACM-aware

upalatucci:
## Summary
- Replace `useK8sModel` in `useIsHyperConvergedV1Available` with a cluster-aware `useK8sListData` probe against pinned `HyperConvergedV1Model` (`hco.kubevirt.io/v1`)
- Accept optional cluster override (ACM cluster param fallback) and pass `cluster` from `usePreviewFeaturesData`
- Keeps Preview Features VMT toggle hidden only when HCO v1 is available on the target cluster

## Links
- https://issues.redhat.com/browse/CNV-92254
- Follow-up to #4379

## Test plan
- [ ] Single-cluster / HCO v1: Preview Features does **not** show “Enable native VirtualMachine templates”
- [ ] Single-cluster / HCO v1beta1 only: toggle is still shown
- [ ] ACM: open Preview Features for a spoke with HCO v1 → toggle hidden for that cluster
- [ ] ACM: spoke with HCO v1beta1 only → toggle still shown

## Summary by CodeRabbit

* **New Features**
  * Improved detection of HyperConverged v1 availability on a per-cluster basis.
  * VM Templates preview availability now reflects the selected cluster.
  * Clusters where access is restricted but the feature is present are handled as available.

* **Bug Fixes**
  * Improved handling of missing, forbidden, and unexpected cluster API responses.

* **Tests**
  * Added coverage for cluster selection, availability states, and Kubernetes error responses.

### DIFF
--- a/src/utils/hooks/useVMTemplateFeatureFlag/useIsHyperConvergedV1Available.ts
+++ b/src/utils/hooks/useVMTemplateFeatureFlag/useIsHyperConvergedV1Available.ts
@@ -1,19 +1,58 @@
+import {
+  isK8sForbiddenError,
+  isK8sNotFoundError,
+} from '@kubevirt-utils/resources/errorStatusChecks';
+import { operatorNamespaceSignal } from '@kubevirt-utils/store/operatorNamespace';
 import { isEmpty } from '@kubevirt-utils/utils/utils';
-import { useK8sModel } from '@openshift-console/dynamic-plugin-sdk';
+import useClusterParam from '@multicluster/hooks/useClusterParam';
+import useK8sWatchData from '@multicluster/hooks/useK8sWatchData';
+import { type K8sResourceCommon } from '@openshift-console/dynamic-plugin-sdk';
 
-import { HYPERCONVERGED_V1_GROUP_VERSION_KIND } from './constants';
+import { HyperConvergedV1GroupVersionKind } from './constants';
+
+type UseIsHyperConvergedV1Available = (clusterOverride?: string) => {
+  isHCOV1: boolean;
+  loading: boolean;
+};
 
 /**
  * HCO v1 exposes slice-based featureGates (Template is Beta / first-class).
  * HCO v1beta1 still needs the Preview Features jsonpatch toggle.
+ * Probes the target cluster (ACM-aware) via a shared watch; does not use useK8sModel.
+ * @param clusterOverride
  */
-const useIsHyperConvergedV1Available = (): { isHCOV1: boolean; loading: boolean } => {
-  const [hcoV1Model, inFlight] = useK8sModel(HYPERCONVERGED_V1_GROUP_VERSION_KIND);
+const useIsHyperConvergedV1Available: UseIsHyperConvergedV1Available = (clusterOverride) => {
+  const clusterParam = useClusterParam();
+  const cluster = clusterOverride ?? clusterParam;
+  const operatorNamespace = operatorNamespaceSignal.value;
 
-  return {
-    isHCOV1: !inFlight && !isEmpty(hcoV1Model),
-    loading: inFlight,
-  };
+  const [, loaded, loadError] = useK8sWatchData<K8sResourceCommon[]>(
+    operatorNamespace
+      ? {
+          cluster,
+          groupVersionKind: HyperConvergedV1GroupVersionKind,
+          isList: true,
+          namespace: operatorNamespace,
+        }
+      : null,
+  );
+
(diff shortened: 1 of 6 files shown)


src/utils/hooks/useVMTemplateFeatureFlag/useIsHyperConvergedV1Available.ts
```diff
@@ -1,19 +1,54 @@
...
+    operatorNamespace
+      ? {
+          cluster,
+          model: HyperConvergedV1Model,
+          namespace: operatorNamespace,
+        }
+      : null,
+  );
+  const loaded = listResult[1];
+  const loadError = listResult[2] as Error | undefined;
```
metalice: ?
why not deconstruct?

src/utils/hooks/useVMTemplateFeatureFlag/useIsHyperConvergedV1Available.ts
```diff
@@ -1,19 +1,54 @@
...
 
-import { HYPERCONVERGED_V1_GROUP_VERSION_KIND } from './constants';
+import { HyperConvergedV1Model } from './constants';
+
+type UseIsHyperConvergedV1Available = (clusterOverride?: string) => {
+  isHCOV1: boolean;
+  loading: boolean;
+};
+
+const isNotFoundError = (error: unknown): boolean => {
```
metalice: util?

src/utils/hooks/useVMTemplateFeatureFlag/useIsHyperConvergedV1Available.ts
upalatucci: as its was to be able to do the cast but i can change that

src/utils/hooks/useVMTemplateFeatureFlag/useIsHyperConvergedV1Available.ts
upalatucci: done

src/utils/hooks/useVMTemplateFeatureFlag/useIsHyperConvergedV1Available.ts
```diff
@@ -1,19 +1,56 @@
+import {
```
metalice, line 1: why new hook?> We don't already have a hook to fetch HCO? That we can expand instead of separate/logic?

src/multicluster/hooks/useK8sListData.ts
```diff
@@ -2,22 +2,33 @@ import { useEffect, useState } from 'react';
...
+import { type K8sModel, type K8sResourceCommon } from '@openshift-console/dynamic-plugin-sdk';
 
 type K8sListDataOptions = {
   cluster?: string;
   model: K8sModel;
   namespace?: string;
 };
 
+const isRecord = (value: unknown): value is Record<string, unknown> =>
+  typeof value === 'object' && value !== null;
```
adamviktora: we can extract this to some utils file. it is duplicated in [errorStatusChecks.ts](https://github.com/kubevirt-ui/kubevirt-plugin/pull/4404/changes#diff-27c0d30cc0eae1bd6bb6fd769931a0791f2de482fa18b2cadb6c74ed88aab1c2)

src/utils/hooks/useVMTemplateFeatureFlag/useIsHyperConvergedV1Available.ts
upalatucci: non-admin users will not go into the settings page and fetch HCO as they have no permissions to do so. But we need this info if we want to understand that a particular feature can be installed or not. 

This hook understands if v1 is installed even tho the user have no permissions to fetch HCO. 
Btw: if u are worried about the amount of requests fired, i can use usek8swathc that will make sure we do not fire multiple requests for the same resource

upalatucci: Now in `useHyperConvergeConfiguration` we fetch v1beta1 HCO for all clusters, and that's okay. k8s automatically convert v1 to v1beta1. At some point we'll have to fetch v1 HCO. But if we fetch v1 when v1 isn't installed in a managed cluster, we'll run into issues. So we need first to understand what we need to fetch and then fetch it. If we are right we'll not fetch two times the same thing. If we are not right and v1 is not installed, we first fetch v1 and then v1beta1

PR #4445, opened 2026-07-29
CNV-85806: Autopilot data-layer

batyana:
## 📝 Description

adds the autopilot data layer for the configuration column in the recommended capabilities table (future pr)
- static registry mapping 6 autopilot supported operators to their CRs and recommended YAML (autopilot doesn't currently expose)
- hook to watch managed CRs and get configuration status (Recommended/Manual)

## 🔗 Links

Jira ticket: https://redhat.atlassian.net/browse/CNV-85806

## Summary by CodeRabbit

* **New Features**
  * Added Autopilot status tracking to the Recommended Capabilities tab, including an enabled/loaded state and per-capability status.
  * Introduced recommended YAML templates for supported capabilities (descheduler, MetalLB, forklift, monitoring, logging, and collection).
  * Added automatic configuration status detection (recommended vs manual) based on Autopilot-related metadata.
* **Bug Fixes / Improvements**
  * Updated capability data loading so it becomes available only after Autopilot status is resolved, with load errors treated as non-blocking.

### DIFF
--- a/src/views/settings/tabs/RecommendedCapabilitiesTab/utils/autopilotRegistry.ts
+++ b/src/views/settings/tabs/RecommendedCapabilitiesTab/utils/autopilotRegistry.ts
@@ -1,0 +1,78 @@
+import { type K8sGroupVersionKind } from '@openshift-console/dynamic-plugin-sdk';
+
+import {
+  CLUSTER_OBSERVABILITY_OPERATOR_PACKAGE,
+  ClusterLogForwarderGVK,
+  DESCHEDULER_OPERATOR_PACKAGE,
+  ForkliftControllerGVK,
+  KubeDeschedulerGVK,
+  LokiStackGVK,
+  MetalLBGVK,
+  UIPluginGVK,
+} from './autopilotConstants';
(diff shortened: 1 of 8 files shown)


src/views/settings/tabs/RecommendedCapabilitiesTab/utils/autopilotRegistry.ts
```diff
@@ -0,0 +1,129 @@
...
+        version: v13
+  tenants:
+    mode: openshift-logging`,
+  },
+  {
+    crGVK: ClusterLogForwarderGVK,
+    crName: 'collector',
+    crNamespace: 'openshift-logging',
+    operatorPackageName: CLUSTER_LOGGING_OPERATOR_NAME,
+    recommendedYAML: `apiVersion: observability.openshift.io/v1
```
upalatucci: We can have recommended yaml into actual yaml files if they do not need to be generated based on variables. Otherwise we can have a separate ts file with generators

upalatucci (approved):

PR #4466, opened 2026-07-30
CNV-93106: disable create vm button when user has no access

mayarub:
## 📝 Description

Show the "Create VirtualMachine" button as disabled with a tooltip ("To create a VM, select a project") instead of hiding it when the user lacks create permissions

Reuse shared access review config and tooltip in the tree view right-click menu (Not mentioned in the bug — added to prevent inconsistency between the button and the right-click menu)

## 🔗 Links

Jira ticket: https://redhat.atlassian.net/browse/CNV-93106
## 🎥 Demo

* logged in as testuser

before:

https://github.com/user-attachments/assets/50440fde-cb79-41d5-9a06-76eb518ac200

after:

https://github.com/user-attachments/assets/337b56a2-4beb-43bc-a904-5eae3e71eccc

## Bug Fixes

* **Virtual Machine Creation**
  * Create actions remain visible when unavailable, with disabled controls and localized explanatory tooltips.
  * Tooltips now distinguish between all-namespaces mode and insufficient project permissions.
  * Right-click creation actions correctly reflect access permissions and disabled states.
  * Access reviews now use the appropriate namespace and cluster context for Virtual Machine creation.


src/views/virtualmachines/list/utils/utils.ts
```diff
@@ -32,3 +35,16 @@ export const getNamespacesWithVMsCount = (
 
   return new Set(vms.map((vm) => getNamespace(vm))).size;
 };
+
+export const getDisabledCreateVMTooltip = (t: TFunction) => t('To create a VM, select a project');
```
upalatucci: If a project is selected but the user have no permission in that project to create vms we should show a different text like 'U don't have permissions to create vms in this project'

And btw i'm not sure we should still use `VM` or `VirtualMachine` or `Virtual machine` . I received different feedback on this

metalice: VirtualMachine as im aware of

src/views/virtualmachines/list/utils/utils.ts
```diff
@@ -32,3 +35,25 @@ export const getNamespacesWithVMsCount = (
 
   return new Set(vms.map((vm) => getNamespace(vm))).size;
 };
+
+export const getDisabledCreateVMTooltip = (t: TFunction, isNoNamespaceSelected: boolean) => {
```
upalatucci: i would do `isNamespaceSelected` . Using negations adds complexity to the logic. Or if u don't want to do `!isEmty` in the caller, we can rename this as `isInAllNamespace`

upalatucci (approved):

PR #4484, opened 2026-07-31
CNV-92668:[release-4.22] Updating js-yaml to 4.3.0 as a fix for CVE-2026-59869

namansharma18899:
## Summary
- Updates `js-yaml` from `^4.1.1` to `4.3.0` to fix CVE-2026-59869
- **CVE-2026-59869**: js-yaml can spend quadratic CPU time parsing a document whose size grows only linearly when a chain of mappings uses merge keys where each mapping merges the previous one
- Fixed in js-yaml versions 3.15.0 and 4.3.0

## Jira
- [CNV-92668](https://redhat.atlassian.net/browse/CNV-92668)

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "immer": "^9.0.12",
     "ipaddr.js": "2.2.0",
     "js-abbreviation-number": "^1.4.0",
-    "js-yaml": "^4.1.1",
+    "js-yaml": "4.3.0",
     "lodash": "^4.18.1",
     "mochawesome-report-generator": "^6.2.0",
     "murmurhash-js": "1.0.0",
@@ -147,7 +147,8 @@
       "react-i18next": "~16.5.8"
     },
     "eslint": "^8.57.1",
-    "delaunator": "4.0.1"
+    "delaunator": "4.0.1",
+    "js-yaml": "4.3.0"
   },
   "private": true,
   "scripts": {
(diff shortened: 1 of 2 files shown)


upalatucci:
@namansharma18899 needs rebase

namansharma18899:
> @namansharma18899: This pull request references [CNV-92668](https://redhat.atlassian.net/browse/CNV-92668) which is a valid jira issue.
> 
> Warning: The referenced jira issue has an invalid target version for the target branch this PR targets: expected the vulnerability to target either version "4.22._" or "openshift-4.22._", but it targets "CNV v4.22.0" instead.
> 
> Details

@upalatucci  the prow wants the targget, fix version to be in format "4.22." or "openshift-4.22" but the kubevirt bot wants it to be in release-4.22. 

Can you clarify which target, fix version to choose ??

package.json
```diff
@@ -33,7 +33,7 @@
     "immer": "^9.0.12",
     "ipaddr.js": "2.2.0",
     "js-abbreviation-number": "^1.4.0",
-    "js-yaml": "^4.1.1",
+    "js-yaml": "4.3.0",
```
upalatucci: can we have 
```suggestion
    "js-yaml": "^4.3.0",
```

? so it will be easiers in the future with `npm audit fix`

upalatucci:
@namansharma18899 don't worry about all those things. we'll merge it anyway

package.json
namansharma18899: Definitely @upalatucci, I will also add this for all other versions for `npm audit` to pass

PR #4507, opened 2026-08-03
CNV-91016: store table sorting in URL

adamviktora:
## 📝 Description

Stores `sortBy` and `direction` params in URL (by passing `searchParams` and `setSearchParams` props to `useDataViewTableSort` -> `useDataViewSort`), so sorting is kept on page refresh.

## 🔗 Links

Jira: https://redhat.atlassian.net/browse/CNV-91016

## 🎥 Demo

After:

https://github.com/user-attachments/assets/6b808239-955a-4cb0-8517-4e17a365a87a

## Summary by CodeRabbit

* **New Features**
  * Table sorting can now be saved in the URL across supported resource lists.
  * Sort column and direction are preserved when refreshing, sharing, or navigating back to a list.
  * URL-based sort persistence is enabled across virtual machines, templates, volumes, migrations, preferences, quotas, networks, and related views.

### DIFF
--- a/src/views/storagemigrations/list/StorageMigrationList.tsx
+++ b/src/views/storagemigrations/list/StorageMigrationList.tsx
@@ -105,6 +105,7 @@ const StorageMigrationList: FC = () => {
           loadError={loadError}
           noDataMsg={t("You don't have any storage migrations yet")}
           pagination={pagination}
+          persistSortInUrl
           unfilteredData={unfilteredData}
         />
       </ListPageBody>
--- a/src/utils/components/KubevirtTable/KubevirtTable.tsx
+++ b/src/utils/components/KubevirtTable/KubevirtTable.tsx
@@ -1,4 +1,5 @@
 import React, { ReactElement, ReactNode, useEffect, useMemo } from 'react';
+import { useSearchParams } from 'react-router';
 
 import { PF_TABLE_CHECK_CLASS } from '@kubevirt-utils/hooks/useDataViewTableSort/constants';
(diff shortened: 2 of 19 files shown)


batyana (approved):

src/views/storagemigrations/list/StorageMigrationList.tsx
```diff
@@ -26,6 +27,7 @@ import '@kubevirt-utils/styles/list-managment-group.scss';
 
 const StorageMigrationList: FC = () => {
   const { t } = useKubevirtTranslation();
+  const [searchParams, setSearchParams] = useSearchParams();
```
upalatucci: why we are doing that and we are not calling the hook inside the kubevirt table and the `useKubevirtDataViewFilters` hook?

adamviktora: This is done in pages which don't use `useKubevirtDataViewFilters` yet, but still the old `useListPageFilter` from plugin-sdk. I will update it when migrating pages to the new filtering system.

upalatucci: okay can we call useSearchParams inside useDataViewTableSort? instead of getting searchParams and setSearchParams as prop?

upalatucci: Will we also fix that after all tables use DataView filters?

adamviktora: good idea, we don't need to pass it

pushed an update

upalatucci (approved):

PR #4525, opened 2026-08-04
CNV-94565: Require CNV v5.0.0 fix version for PRs targeting main

upalatucci:
## 📝 Description

Floor Jira fix-version validation on `main` at **5.0** (`CNV v5.0.0`).

With the highest release branch still at `release-4.22`, auto-detection was deriving `4.23` for main. After the major bump, PRs to main should require fix version `>= 5.0`. Once a `release-5.0` branch exists, the computed next version (`5.1+`) takes over again.

Also updates the coverage-analyst Jira version filter to `CNV v5.0.0` / `CNV v5.0.z`.

## 🔗 Links

- Jira: https://redhat.atlassian.net/browse/CNV-94565

## Test plan

- [x] Unit tests for `getExpectedVersionForBranch` (`npm test` in `.github/scripts`)
- [ ] Open a PR to `main` with fix version `CNV v4.23.0` → Jira validation fails (below floor)
- [ ] Open a PR to `main` with fix version `CNV v5.0.0` → fix-version check passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated enriched Jira reporting to use the current CNV 5.0 fix versions.
  * Ensured main-branch version calculations do not fall below version 5.0.

* **Tests**
  * Added coverage for release branches, minimum-version handling, 5.x version progression, empty release lists, and unrelated branches.

### DIFF
--- a/.github/scripts/src/version-compare.ts
+++ b/.github/scripts/src/version-compare.ts
@@ -4,6 +4,13 @@ import { extractVersionNumber, parseVersionToNumber } from './version-parse';
 const RELEASE_BRANCH_REGEX = /^release-(\d+\.\d+)$/;
 const RELEASE_SUMMARY_PREFIX_REGEX = /^\[release-\d+\.\d+\]\s*/i;
 
+/**
+ * Minimum fix version accepted on `main` (matches Jira "CNV v5.0.0" → "5.0").
+ * Used as a floor when release-branch auto-detection still points at a 4.x next
+ * minor (e.g. highest branch is release-4.22 → computed 4.23) after a major bump.
+ */
+export const MAIN_MIN_FIX_VERSION = '5.0';
+
 /** Extract "4.21" from "release-4.21". Returns null for non-release branches. */
 export const extractVersionFromBranch = (branchName: string): null | string => {
   const match = RELEASE_BRANCH_REGEX.exec(branchName);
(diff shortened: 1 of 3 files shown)


.github/scripts/src/version-compare.ts
```diff
@@ -4,6 +4,13 @@ import { extractVersionNumber, parseVersionToNumber } from './version-parse';
 const RELEASE_BRANCH_REGEX = /^release-(\d+\.\d+)$/;
 const RELEASE_SUMMARY_PREFIX_REGEX = /^\[release-\d+\.\d+\]\s*/i;
 
+/**
+ * Minimum fix version accepted on `main` (matches Jira "CNV v5.0.0" → "5.0").
+ * Used as a floor when release-branch auto-detection still points at a 4.x next
+ * minor (e.g. highest branch is release-4.22 → computed 4.23) after a major bump.
+ */
+export const MAIN_MIN_FIX_VERSION = '5.0';
```
metalice, line 12: i dont like when its hard coded, can we say something like, if the fix version is bigger then latest release-XXX branch, so it suppose to be in main? if the fix version already has release-XXX branch, fix version is wrong?

upalatucci: Yes this is already done. The only issue that we have is the suggested fix version in the bot comment . It will always be 4.23. If greater it will be fine for the action.

upalatucci: After we'll have the release-5.0 branch we can close this as the bot will suggest the 5.1 for main all the time

upalatucci: Screenshot 2026-08-05 at 10 10 34
the expected version is 4.23 as the latest release branch is 4.22. so it does 4.22 + 0.01 :D

upalatucci: if its not a big issue for u, we can close this pr

PR #4544, opened 2026-08-06
CNV-94189: Migrate from webpack to rspack for faster builds

vojtechszocs:
Minimal PR to replace [webpack](https://webpack.js.org/) with [Rspack](https://rspack.rs/) for faster builds.

Additional improvements like using `"jsx": "react-jsx"` (modern JSX transform) can be done as follow-ups.

Reference: https://github.com/openshift/console-plugin-template/pull/115

## Config file

- `webpack.config.ts` renamed to `rspack.config.mts`
- The `.mts` extension is used instead of `.ts` to explicitly mark the file as an ES module. Rspack CLI loads the config via native `import()` (unlike webpack CLI which relied on `require()` through ts-node). Without the `.mts` extension, Node.js emits a `[MODULE_TYPELESS_PACKAGE_JSON]` warning because the `package.json` does not have `"type": "module"`.
- Rspack auto-detects `rspack.config.mts` without needing `--config` in scripts.

### Removed

| Package | Reason |
|---|---|
| `webpack` | Replaced by `@rspack/core` |
| `webpack-cli` | Replaced by `@rspack/cli` |
| `webpack-dev-server` | Replaced by `@rspack/dev-server` |
| `esbuild` | No longer needed, rspack uses built-in SWC loader |
| `esbuild-loader` | Replaced by `builtin:swc-loader` |
| `copy-webpack-plugin` | Replaced by `CopyRspackPlugin` from `@rspack/core` |
| `fork-ts-checker-webpack-plugin` | Replaced by `ts-checker-rspack-plugin` |
| `tsconfig-paths-webpack-plugin` | Replaced by rspack's native `resolve.tsConfig` option |
| `mini-svg-data-uri` | No longer used after simplifying asset rules |

### Added

| Package | Purpose |
|---|---|
| `@rspack/core` | Core Rspack bundler |
| `@rspack/cli` | Rspack CLI |
| `@rspack/dev-server` | Rspack development server |
| `ts-checker-rspack-plugin` | TypeScript type checking (rspack-compatible fork of fork-ts-checker-webpack-plugin) |

### Retained

| Package | Reason |
|---|---|
| `style-loader` | CSS injection into JS at runtime (see CSS Strategy below) |
| `css-loader` | Upgraded from `^6.11.0` to `^7.1.4` for `@rspack/core@2.x` peer dependency compatibility |
| `resolve-url-loader` | Required by sass-loader for resolving relative URLs in SCSS |
| `sass-loader` | SCSS compilation |
| `sass` | Sass compiler used by sass-loader |

## CSS Strategy

The original webpack config used `style-loader` to inject CSS into generated JavaScript at runtime.

Rspack's native CSS handling (`type: 'css'` with `builtin:lightningcss-loader`) extracts CSS into static files at build time instead.

We chose to **retain the `style-loader` approach** (`type: 'javascript/auto'`) for this migration because:

- It preserves the existing runtime behavior, reducing migration risk.
- Switching to CSS extraction caused "Conflicting order" warnings due to SCSS files being imported in different orders across different code paths within the same chunk (primarily in the Settings page components). With `style-loader`, ordering is determined at runtime by JS execution, so these conflicts don't surface.
- CSS extraction can be adopted later as a separate improvement, along with fixing the import ordering issues.

## Assets

Simplified SVG/font/image handling into a single `asset/resource` rule. Used `[contenthash]` in dev mode filenames to avoid conflicts when multiple files share the same name in different directories (e.g. `fedora.svg`, `ubuntu.svg`, `windows.svg`).

## TypeScript tweaks

Removed `sourceMap: true` in `tsconfig.json`.

Source map generation is now fully handled by Rspack (`devtool`) and SWC (`sourceMaps: true` in `builtin:swc-loader`). Since `tsc` is no longer used for compilation but only for type-checking via `TsCheckerRspackPlugin` the `sourceMap` option in `tsconfig.json` was dead config.

## Rspack config tweaks

Don't use `minify` option in Rspack `builtin:swc-loader` i.e. turn off module-level minification.

- Loader-level `minify` operates on individual modules *before* bundling.
- `optimization.minimize: isProd` triggers `SwcJsMinimizerRspackPlugin` on the final bundles *after* tree-shaking and dead code elimination.
- Bundle-level minification is strictly better because it has cross-module visibility.
- Having both enabled results in redundant double-minification with no benefit.

Minification is now solely handled by `optimization.minimize: isProd`, which is the Rspack-recommended approach.

## Notes

- Use Rspack latest version which adds support for `ppc64` and `s390x` archs to avoid potential arch-related build issues

## Summary by CodeRabbit

* **Build & Development**
  * Migrated the application’s build and development tooling to Rspack.
  * Updated build, development, and start commands for improved compatibility and performance.
  * Refined handling of TypeScript, React, CSS, and static assets.

* **Chores**
  * Updated project configuration and formatting exclusions for local development settings.


rspack.config.mts
```diff
@@ -50,22 +37,30 @@ const config: Configuration = {
...
-          {
-            loader: 'esbuild-loader',
-            options: {
-              tsconfig: path.resolve(__dirname, 'tsconfig.json'),
+        use: {
+          loader: 'builtin:swc-loader',
+          options: {
+            detectSyntax: 'auto',
+            jsc: {
+              transform: {
```
logonoff, line 54: this wasn't needed?

```
              parser: {
                syntax: 'typescript',
                tsx: true,
              },
              transform: {
```

rspack.config.mts
```diff
@@ -50,22 +37,30 @@ const config: Configuration = {
...
+                react: {
+                  runtime: 'automatic',
```
logonoff: no react refresh?

rspack.config.mts
```diff
@@ -50,22 +37,30 @@ const config: Configuration = {
...
-            loader: 'esbuild-loader',
-            options: {
-              tsconfig: path.resolve(__dirname, 'tsconfig.json'),
+        use: {
+          loader: 'builtin:swc-loader',
+          options: {
+            detectSyntax: 'auto',
+            jsc: {
+              transform: {
+                react: {
```
logonoff, line 55: Also consider using react compiler in the future for performance (note it automatically adds `useMemo`/`memo` to things, this will need additional testing: https://rspack.rs/blog/announcing-2-1#react-compiler)

rspack.config.mts
logonoff, line 14: you can use the `defineConfig` helper in rspack. I don't think it matters but it is how they recommend you do it https://rspack.rs/config/

vojtechszocs:
Marking as WiP work-in-progress for now.

To avoid potential issues with multi arch builds, we should use upcoming Rspack version 2.1.9 which will include support for `ppc64` and `s390x` archs.

See Rspack PR https://github.com/web-infra-dev/rspack/pull/14962 for details.

rspack.config.mts
vojtechszocs: Seems like an explicit parser config is not needed.

> The config uses `detectSyntax: 'auto'` - that's the key. When set to `'auto'`, Rspack's builtin SWC loader automatically infers the parser config from the file extension:
> - `.ts` → `parser: { syntax: 'typescript', tsx: false }`
> - `.tsx` → `parser: { syntax: 'typescript', tsx: true }`
> - `.js`/`.jsx` → `parser: { syntax: 'ecmascript', jsx: true }`
>
> So yes, the explicit `parser: { syntax: 'typescript', tsx: true }` config is unnecessary here.
>
> `detectSyntax: 'auto'` is actually better because it correctly sets `tsx: false` for plain `.ts` files (slightly faster parsing).

rspack.config.mts
vojtechszocs: Thanks, good catch, I'll update it.

rspack.config.mts
vojtechszocs: Good point - we can use React Compiler in the future, however doing so requires more testing and a consistent code change to remove manual memoization boilerplate code across the codebase.

rspack.config.mts
vojtechszocs: I'd prefer use the `Configuration` type for now, we can always change it later.

> `defineConfig` exists primarily for JavaScript config files (`.js`/`.mjs`) where you don't have TypeScript types - it gives you IDE autocomplete via JSDoc inference without needing a `@type` annotation.
>
> Since your config is already a TypeScript file (`rspack.config.mts`) with an explicit `Configuration` type, `defineConfig` adds nothing. Both approaches give you the same type checking and autocomplete.
>
> It's "recommended" in the docs because it's the simplest path for the broadest audience (many users use `.js` configs). Not every recommendation applies to every setup.

src/views/lightspeed/utils/constants.ts
logonoff, line 1: This seems like a runtime change? Is it relevant to a PR focused on build tooling changes?

package.json
```diff
@@ -74,6 +74,10 @@
     "@openshift-console/dynamic-plugin-sdk-internal": "4.23.0-prerelease.4",
     "@openshift-console/dynamic-plugin-sdk-webpack": "4.23.0-prerelease.4",
     "@playwright/test": "^1.59.1",
+    "@rspack/cli": "^2.1.9",
+    "@rspack/core": "^2.1.9",
+    "@rspack/dev-server": "^2.2.0",
+    "@rspack/plugin-react-refresh": "^2.0.2",
```
logonoff: This has a peer dep on `react-refresh` - if y'all weren't using this before, there is a chance you don't need it now...

package.json
```diff
@@ -152,13 +148,13 @@
...
   "scripts": {
-    "build": "npm run clean && NODE_ENV=production NODE_OPTIONS=--max-old-space-size=8192 npm run ts-node -- node_modules/.bin/webpack",
+    "build": "npm run clean && NODE_ENV=production NODE_OPTIONS=--max-old-space-size=8192 npm run ts-node -- node_modules/.bin/rspack",
     "check-types": "tsc",
     "check-types:playwright": "tsc --noEmit --project playwright/tsconfig.json",
     "clean": "rm -rf dist",
     "playwright-install": "playwright install chromium",
     "playwright-report": "playwright show-report ./playwright/test-results/html-report",
-    "dev": "npm run clean && NODE_ENV=development NODE_OPTIONS=--max-old-space-size=8192 npm run ts-node -- node_modules/.bin/webpack serve --progress",
+    "dev": "npm run clean && REACT_REFRESH=true NODE_ENV=development NODE_OPTIONS=--max-old-space-size=8192 npm run ts-node -- node_modules/.bin/rspack serve",
```
logonoff: REACT_REFRESH wasn't used before..?

vojtechszocs: Nope, AFAIK we used `esbuild-loader` in our webpack config without React Fast Refresh.

package.json
vojtechszocs: I'll test this locally and see if there's a benefit to using React Fast Refresh (swc-loader option `jsc.transform.react.refresh`) when running the `dev` script.

src/views/lightspeed/utils/constants.ts
vojtechszocs: I guess I could take it out of this PR if we intend to use it as some Rspack migration reference.

rspack.config.mts
vojtechszocs: Update - using React Fast Refresh (component-level HMR) with `@rspack/plugin-react-refresh` and `builtin:swc-loader` option `jsc.transform.react.refresh` does not seem to have impact on dynamic plugins.

I'm running a local Console/Bridge and Kubevirt plugin - regardless of Kubevirt plugin Rspack `jsc.transform.react.refresh` option, recompile of Kubevirt React component always triggers Console page reload.

Maybe there's a way to make it work but I'll leave it out for now.

PR #4551, opened 2026-08-11
CNV-92611: [release-4.18] Non-admin users trigger excessive per-namespace watches

galkremer1:
## 📝 Description

Non-admin users were seeing a constant flood of 403 Forbidden errors (and slow, sometimes broken pages) because the plugin was opening live `Kubernetes` watches for `VMs`, `Templates`, `Nodes`, `NetworkAttachmentDefinitions`, and a few other resources before checking whether the user was actually allowed to list/watch them - being able to see a Project didn't mean you could list resources inside it, and "All Projects" watches silently fell back to cluster-wide requests that non-admins rarely have access to. 
This PR fixes the root cause everywhere it showed up: every affected hook now runs a `SelfSubjectAccessReview` check first (via `useAccessReview`/`useMultipleAccessReviews`) and only opens a watch for `namespaces`/`resources` the user is actually permitted to access, instead of firing first and discovering the 403 at runtime. 
It also fixes a few related issues found along the way - a VM Tree View that linked into forbidden namespaces, watches that fired on still-loading placeholder objects, and RBAC-write calls that could throw uncaught errors, plus adds unit tests covering the non-admin behavior for the hardened hooks.

## 🔗 Links

Jira ticket: [CNV-92611](https://redhat.atlassian.net/browse/CNV-92611)

## 🎥 Demo

Before:

https://github.com/user-attachments/assets/e136da53-1099-4858-97fd-a436965b2da5

After:

https://github.com/user-attachments/assets/c6cc8c4c-37ed-4aaa-8de0-8b253d099a0e

## Summary by CodeRabbit

* **New Features**
  * Resource listings and inventory views now respect access permissions, showing only namespaces and resources available to the current user.
  * Feature settings are managed according to administrator permissions, with sensible defaults when settings are unavailable.
  * Node, virtual machine, migration, and network resource views now use more consistent loading and access handling.

* **Bug Fixes**
  * Prevented unscoped resource requests while names or namespaces are still loading.
  * Improved handling of unavailable monitoring and Alertmanager configuration, avoiding startup errors.

### DIFF
--- a/src/utils/hooks/useFeatures/useFeatures.ts
+++ b/src/utils/hooks/useFeatures/useFeatures.ts
@@ -6,6 +6,7 @@ import {
   IoK8sApiRbacV1Role,
   IoK8sApiRbacV1RoleBinding,
 } from '@kubevirt-ui/kubevirt-api/kubernetes';
+import { kubevirtConsole } from '@kubevirt-utils/utils/utils';
 import { k8sCreate, k8sPatch } from '@openshift-console/dynamic-plugin-sdk';
 
 import { featuresConfigMapInitialState, featuresRole, featuresRoleBinding } from './constants';
@@ -14,6 +15,36 @@ import useFeaturesConfigMap from './useFeaturesConfigMap';
 
 type UseFeatures = (featureName: string) => UseFeaturesValues;
 
+const ensureFeaturesRBAC = async () => {
+  try {
+    await k8sCreate<IoK8sApiRbacV1Role>({
+      data: featuresRole,
+      model: RoleModel,
+    });
+  } catch (roleError) {
+    if (roleError?.code !== 409) {
+      kubevirtConsole.warn('Failed to create kubevirt-ui-features RBAC Role', roleError);
+    }
+  }
+
+  try {
+    await k8sCreate<IoK8sApiRbacV1RoleBinding>({
+      data: featuresRoleBinding,
+      model: RoleBindingModel,
+    });
+  } catch (roleBindingError) {
+    if (roleBindingError?.code !== 409) {
+      kubevirtConsole.warn(
+        'Failed to create kubevirt-ui-features RBAC RoleBinding',
+        roleBindingError,
+      );
+    }
+  }
+};
+
+// Shared so the bootstrap runs once per page load, not once per mounted useFeatures consumer.
+let featuresRBACBootstrap: null | Promise<void> = null;
+
 export const useFeatures: UseFeatures = (featureName) => {
   const { featuresConfigMapData, isAdmin } = useFeaturesConfigMap();
   const [featureConfigMap, loaded, loadError] = featuresConfigMapData;
@@ -21,37 +52,50 @@ export const useFeatures: UseFeatures = (featureName) => {
   const [loading, setLoading] = useState(true);
   const [error, setError] = useState<Error>(null);
 
+  // Ensure the RBAC independently of configmap existence -- HCO may pre-create the configmap
+  // itself, which would otherwise skip this bootstrap and leave non-admins 403ing forever.
   useEffect(() => {
+    if (!isAdmin) {
+      return;
+    }
+
+    if (!featuresRBACBootstrap) {
+      featuresRBACBootstrap = ensureFeaturesRBAC();
+    }
+  }, [isAdmin]);
+
+  useEffect(() => {
+    let ignore = false;
+
     if (loadError?.code === 404) {
       setError(loadError);
 
-      const createResources = async () => {
-        await k8sCreate<IoK8sApiCoreV1ConfigMap>({
-          data: featuresConfigMapInitialState,
-          model: ConfigMapModel,
-        });
+      // Only admins can bootstrap the shared configmap; non-admins simply fall back to defaults.
+      if (isAdmin) {
+        const createConfigMap = async () => {
+          try {
+            await k8sCreate<IoK8sApiCoreV1ConfigMap>({
+              data: featuresConfigMapInitialState,
+              model: ConfigMapModel,
+            });
+          } catch (createError) {
+            // Don't reinstate a stale error once a later run of this effect has moved on.
+            if (!ignore) {
+              setError(createError);
+            }
+          }
+        };
 
-        await k8sCreate<IoK8sApiRbacV1Role>({
-          data: featuresRole,
-          model: RoleModel,
-        });
-
-        await k8sCreate<IoK8sApiRbacV1RoleBinding>({
-          data: featuresRoleBinding,
-          model: RoleBindingModel,
-        });
-      };
-
-      try {
-        createResources();
-        setFeatureEnabled(featuresConfigMapInitialState.data[featureName] === 'true');
-        setLoading(false);
-        setError(null);
-      } catch (createError) {
-        setError(createError);
+        createConfigMap();
       }
 
-      return;
+      setFeatureEnabled(featuresConfigMapInitialState.data[featureName] === 'true');
+      setLoading(false);
+      setError(null);
+
+      return () => {
+        ignore = true;
+      };
     }
 
     if (!loaded && loadError) {
@@ -71,26 +115,30 @@ export const useFeatures: UseFeatures = (featureName) => {
         // In case of features config-map exists but there is a new feature to enter that is missing
         case undefined:
         case null: {
-          const applyMissingFeatures = async () => {
-            await k8sPatch({
-              data: [
-                {
-                  op: 'replace',
-                  path: `/data/${featureName}`,
-                  value: featuresConfigMapInitialState.data[featureName],
-                },
-              ],
-              model: ConfigMapModel,
-              resource: featureConfigMap,
-            });
-          };
+          // Only admins can patch the shared configmap; non-admins simply fall back to defaults.
+          if (isAdmin) {
+            const applyMissingFeatures = async () => {
+              try {
+                await k8sPatch({
+                  data: [
+                    {
+                      op: 'replace',
+                      path: `/data/${featureName}`,
+                      value: featuresConfigMapInitialState.data[featureName],
+                    },
+                  ],
+                  model: ConfigMapModel,
+                  resource: featureConfigMap,
+                });
+              } catch (updateError) {
+                setError(updateError);
+              }
+            };
 
-          try {
             applyMissingFeatures();
-            setFeatureEnabled(featuresConfigMapInitialState.data[featureName] === 'true');
-          } catch (updateError) {
-            setError(updateError);
           }
+
+          setFeatureEnabled(featuresConfigMapInitialState.data[featureName] === 'true');
           break;
         }
         default:
@@ -99,7 +147,8 @@ export const useFeatures: UseFeatures = (featureName) => {
       setLoading(false);
       return;
     }
-  }, [loadError, featureConfigMap, loaded, featureName, featureEnabled]);
+    // featureEnabled is excluded: it's written by this effect, so including it would re-trigger it.
+  }, [loadError, featureConfigMap, loaded, featureName, isAdmin]);
 
   const toggleFeature = useCallback(
     async (value: boolean) => {
--- a/src/utils/hooks/useHyperConvergeConfiguration.ts
+++ b/src/utils/hooks/useHyperConvergeConfiguration.ts
@@ -1,10 +1,16 @@
 import { useMemo } from 'react';
 
 import { HyperConvergedModelGroupVersionKind } from '@kubevirt-ui/kubevirt-api/console';
+import HyperConvergedModel from '@kubevirt-ui/kubevirt-api/console/models/HyperConvergedModel';
 import { V1LabelSelector } from '@kubevirt-ui/kubevirt-api/containerized-data-importer/models';
 import { V1MigrationConfiguration } from '@kubevirt-ui/kubevirt-api/kubevirt';
-import { isEmpty } from '@kubevirt-utils/utils/utils';
-import { K8sResourceCommon, useK8sWatchResource } from '@openshift-console/dynamic-plugin-sdk';
+import { DEFAULT_OPERATOR_NAMESPACE, isEmpty } from '@kubevirt-utils/utils/utils';
+import {
+  K8sResourceCommon,
+  K8sVerb,
+  useAccessReview,
+  useK8sWatchResource,
+} from '@openshift-console/dynamic-plugin-sdk';
 
 export type HyperConverged = K8sResourceCommon & {
   spec: {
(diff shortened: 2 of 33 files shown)


galkremer1:
@coderabbitai review

start-console.sh
```diff
@@ -49,8 +49,8 @@ BRIDGE_K8S_MODE="off-cluster"
 BRIDGE_K8S_AUTH="bearer-token"
 BRIDGE_K8S_MODE_OFF_CLUSTER_SKIP_VERIFY_TLS=true
 BRIDGE_K8S_MODE_OFF_CLUSTER_ENDPOINT=$(oc whoami --show-server)
```
galkremer1, line 51: Enables us to start the console with a non-privileged user

src/utils/hooks/useFeatures/useFeatures.ts
```diff
@@ -14,44 +14,86 @@ import useFeaturesConfigMap from './useFeaturesConfigMap';
 
 type UseFeatures = (featureName: string) => UseFeaturesValues;
 
+const ensureFeaturesRBAC = async () => {
```
galkremer1, line 18: Creates the Role/RoleBinding that grant non-admins read access to the shared `kubevirt-ui-features` ConfigMap. A 409 error (already exists) is treated as success, since another admin session  or a concurrent call to this same function may have already created it.

src/utils/hooks/useKubevirtUserSettings/useKubevirtUserSettings.ts
```diff
@@ -19,21 +15,60 @@ const useKubevirtUserSettings: UseKubevirtUserSettings = (key) => {
...
-    useK8sWatchResource<IoK8sApiCoreV1ConfigMap>(
-      userName && {
-        groupVersionKind: modelToGroupVersionKind(ConfigMapModel),
-        name: KUBEVIRT_USER_SETTINGS_CONFIG_MAP_NAME,
-        namespace: DEFAULT_OPERATOR_NAMESPACE,
-      },
-    );
+  // Regular users only have `get` (no `watch`/`list`) here, so fetch once instead of watching.
+  useEffect(() => {
+    let isMounted = true;
```
galkremer1, line 28: `k8sGet` has no cancellation of its own, so `isMounted` guards against setting state on an unmounted component if the component unmounts before the request resolves.

src/utils/hooks/useKubevirtUserSettings/useKubevirtUserSettings.ts
galkremer1, line 12: Fetches the current User and their settings ConfigMap with one-time `k8sGet` calls instead of
`useK8sWatchResource`: regular users are only granted `get` on these two resources, not
`watch`/`list`, so a watch would `403` repeatedly instead of just fetching once.

src/utils/hooks/useMultipleAccessReviews.ts
```diff
@@ -0,0 +1,42 @@
+import { useEffect, useState } from 'react';
+
+import { AccessReviewResourceAttributes, checkAccess } from '@openshift-console/dynamic-plugin-sdk';
+import { ImpersonateKind } from '@openshift-console/dynamic-plugin-sdk/lib/app/redux-types';
+
+type UseMultipleAccessReviews = (
+  multipleResourceAttributes: AccessReviewResourceAttributes[],
+  impersonate?: ImpersonateKind,
+) => [{ allowed: boolean; resourceAttributes: AccessReviewResourceAttributes }[], boolean];
+
```
galkremer1, line 11: Moved here from `src/views/cdi-upload-provider/hooks` - it's a shared, view-agnostic hook, so it
shouldn't live under a specific view while being imported by utils/ (e.g. useKubevirtWatchResource).
The old path now just re-exports this module for backwards compatibility.

src/utils/hooks/useNodes.ts
```diff
@@ -0,0 +1,38 @@
+import { modelToGroupVersionKind, NodeModel } from '@kubevirt-ui/kubevirt-api/console';
```
galkremer1, line 1: Shared hook for reading Nodes safely. Many scheduling-related components (node selector,
affinity, tolerations, dedicated resources, etc.) need the Node list, but most non-admin
users aren't allowed to watch Nodes. Without this check, each of those components would
independently open a Node watch that just 403s. This hook checks access once and reuses
the same safe pattern everywhere Nodes are needed.

src/utils/hooks/useFeatures/useFeatures.ts
```diff
@@ -14,44 +14,86 @@ import useFeaturesConfigMap from './useFeaturesConfigMap';
...
+  try {
+    await k8sCreate<IoK8sApiRbacV1Role>({
+      data: featuresRole,
+      model: RoleModel,
+    });
+  } catch (roleError) {
+    if (roleError?.code !== 409) {
+      // eslint-disable-next-line no-console
+      console.warn('Failed to create kubevirt-ui-features RBAC Role', roleError);
```
metalice: KubevirtConsole

src/utils/hooks/useFeatures/useFeatures.ts
metalice, line 18: why its nested in this hook?

Please make sure to carefully check those changes in this file

src/utils/hooks/useKubevirtUserSettings/useKubevirtUserSettings.ts
metalice: U sure? Non-admin doesn't have a watch?

src/utils/hooks/useKubevirtWatchResource.ts
```diff
@@ -1,15 +1,23 @@
 import { useEffect, useMemo, useState } from 'react';
 
 import {
+  AccessReviewResourceAttributes,
```
metalice, line 4: im concern regarding those changes. This hook is a core part, there are a lot of changes here.
Did everything pass local e2e?

src/views/cdi-upload-provider/hooks/useMultipleAccessReviews.ts
```diff
@@ -1,42 +1,2 @@
...
-        console.warn('SelfSubjectAccessReview failed', e);
-        setLoading(false);
-      });
-  }, [impersonate, multipleResourceAttributes]);
-
-  return [allowedResourceAttributes, loading];
-};
-
-export default useMultipleAccessReviews;
+// Re-exported for backwards compatibility -- this is a shared hook, now moved to the utils layer.
```
metalice: what backwards compatibility ?

src/views/virtualmachines/details/VirtualMachineNavPageTitle.tsx
```diff
@@ -35,12 +35,17 @@ const VirtualMachineNavPageTitle: FC<VirtualMachineNavPageTitleProps> = ({
...
-    groupVersionKind: VirtualMachineInstanceModelGroupVersionKind,
-    isList: false,
-    name: vm?.metadata?.name,
-    namespace: vm?.metadata?.namespace,
-  });
+  // `vm` starts as `{}` before the VM watch resolves; without this guard, undefined
+  // name/namespace would fall back to a cluster-wide VMI list.
+  const [vmi] = useK8sWatchResource<V1VirtualMachineInstance>(
+    vm?.metadata?.name &&
+      vm?.metadata?.namespace && {
```
adamviktora: nit: we could have variables like `const name = getName(vm)`

src/utils/resources/udn/hooks/useNamespaceUDN.ts
```diff
@@ -30,11 +35,21 @@ const useNamespaceUDN = (
     namespace,
   });
 
-  const [clusterUDNs] = useK8sWatchResource<ClusterUserDefinedNetworkKind[]>({
-    groupVersionKind: ClusterUserDefinedNetworkModelGroupVersionKind,
-    isList: true,
+  // ClusterUserDefinedNetwork is cluster-scoped; non-admins are frequently denied `watch` on it.
+  const [canWatchClusterUDNs, canWatchClusterUDNsLoading] = useAccessReview({
+    group: ClusterUserDefinedNetworkModelGroupVersionKind.group,
+    resource: 'clusteruserdefinednetworks',
```
adamviktora: can be `ClusterUserDefinedNetworkModel.plural`

src/utils/hooks/useKubevirtWatchResource.ts
adamviktora, line 1: would it make sense to reuse and upgrade the `useAccessibleResources` hook, which we have in newer CNV versions? so this `useKubevirtWatchResource` hook doesn't become too complicated. 

Anyways this is pre-ACM, so I guess this fix will look quite different in newer releases, so more of a reminder than something to do in 4.18 fix

src/utils/hooks/useHyperConvergeConfiguration.ts
```diff
@@ -49,18 +54,39 @@ type UseHyperConvergeConfigurationType = () => [
 ];
 
 const useHyperConvergeConfiguration: UseHyperConvergeConfigurationType = () => {
+  // Scoped to DEFAULT_OPERATOR_NAMESPACE (was cluster-wide) and gated on `watch`, which
+  // non-admins are rarely granted.
+  const [canWatchHyperConverged, canWatchHyperConvergedLoading] = useAccessReview({
+    group: HyperConvergedModelGroupVersionKind.group,
+    namespace: DEFAULT_OPERATOR_NAMESPACE,
+    resource: 'hyperconvergeds',
```
adamviktora: nit: `HyperConvergedModel.plural`

src/views/virtualmachines/tree/hooks/useTreeViewData.ts
```diff
@@ -27,17 +34,45 @@ export const useTreeViewData = (): UseTreeViewData => {
   const { featureEnabled: treeViewFoldersEnabled } = useFeatures(TREE_VIEW_FOLDERS);
   const [projectNames, projectNamesLoaded, projectNamesError] = useProjects();
 
-  const [allVMs, allVMsLoaded] = useK8sWatchResource<V1VirtualMachine[]>({
-    groupVersionKind: VirtualMachineModelGroupVersionKind,
-    isList: true,
-    limit: OBJECTS_FETCHING_LIMIT,
-  });
+  // Only admins can list VMs cluster-wide; non-admins rely on the per-namespace watches below.
+  const [allVMs, allVMsLoaded] = useK8sWatchResource<V1VirtualMachine[]>(
```
adamviktora, line 38: Can't we use the updated `useKubevirtWatchResource.ts` for this?

src/utils/hooks/useFeatures/useFeatures.ts
galkremer1: Thanks, I missed it.

src/utils/hooks/useKubevirtUserSettings/useKubevirtUserSettings.ts
galkremer1: It depends on the non admin user privileges. All admin users can watch resources, but not all non admin users have this option.

src/utils/hooks/useKubevirtWatchResource.ts
galkremer1: I will verify that.

src/utils/hooks/useKubevirtWatchResource.ts
galkremer1: Yes I am not sure if we want to introduce more changes.

src/views/cdi-upload-provider/hooks/useMultipleAccessReviews.ts
galkremer1: It was used to keep the same imports (I agree the "backwards compatibility" wording is confusing). I am changing it to import the new file directly from the shared folder rather than re-exporting it.

src/views/virtualmachines/tree/hooks/useTreeViewData.ts
galkremer1: We can't simple replace it without a further code refactoring. 

`useKubevirtWatchResource` only returns the flattened VM data ([data, loaded, error]) - it doesn't expose the `allowedNamespaces` list it computes internally.

`useTreeViewData` needs that namespace list itself (not just the VMs) to filter which tree nodes get rendered, so it can't get everything it needs from that hook without running a second, duplicate `useMultipleAccessReviews` call anyway.

galkremer1:
@metalice I can confirm that Cypress gating tests passed with the changes of this PR: 

# [CNV-92611](https://redhat.atlassian.net/browse/CNV-92611): E2E Gating Test Results

Full-suite Cypress gating regression run against a real `release-4.18` cluster, to verify this
RBAC-hardening fix introduces no regressions.

## Summary

Both suites were run **with** the fix (commit `f848eaf4a`) and **without** it (`release-4.18` branch before this change), against the same cluster, back-to-back,
to get a true before/after comparison rather than just a single-sided pass:

| Suite | With fix (`f848eaf4a`) | Without fix (`release-4.18` branch) | 403 / Forbidden (either) |
| --- | --- | --- | --- |
| `tier1-gating` | **47 / 51** passing, 0 failing, 4 pending | **47 / 51** passing, 0 failing, 4 pending | **0** |
| `tier2-gating` | **15 / 26** passing, 0 failing, 11 pending | **16 / 26** passing, 0 failing, 10 pending | **0** |

Zero test failures in every run, with or without the fix. All four run logs were grepped in full
for `403`/`Forbidden` -- zero matches in any of them.

**Why with/without look almost identical:** these gating suites always run as `kube:admin`, and
this fix is exclusively about *non-admin* RBAC gating -- admin behavior is explicitly unchanged
everywhere in the fix (see `CNV-92611-RBAC-HARDENING-NOTES.md`). The one-test difference in
`tier2-gating` (16 vs 15 passing) is "Test dynamic columns management," a test already documented
as flaky on UI timing in this suite, independent of the fix -- it happened to pass in the
without-fix run and land in the pending/auto-skip bucket in one of the with-fix runs. This is exactly
the expected outcome: the fix produces no behavioral change for admin users, so the admin-only
gating suites are unaffected either way.

**Reproducibility:** the with-fix runs were additionally repeated twice, back-to-back, against the
same cluster and commit. Both produced identical results -- same pass/pending counts, same exact
pending tests, 0 failures and 0 403s each time -- confirming the with-fix numbers above are a
stable baseline, not a one-off.

## How this was run

- Real OpenShift `release-4.18` cluster, with the fix branch built and served through the
  plugin's local webpack dev server (`npm run start`) via the console's plugin-proxy mechanism.
- Full `tier1-gating` and `tier2-gating` composite suites from the `kubevirt-ui-e2e` repo, run
  end-to-end (VM/template/instance-type creation, catalog filters, tabs, migration policies,
  bootable volumes, dynamic columns, NAD/OVN, SSH keys, etc.) -- not a partial or hand-picked
  subset.
- Environment issues unrelated to this fix (no default `StorageClass`, worker memory pressure
  from leftover test VMs) were found and fixed first, so the run reflects the suites' real
  baseline rather than infrastructure noise.

## Pending test breakdown (none are regressions from this fix)

**`tier1-gating` (4 pending):**

- 2 are permanently-disabled `xit()` tests in `overview.cy.ts`, unrelated to this change.
- 2 are conditional "close/ensure welcome modal is closed" steps that correctly no-op once the
  modal was already dismissed earlier in the same run.

**`tier2-gating` (11 pending):**

All auto-skipped by the suite's own infrastructure-flakiness harness (timeouts/detached-DOM/
element-covered errors are reclassified as skips, not failures), and traced to cluster-specific
gaps rather than this fix:

- OVN overlay network tests need a real OVN localnet attachment not present on this test
  cluster (2 downstream "VM with OVN overlay" tests skip in turn, since they depend on that NAD).
- Live-migration-dependent network-hotplug tests are flaky on this cluster's node layout.
- Migration-policy edit/delete and dynamic-column-management are known-flaky UI-timing tests in
  this suite, independent of this fix.

## Additional targeted validation

A dedicated non-admin RBAC smoke test (loading the all-namespaces VM list and the Virtualization
Tree View as a restricted, non-admin user) was also run and passed with **0 in-scope 403s**,
directly exercising the scenarios this fix addresses.

metalice (approved):

upalatucci (approved):

PR #4559, opened 2026-08-14
CNV-95299: Virtual Machines list: Select page selects the wrong VMs after sorting the table

galkremer1:
## 📝 Description

On the Virtual Machines list, after sorting the table by a column, using bulk select "Select page" selects different virtual machines than the ones currently shown on the page.

## 🔗 Links

Jira ticket: [CNV-95299](https://redhat.atlassian.net/browse/CNV-95299)

## 🎥 Demo

Before: 

https://github.com/user-attachments/assets/4763ca02-11d9-4dee-85bc-49ddbe3b4668

After:

https://github.com/user-attachments/assets/5dd775ed-2e4f-43d8-b74b-b4b969e259f4

## Summary by CodeRabbit

- **New Features**
  - Added URL-based sorting for the virtual machine list, preserving sort choices across navigation and refreshes.
  - Virtual machine rows now display in the selected ascending or descending order.

- **Bug Fixes**
  - Selection resets appropriately when filters, search criteria, or pagination context changes.
  - Pagination no longer triggers unnecessary updates when already on the first page.
  - Sorting behavior is now consistent across supported column types and custom sorting rules.

### DIFF
--- a/src/views/virtualmachines/list/VirtualMachinesList.tsx
+++ b/src/views/virtualmachines/list/VirtualMachinesList.tsx
@@ -13,6 +13,7 @@ import { runningTourSignal } from '@kubevirt-utils/components/GuidedTour/utils/g
 import KubevirtTable from '@kubevirt-utils/components/KubevirtTable/KubevirtTable';
 import { TableToolbarActionsFlex } from '@kubevirt-utils/components/TableToolbarActions/TableToolbarActionsFlex';
 import { PageTitles } from '@kubevirt-utils/constants/page-constants';
+import { sortDataViewTableData } from '@kubevirt-utils/hooks/useDataViewTableSort/useDataViewTableSort';
 import { KUBEVIRT_APISERVER_PROXY } from '@kubevirt-utils/hooks/useFeatures/constants';
 import { useFeatures } from '@kubevirt-utils/hooks/useFeatures/useFeatures';
 import { type KubevirtFilterState } from '@kubevirt-utils/hooks/useKubevirtDataViewFilters/types';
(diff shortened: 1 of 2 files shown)


galkremer1:
Pending confirmation of the changed behavior of keeping selected VMs when changing table sorting from @yfrimanm

src/views/virtualmachines/list/VirtualMachinesList.tsx
```diff
@@ -155,7 +171,11 @@ const VirtualMachinesList: FC<VirtualMachinesListProps> = ({
 
   useEffect(() => {
     deselectAllVMs();
-  }, [namespace, cluster, query]);
+  }, [namespace, cluster, selectionResetKey]);
+
+  useEffect(() => {
+    resetPagination();
+  }, [resetPagination, sortBy, sortDirection]);
```
adamviktora, line 178: Do we want to reset pagination on every sort change? It is something we don't do in other tables.

adamviktora (commented):
Looking good, just one question.

We could also later refactor / unify the code, so all tables have the same behavior.

src/views/virtualmachines/list/VirtualMachinesList.tsx
galkremer1: I think that we should do that. @yfrimanm  - WDYT?

yfrimanm: @galkremer1 @adamviktora 
I think we should not reset pagination on sort, b/c sort only reorders the same result set. Users expect to stay on the current page. 
TMO reset pagination should apply on filter and search changes, but not on column sort.

adamviktora (approved):

PR #4563, opened 2026-08-18
CNV-95611: Add create-std-doc Cursor skill for generating Playwright STD docs

galkremer1:
## 📝 Description

Add `create-std-doc` Cursor skill and custom command (`/create-std-doc`) for generating Playwright STD documentation

## 🔗 Links

Jira ticket: [CNV-95611](https://redhat.atlassian.net/browse/CNV-95611)

## 🎥 Demo

N/A

## Summary by CodeRabbit

* **New Features**
  * Added a `/create-std-doc` command to generate or update colocated Software Test Description documents.
  * Supports specification paths and feature names, requesting clarification when no unique specification is identified.
  * Documents test coverage, skipped or deferred tests, traceability, Jira references, release versions, and validation checks.
  * Preserves manually maintained content while applying standardized placement, formatting, and safe handling of sensitive information.

### DIFF
--- a/.cursor/skills/create-std-doc/SKILL.md
+++ b/.cursor/skills/create-std-doc/SKILL.md
@@ -1,0 +1,57 @@
+---
+name: create-std-doc
+description: >-
+  Create or update a Software Test Description (STD) markdown file for a
+  KubeVirt Playwright E2E spec, colocated next to the .spec.ts file and
+  following playwright/docs/STD-TEMPLATE.md.
+disable-model-invocation: true
+---
+
+# Create STD Doc
+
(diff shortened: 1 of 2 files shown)


.cursor/skills/create-std-doc/SKILL.md
adamviktora, line 1: I think the skill can be much shorter, so we don't waste tokens. I think that a few short sentences similar to what is in the description should be enough.

If we find out in practice it does mistakes, we can extend it later.

galkremer1: Yes - it kept getting bigger after I got more comments about it from CodeRabbit. Let me try to make it more concise.

adamviktora (approved):

PR #4579, opened 2026-08-24
CNV-90382: add sonarjs linters part 1

benbassatido:
## 📝 Description

Adds sonarjs rules to the code base, part - 1

## 🔗 Links

### DIFF
--- a/src/utils/components/DiskModal/components/DiskNameInput/DiskNameInput.tsx
+++ b/src/utils/components/DiskModal/components/DiskNameInput/DiskNameInput.tsx
@@ -34,7 +34,7 @@ const DiskNameInput: FC<{ isDisabled?: boolean }> = ({ isDisabled }) => {
     | undefined;
 
   const debouncedHandler = debounce((event, newName) => {
-    void registered.onChange(event);
+    registered.onChange(event).catch(() => {});
     setValue(DISK_NAME_FIELD, newName);
   }, 300);
 
--- a/src/utils/components/ExportModal/hooks/parseUploaderLog.ts
+++ b/src/utils/components/ExportModal/hooks/parseUploaderLog.ts
@@ -6,7 +6,7 @@ const UPLOADING_MSG = 'Uploading image';
 const extractErrorDetails = (raw: string | undefined): string | undefined => {
   if (!raw) return undefined;
 
-  const jsonMatch = raw.match(/\{[^}]*"details"\s*:\s*"([^"]+)"[^}]*\}/);
+  const jsonMatch = /\{[^}]*"details"\s*:\s*"([^"]+)"[^}]*\}/.exec(raw);
   if (jsonMatch?.[1]) return jsonMatch[1];
 
   return raw;
--- a/src/utils/components/PodSelectorModal/selectorRequirements.ts
+++ b/src/utils/components/PodSelectorModal/selectorRequirements.ts
@@ -10,7 +10,7 @@ import {
   SELECTOR_OPERATOR_NOT_EQUALS_REGEX,
   SELECTOR_OPERATOR_NOT_IN_REGEX,
 } from './constants';
-import { Requirement } from './types';
+import { type Requirement } from './types';
 
 export const requirementFromString = (inputString: string): Requirement | undefined => {
   const string = inputString.trim();
@@ -56,7 +56,7 @@ export const requirementFromString = (inputString: string): Requirement | undefi
     const values = valuesString
       .slice(1, -1)
       .split(',')
-      .map((v) => v.trim());
+      .map((val) => val.trim());
 
     return {
       key,
@@ -70,7 +70,7 @@ export const requirementFromString = (inputString: string): Requirement | undefi
     const values = valuesString
       .slice(1, -1)
       .split(',')
-      .map((v) => v.trim());
+      .map((val) => val.trim());
 
     return {
       key,
--- a/src/utils/components/SSHSecretModal/components/SSHOptionUseExisting/SSHOptionUseExisting.tsx
+++ b/src/utils/components/SSHSecretModal/components/SSHOptionUseExisting/SSHOptionUseExisting.tsx
@@ -1,7 +1,6 @@
-import React, { Dispatch, FC, SetStateAction, useCallback, useEffect, useState } from 'react';
+import React, { type ReactNode, useCallback, useEffect, useState } from 'react';
 
 import { modelToGroupVersionKind, ProjectModel } from '@kubevirt-ui-ext/kubevirt-api/console';
-import { IoK8sApiCoreV1Secret } from '@kubevirt-ui-ext/kubevirt-api/kubernetes';
 import InlineFilterSelect from '@kubevirt-utils/components/FilterSelect/InlineFilterSelect';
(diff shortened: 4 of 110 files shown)


src/utils/components/PodSelectorModal/selectorRequirements.ts
```diff
@@ -52,11 +52,13 @@ export const requirementFromString = (inputString: string): Requirement | undefi
   }
 
   if (SELECTOR_OPERATOR_IN_REGEX.test(string)) {
-    const [key, valuesString] = string.split(/\s+in\s+/);
+    const inIndex = string.search(/ in /i);
+    const key = string.substring(0, inIndex).trim();
+    const valuesString = string.substring(inIndex + 4).trim();
```
avivtur: magic number 4

src/utils/components/PodSelectorModal/selectorRequirements.ts
```diff
@@ -66,11 +68,13 @@ export const requirementFromString = (inputString: string): Requirement | undefi
   }
 
   if (SELECTOR_OPERATOR_NOT_IN_REGEX.test(string)) {
-    const [key, valuesString] = string.split(/\s+notin\s+/);
+    const notinIndex = string.search(/ notin /i);
+    const key = string.substring(0, notinIndex).trim();
+    const valuesString = string.substring(notinIndex + 8).trim();
```
avivtur: magic number 8

src/utils/components/SSHSecretModal/components/SSHOptionUseExisting/SSHOptionUseExisting.tsx
```diff
@@ -100,27 +86,29 @@ const SSHOptionUseExisting: FC<SSHOptionUseExistingProps> = ({
...
   return (
     <>
       <Alert
+        isInline
         title={t(
           'Select a secret from a different project to copy the secret to your current project.',
         )}
-        isInline
-        variant={AlertVariant.info}
+        variant="info"
```
avivtur, line 100: use AlertVariant

src/utils/components/PodSelectorModal/selectorRequirements.ts
benbassatido: Fixed

src/utils/components/PodSelectorModal/selectorRequirements.ts
benbassatido: Fixed

avivtur (approved):

src/utils/components/DiskModal/components/DiskNameInput/DiskNameInput.tsx
```diff
@@ -34,7 +34,7 @@ const DiskNameInput: FC<{ isDisabled?: boolean }> = ({ isDisabled }) => {
     | undefined;
 
   const debouncedHandler = debounce((event, newName) => {
-    void registered.onChange(event);
+    registered.onChange(event).catch(() => {});
```
upalatucci, line 37: we have to handle it in some way otherwise the rule does not make sense

src/utils/components/PodSelectorModal/selectorRequirements.ts
```diff
@@ -25,14 +25,14 @@ export const requirementFromString = (inputString: string): Requirement | undefi
 
   if (SELECTOR_OPERATOR_NEGATED_KEY_REGEX.test(string)) {
     return {
-      key: string.split(/!\s*/)[1],
+      key: string.split('!')[1].trim(),
```
upalatucci: why we are changing all those regex? Will they still work?

src/utils/components/ExportModal/hooks/parseUploaderLog.ts
```diff
@@ -4,10 +4,15 @@ const DOWNLOADING_MSG = 'Downloading disk image';
...
 
 const extractErrorDetails = (raw: string | undefined): string | undefined => {
-  if (!raw) return undefined;
+  if (!raw) {
+    return undefined;
+  }
 
-  const jsonMatch = raw.match(/\{[^}]*"details"\s*:\s*"([^"]+)"[^}]*\}/);
-  if (jsonMatch?.[1]) return jsonMatch[1];
+  const detailsRegex = /"details"\s*:\s*"([^"]+)"/;
```
upalatucci: also this regex. are we sure will still work?

src/utils/hooks/useKubevirtUserSettings/useKubevirtUserSettingsTopConsumerCards.ts
```diff
@@ -40,9 +40,7 @@ const useKubevirtUserSettingsTopConsumerCards: UseKubevirtUserSettingsTopConsume
     }
 
     if (!isEqualObject(topConsumerSettingsLocalStorage, cards?.[TOP_CONSUMERS_CARD])) {
-      void setCards?.({
-        [TOP_CONSUMERS_CARD]: topConsumerSettingsLocalStorage,
-      } as unknown as typeof cards);
+      setCards?.({ [TOP_CONSUMERS_CARD]: topConsumerSettingsLocalStorage }).catch(() => {});
```
upalatucci: empty catch also here

src/utils/store/operatorNamespace.ts
```diff
@@ -34,6 +34,8 @@ const resolveOperatorNamespace = async (): Promise<string> => {
...
 };
 
-void resolveOperatorNamespace().then((ns) => {
-  operatorNamespaceSignal.value = ns;
-});
+resolveOperatorNamespace()
+  .then((ns) => {
+    operatorNamespaceSignal.value = ns;
+  })
+  .catch(() => {});
```
upalatucci:
```suggestion
  .catch(() => {
    operatorNamespaceSignal.value = DEFAULT_OPERATOR_NAMESPACE;
  });
```

upalatucci: and at this point we should remove the try catch in the `resolveOperatorNamespace` function .

src/utils/components/PodSelectorModal/selectorRequirements.ts
benbassatido: It was flagged by the rule sonarjs/super-linear-regex. 
The fix should work the same as with the regex.

src/views/settings/tabs/UserTab/components/ManageSSHKeySection/SSHAuthKeysList/hooks/useSSHAuthProjects.ts
```diff
@@ -89,6 +89,7 @@ const useSSHAuthProjects: UseSSHAuthProjects = (authKeyRows) => {
 
         setSelectableProjects(projectsAllowedCreateSecret);
       })
+      .catch(() => {})
```
upalatucci: Promise.allSettled creates a promise that cannot throw an error . Only if the input is not iterable but i think typescript would warn u about it.

This function return all the promises with information on which of them failed and which passed 

i would turn off this rule for allSettled if possible or add an ignore rule comment

upalatucci (approved):

PR #4581, opened 2026-08-24
CNV-95837: wait for column settings save before closing management modal

upalatucci:
## Summary
- `setActiveColumns` now returns the user-settings update Promise instead of discarding it.
- Column management Save awaits that Promise so the modal is not unmounted before the ConfigMap patch is scheduled.

Jira ID is a placeholder (`TEMP-0`); replace it in the title and commit when you have the ticket.

## Test plan
- [ ] Open a table with column management (e.g. VirtualMachines list) and change visible columns, then Save
- [ ] Confirm a ConfigMap patch for user settings is sent and the table columns update after the modal closes
- [ ] Confirm Save still works if you close/reopen the modal immediately after saving
- [ ] Confirm a failed save keeps the modal open (or at least does not silently drop the request)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved column management so the modal closes only after column updates are successfully applied.
  * Added more reliable handling when saving table column preferences.
  * Improved validation of selected column identifiers to prevent invalid values.

### DIFF
--- a/src/utils/components/ColumnManagementModal/ColumnManagementModal.tsx
+++ b/src/utils/components/ColumnManagementModal/ColumnManagementModal.tsx
@@ -64,12 +64,12 @@ export const ColumnManagementModal: FC<ColumnManagementModalProps> = ({
     setCheckedColumns(updatedCheckedColumns);
   };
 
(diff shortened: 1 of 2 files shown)


galkremer1 (approved):
/lgtm

PR #4604, opened 2026-08-28
CNV-94189: stop spreading React keys onto remaining select options

upalatucci:
## 📝 Description

Follow-up to #4596 and #4603. After the rspack / automatic JSX transform, a React `key` inside a props object that is spread onto a child is not treated as a list key. React then tries to warn, and that warning path crashes Console with `getStackAddendum`.

This change:
- Removes `key` from network and environment select option objects
- Stops spreading `key` onto `SelectOption` in `InlineFilterSelectOption` and `CheckboxSelect`

`SelectTypeahead` was already fixed in #4603.

## 🔗 Links

- Jira: [CNV-94185](https://issues.redhat.com/browse/CNV-94185)
- Related: [CNV-96190](https://issues.redhat.com/browse/CNV-96190)
- Related PRs: https://github.com/kubevirt-ui/kubevirt-plugin/pull/4596, https://github.com/kubevirt-ui/kubevirt-plugin/pull/4603

## 🎥 Demo

WIP — still need to verify:
- [ ] Add network interface → Network dropdown
- [ ] VM environment editor resource dropdown
- [ ] Checkbox filter selects (no regression)

## Test plan

- [ ] Open Add network interface and expand the Network dropdown (dev mode, no overlay)
- [ ] Open a VM environment editor and expand the resource select
- [ ] Toggle checkbox filters on list pages

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved option handling across environment, network, and filter selectors for more reliable selection behavior.
  * Corrected option identifiers and values used by inline filters, improving consistency for accessibility and automated interactions.
  * Preserved disabled and favorite states while maintaining expected selection behavior.
  * Improved error reporting when environment options fail to load, making loading issues easier to handle.
  * Standardized option values and identifiers to prevent duplicate or conflicting selection metadata.

### DIFF
--- a/src/utils/components/FilterSelect/components/InlineFilterSelectOption.tsx
+++ b/src/utils/components/FilterSelect/components/InlineFilterSelectOption.tsx
@@ -1,9 +1,8 @@
-import React, { FC } from 'react';
+import React, { type FC } from 'react';
 
 import { SelectOption } from '@patternfly/react-core';
 
-import { EnhancedSelectOptionProps } from '../utils/types';
-
+import { type EnhancedSelectOptionProps } from '../utils/types';
 import InlineFilterSelectOptionContent from './InlineFilterSelectOptionContent';
 
 type InlineFilterSelectOptionProps = {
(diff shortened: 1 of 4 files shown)


src/utils/components/FilterSelect/components/InlineFilterSelectOption.tsx
```diff
@@ -1,23 +1,24 @@
...
+import { type EnhancedSelectOptionProps } from '../utils/types';
 import InlineFilterSelectOptionContent from './InlineFilterSelectOptionContent';
 
 type InlineFilterSelectOptionProps = {
   isFocused: boolean;
   option: EnhancedSelectOptionProps;
 };
 const InlineFilterSelectOption: FC<InlineFilterSelectOptionProps> = ({ isFocused, option }) => {
+  const optionValue = String(option.value);
+  const { key: _key, ...optionWithoutKey } = option;
```
galkremer1: I think we should stick to the `rest` convention, especially in a small component like this one. If it's a big one with many consts we might need to use a different name (like `optionWithoutKey`).

upalatucci: done

galkremer1 (approved):
/lgtm

PR #4415, opened 2026-07-23
 CNV-91273: [release-4.18]  sanitize-html: Stored Cross-Site Scripting via HTML sanitizer bypass

UdayYendva:
Fix raised to bump vulnerable sanitize-html package to patched version

Before:
npm ls --all sanitize-html
kubevirt-plugin@0.0.0 /home/uyendava/kubevirt-plugin
└─┬ @openshift-console/plugin-shared@0.0.3
└── sanitize-html@2.13.1

After:
npm ls --all sanitize-html
kubevirt-plugin@0.0.0 /home/uyendava/kubevirt-plugin
└─┬ @openshift-console/plugin-shared@0.0.3 overridden
└── sanitize-html@2.17.4 overridden

PR #4611, opened 2026-08-31
CNV-91273: [release-4.18] bump sanitize-html to 2.17.7 to fix stored XSS sanitizer bypass

upalatucci:
## 📝 Description

Bump transitive `sanitize-html` from 2.13.1 to 2.17.7 and add an npm override so the patched version stays pinned. This addresses CVE-2026-44990 (stored XSS via HTML sanitizer bypass in versions prior to 2.17.4).

Supersedes #4415, which rewrote a large portion of `package-lock.json` and hit build issues.

https://issues.redhat.com/browse/CNV-91273

manual tests have been done to ensure that all the pages are still working after the override


galkremer1 (approved):
/lgtm

PR #4621, opened 2026-09-01
[Release-5.0] CNV-96347: VirtualMachines page crashes when the guided tour is running

galkremer1:
## 📝 Description

While the guided tour is running, VirtualMachines fails with "Something wrong happened" instead of showing the VM list and tree.

## 🔗 Links

Jira ticket: [CNV-96347](https://redhat.atlassian.net/browse/CNV-96347)

## 🎥 Demo

Before:

https://github.com/user-attachments/assets/7f0a8732-3431-4a3d-8610-2a0e2763f9a4

After:

https://github.com/user-attachments/assets/86c70583-ca7e-4591-8fda-ad6f290baa22

### DIFF
--- a/src/views/virtualmachines/tree/utils/treeItemBuilders.tsx
+++ b/src/views/virtualmachines/tree/utils/treeItemBuilders.tsx
@@ -34,9 +34,7 @@ const createVMTreeItem = (
   const vmNamespace = getNamespace(vm);
   const vmName = getName(vm);
   const vmCluster = getCluster(vm);
-  const VMStatusIcon = vm?.status?.printableStatus
-    ? statusIcon[vm.status.printableStatus]
-    : undefined;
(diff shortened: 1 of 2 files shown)


src/views/virtualmachines/tree/utils/treeItemBuilders.tsx
```diff
@@ -34,9 +34,7 @@ const createVMTreeItem = (
   const vmNamespace = getNamespace(vm);
   const vmName = getName(vm);
   const vmCluster = getCluster(vm);
-  const VMStatusIcon = vm?.status?.printableStatus
-    ? statusIcon[vm.status.printableStatus]
-    : undefined;
+  const VMStatusIcon = statusIcon[vm?.status?.printableStatus ?? ''];
```
galkremer1, line 37: The problem here is that `undefined` crashed the view.

upalatucci (approved):

PR #4644, opened 2026-09-04
CNV-96404: Edit hostname modal- invalid hostname error doesn't propagated when clicking Enter on keyboard 

vojtechportes:
## 📝 Description

Fixes flaky validation within hostname modal by refactoring the modal to use react hook form and to have soft and strict validation in the modal UI. Strict validation is always applied but displayed only after form submission or input blur.

Added empty value validation message to getDNS1123LabelError to distinguish between empty and invalid value and therefore to communicate correct error message to user.

## 🔗 Links

Jira: https://redhat.atlassian.net/browse/CNV-96404

## 🎥 Demo

Before:

https://github.com/user-attachments/assets/b273a065-4c20-4374-a171-5d01cf2a3b52

After:

https://github.com/user-attachments/assets/8fcd988f-d4b1-49dc-8628-fe15b82bb8e9

## Bug Fixes

- Improved hostname validation in the hostname editor using DNS-compliant rules.
- Empty or whitespace-only hostnames now display a required-field error.
- Validation messaging is less intrusive while users enter a hostname and becomes stricter on submission.
- Hostname changes are applied only after successful form submission, with the modal closing afterward.
- Empty submissions now preserve the submitted hostname value rather than clearing it unexpectedly.

### DIFF
--- a/src/utils/components/HostnameModal/HostnameModal.tsx
+++ b/src/utils/components/HostnameModal/HostnameModal.tsx
@@ -1,4 +1,5 @@
-import React, { type FC, useMemo, useState } from 'react';
+import React, { type FC, useCallback } from 'react';
+import { Controller, useForm, useWatch } from 'react-hook-form';
 import produce from 'immer';
 
 import {
@@ -10,7 +11,11 @@ import TabModal from '@kubevirt-utils/components/TabModal/TabModal';
 import { useKubevirtTranslation } from '@kubevirt-utils/hooks/useKubevirtTranslation';
 import { getHostname } from '@kubevirt-utils/resources/vm';
 import { ensurePath } from '@kubevirt-utils/utils/utils';
-import { getNameValidationMessage } from '@kubevirt-utils/utils/validation';
+import {
+  getDNS1123LabelError,
+  getDNS1123LabelErrorLenient,
+  getFieldRequiredMessage,
+} from '@kubevirt-utils/utils/validation';
 import { FormGroup, TextInput, ValidatedOptions } from '@patternfly/react-core';
 
 import FormGroupHelperText from '../FormGroupHelperText/FormGroupHelperText';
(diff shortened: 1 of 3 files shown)


src/utils/components/HostnameModal/HostnameModal.tsx
```diff
@@ -23,44 +27,77 @@ type HostnameModalProps = {
...
 
-  const hostnameError = getNameValidationMessage(t, newHostname);
-  const isHostnameInvalid = isTouched && !!hostnameError;
+  const handleSubmit = useCallback(
+    () =>
+      submitForm(async ({ hostname }) => {
+        const updatedVM = produce<V1VirtualMachine>(vm, (vmDraft: V1VirtualMachine) => {
+          ensurePath(vmDraft, ['spec.template.spec']);
+
+          if (vmDraft.spec.template.spec) {
```
galkremer1: `ensurePath(vmDraft, ['spec.template.spec']);` is already making sure that `vmDraft.spec.template.spec` exists, therefor `if (vmDraft.spec.template.spec) {` is redundant.

src/utils/utils/validation.test.ts
```diff
@@ -63,6 +62,10 @@ describe('getDNS1123LabelError', () => {
 });
 
 describe('getDNS1123LabelErrorLenient', () => {
+  it('should reject empty value', () => {
+    expect(getDNS1123LabelErrorLenient('')(t)).toBe('This field is required');
```
galkremer1: You can use `getFieldRequiredMessage` to get this message.

src/utils/utils/validation.ts
```diff
@@ -22,6 +22,10 @@ const maxNameLengthErrorMsg = (t: TFunction): string =>
 export const isDNS1123Label = (value: string): boolean => !getDNS1123LabelError(value);
 
 export const getDNS1123LabelError = (value: string): ((t: TFunction) => string) | undefined => {
+  if (!value?.trim()) {
```
galkremer1: I know this might be out of the scope of the PR, but if we are already refactoring it - please check if we can simply move the required validation to the rules, instead of mixing it with `getDNS1123LabelError`:
```
<Controller
  control={control}
  name="hostname"
  rules={{ 
    required: getFieldRequiredMessage(t), 
    validate: (value) => getDNS1123LabelError(value)?.(t) 
  }}
  // ...
/>```

Comment (galkremer1) [reply src/utils/utils/validation.ts]:
```
export const getDNS1123LabelError = (value: string): ((t: TFunction) => string) | undefined => {
  // If empty, return undefined so it doesn't fail format validation. 
  // Let the RHF `required` rule handle empty states.
  if (!value?.trim()) {
    return undefined; 
  }
  
  if (value?.length > DNS1123LabelMaxLength) {
    return maxNameLengthErrorMsg;
  }
  // ... rest of the regex validations
```

**Review by galkremer1 — COMMENTED** [review]
Congrats on your first PR @vojtechportes! Please see my comments.

File: src/utils/components/HostnameModal/HostnameModal.tsx (hunk 1 above)
Comment (vojtechportes) [reply src/utils/components/HostnameModal/HostnameModal.tsx]: That is correct, though ensurePath is not asserting that and therefore from Typescript point of view, it remains potentially undefined and Typescript is complaining about that.

Something like this would fix it

```
type EnsuredPath<Path extends string> = string extends Path
  ? unknown
  : Path extends `${infer Head}.${infer Tail}`
    ? { [Key in Head]: EnsuredPath<Tail> }
    : { [Key in Path]: NonNullable<unknown> };

type EnsuredPaths<Paths extends string | readonly string[]> = Paths extends string
  ? EnsuredPath<Paths>
  : Paths extends readonly [infer First extends string, ...infer Rest extends readonly string[]]
    ? EnsuredPath<First> & EnsuredPaths<Rest>
    : unknown;

export function ensurePath<T extends object, const Paths extends string | readonly string[]>(
  data: T,
  paths: Paths,
): asserts data is T & EnsuredPaths<Paths> {
```

I removed the condition, but it would be good to change the signature of the ensurePath function.

src/utils/utils/validation.ts
vojtechportes: > I know this might be out of the scope of the PR, but if we are already refactoring it - please check if we can simply move the required validation to the rules, instead of mixing it with `getDNS1123LabelError`:
> 
> ```
> <Controller
>   control={control}
>   name="hostname"
>   rules={{ 
>     required: getFieldRequiredMessage(t), 
>     validate: (value) => getDNS1123LabelError(value)?.(t) 
>   }}
>   // ...
> />```
> ```

Good point, I moved the requiredness to the controller and reverted validation.ts and validation.spec.ts.

src/utils/components/HostnameModal/HostnameModal.tsx
galkremer1: I see. We can address it later then.

src/utils/utils/validation.ts
galkremer1: Great. Please squash / rebase all the changes to a single commit. I assume you tested the component with the new changes, right?

vojtechportes: I did test it after the latest changes. The changes are rebased and squashed into a single commit now.

galkremer1 (approved):
/lgtm

PR #4645, opened 2026-09-04
CNV-87425: skip template-from-VM gating test until CNV 4.22+

upalatucci:
## 📝 Description

Skip the "Create a template from a virtual machine" gating Playwright test until the hot cluster used in upstream CI is upgraded to CNV 4.22+. The test depends on `template.kubevirt.io` CRDs that are not yet available on the current cluster.

Also stabilizes the **Create a VM via YAML import** gating test by scoping the Create dropdown to `page-heading`, waiting for the button to become enabled after access review, and waiting for the **With YAML** menu item before clicking. This reduces flakiness when the empty-state Create button or a still-disabled header control was clicked instead.

## 🔗 Links

Jira: https://issues.redhat.com/browse/CNV-87425

## 🎥 Demo

N/A — test-only change (skips a failing gating test and hardens YAML create navigation).

Before: Gating test fails when native VM template CRDs are missing on the hot cluster; YAML import test could intermittently stay on the VM list.

After: Template-from-VM test is skipped with a comment documenting when to re-enable it; YAML import uses a more reliable Create → With YAML flow.

## Test plan

- [x] Verified diff is limited to skipping one gating test with an explanatory comment
- [x] VM list YAML create page object waits for enabled header Create button and visible With YAML menu item
- [ ] CI gating suite passes with the test skipped

## Summary by CodeRabbit

* **Improvements**
  * Improved VM creation workflows by waiting for controls to become visible and enabled before opening creation options and selecting YAML creation.

* **Tests**
  * Temporarily skipped the VM-to-template creation gating scenario while required platform support and template definitions are unavailable.
  * Existing feature-flag conditions and test workflow remain unchanged.
  * The scenario is expected to be re-enabled after upgrading to CNV 4.22 or later and installing the required template definitions.

### DIFF
--- a/playwright/src/components/vm/vm-list-template-create-component.ts
+++ b/playwright/src/components/vm/vm-list-template-create-component.ts
@@ -1,10 +1,10 @@
 import BaseComponent from '@/components/shared/base-component';
 import { TestTimeouts } from '@/utils/test-config';
-import type { Page } from '@playwright/test';
+import { expect, type Page } from '@playwright/test';
 
 /** VM list split Create control, template quick-create form, and catalog navigation from the list page. */
 export default class VmListTemplateCreateComponent extends BaseComponent {
-  private readonly _createButton = this.testId('item-create');
+  private readonly _pageHeaderCreateButton = this.testId('page-heading').getByTestId('item-create');
   private readonly _quickCreateVmButton = this.testId('quick-create-vm-btn');
   private readonly _roleMenuitem = this.locator('[role="menuitem"]');
   private readonly _startAfterCreateCheckbox = this.locator('#start-after-create-checkbox');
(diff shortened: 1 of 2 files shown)


upalatucci:
Pushed an additional commit to reduce flakiness in **Create a VM via YAML import**:

- Scope `item-create` under `page-heading` (avoids clicking the empty-state Create button, which has no YAML dropdown)
- Wait for the header Create control to be **enabled** after fleet access review
- Wait for the **With YAML** menu item to be visible before clicking

This targets the intermittent failure where the test stayed on the VM list with \`YAML editor heading not visible after 2 attempts\`.

galkremer1 (approved):
/lgtm

PR #4646, opened 2026-09-04
CNV-96686: Use cluster VM template API version in ACM

upalatucci:
## 📝 Description

In ACM multicluster environments, managed clusters on older CNV versions may only expose the `VirtualMachineTemplate` v1alpha1 API, which does not support template categories. This change detects HCO v1 availability per cluster and:

- Uses the correct `VirtualMachineTemplateRequest` API version (`v1beta1` on HCO v1 clusters, `v1alpha1` otherwise) when saving a VM as a template
- Hides the Category field in the Save as template modal when categories are not supported on the target cluster
- Relocates HCO API discovery hooks to `src/utils/resources/hyperconverged/hooks/` for reuse across features

## 🔗 Links

Jira: https://issues.redhat.com/browse/CNV-96686

## 🎥 Demo

Before:

- Category field was shown on managed clusters without category support, but could not be saved
- Template requests always used the v1beta1 API regardless of cluster capabilities. depending on HCO version installed we can have v1alpha1 or v1beta1 vm templates

test-failed-1

After:

- Category field is hidden while HCO version is loading, and only shown when HCO v1 is available
- Template requests use the API version matching the managed cluster's installed template CRD. If cluster has v1alpha1 we cannot use `templateLabels` and so the category field cannot be set on creation

FOR v1alpha1 vm template 

Screenshot 2026-09-04 at 18 34 52

For v1beta1 vm template (untouched)

Screenshot 2026-09-04 at 18 48 36

## Test plan

- [ ] On an ACM hub managing a CNV 4.21/4.22 spoke, open Save as template from a VM — Category field should be hidden
- [ ] On a hub/local cluster with HCO v1, Save as template should show Category and create a v1beta1 template request
- [ ] On an older managed cluster, saving as template should succeed using the v1alpha1 API
- [ ] Preview Features tab should hide the VM templates toggle on HCO v1 clusters
- [ ] Unit tests pass: `useHyperConvergedAPIDiscovery`, `useIsHyperConvergedV1Available`

## New Features

- Save-as-template now supports HyperConverged environments with the appropriate template API version.
- Category selection is shown only when categories are available.
- A loading skeleton appears while category availability is being determined.
- Template submission is disabled while category availability loads.
- HyperConverged version availability is detected automatically for the selected cluster.


src/utils/resources/template/utils/vmTemplateRequest.ts
```diff
@@ -10,22 +13,28 @@ import { TEMPLATE_CATEGORY_LABEL } from './constants';
...
+  isHCOV1,
   templateName,
   templateNamespace,
   vm,
 }: CreateVMTemplateRequestOptions): Promise<V1beta1VirtualMachineTemplateRequest> => {
   const cluster = getCluster(vm);
 
+  const templateModel = isHCOV1
+    ? VirtualMachineTemplateRequestModel
+    : VirtualMachineTemplateV1Alpha1Model;
```
adamviktora: this should be `VirtualMachineTemplateRequestV1Alpha1Model` and I'd rename the variable to `templateRequestModel`

mayarub (commented):
I reviewed it and it looks good to me. I can see @adamviktora left a comment above

adamviktora (approved):

PR #4648, opened 2026-09-06
CNV-96224: Compute resources step - next button validation

mayarub:
## 📝 Description

* Go to next step from compute resources step to customization step only if vm signal value exists
* Save description and folder in customization step to react hook form in order to not override their values.
* Added organize imports to settings json
## 🔗 Links

Jira: https://redhat.atlassian.net/browse/CNV-96224
## 🎥 Demo

Before:
https://github.com/user-attachments/assets/7058ae07-2a3a-41ed-9d89-68bb0fafa34e

After:
https://github.com/user-attachments/assets/07f598f7-d1df-4813-9791-182361adfd84

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved VM wizard progression so the Customization step remains unavailable until Compute resources are completed.
  * Ensured description and folder selections are retained when submitted during VM customization.
  * Improved step-transition handling to provide more reliable navigation.

* **Tests**
  * Added coverage for customization-step navigation and completed VM creation wizard flows.
  * Expanded automated checks for deployment details, guest OS, boot source, and compute resource selections.

### DIFF
--- a/playwright/src/utils/vm-wizard-test-helpers.ts
+++ b/playwright/src/utils/vm-wizard-test-helpers.ts
@@ -1,0 +1,82 @@
+import { expect } from '@playwright/test';
+
+import type {
+  VmWizardBootSourcePage,
+  VmWizardComputeCustomizationPage,
+  VmWizardNavigationPage,
+} from '@/page-objects';
+
+export async function completeDeploymentDetailsStep(
+  vmWizardNavigationPage: VmWizardNavigationPage,
+): Promise<void> {
+  const tilesVisible = await vmWizardNavigationPage.verifyCreationMethodTilesVisible();
+  expect.soft(tilesVisible, 'Creation method tiles should be visible').toBe(true);
+
+  const isCustomSelected = await vmWizardNavigationPage.verifyCreationMethodCardSelected('newVm');
+  expect.soft(isCustomSelected, 'Custom configuration should be selected by default').toBe(true);
+
+  await vmWizardNavigationPage.generateVmName();
+  await vmWizardNavigationPage.clickNext();
+}
+
+export async function completeGuestOsStep(
+  vmWizardNavigationPage: VmWizardNavigationPage,
+): Promise<void> {
+  const osTilesVisible = await vmWizardNavigationPage.verifyOsTilesVisible();
+  expect.soft(osTilesVisible, 'OS tiles (RHEL, Windows, Other Linux) should be visible').toBe(true);
+
+  const osDropdownVisible = await vmWizardNavigationPage.verifyOsTypeDropdownVisible();
+  expect.soft(osDropdownVisible, 'OS type dropdown should be visible').toBe(true);
+
+  const selectedOs = await vmWizardNavigationPage.getSelectedOsType();
+  expect.soft(selectedOs.length, 'An OS type should be pre-selected').toBeGreaterThan(0);
+
+  await vmWizardNavigationPage.clickNext();
+}
+
+export async function completeBootSourceStep(
+  vmWizardBootSourcePage: VmWizardBootSourcePage,
+  vmWizardNavigationPage: VmWizardNavigationPage,
+): Promise<void> {
+  const bootStepVisible = await vmWizardBootSourcePage.verifyBootSourceStepVisible();
+  expect.soft(bootStepVisible, 'Boot source step should be visible').toBe(true);
+
+  const tableVisible = await vmWizardBootSourcePage.verifyBootVolumeTableOrEmptyState();
+  expect.soft(tableVisible, 'Boot volume table or empty state should be visible').toBe(true);
+
+  const volumeCount = await vmWizardBootSourcePage.getBootVolumeCount();
+  if (volumeCount > 0) {
+    const columnsVisible = await vmWizardBootSourcePage.verifyBootVolumeTableColumnsVisible();
+    expect.soft(columnsVisible, 'Boot volume table columns should be visible').toBe(true);
+
+    await vmWizardBootSourcePage.selectBootVolumeByName('rhel');
+  } else {
+    await vmWizardBootSourcePage.selectNoBootSource();
+  }
+
+  await vmWizardNavigationPage.clickNext();
+}
+
+export async function verifyComputeResourcesStep(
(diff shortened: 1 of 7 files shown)


src/views/virtualmachines/wizard/steps/InstanceTypesSteps/ComputeResourcesStep/components/ComputeResourcesStepFooter.tsx
```diff
@@ -1,22 +1,41 @@
...
+    void goToNextStep();
+  }, [goToNextStep, vm]);
 
   const handleGoToNextStep = (): void => {
+    if (isEmpty(generatedVM)) {
+      return;
+    }
     setCustomizeWizardVMSignal(generatedVM);
-    void goToNextStep();
+    isAbleToGoToNextStepRef.current = true;
```
upalatucci, line 38: why we need th useEffect?

.vscode/settings.json
```diff
@@ -4,7 +4,8 @@
     "package-lock.json": true
   },
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": "always"
+    "source.fixAll.eslint": "always",
+    "source.organizeImports": "always"
```
adamviktora: just a note that this sometimes clashes on save with `simple-import-sort/imports` in [eslintv10.config.mjs](https://github.com/kubevirt-ui/kubevirt-plugin/blob/main/eslintv10.config.mjs#L106) - both sort imports differently, so I removed it. I think we should go with one or the other

playwright/tests/tier2/create-vm/create-vm-wizard-customization-step-navigation.spec.ts
galkremer1, line 1: Please create the STD file for this spec.

playwright/tests/tier1/create-vm/create-vm-wizard-custom-config.spec.ts
galkremer1, line 1: Please update the STD file for this spec.

src/views/virtualmachines/wizard/steps/InstanceTypesSteps/ComputeResourcesStep/components/ComputeResourcesStepFooter.tsx
```diff
@@ -1,22 +1,41 @@
...
   const closeWizard = useCloseWizard();
   const { isNextDisabledForStep } = useWizardStepValidation();
+  const isAbleToGoToNextStepRef = useRef(false);
+
+  useEffect(() => {
+    if (!isAbleToGoToNextStepRef.current || isEmpty(vm)) {
+      return;
+    }
+    isAbleToGoToNextStepRef.current = false;
+    void goToNextStep();
```
galkremer1, line 30: I think `isAbleToGoToNextStepRef.current = false;` is redundant here as the component unmounts after `goToNextStep` is called. 
Regardless, and more important, the use of `useEffect` as @upalatucci  mentioned below seems like a workaround. 
We should be able to fix this bug without `useEffect` and `useRef` by using
```
  const { activeStep, goToPrevStep, goToStepByIndex } = useWizardContext();

  const handleGoToNextStep = (): void => {
    if (isEmpty(generatedVM)) {
      return;
    }
    setCustomizeWizardVMSignal(generatedVM);
    goToStepByIndex(activeStep.index + 1);
  };

```

galkremer1 (changes requested):

playwright/src/utils/vm-wizard-test-helpers.ts
```diff
@@ -0,0 +1,82 @@
+import { expect } from '@playwright/test';
+
+import type {
+  VmWizardBootSourcePage,
+  VmWizardComputeCustomizationPage,
+  VmWizardNavigationPage,
+} from '@/page-objects';
+
+export async function completeDeploymentDetailsStep(
```
bmaio-redhat, line 9: we should keep expects / assertions on the actual spec file and avoid mixing validation with behavior so in this case, we could return the validation results and return a Promise<boolean> (method would return tilesVisible && isCustomSelected) and keep the page object methods to generate name and leave 

await vmWizardNavigationPage.generateVmName();
await vmWizardNavigationPage.clickNext();

to the spec file which already has access to these through fixture injection

playwright/src/utils/vm-wizard-test-helpers.ts
```diff
@@ -0,0 +1,82 @@
...
+  expect.soft(tilesVisible, 'Creation method tiles should be visible').toBe(true);
+
+  const isCustomSelected = await vmWizardNavigationPage.verifyCreationMethodCardSelected('newVm');
+  expect.soft(isCustomSelected, 'Custom configuration should be selected by default').toBe(true);
+
+  await vmWizardNavigationPage.generateVmName();
+  await vmWizardNavigationPage.clickNext();
+}
+
+export async function completeGuestOsStep(
```
bmaio-redhat, line 22: similar as above

playwright/src/utils/vm-wizard-test-helpers.ts
```diff
@@ -0,0 +1,82 @@
...
+  const osDropdownVisible = await vmWizardNavigationPage.verifyOsTypeDropdownVisible();
+  expect.soft(osDropdownVisible, 'OS type dropdown should be visible').toBe(true);
+
+  const selectedOs = await vmWizardNavigationPage.getSelectedOsType();
+  expect.soft(selectedOs.length, 'An OS type should be pre-selected').toBeGreaterThan(0);
+
+  await vmWizardNavigationPage.clickNext();
+}
+
+export async function completeBootSourceStep(
```
bmaio-redhat, line 37: similar as above

playwright/src/utils/vm-wizard-test-helpers.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    await vmWizardBootSourcePage.selectBootVolumeByName('rhel');
+  } else {
+    await vmWizardBootSourcePage.selectNoBootSource();
+  }
+
+  await vmWizardNavigationPage.clickNext();
+}
+
+export async function verifyComputeResourcesStep(
```
bmaio-redhat, line 60: same as above

PR #4650, opened 2026-09-06
CNV-96723: Throws an error when selected vm to clone no longer exists

mayarub:
## 📝 Description

Throws an error when vm to clone no longer exists
## 🔗 Links

Jira: https://redhat.atlassian.net/browse/CNV-96723

## 🎥 Demo

Before:
https://github.com/user-attachments/assets/945b3443-2b3f-4700-a1b2-a38b48398c52

After:
https://github.com/user-attachments/assets/0caaae5e-1315-4aba-90be-a964ad501f29

### DIFF
--- a/src/views/virtualmachines/wizard/hooks/useCloneVM.ts
+++ b/src/views/virtualmachines/wizard/hooks/useCloneVM.ts
@@ -14,6 +14,7 @@ import useLocalStorage from '@kubevirt-utils/hooks/useLocalStorage';
 import { getName, getNamespace } from '@kubevirt-utils/resources/shared';
 import { RUNSTRATEGY_HALTED } from '@kubevirt-utils/resources/vm';
 import { customizeWizardVMSignal } from '@kubevirt-utils/signals/customizeWizardVMSignal';
+import { isEmpty } from '@kubevirt-utils/utils/utils';
 import { getCluster } from '@multicluster/helpers/selectors';
 import { isACMPath } from '@multicluster/urls';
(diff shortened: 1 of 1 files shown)


src/views/virtualmachines/wizard/hooks/useCloneVM.ts
```diff
@@ -82,10 +83,17 @@ const useCloneVM: UseCloneVM = () => {
...
-      if (vmSameName) {
+      if (isEmpty(vmToClone)) {
+        throw new Error(t('{{name}} VirtualMachine no longer exists', { name: sourceName }));
+      }
+
+      const targetVMAlreadyExists = await vmExists(name, targetNamespace, targetCluster);
+
+      if (targetVMAlreadyExists) {
         throw new Error(t('VirtualMachine with this name already exists'));
       }
```
upalatucci, line 103: Can we fire those requests in parallel? liek
```suggestion
      const targetCluster = getCluster(source) ?? cluster;
      const sourceName = getName(source) ?? '';
      const vmToCloneExistPromise = vmExists(sourceName, getNamespace(source) ?? '', targetCluster);
      const targetVMAlreadyExistsPromise =  vmExists(name, targetNamespace, targetCluster);
      
      const [vmToCloneExist, targetVMAlreadyExists] = new Promise.all([vmToCloneExistPromise, ]targetVMAlreadyExistsPromise)

      if (isEmpty(vmToClone)) {
        throw new Error(t('{{name}} VirtualMachine no longer exists', { name: sourceName }));
      }

      if (targetVMAlreadyExists) {
throw new Error(t('VirtualMachine with this name already exists'));
}
```

upalatucci (approved):

PR #4654, opened 2026-09-07
CNV-96421: Add spacing between volume project and name fields

upalatucci:
## 📝 Description

Fixes missing spacing between the **Volume project** and **Volume name** fields in the Add volume form when source type is **Use existing → Volume**.

## 🔗 Links

Jira: https://issues.redhat.com/browse/CNV-96421

## 🎥 Demo

Before: Volume project and Volume name fields rendered with no vertical gap.

Screenshot 2026-09-07 at 16 07 18

After: Standard medium spacing between the two fields, matching other form groups in the modal.

Screenshot 2026-09-07 at 16 09 21

## Test plan

- [ ] Open Bootable volumes → Add volume → With form
- [ ] Select Source type: Use existing → Volume
- [ ] Verify spacing between Volume project and Volume name fields matches other fields in the form

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved bootable volume selection by excluding PVC entries without names.
  - Prevented errors when optional disk-size or namespace selection controls are unavailable.
  - Improved stability when selecting PVC-based disk sources.

- **UI Improvements**
  - Simplified the volume source selection layout without changing its visible functionality.

### DIFF
--- a/src/utils/components/AddBootableVolumeModal/components/VolumeSource/components/DiskSourcePVCSelectNamespace.tsx
+++ b/src/utils/components/AddBootableVolumeModal/components/VolumeSource/components/DiskSourcePVCSelectNamespace.tsx
@@ -1,4 +1,4 @@
-import React, { Dispatch, FC, SetStateAction } from 'react';
+import React, { type Dispatch, type FC, type SetStateAction } from 'react';
 
 import { modelToGroupVersionKind, ProjectModel } from '@kubevirt-ui-ext/kubevirt-api/console';
 import InlineFilterSelect from '@kubevirt-utils/components/FilterSelect/InlineFilterSelect';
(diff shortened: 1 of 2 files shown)


src/utils/components/AddBootableVolumeModal/components/VolumeSource/components/DiskSourcePVCSelectNamespace.tsx
```diff
@@ -28,7 +28,7 @@ const DiskSourcePVCSelectNamespace: FC<DiskSourcePVCSelectNamespaceProps> = ({
 
   return (
     <FormGroup
-      className="pvc-selection-formgroup"
+      className="pf-v6-u-mb-md"
```
adamviktora: I think a more proper fix would be to change the wrapping `div` in the `DiskSourcePVCSelect` to a fragment. So that the Form classes are applied (and then we don't need this margin).

Anyways it might be worth to do some layout changes in a followup, to better split the form groups visually

upalatucci: @adamviktora done

adamviktora (approved):

PR #4674, opened 2026-09-08
CNV-96509: disable cross-cluster migration for unsupported clusters and non-priv users

upalatucci:
## 📝 Description

Fixes cross-cluster migration staying enabled for non-privileged users on non-ACM clusters when opening the action from the VM tree project context menu.

The bulk/tree-view menu previously only disabled the action when `providerLoaded && isEmpty(provider)`. When the MTV provider watch fails or never loads (e.g. Forbidden), the action stayed enabled and opened a modal that errors.

This PR:
- Disables cross-cluster migration when provider data is unavailable, errored, or empty
- Adds MTV `Plan` create `accessReview` so non-privileged users get the standard permission tooltip
- Shows context-specific disabled tooltips (not running, unsupported cluster, no permission)

## 🔗 Links

Jira: https://issues.redhat.com/browse/CNV-96509

## 🎥 Demo

Before:
- Non-privileged user on non-ACM cluster: Cross-cluster migration enabled, modal errors on click
- Admin on same cluster: action disabled with "not supported" tooltip

https://github.com/user-attachments/assets/0839c6ec-99f2-4fe6-afa6-02e7175e1f55

After:
- Non-privileged user matches admin behavior: action disabled with appropriate tooltip

Screenshot 2026-09-08 at 18 17 01

## Test plan

- [ ] On a non-ACM cluster, log in as a non-privileged user without MTV provider list permissions
- [ ] Open VirtualMachines list tree view, right-click a project with running VMs
- [ ] Verify Migration → Cross-cluster migration is disabled with tooltip
- [ ] Repeat as admin and confirm same disabled state/tooltip
- [ ] On a supported ACM cluster with MTV provider, verify cross-cluster migration remains enabled for users with Plan create permission

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved cross-cluster migration availability handling while provider information is loading, unavailable, or fails to load.
  - Migration actions are now correctly disabled for unsupported clusters, insufficient permissions, and virtual machines that are not running.
  - Added clearer guidance explaining why migration cannot be started.
  - Improved provider-error handling so migration options remain consistent and predictable.
  - Added access checks for migration namespaces, plans, clusters, and plan creation.

### DIFF
--- a/src/views/virtualmachines/actions/bulkMigrationActionConfigs.tsx
+++ b/src/views/virtualmachines/actions/bulkMigrationActionConfigs.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { type TFunction } from 'i18next';
 
+import { PlanModel } from '@forklift-ui/types';
 import { VirtualMachineInstanceMigrationModel } from '@kubevirt-ui-ext/kubevirt-api/console';
 import { type V1VirtualMachine } from '@kubevirt-ui-ext/kubevirt-api/kubevirt';
 import { type ActionDropdownItemType } from '@kubevirt-utils/components/ActionsDropdown/constants';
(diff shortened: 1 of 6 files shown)


src/views/virtualmachines/actions/bulkMigrationActionConfigs.tsx
```diff
@@ -29,19 +31,34 @@ export const createCrossClusterMigrationConfig =
...
   ): ActionDropdownItemType => {
     const allRunning = vms?.every(isRunning);
 
+    const getDisabledTooltip = (): string | undefined => {
+      if (!allRunning) {
+        return t('All VirtualMachines must be running');
+      }
+      if (isClusterUnsupported) {
+        return t('Cross-cluster migration is not supported on this cluster.');
+      }
```
batyana, line 44: Is this a new string that should be added to the translation?

batyana (approved):

PR #4681, opened 2026-09-08
CNV-96788: hide Windows OS filter on s390x clusters

upalatucci:
## Summary
- Exclude Windows from the template operating system filter on clusters that do not support Windows workloads (e.g. s390x)
- Uses existing `useIsWindowsSupportedArchitecture()` hook, consistent with other Windows UI hiding on unsupported architectures

## Links

Jira: https://issues.redhat.com/browse/CNV-96788

## Test plan
- [ ] On an **s390x** cluster: open VM wizard → Create from template → Windows is not shown in the OS filter
- [ ] On **amd64/arm64** cluster: Windows still appears in the OS filter
- [ ] Templates list page OS filter behaves the same on both architectures

**BEFORE**

Screenshot 2026-09-08 at 23 55 23

Screenshot 2026-09-09 at 09 22 14

**AFTER**

Screenshot 2026-09-08 at 23 56 05

Screenshot 2026-09-09 at 09 23 02

## Summary by CodeRabbit

- **Bug Fixes**
  - Windows is now hidden from operating system filter options when it isn’t supported by the current system architecture.
  - This behavior is applied consistently to VM lists, VM templates, and clone-source selections.
  - Operating system filter options now refresh correctly when architecture support changes, helping prevent unsupported Windows options from appearing.

### DIFF
--- a/src/views/templates/list/filters/useOSFilter.ts
+++ b/src/views/templates/list/filters/useOSFilter.ts
@@ -1,6 +1,7 @@
 import { useMemo } from 'react';
 
-import { KubevirtFilter } from '@kubevirt-utils/hooks/useKubevirtDataViewFilters/types';
+import useIsWindowsSupportedArchitecture from '@kubevirt-utils/hooks/useIsWindowsSupportedArchitecture';
+import { type KubevirtFilter } from '@kubevirt-utils/hooks/useKubevirtDataViewFilters/types';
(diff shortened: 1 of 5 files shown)


adamviktora (commented):
@upalatucci We should probably also remove Windows option from the VirtualMachines OS filter `src/views/virtualmachines/list/filters/getOSFilter.ts`

upalatucci:
@adamviktora fixed

adamviktora (approved):

PR #4691, opened 2026-09-09
CNV-90382: Improve ensurePath typings with type assertion

vojtechportes:
## 📝 Description

Improves the `ensurePath` signature with a type assertion. After calling the function, TypeScript now recognizes the ensured paths as defined on the provided object.

Updates the runtime logic to explicitly distinguish between string and array inputs, avoiding incorrect handling of unsupported values. Throws a `TypeError` when `paths` is neither a string nor an array of strings.

## 🔗 Links

Jira: https://redhat.atlassian.net/browse/CNV-90382

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved nested configuration updates by creating missing paths and preserving existing values.
  * CPU and memory settings now apply correctly when configuration sections are initially missing.
  * Disk resize updates are skipped when no expansion size is provided.
  * Pause settings now correctly add or remove the paused state.
  * Template, workload, and bootloader settings update reliably, without undefined values.
  * Added safer handling for missing metadata during data import updates.

* **Tests**
  * Expanded validation coverage for nested path handling, including invalid input and nested paths.

### DIFF
--- a/src/views/virtualmachines/wizard/steps/TemplateStep/hooks/utils.ts
+++ b/src/views/virtualmachines/wizard/steps/TemplateStep/hooks/utils.ts
@@ -84,15 +84,25 @@ export const getVMObjectFromTemplate = ({
   vmName?: string;
 }) => {
   const generatedVM = produce(vm, (draftVM) => {
-    ensurePath(draftVM, 'metadata.labels');
-    ensurePath(draftVM, 'metadata.annotations');
+    ensurePath(draftVM, ['metadata.labels', 'metadata.annotations']);
 
     if (!isEmpty(description)) {
       draftVM.metadata.annotations.description = description;
     }
 
-    draftVM.metadata.labels[LABEL_USED_TEMPLATE_NAME] = selectedTemplate?.metadata?.name;
-    draftVM.metadata.labels[LABEL_USED_TEMPLATE_NAMESPACE] = selectedTemplate?.metadata?.namespace;
+    const selectedTemplateName = selectedTemplate?.metadata?.name;
+    if (selectedTemplateName) {
+      draftVM.metadata.labels[LABEL_USED_TEMPLATE_NAME] = selectedTemplateName;
+    } else {
+      delete draftVM.metadata.labels[LABEL_USED_TEMPLATE_NAME];
+    }
+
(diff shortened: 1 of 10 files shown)


src/views/virtualmachines/wizard/steps/TemplateStep/hooks/utils.ts
```diff
@@ -84,15 +84,25 @@ export const getVMObjectFromTemplate = ({
...
-    ensurePath(draftVM, 'metadata.annotations');
+    ensurePath(draftVM, ['metadata.labels', 'metadata.annotations']);
 
     if (!isEmpty(description)) {
       draftVM.metadata.annotations.description = description;
     }
 
-    draftVM.metadata.labels[LABEL_USED_TEMPLATE_NAME] = selectedTemplate?.metadata?.name;
-    draftVM.metadata.labels[LABEL_USED_TEMPLATE_NAMESPACE] = selectedTemplate?.metadata?.namespace;
+    const selectedTemplateName = selectedTemplate?.metadata?.name;
```
upalatucci: use getName

src/views/virtualmachines/wizard/steps/TemplateStep/hooks/utils.ts
```diff
@@ -84,15 +84,25 @@ export const getVMObjectFromTemplate = ({
...
+    if (selectedTemplateName) {
+      draftVM.metadata.labels[LABEL_USED_TEMPLATE_NAME] = selectedTemplateName;
+    } else {
+      delete draftVM.metadata.labels[LABEL_USED_TEMPLATE_NAME];
+    }
+
+    const selectedTemplateNamespace = selectedTemplate?.metadata?.namespace;
```
upalatucci: use getNamespace

src/views/virtualmachines/wizard/steps/TemplateStep/hooks/utils.ts
vojtechportes: Updated, but it would be then good to fix typing of most of the helpers so they accept also null, otherwise, we are still getting an error.

adamviktora (approved):

PR #4700, opened 2026-09-09
CNV-96639: prevent restricted-access flash on VM utilization charts

upalatucci:
## 📝 Description

Fixes a flash of the "Restricted access" error on VM overview utilization charts for non-privileged users.

Prometheus queries were firing before the VMI was available, producing invalid queries that returned 403 responses. This change:

- Skips Prometheus queries when VMI metadata (`name`/`namespace`) is missing
- Shows a section-level loading state while the VMI is being fetched
- Surfaces VMI fetch errors at the utilization section level

## 🔗 Links

Jira: https://issues.redhat.com/browse/CNV-96639

## 🎥 Demo

Before: utilization charts briefly flash "Restricted access" before rendering correctly for non-priv users.

prometheus call (instead of prometheus-tenancy for namespaced calls)

https://github.com/user-attachments/assets/565ed619-6b21-41a3-be28-171d0e9765df

After: utilization section shows a spinner while VMI loads, then renders charts without the error flash.

no prometheus call but only prometheus-tenancy

https://github.com/user-attachments/assets/ad437363-366d-46ff-844a-1a31d2325395

## Test plan

- [ ] As a non-privileged user, open a running VM overview and confirm utilization charts do not flash "Restricted access"
- [ ] Confirm a spinner appears briefly while VMI loads, then charts render with data
- [ ] Confirm stopped VMs still show "VirtualMachine is not running"
- [ ] Confirm utilization charts still work on the VM metrics tab

## Bug Fixes
* Improved utilization views to remain in a loading state until required VM data and monitoring metrics are available.
* Prevented charts from rendering with incomplete VM information.
* Improved handling of missing VM, namespace, or query data.
* Improved display of loading and error states while VM instance details are being retrieved.

## Tests
* Added coverage for utilization queries when VM data is absent or incomplete.


src/utils/components/Charts/CPUUtil/CPUThresholdChart.tsx
```diff
@@ -58,7 +58,7 @@ const CPUThresholdChart: FC<CPUThresholdChartProps> = ({ vmi }) => {
     query,
   });
 
-  const isLoading = !loaded;
+  const isLoading = !vmi || !loaded;
```
galkremer1, line 61: If vmi is not provided - wouldn't it be stuck on a loaded state?

src/utils/components/Charts/CPUUtil/CPUThresholdChart.tsx
```diff
@@ -35,7 +35,7 @@ import {
 } from '../utils/utils';
 
 type CPUThresholdChartProps = {
-  vmi: V1VirtualMachineInstance;
+  vmi?: V1VirtualMachineInstance;
 };
 
 const CPUThresholdChart: FC<CPUThresholdChartProps> = ({ vmi }) => {
```
galkremer1, line 41: `vmi` is optional for `CPUThresholdChart` however it's required for `useVMQuery`

galkremer1 (commented):
I didn't see the original [Jira](https://redhat.atlassian.net/browse/CNV-96639) issue in the before video. Did you test it with a non priv user or there is a different issue?

upalatucci:
@galkremer1 i was not able to create the right user with missing permissions that we need for the bug. but as u can see fromthe video, when the browser loads the page initially, we have an initiall Prometheus all-namespaces call as vmi is not loaded yet. And then we do the correct call after vmi is loaded. 
We have to remove that initial call in order to remove the flash.

galkremer1 (approved):
/lgtm

PR #4711, opened 2026-09-10
CNV-96424: Diagnostics show warning for unsupported ephemeral disks

vojtechportes:
## 📝 Description

Removes ephemeral volume disk warnings from VM detail's indicators on tab and diagnostics overview cards by excluding containerDisk volumes from list of counted items as warnings that snapshots are not being available for this type of storage is expected state and not a diagnostic issue.

## 🔗 Links

Jira: https://redhat.atlassian.net/browse/CNV-96424

## 🎥 Demo

Before:

[CNV-96424](https://redhat.atlassian.net/browse/CNV-96424)-before

After:

[CNV-96424](https://redhat.atlassian.net/browse/CNV-96424)-after

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved virtual machine diagnostic reporting for volume and snapshot statuses.
  - Container-disk and cloud-init volumes are now correctly shown as healthy when appropriate.
  - Diagnostic messages are displayed more consistently, including reasons with or without additional details.
  - Status conditions and volume information are now reflected more accurately in the diagnostics view.


src/views/virtualmachines/details/tabs/diagnostic/hooks/useDiagnosticData.ts
```diff
@@ -42,21 +53,23 @@ const buildDataVolumeResources = (vm: V1VirtualMachine): Record<string, unknown>
...
     const hasColon = index !== -1;
 
-    const reason = hasColon ? vss.reason.slice(0, index) : vss?.name;
-    const message = hasColon ? vss.reason.slice(index + 1) : vss?.reason;
+    const reason = hasColon && vss.reason ? vss.reason.slice(0, index) : vss?.name;
+    const message = hasColon && vss.reason ? vss.reason.slice(index + 1) : vss?.reason;
 
     return {
       ...vss,
+      excludeFromWarningCount: vss?.enabled === false && isContainerDiskVolume(vss?.name, volumes),
```
adamviktora, line 67: is it guaranteed that a name of `V1VolumeSnapshotStatus` matches the corresponding name of `V1Volume` ?

src/views/virtualmachines/details/tabs/diagnostic/utils/utils.ts
```diff
@@ -51,6 +52,9 @@ export const getDVSeverity = (phase: string): DiagnosticSeverity => {
 export const getSnapshotSeverity = (enabled: boolean): DiagnosticSeverity =>
   enabled ? 'healthy' : 'warning';
 
+export const isContainerDiskVolume = (volumeName: string, volumes: V1Volume[] = []): boolean =>
```
adamviktora, line 55: the Jira ticket talks about `ephemeral` disks - so I think we should exclude these rather than `containerDisk`  volumes

vojtechportes: From what I can see in the volumes object, ephemeral disks are mapped to containerDisk, but I will double check it next week.

src/views/virtualmachines/details/tabs/diagnostic/hooks/useDiagnosticCounts.test.ts
upalatucci, line 1: no need for this test

src/views/virtualmachines/details/tabs/diagnostic/utils/utils.ts
upalatucci: for ephemeral we mean containerDisk . containerDisks are ephemeral per se

src/views/virtualmachines/details/tabs/diagnostic/hooks/useDiagnosticData.ts
upalatucci, line 76: with this line we exclude the cloudinitdisks from the warning at the top. We can do the same when `isContainerDiskVolume` is `true`  without adding the new prop `excludeFromWarningCount`

upalatucci: it would be great to have a general way to exclude things like them. Even tho probably we would exlude only those two cases: cloudinit and containerDisk

src/views/virtualmachines/details/tabs/diagnostic/hooks/useDiagnosticData.ts
```diff
@@ -42,21 +53,23 @@ const buildDataVolumeResources = (vm: V1VirtualMachine): Record<string, unknown>
...
       id: uuidv4(),
       message,
       metadata: {
         condition: 'Other',
-        name: hasColon ? vss.reason.slice(0, index) : (vss?.reason ?? vss?.name),
+        name: hasColon && vss.reason ? vss.reason.slice(0, index) : (vss?.reason ?? vss?.name),
```
galkremer1: `hasColon && vss.reason ? vss.reason.slice(0, index)` logic is duplicated - we can extract it to a util function.

upalatucci (approved):

PR #4728, opened 2026-09-15
CNV-96685: Cannot enable "SSH over NodePort service" before filling node address

vojtechportes:
## 📝 Description

Fixes unclear NodePort service dependency on node address by aligning the text input and switch next to each other.

Adds e2e tests covering this dependency.

## 🔗 Links

Jira: https://redhat.atlassian.net/browse/CNV-96685

## 🎥 Demo

Before:

https://github.com/user-attachments/assets/e1b822eb-4abe-463b-b54b-bafe2a9cbd17

After:

https://github.com/user-attachments/assets/1edbef02-368b-48ef-8304-6782c1107884

## Summary by CodeRabbit

- **New Features**
  - Added SSH over NodePort service configuration to cluster settings.
  - Administrators can configure a node address and enable or disable SSH over NodePort services.
  - The enable switch remains unavailable until a node address is provided and automatically turns off when the address is cleared.
  - Added SSH over LoadBalancer service settings with controls reflecting cluster capabilities and permissions.
  - Added SSH service entries to settings navigation and search.
- **Tests**
  - Added coverage for SSH configuration behavior.


src/views/settings/tabs/ClusterTab/components/GeneralSettings/SSHConfiguration/components/SSHConfirgurationProvider.tsx
```diff
@@ -0,0 +1,80 @@
...
+  onChange?: (val: string, field: string) => void;
+  setLoadBalancerIsLoading?: Dispatch<SetStateAction<boolean>>;
+  setNodePortIsLoading?: Dispatch<SetStateAction<boolean>>;
+};
+
+export type SSHConfigurationContextValue = Required<SSHConfigurationContextType>;
+
+export const SSHConfigurationContext = createContext<SSHConfigurationContextType>({});
+
+export const SSHConfigurationProvider: FC<PropsWithChildren> = ({ children }) => {
```
upalatucci, line 31: hei @vojtechportes  can we switch this provider into a hook ? 

We can have a useChangeFeaturesConfigMap hook that expose a function with a field .
This will have a couple of benefits in my view:

- no context, more simple to manage
- more general behavior. it means that other components could use the same hook to change other settings in the future. 

we can even have this function in the `useFeaturesConfigMap` hook but i have the impression it would become more complex. Not sure about that

vojtechportes: @upalatucci I refactored it into useUpdateSSHConfiguration hook that is arbitrary to the SSH Configuration section to avoid inventing some abstract logic and to stay within the scope of this PR.

playwright/tests/settings/ssh-configuration.spec.md
```diff
@@ -0,0 +1,82 @@
...
+**Coverage Type values:**
+
+- `Feature coverage` — test validates a feature delivered by the ticket
+- `Bugfix regression guard` — test asserts the specific bug fixed by the ticket does not regress
+- `Functional smoke` — test validates baseline behavior; no specific Jira ticket drives it
+
+## 6. Approvals
+
+- **Prepared By:** Test automation / QE
+- **Reviewed By:** Pending
```
galkremer1: Enter your name here instead of `Pending`

src/views/settings/tabs/ClusterTab/components/GeneralSettings/SSHConfiguration/components/SSHOverNodePortService/SSHOverNodePortService.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const { updateSSHConfiguration } = useUpdateSSHConfiguration(NODE_PORT_ADDRESS);
+
+  const {
+    featuresConfigMapData: [featureConfigMap, loaded],
+    isAdmin,
+  } = useFeaturesConfigMap(cluster);
+
+  const onTextChange = useMemo(
+    () =>
+      debounce((value: string) => {
```
galkremer1, line 33: You can use `useDebounceCallback`

src/views/settings/tabs/ClusterTab/components/GeneralSettings/SSHConfiguration/components/SSHOverNodePortService/SSHOverNodePortService.tsx
```diff
@@ -0,0 +1,78 @@
...
+          name="node-address"
+          onChange={(_event, value: string) => {
+            setUrl(value);
+            onTextChange(value);
+          }}
+          placeholder={t('Enter node address')}
+          value={url ?? featureConfigMap?.data?.[NODE_PORT_ADDRESS]}
+        />
+
+        <NodePortSwitch />
```
galkremer1, line 72: There is still no tooltip that explains why the switch is disabled and cannot be enabled. 

I suggest that the switch should always be enabled:

- If the input is empty and the user toggles on the switch, it will focus on the input and allow the user to type in the value. If the user does not type anything the switch will be toggled off again when the input loses focus (So there will be no state where the input field is empty and the switch is toggled on). 

- If the user enters their input we should not automatically toggle on the switch (like the current UX) as there is an option to toggle off the switch and still have value for `SSH over NodePort service` (The current fix changes this UX and automatically turns on the switch once a value is provided). 

@upalatucci  / @yfrimanm  WDYT?

src/views/settings/tabs/ClusterTab/components/GeneralSettings/SSHConfiguration/components/SSHOverNodePortService/SSHOverNodePortService.tsx
vojtechportes: useDebounceCallback doesn't have the option to unsubscribe, that is why I used debounce from lodash instead

src/views/settings/tabs/ClusterTab/components/GeneralSettings/SSHConfiguration/components/SSHOverNodePortService/SSHOverNodePortService.tsx
vojtechportes: Regarding the tooltip, what about including the information below the text input
Screenshot 2026-09-16 at 10 24 17

upalatucci: for me the information belowthe text input is better than the tooltip

upalatucci: i'll approve the pr. If we get some feedback from yifat, we can change it later

upalatucci (approved):

PR #4729, opened 2026-09-15
CNV-96482: Bulk VM deletion should show cluster information

vojtechportes:
## 📝 Description

Adds information which cluster VM being deleted belongs to, to a bulk deletion modal

## 🔗 Links

Jira: https://redhat.atlassian.net/browse/CNV-96482 

## 🎥 Demo

Before:

Screenshot 2026-09-15 at 13 23 47 before

After:

Screenshot 2026-09-15 at 13 23 47

## Summary by CodeRabbit

- **Enhancements**
  - VM deletion lists now display each virtual machine’s cluster name alongside its VM name.
  - The cluster label remains blank when cluster information is unavailable.


upalatucci:
@vojtechportes please talk to @yfrimanm for this. I think we should add cluster into the paragraph wheere we say: "2 VirtualMachines in the project **default**"

yfrimanm:
Following @upalatucci comment, I think that in a multi cluster view, we should add the cluster info, before the project. So something like "2 VirtualMachines in cluster [**cluster name**] project [**project name**]".
If more than one cluster is involved, follow the pattern that we have for more than one project, as described in the UX doc.
https://docs.google.com/document/d/1eh0Cz5V0135pMpADYg1yKN1JhJ0uI56CzLFYniKhnrg/edit?tab=t.0

src/views/virtualmachines/actions/components/DeleteAllConfirmationModal/components/DeleteAllVmsList.tsx
```diff
@@ -34,7 +35,7 @@ const DeleteAllVMsList: FC<DeleteAllVMsListProps> = ({
   const vmsList = visibleVMs.map((vm) => (
     <StackItem key={`${getNamespace(vm)}/${getName(vm)}`}>
       {hasMultipleNamespaces ? `${getNamespace(vm)}/ ` : ''}
-      {getName(vm)}
+      {getName(vm)} {`(${t('Cluster: {{clusterName}}', { clusterName: getCluster(vm) ?? '-' })})`}
```
galkremer1, line 38: We need to tackle a few scenarios: 
1. Same ns, same cluster:

Image

2. Different namespaces, same cluster:

Image

3. Different namespaces, different clusters:

Image

@upalatucci  / @yfrimanm  are you good with that?

galkremer1: Also we should only show the cluster name of course in ACM mode.

yfrimanm: @galkremer1 @upalatucci  thinking over, WDYT about adding the icons we use in the tree for `Cluster`, `Project`, `VM` before the names, so they can replace the `/` ?

galkremer1: Yes that's a good idea - I don't like these `/`. Is it possible for you to provide a mockup for @vojtechportes to implement?

yfrimanm: @galkremer1 @vojtechportes 
Delete all Bulk collapse Affected VM-icons-1B
The icons in this mockup are just place holders. Please use the correct icons as shown in the tree view for: Cluster, Project, VM.

galkremer1: Great, thank you. I assume we should only show the cluster names in case multiple VMs are deleted across different clusters, correct @yfrimanm ?

PR #4736, opened 2026-09-15
CNV-95613: Fix stale Last Restored time for recreated snapshots

upalatucci:
## Summary
- Fix VM Snapshot list showing stale "Last Restored" time when a snapshot is recreated with the same name
- Match `VirtualMachineRestore` objects to snapshot instances by snapshot UID and ensure restore time is after snapshot creation
- Scope restores to the current VM so same-named snapshots on other VMs are not affected

Fixes [CNV-95613](https://issues.redhat.com/browse/CNV-95613)

## Root cause
`buildRestoresMap` keyed restores by snapshot name only, so old restore objects from deleted snapshots (or restores on other VMs) were incorrectly associated with new snapshots sharing the same name.

## Test plan
- [x] Unit tests for stale restore, cross-VM restore, and latest valid restore scenarios
- [ ] Create VM, snapshot `Snapshot1`, restore, delete snapshot, recreate `Snapshot1` — Last Restored should be empty
- [ ] Create second VM from same template, snapshot `Snapshot1` — Last Restored should not show restore time from first VM

**BEFORE**

https://github.com/user-attachments/assets/6e49b27a-3ba3-4e85-8c9c-c545ede17d94

**AFTER**

https://github.com/user-attachments/assets/e8945107-9300-417a-a969-784d45f320e3

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved snapshot restore status accuracy by matching restores to the correct virtual machine and snapshot.
  - Prevented stale, cross-VM, and pre-creation restores from being associated with snapshots.
  - Ensured the most recent valid restore is displayed for each snapshot.
- **Tests**
  - Added coverage for stale restores, cross-VM restores, and selecting the latest valid restore.


adamviktora (approved):

src/views/virtualmachines/details/tabs/snapshots/components/list/snapshotListDefinition.tsx
```diff
@@ -55,10 +55,8 @@ const StatusCell: FC<SnapshotCellProps> = ({ row }) => (
 );
 
 const LastRestoredCell: FC<SnapshotCellProps> = ({ callbacks, row }) => {
-  const snapshotName = getName(row);
-  const relevantRestore = snapshotName ? callbacks?.restores?.[snapshotName] : undefined;
-  const restoreTime = relevantRestore?.status?.restoreTime;
-  return <Timestamp timestamp={restoreTime} />;
+  const relevantRestore = callbacks?.restores?.[getUID(row) ?? ''];
```
vojtechportes, line 58: I think rather than introducing artificial key, it would be better to make timestamp prop on Timestamp component optional and rather go with something like this which is more explicit.
```
const uid = getUID(row);
const relevantRestore =
  uid != null ? callbacks?.restores?.[uid] : undefined;

return <Timestamp timestamp={relevantRestore?.status?.restoreTime} />;
```

The Timestamp component is already counting with the timestamp prop to be undefined, even though the component typings do not reflect it.

upalatucci: oh sorry i thought ci would remove lgtm labels after approvers review. strange

adamviktora: @upalatucci this seems to be because we have two OWNERS files, the other one in `.github/OWNERS` has only these approvers:
```
approvers:
  - upalatucci
  - metalice
  - galkremer1
  - adamviktora
```

PR #4740, opened 2026-09-16
CNV-96478: Remove view links from toast message if the targeted resource is deleted

galkremer1:
## 📝 Description

When uploading a file, in case the targeted resource (VM or DV) is deleted, the resource link in the toast message remained intact, leading to a not found resource. 

## 🔗 Links

Jira ticket: [CNV-96478](https://redhat.atlassian.net/browse/CNV-96478)

## 🎥 Demo

Before:

https://github.com/user-attachments/assets/d005c84b-836a-4523-af70-ee005aba9f9e

After:

https://github.com/user-attachments/assets/2af27eaf-c038-4cf9-be3f-a9cc198e3bde

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved upload progress handling to prevent stale retries, cancellations, or delayed errors from changing newer uploads.
  * Upload completion links now accurately reflect deleted, deleting, or unavailable virtual machines, disks, and bootable volumes.
  * Removed outdated resource links automatically after related storage resources are deleted, including cluster-specific resources.
  * Upload failures and completions now remain correctly associated with the active upload attempt.

* **User Experience**
  * Empty upload-link areas are no longer displayed.
  * Upload progress and success states are preserved during resource cleanup and cancellation.
  * Links are automatically updated when linked resources are detected as deleted.


upalatucci:
hei @galkremer1 what do u think about this strategy: 

Instead of passing a generic link, we pass the resource informations so we can watch the resource and see if it has beeing deleted (maybe in future we can even track other things) . 
If we need it, we can still have an option to pass a generic link so we can navigate to non-resources pages using the toaster link. 

More specifically, we can have `UploadSuccessLink` like this

```
export type UploadSuccessLink = {
  label: ReactNode;
  url?: string;
  resource?: {name: string; namespace: string; model: K8sModel}
};
```

or instead of `resourceUrl` and `resourceName`, we can jsut have `resource: {name, namespace, model}` 



This way, we don't have to handle anything on the function that call the toaster

galkremer1:
> hei @galkremer1 what do u think about this strategy:
> 
> Instead of passing a generic link, we pass the resource informations so we can watch the resource and see if it has beeing deleted (maybe in future we can even track other things) . If we need it, we can still have an option to pass a generic link so we can navigate to non-resources pages using the toaster link.
> 
> More specifically, we can have `UploadSuccessLink` like this
> 
> ```
> export type UploadSuccessLink = {
>   label: ReactNode;
>   url?: string;
>   resource?: {name: string; namespace: string; model: K8sModel}
> };
> ```
> 
> or instead of `resourceUrl` and `resourceName`, we can jsut have `resource: {name, namespace, model}`
> 
> This way, we don't have to handle anything on the function that call the toaster

Thanks @upalatucci , good idea. Let me look into that and see if it makes things more simple without losing functionality.

src/utils/hooks/useUploadProgressToast/completion/uploadLinks.ts
```diff
@@ -1,61 +1,126 @@
...
+  const filtered = links.filter((link) => !urlsToOmit.has(link.url));
+  return filtered.length === links.length ? links : filtered;
 };
 
-const getDataVolumeUrl = (dataVolumeName: string, namespace: string): string =>
+export const getDataVolumeUrl = (dataVolumeName: string, namespace: string): string =>
   getResourceUrl({
     model: DataVolumeModel,
     resource: { metadata: { name: dataVolumeName, namespace } },
   });
```
upalatucci: we need also the clsuter here right?  and do what we already do in `getBootableVolumeUrl`

galkremer1: @upalatucci good catch! Yes it was needed for showing the right links in ACM. Fixed.

PR #4761, opened 2026-09-18
[release-4.20] CNV-97824: Fix clearing SSH key to None in VM wizard

upalatucci:
## 📝 Description

When creating a VM via the wizard customization step, selecting **None** in the Public SSH key modal failed with an error instead of clearing the SSH key.

The **None** path called `detachVMSecret`, which PATCHed the VM in the cluster. In the wizard, the VM only exists in memory (`customizeWizardVMSignal`), so the API call failed.

This change routes the **None** path through `updateVM(removeSecretFromVM(vm))`, matching how **Use existing** and **Add new** already work. It also removes the now-unused `detachVMSecret` helper.

## 🔗 Links

Jira: https://issues.redhat.com/browse/CNV-97824

## 🎥 Demo

Before: Selecting None after adding an SSH key shows an error and does not clear the key.

After: Selecting None clears the SSH key and closes the modal without error.

## Test plan

- [ ] Start VM creation wizard with custom configuration
- [ ] On Customization step, add a Public SSH key and save
- [ ] Re-open Public SSH key modal, select **None**, and save
- [ ] Verify the SSH key is cleared and no error is shown
- [ ] Verify clearing SSH key to None still works on an existing VM details page


upalatucci:
Superseded by main-targeted PR.

PR #4763, opened 2026-09-18
CNV-96250: disable bulk VM actions when user lacks namespace permissions

upalatucci:
## 📝 Description

Non-privileged users could still use folder-level and bulk VM actions (Control, Take snapshot, Move to group, Edit VM labels) in namespaces where they only have view access. Those actions failed with Forbidden or silently in the console.

This adds `accessReview` checks to all bulk VM action configs, matching the existing single-VM action patterns. When the user lacks permission, actions are disabled with the standard *"You don't have permission to perform this action"* tooltip.

Any other silent failures will be hanlded in this epic: https://redhat.atlassian.net/browse/CNV-95267

## 🔗 Links

Jira: https://issues.redhat.com/browse/CNV-96250

Related: [CNV-96236](https://redhat.atlassian.net/browse/CNV-96236) (action tooltips), [CNV-96238](https://redhat.atlassian.net/browse/CNV-96238) (snapshot), [CNV-96242](https://redhat.atlassian.net/browse/CNV-96242) (config patch)

**BEFORE**

https://github.com/user-attachments/assets/8dedadb0-313f-4f63-a7f2-9fb134965843

**AFTER**

https://github.com/user-attachments/assets/bb293db7-95a9-4413-a8fc-e10884913dd9

## Test plan

- [ ] Log in as a non-privileged user with view-only access to a namespace containing VMs
- [ ] Select a folder in the VM tree (or select multiple VMs) and open the Actions menu
- [ ] Verify Control, Take snapshot, Move to group, and Edit VM labels are disabled with permission tooltip
- [ ] Log in as a user with edit access and verify the same actions remain enabled
- [ ] Confirm row-level kebab actions still behave as before

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved access checks for bulk virtual machine lifecycle actions, including start, stop, restart, pause, unpause, and reset.
  * Improved authorization handling for bulk delete, label editing, snapshot, run-strategy, and folder-move actions.
  * Preserved action-specific confirmations and warnings for applicable bulk lifecycle operations.


src/views/virtualmachines/actions/accessReviewUtils.ts
```diff
@@ -0,0 +1,23 @@
...
+import { type K8sModel, type K8sVerb } from '@openshift-console/dynamic-plugin-sdk';
+import { type FleetAccessReviewResourceAttributes } from '@stolostron/multicluster-sdk';
+
+export const asBulkAccessReview = (
+  model: K8sModel,
+  vms: V1VirtualMachine[],
+  verb: K8sVerb,
+  subresource?: string,
+): FleetAccessReviewResourceAttributes | undefined => {
+  const accessReview = asAccessReview(model, vms?.[0], verb, subresource);
```
galkremer1, line 12: I agree with the Rabbit - we need to check that the user has access to all VMs, not just the first one. We need to have two disabled tooltips - one for having no access to all vms and one to some of them.

upalatucci: oh yea okay. it would be more precise . when i startedworking on it seemed not possible to check for start and stop for the whole folder. But actually folder will have same permissins as the namespace

upalatucci: the thing is: those are the same actions and permissions checks done for the selected vms and if the user select like 1k vms we can't checks for all of them. i'll try to propose a different solution then

upalatucci: So now we group the vms and if the vms are in the same namespace we would check for that namespace permissions. if they comes from different namespaces we'll not check for permissions.

galkremer1: Ok we need to think about a proper solution later on then.

galkremer1 (approved):
/lgtm

